PR #95, opened 2012-03-22
Add ecat to the list of known imageclasses to ease loading with .load()

yarikoptic:
I am not sure if that is a logical thing to do -- just did it while testing nib-ls  so decided to "ask" with this PR

### DIFF
--- a/nibabel/imageclasses.py
+++ b/nibabel/imageclasses.py
@@ -14,6 +14,7 @@ from .nifti1 import Nifti1Pair, Nifti1Image
 from .minc import MincImage
 from .freesurfer import MGHImage
 from .volumeutils import Recoder
+from .ecat import EcatImage
 
 # If we don't have scipy, then we cannot write SPM format files
 try:
@@ -52,11 +53,16 @@ class_map = {
     'mgh':{'class': MGHImage,
            'ext': '.mgh',
            'has_affine': True,
-           'rw':True},
+           'rw': True},
     'mgz':{'class': MGHImage,
            'ext': '.mgz',
            'has_affine': True,
-           'rw':True}}
+           'rw': True},
+    'ecat':{'class': EcatImage,
+           'ext': '.v',
+           'has_affine': True,
+           'rw': False},
+    }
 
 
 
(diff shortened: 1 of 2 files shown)


bcipolli:
@matthew-brett what about allowing `load` to return a list of images? People using Ecat would then simply need to be ready for that.

matthew-brett:
Hmm - yes - I'd been thinking about that for a looooooooong time.

My best idea was one that Fernando suggested a while ago - to have a 'load_multi' function that could load things like ECAT as a list of images.   Maybe we could load things like 4D niftti with this interface too, returning a list of 3D volumes, each with their own affines (as we would need to do, for motion corrected runs with individual affines, as saved by SPM. 

How about a google hangout in the next few days?  Tomorrow?    If you're around I propose on the list and do a doodle for it.

bcipolli:
@matthew-brett I'm away for the weekend, but OK to hangout next week. I'm fine if you simply have a discussion on the list as well--whatever works for you!

matthew-brett:
Away for the weekend ?  Crazy !   Surely the weekends are for intensive development work :)

OK - I'll have a think and write a small BIAP, to get my slower gears spinning.  I'll try to get that done tomorrow for a discussion of some sort next week.

bcipolli:
Sounds good. To tip my hand, my question is going to be: what's the benefit of `load_multi` over putting this into `load` and letting class functions decide whether to return an image or a list of images? In addition, I am concerned: would you have to know the contents of the `Ecat` file before loading it, in order to know which function to call?

Looking forward to your BIAP (whatever that means :) )

matthew-brett:
Yes, sure, these are the important questions.

For ECAT, the idea would be you would always use load_multi (deprecation warning from load).

Obviously for nifti you sometimes don't care about multiple affines and do want a single 4D image, other times you may want separate images.

I'm touched you are looking forward to the mysterious BIAP - A BIAP is a niBabel Increased Awesomeness Proposal : https://github.com/nipy/nibabel/wiki

bcipolli:
LOL :+1:

PR #215, opened 2013-11-23
WIP: Save SpatialImages into HDF5 files.

alexsavio:
First commit: functions to save a Nifti1Image into a HDF5 file group and retrieve it back.

### DIFF
--- a/nibabel/hdfloadsave.py
+++ b/nibabel/hdfloadsave.py
@@ -38,23 +38,28 @@ def nifti1img_to_hdf(fname, spatial_img, h5path='/img', append=True):
     append: bool
     True if you don't want to erase the content of the file
     if it already exists, False otherwise.
+
+    @note:
+    HDF5 open modes
+    >>> 'r' Readonly, file must exist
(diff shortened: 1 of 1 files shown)


bcipolli:
I looked this over, and this seems more useful for defining datasets, rather than individual images. I think of `nibabel` as working on individual images, not datasets. My thought is to close this...

Perhaps something useful would be to have a website example of how to use HDF5 to store an entire dataset in a single HDF5 file (which is what the pymvpa folks did for the Haxby 2001 and 2011 datasets).

matthew-brett:
Ben - can you give links to the PyMVPA hdf5 datasets?

bcipolli:
@matthew-brett please use http://data.pymvpa.org/datasets/hyperalignment_tutorial_data/hyperalignment_tutorial_data_2.4.hdf5.gz.

If you'd like code for downloading and loading the file, please install [nidata](http://github.com/nidata/nidata) and run:
`ipython nidata/atlas/haxby_etal_2011/example1.py`

PR #216, opened 2013-12-03
Support for BrainVoyager VTC, MSK, VMR and NR-VMP files

thomastweets:
After talking to Emanuele Olivetti and Yaroslav Halchenko about it I submit this pull request for implementing support for BrainVoyager QX files. I started out with VTC, MSK, and NR-VMP files and read and write does work. As this is my first major pull request I would really appreciate some comments or help. I did not write nose tests yet and the code is not thoroughly tested yet either.

I hope that my implementation is not too "dirty" as it was a little difficult for me to implement the variable-length header of BrainVoyager files...


flrgsr:
Hi. Would be really great if this would be usable already. Can I somehow manually merge these changes locally to make this functionality available?

matthew-brett:
Excellent timing - I have just finished teaching for this semester, and have much more time for nibabel.

@thomastweets - can I help with this?  It would be very good to get this merged.

thomastweets:
I am very sorry to seemingly have abandoned this PR. There was simply too much other work going on.
However, I am still very interested in getting this out.
As a first measure I squashed and rebased my fork to work with an up-to-date nibabel. I used this guide for doing so, please tell me if you think that it is not a good idea to do so: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

I will try to take time this week to implement a more clean way of parsing the header using OrderedDict and writing tests.

@matthew-brett Thank you for your documentation work in the developers section and the clear test_image_api!

marklescroart:
Hello, 
I'm interested in getting this functionality into nibabel, too - what's the status of this project? It seems to have been stalled for a while, any chance of it being resuscitated / finished? Can I offer some cheerleading to move it along? (Go go go! Your efforts are appreciated!)

matthew-brett:
I'm very sorry to be slow - Thomas - can I help?  I am mainly free at the moment, and will be for the next few weeks - is there anything I can do?

matthew-brett:
For example - would it be worth doing another short hangout?   I'm on European time, it that helps.

matthew-brett:
I added a beginning doc on BV formats : https://github.com/nipy/nibabel/pull/327

thomastweets:
Also from my side: sorry to be so slow. Great that you started adding some file format documentation. I will definitely help out there as some things are not completely documented.
As for a hangout: actually, that would be great! Maybe we could set the steps that are still necessary to complete this and also shortly discuss the header parsing options (OrderedDict etc.). I am on European time, too (GMT+2) and would be pretty available today and tomorrow.

matthew-brett:
I'm should be around tomorrow (Saturday) at various times - can you email me at matthew dot brett  gmail to give me your address?

It looks like there's some energy for a hangout early next week - we could either do two (fine by me) or combine them.

matthew-brett:
From our phone call today:

```
In [4]: nib.load('T1.nii.gz').shape
Out[4]: (80, 80, 10)

In [5]: nib.load('T2.nii.gz').shape
Out[5]: (80, 80, 10)

In [6]: nib.load('T2_.nii.gz').shape
Out[6]: (80, 80, 10, 2)

In [7]: pwd
Out[7]: u'/Users/mb312/dev_trees/nibabel/nibabel-data/nitest-balls1/NIFTI'
```

So - the `nitest-balls1` submodule has time courses with two time points, and includes DICOM.  You could maybe add the brainvoyager conversions into that repository and make a pull request?

matthew-brett:
For `OrderedDict`:

```
In [1]: from collections import OrderedDict
In [2]: d = OrderedDict()
In [3]: d['z'] = 'z'
In [4]: d['a'] = 'a'
In [5]: list(d)
Out[5]: ['z', 'a']
In [6]: list(dict(d))
Out[6]: ['a', 'z']
In [7]: list(dict(d))
Out[7]: ['a', 'z']
```

Also : http://pymotw.com/2/collections/ordereddict.html

matthew-brett:
I merged my sketch of the BV format document.

thomastweets:
Here is a refactored implementation using OrderedDict to parse the header. Right now only VTC files work. There is a very tiny test file in nibabel/tests/data/test.vtc (it is random data).

I think this new header parsing is indeed much cleaner and closer to original 'feel' of the data format.
I also implemented some tests for the new header parsing. And I implemented the test_image_api tests for VTC files.

@matthew-brett If you agree on this refactoring then I would continue to implement the remaining file formats (msk, vmp) this way and add tests for them.

One more question: One test_image_api test is still failing (validate_header) because of the order of dimensions in BV files (zyxt). Could you give me a hint where to find the 'on-the-fly' reslicing of proxyArrays as implemented for other file formats (you mentioned minc). I somehow did not find it. Thanks!

thomastweets:
BTW I tried to set affine to None for the tests but that did not work. I initially implemented some kind of basic affine transformation to at least get the axes in the right order and orientation (RAS). See bv.py.
What do you think of it?

Once we agree on the new implementation I will also add more BV-specific tests (apart from the test_image_api tests for the other file formats).

thomastweets:
@matthew-brett Did you have time to look at the implementation? I don't want to be pushy (after this already took so long) but these weeks I have some time to complete this :)

matthew-brett:
Thanks for reminding me.   I will look this evening or tomorrow.   Now I think about it, setting the affine to the best shot you can give doesn't seem like a bad idea.

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+    startPos: int, optional
+       Define the start file position from which to search. If None then start
+       where the file object currently points to. Default is None.
+    strip : bool, optional
+       Whether to strip the trailing zero from the returned string.
+       Default is True.
+    rewind: bool, optional
+       Whether the fileobj f should be returned to the initial position after
+       reading. Default is False.
+    Returns
```
matthew-brett: Blank line between parameters section and returns section.

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+    -------
+    [str] : list of string(s)
```
matthew-brett: Suggest something like `str_list : list of string(s)`

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+    lines = data.split('\x00')
+    if rewind:
+        f.seek(currentPos)
+    else:
+        offset = 0
+        for s in range(nStrings):
+            offset += len(lines[s])+1
+        f.seek(currentPos+offset)
+    for s in range(nStrings):
+        yield lines[s] + suffix
```
matthew-brett: You said in the returns section this returns a list, but in fact this is a generator.   I guess no real need for generator here?  But if you do want a generator, you could rename the function and rewrite docstring.

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+
+
+def parse_BV_header(hdrDict, fileobj):
+    """Parse the header of a BV file format.
+
+    This function can be (and is) called recursively to iterate through nested
+    fields (e.g. the prts field of the VTC header).
```
matthew-brett: Consider using double backticks to indicate `prts` is a technical BV term?  As in

```
the ``prts`` field of the VTC header
```

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+    else:
+        offset = 0
+        for s in range(nStrings):
+            offset += len(lines[s])+1
+        f.seek(currentPos+offset)
+    for s in range(nStrings):
+        yield lines[s] + suffix
+
+
+def parse_BV_header(hdrDict, fileobj):
```
matthew-brett: `hdr_dict`

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+    Parameters
+    ----------
+    hdrDict: OrderedDict
+       Pre-populated hdrDict that contains the fields to expect for the
+       respective BV file format.
+    fileobj : fileobj
+       File object to use. Make sure that the current position is at the
+       beginning of the header (e.g. at 0).
+    """
+    for key in hdrDict.keys():
```
matthew-brett: You can also write `for key in hdrDict:` here.

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+
+def parse_BV_header(hdrDict, fileobj):
+    """Parse the header of a BV file format.
+
+    This function can be (and is) called recursively to iterate through nested
+    fields (e.g. the prts field of the VTC header).
+
+    Parameters
+    ----------
+    hdrDict: OrderedDict
```
matthew-brett: Maybe some notes on the format of `hdrDict`?

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+                # recusively iterate through the fields of all items
+                # in the array
+                hdrSize += calc_BV_header_size(hdrDict[key]['value'][i])
+        else:
+            hdrSize += calcsize(hdrDict[key]['dt'])
+    return hdrSize
+
+
+class BvError(Exception):
+
```
matthew-brett: PEP8 - no need for blank lines here.

nibabel/bv.py
```diff
@@ -0,0 +1,691 @@
...
+        ----------
+        binaryblock : {None, string} optional
+            binary block to set into header.  By default, None, in
+            which case we insert the default empty header block
+        endianness : {None, '<','>', other endian code} string, optional
+            endianness of the binaryblock.  If None, guess endianness
+            from the data.
+        check : bool, optional
+            Whether to check content of header in initialization.
+            Default is True.
```
matthew-brett: Description for `offset`

nibabel/bv.py
thomastweets: I ran a pep8 linter: https://www.python.org/dev/peps/pep-0008/#blank-lines

matthew-brett: The PEP8 linter was OK with these blank lines?     My understanding of PEP8 is to avoid blank lines unless there is a good reason for them - for example to separate parts of a function that do different tasks.  In this case I think the blank lines aren't useful because they make the eye jump, between parts of one thing, which is a minimal class with a docstring.

thomastweets: The linter even asked for these lines. As in the link I pasted: "Surround top-level function and class definitions with two blank lines."

matthew-brett: Ah - no - that just means that there should be two blank lines between the end of one function / class definition and the beginning of the next one.   I think your PEP8 linter is wrong here - my `pep8` checker allows either version.

nibabel/bv.py
```diff
@@ -0,0 +1,703 @@
...
+                # recusively iterate through the fields of all items
+                # in the array
+                hdr_size += calc_BV_header_size(hdr_dict[key]['value'][i])
+        else:
+            hdr_size += calcsize(hdr_dict[key]['dt'])
+    return hdr_size
+
+
+class BvError(Exception):
+
```
matthew-brett: For example see `Photo` class in https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-classes for lack of blank line after class line and before docstring.

thomastweets: OK, I removed the blank lines and switched from pep8 to pylint (which accepts these blank lines).

matthew-brett: How odd - my `pep8` is OK with either version.  I was using `pep8` version 1.5.6, but I've just upgraded to 1.6.2, same result - i.e. `pep8 nibabel/bv.py` gives no output, with or without the blank lines.

thomastweets: Precisely, I use https://github.com/klen/pylama (even more precisely the atom.io package for it) and I switched from PEP8 to PyLint _in there_ so maybe they use different versions of the packages than the most recent updates...
However, for now I eliminate blank lines ;)

matthew-brett: But in general, as PEP8 says - readability and consistency are more important than strict PEP8:

```
A style guide is about consistency. Consistency with this style guide is
important. Consistency within a project is more important. Consistency within
one module or function is most important.

But most importantly: know when to be inconsistent -- sometimes the style
guide just doesn't apply. When in doubt, use your best judgment. Look at other
examples and decide what looks best. And don't hesitate to ask!

In particular: do not break backwards compatibility just to comply with this
PEP!

Some other good reasons to ignore a particular guideline:

1) When applying the guideline would make the code less readable, even for
someone who is used to reading code that follows this PEP.
```

matthew-brett:
About the header parsing:

Am I right in saying you are using two header definition / parsing schemes
here?

In `bv_vmp`, I think you are compiling a numpy datatype dynamically, based on
calculated null string lengths and given number of sub-maps.

`bv_msk` uses a standard numpy datatype definition.

`bv_vtc` uses a new custom `OrderedDict` prototype format.

The format has (key, value) pairs with key == data field name, value == a
dict.  Call this value dictionary valdict.

The valdict has key, value pairs with keys usually ==  (`dt`, `default`).

`dt` is the data type in `struct.pack` format, except that `s` is a code for a
zero-terminated string, or `dt` may contain the special value `multi`.

`default` is the default value of the field.

If `dt` == "multi" then valdict also has key, value (`nField`, `<field name>`)
where `<field name>` is the name of the field in the enclosing ordered dict,
that contains the number of values for this multiple-value field.

In the case `dt` == "multi", the `default` is a list containing something that
looks like another header protocol, but also defining a `value`.

Reading the header involves creating a new ordered dictionary that is a copy
of the prototype, but where each valdict has key, value of (`value`, `<val>`)
where `<val>` is the value for key `default`.

The reading proceeds by reading from the header file using the `dt` information, and filling in the `value` field.

Is that right?

nibabel/bv_vtc.py
```diff
@@ -0,0 +1,250 @@
...
+                OrderedDict([
+                    ('filename', {'dt': '<s', 'default': '', 'value': ''})
+                ])
+            ]}
+         ),
+        ('currentPrt', {'dt': '<h', 'default': 0}),
+        ('datatype', {'dt': '<h', 'default': 2}),
+        ('volumes', {'dt': '<h', 'default': 0}),
+        ('Resolution', {'dt': '<h', 'default': 3}),
+        ('XStart', {'dt': '<h', 'default': 57}),
```
matthew-brett: These values seem terribly arbitrary.  Is it easy to explain why 57 would be a sensible default value here?

thomastweets: Yes, there (kind of) is: http://support.brainvoyager.com/installation-introduction/23-file-formats/379-users-guide-23-the-format-of-vtc-files.html

Look at the 'default' column. Also, NeuroElf implements it this way.

thomastweets:
OK, let me comment:

> Am I right in saying you are using two header definition / parsing schemes here?

Sorry, I did only mention it in the commit message: I only implemented bv_vtc.py in the 'new' header definition/parsing scheme because I wanted to check with you whether this new implementation is OK before using it for every file format.

> The format has (key, value) pairs with key == data field name, value == a dict. Call this value dictionary valdict.

Correct.

> The valdict has key, value pairs with keys usually == (dt, default).

It also has the key 'value' with a value that is copied from 'default' during the prototype creation (to make it more readable/less repetitive).

> dt is the data type in struct.pack format, except that s is a code for a zero-terminated string, or dt may contain the special value multi.

> default is the default value of the field.

> If dt == "multi" then valdict also has key, value (nField, <field name>) where <field name> is the name of the field in the enclosing ordered dict, that contains the number of values for this multiple-value field.

Correct. I will add more documentation for that in the _make_vtc_hdrDict() function docstring.

> In the case dt == "multi", the default is a list containing something that looks like another header protocol, but also defining a value.

Correct. I did that to be able to recursively invoke parse_BV_header(). In (at least) the VMP file format there are even more nested headers so that a recursive way of parsing becomes handy.
The explicit definition of the key (and value) 'value' was laziness, to be honest, a recursive copy operation at the end of _make_vtc_hdrDict() would indeed be more consistent. I will implement it.

> Reading the header involves creating a new ordered dictionary that is a copy of the prototype, but where each valdict has key, value of (value, <val>) where <val> is the value for key default.

Correct (see above).

> The reading proceeds by reading from the header file using the dt information, and filling in the value field.

Correct.

matthew-brett:
The the `defaults` - would it be enough to specify a single default full header, including sub-parts such as `prts` in VTC files?  Or do you need to know the default separately for each sub-part?

thomastweets:
In the prototype I would always only have one sub-part of a multi-field (as far as I know BV file formats). In any way I would keep the default between them the same and copy from it as new (still empty) sub-parts are generated (by the user).
Again, the similarity of the enclosing OrderedDict and the sub-part OrderedDict is to be able to work recursively. This is why I would keep defaults there, too (if was your point).

matthew-brett:
I am trying to see if there is a way to simplify the prototype specification (for example, omitting the default) and / or separating the prototype and the generated value.

Thinking about it, you probably do need 'defaults' for each value, but  maybe you could simplify the specification.

For example, maybe you could have `z` for zero-terminated string (to make
clear the distinction between `s` (which must have a length) and the
zero-terminated string.  Then have the specification something like:

```
"""
https://docs.python.org/3/reference/introduction.html#notation

(except using double parens indicating a group, single parens indicating a tuple).

elemement_proto ::= (name, pack-format, default)  |
                    name, (n_fields_name, <element_proto>)
proto ::= ((element_proto))*
"""

VTC_HEADER_PROT0 = (
    ('version', 'h', 3),
    ('fmr', 'z', b''),
    ('nPrt', 'h', 0),
    ('prts', ('nPrt',
              ('filename', 'z', b''))),
    ('currentPrt', 'h', 0),
    ...
    ('TR', 'f', 2000.0)
    )
```

Then you pass in this read-only specification, returning a read header in the form of an OrderedDict.

I guess, for speed, you could even pre-parse this thing, to give blocks that can be read together, for example between the variable length fields, but that's a detail.

What do you think?

thomastweets:
I implemented your suggestion about the header prototype definition. It is indeed much nicer!

If you agree on the rest (?) I will go ahead and implement the remaining file formats and tests.

nibabel/bv.py
```diff
@@ -0,0 +1,738 @@
...
+    Parameters
+    ----------
+    hdrDict: OrderedDict
+       Pre-populated hdrDict that contains the fields to expect for the
+       respective BV file format.
+    fileobj : fileobj
+       File object to use. Make sure that the current position is at the
+       beginning of the header (e.g. at 0).
+    """
+    for key in hdr_dict:
```
matthew-brett: How about combining `_make_hdr_dict` with this routine?   Then I think you can get rid of the `dt`, `value` fields and just have numbers, strings or lists (for multi values)?

thomastweets: Getting rid of default would not allow me to create valid empty ('new') BV files. For the same reason I though it is nice to separate ```_make_hdr_dict``` and ```parse_BV_header``` because one would only generate a default header from a given prototype and the other would be used for actual parsing. I can still of course combine it.
Getting rid of dt would break parsing and/or writing as some header fields have different datatypes for numbers (_e.g._ ```RefSpace``` in ```VTC``` files). Or how would you solve this without ```dt```?

matthew-brett: Ah - so `_make_hdr_dict` has two roles.  The first is to make a template into which you will insert all values, from a read header.  The second is to make a default set of header values.

Maybe you could replace `_make_hdr_dict` with an `empty_header` method that would compile the default from the prototype?

Is there any example of a header field name that is duplicated within a header?  For example `myfield` in the main header, than a subheader which also has `myfield`?   If not, I suppose you could store the datatype information in a dictionary with keys by name, and use that during writing?

thomastweets: OK, I will rename the ```_make_hdr_dict``` to ```empty_header```.

I can include such a lookup table, however, it would largely replicate the ```proto``` parameter for (now) ```empty_header``` (_e.g._ ```VTC_HDR_DICT_PROTO``` in bv_vtc.py) as I would need to provide datatypes for every single field. Although it is quite likely that there would not be a duplicate fieldname within one filetype, I am not 100% certain about that, either.
If you think that way it becomes clearer, I will do it, anyway ;)

matthew-brett: Sorry to be demanding ...

I was going to write a longish reply, but then I realized I needed to think it out myself.  I ended up writing this PR to your repo:

https://github.com/thomastweets/nibabel/pull/2

I think the simplest thing to do, is to separate the generation of the empty (default) header from the reading and writing. 

I rewrote the `_make_hdr_dict` into a function `_proto2default` that generates a default header from the prototype description.  To keep track of the format information, I passed the prototype to the parse, pack and calc functions.

In writing, I realized the prototype description was a little clumsy to work with, so I refactored it to be of form `name, format, other`, where `format` is a prototype for a multivalue field, and `other` is either the default value, or the name of the field giving the number of values.

I think that makes the resulting read header cleaner, because you don't need to store a dict value for each field.

What do you think?

matthew-brett:
I also found myself thinking about reading zero terminal strings, and thinking that it would be good to allow strings of arbitrary length.   Once I got into that, I found I couldn't help myself and ended up with:

https://github.com/nipy/nibabel/pull/334

It's based on your code - would you mind taking a look?

thomastweets:
Short status update:
The last weeks have been busy scanning and finishing some project and now I am going on vacation until the end of September. As you see, I implemented the MSK file format using the new header parsing. I got a little stuck, however, trying to implement the VMP file format. There is another need for a change in the hdr_proto_dict format as in the VMP header there are fields that are only existing conditionally depending on the content of other fields (see http://support.brainvoyager.com/automation-aamp-development/23-file-formats/377-users-guide-23-the-format-of-nr-vmp-files.html for example: NrOfLags - only stored if "TypeOf Map" = 3). I started adding functionality for that, however, did not finish yet. I will continue once I am back in October.

matthew-brett:
Thanks for the update.  Please do post anything you're working on, if you want feedback, even if it not finished.   Thanks for persisting with this.

bcipolli:
@thomastweets There were some changes to the `load`/`save` code since you last updated this PR. I've created a PR against your branch that merges in the latest master and makes the necessary changes.

Hard to know if anything was broken, as there don't seem to yet be tests, but I believe it's right and hope it helps!

thomastweets:
A short update: I am pretty busy these days finishing my PhD thesis and writing some papers. However, I will start to work more closely with the company BrainVoyager soon and finishing this PR (finally!) is a high priority. I expect to work on this in April/May. Sorry for all the delays once more!

matthew-brett:
Thanks for the update.  Good luck with the thesis.

I guess you don't mind others working on this if they get the time, before you do?  I don't know if anyone can do that, but I wanted to check with you that that was OK.

thomastweets:
Thanks!
And sure, I would be happy if other people work on it, too! Of course I will assist with any questions that arise.

bcipolli:
I am putting this on the current master, making it `flake8`-compatible, and will try to migrate to the current object/file structure.

bcipolli:
@thomastweets I just created a PR against your branch, to help sync this up a bit and make further development easier.

matthew-brett:
Thomas - any chance you could have a look at Ben's PR ?

thomastweets:
Hello there! I am back from thesis nirvana :)
So I finally merged Ben's PR in a way that effects this PR. However, this overwrites some of the progress we made last summer/fall. So I will in the next days work on updating this again and applying the new prototype header to the VMP file format.
One more update: I started working on a temporary position at BrainInnovation (company behind BrainVoyager) and as Python interfacing becomes more important here, I can spend more time on this, finally!

matthew-brett:
Congratulations on finishing the thesis - and on your new job.

Please let us know if we can help with review or a hangout.    Happy hunting, and thanks for the update.

matthew-brett:
Thomas - put a message here when stuff is ready for review?  I will try very hard to be timely.

thomastweets:
Hi Matthew! Yes, I will put a message here as soon as things are review-ready. Thank you for your patience!

thomastweets:
@matthew-brett I am making a lot of progress and we are close to review. I now implemented test_image_api for all BrainVoyager file formats. However, I cannot figure out why AppVeyor and TravisCI fail to import from the brainvoyager (subfolder) module (see the logs). When I import nibabel locally everything works fine. Do you have any idea what is going wrong?

matthew-brett:
I think you have to add `nibabel.brainvoyager` to `setup.py` `packages` so Python knows to install the brainvoyager directory.

thomastweets:
Thanks for the hint! As you see, however, now it does not build at all. Any further ideas?

matthew-brett:
I think the problem is that your edits to setup.py conflict with recent changes in nibabel, so Appveyor is complaining that it can't merge this PR before testing it.

How are your git skills?   I think you or I will need to do a merge (from current master) or a rebase (onto current master).

thomastweets:
OK, I will try a rebase (I did so last year, too. See my comment above on Jun 15, 2015).

thomastweets:
We are very close to the finish here. There some more checks and filetype-depending tests I want to write but the main functionality is ready and tested. There is one more error happening for python 3 tests and I cannot figure it out. Maybe you could help me out here? I included some print statements into the respective test (validate_header_deprecated in test_image_api). Somehow, BvMskImage and BvVmrImage raise an additional warning here that fails the test. Maybe it depends on their dtype?
Thanks for looking into it!

matthew-brett:
Great to hear we are close - the marathon is nearly over!

Do you mean the failing tests for the matrix entry with numpy ``--pre`` ?  That installs and tests against current numpy master - or a recent build of numpy master.  Can you set up a virtualenv and compile numpy from git source, and run the tests, to debug locally?

thomastweets:
I actually meant the failing appveyor tests.
However, I will also try to figure out the `--pre` flag test fails and correct for the linting errors. Thanks for the hint!

thomastweets:
Great! All TravisCI tests are OK now (other than the style checks that fail in the parrec and nifti1 files). The python3 tests on Windows/AppVeyor still fail with the same issues I described above. I will continue to search for a solution but would be happy about a hint :)

thomastweets:
Actually, this may refer to https://github.com/numpy/numpy/issues/6685 as the second warning that fail `validate_header_deprecated` results from an error in np.rint. Any clues? Can we ignore this?

matthew-brett:
Hmm - I don't understand that error.   Why not try merging or (even better) rebasing on current master - I've just merged some fixes that Ariel Rokem just made to the Appveyor setup.

matthew-brett:
I also did some flake8 fixes on the master branch.

thomastweets:
Yay! Green light on TravisCI for the first time! :)
There are still the same issues in python3 tests on AppVeyor and the Python35-conda32 environment seems to have issues in general. I will continue to search for a solution here and also include some more file-specific tests.

matthew-brett:
Thomas - I'm planning a release soon, and I would like to leave these changes out of the master branch until we've finished review, and so leave this PR out of the next release.  We can do another release as soon as these changes are ready.  Is that OK with you?

thomastweets:
@matthew-brett Of course, that's OK with me!

thomastweets:
OK, here is what I found about the python3/Windows error:
`np.rint` freaks out on python3/windows when it tries to round `inf` or `-inf` and raises an `*** FloatingPointError: invalid value encountered in rint`. This happens during casting `float32` data arrays in the `validate_header_deprecated` tests for the VMR and MSK file formats that want `uint8` data arrays. This does not happen on any other environment. I am not sure what the best way around this would be. I could set a `mn` and `mx` parameter for the `array_to_file` call for these filetypes but this is an ugly workaround for an issue that might affect other filetypes, too. Indeed, if you look at the AppVeyor log (https://ci.appveyor.com/project/Eric89GXL/nibabel/build/1.0.812/job/8l5fyj2m31xd5etc) you see that the Warning appears a lot of times for other tests, too.
@matthew-brett Do you have an idea here? It seems that [you](https://github.com/numpy/numpy/issues/6685) [dealt](http://numpy-discussion.10968.n7.nabble.com/Nice-float-gt-integer-conversion-td8408.html) [with integer rounding](https://github.com/numpy/numpy/issues/7004) in numpy a lot before...

matthew-brett:
Hum - I can replicate this.  Specifically, in a conda enviroment, with this test file:

```
# test_rint.py
import numpy as np
import warnings

with warnings.catch_warnings(record=True) as w:
    np.rint(np.inf)
    np.rint(np.inf)
print("Number of warnings", len(w))
```

The output of `python test_rint.py` is:

```
Number of warnings 2
```

I get the same answer using the Python.org Python installers on Windows, latest numpy from pip.

For OSX or Linux, I get `Number of warnings 0`.

So, I guess the answer is that, on Windows, np.rint raises a warning for np.inf, but not on other platforms.  I'll put up a numpy issue for that now.

For our purposes, I guess we need to work out a way of filtering the `rint` warnings for the tests.

thomastweets:
I added a filter for the numpy.rint warning into the test_image_api test. Now, all tests pass :)

matthew-brett:
Nice!

I noticed that coverage thinks that the code coverage drops a bit with this PR - did you have time to look into that?

matthew-brett:
Is this one ready for review now?   Did you have time to look into the coverage drop?

thomastweets:
I will spend the rest of the week trying to increase coverage a bit. I will post once more here once this is done and it is ready for review.

matthew-brett:
Thanks - that is helpful.

nibotmi:
:umbrella: The latest upstream changes (presumably #479) made this pull request unmergeable. Please resolve the merge conflicts.

thomastweets:
OK, so I added a lot of BrainVoyager specific tests now. The file coverage definitely increased, however, I am not sure whether coveralls will agree :)
From my side, this is now ready for review. Phew!

thomastweets:
@matthew-brett OK, finally coverage increased (and is very high for the BV-related files). This is now definitely ready for review :)

matthew-brett:
Thanks - I will try and get to this ASAP.

thomastweets:
@matthew-brett I know that this became quite a monster of a PR, but do you think it can be merged for the next release? Just to give you a heads-up: I will start a new job in September and might be (even) more unresponsive in the future unfortunately because of time constraints.

matthew-brett:
Sorry to take a while to get back to you, I just had the first day of a course I am teaching, and I got a bit overwhelmed.   Have time now, but, is it too late for you?  Do you have a day or so to respond to stuff, or will you be maxed out?

thomastweets:
Now it took _me_ a while to get back to you ;)
I am now a bit more settled already (move + new job). Over the next week I will be travelling but next Tuesday I should be ready to respond to comments. I will probably respond slower as I can only work on this during the night, though.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    rewind: bool, optional
+       Whether the fileobj f should be returned to the initial position after
+       reading. Default is False.
+
+    Returns
+    -------
+    str_list : generator of string(s)
+    """
+    currentPos = f.tell()
+    if strip:
```
matthew-brett: Consider `suffix = b'' if strip else b'\x00'`

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+       Default is True.
+    rewind: bool, optional
+       Whether the fileobj f should be returned to the initial position after
+       reading. Default is False.
+
+    Returns
+    -------
+    str_list : generator of string(s)
+    """
+    currentPos = f.tell()
```
matthew-brett: PEP8 naming conventions `current_pos` for `currentPos` etc.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+        suffix = b'\x00'
+    if startPos is not None:
+        f.seek(startPos)
+    data = f.read(bufsize)
+    lines = data.split(b'\x00')
+    str_list = []
+    if rewind:
+        f.seek(currentPos)
+    else:
+        offset = 0
```
matthew-brett: How about:

```
    offsets = [len(lines[s]) + 1 for s in range(nStrings)]
    f.seek(current_pos + sum(offsets))
```

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    ('x_start', 'h', 57),
+    ('x_end', 'h', 231),
+    ('y_start', 'h', 52),
+    ('y_end', 'h', 172),
+    ('z_start', 'h', 59),
+    ('z_end', 'h', 197),
+)
+
+
+def readCString(f, nStrings=1, bufsize=1000, startPos=None, strip=True,
```
matthew-brett: Consider `read_c_string` for PEP8 naming?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+                rewind=False):
+    """Read a zero-terminated string from a file object.
+
+    Read and return a zero-terminated string from a file object.
+
+    Parameters
+    ----------
+    f : fileobj
+       File object to use
```
matthew-brett: Mention that object should implement `tell, seek, read`.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+
+    This function can be (and is) called recursively to iterate through nested
+    fields (e.g. the ``prts`` field of the VTC header).
+
+    Parameters
+    ----------
+    hdr_dict_proto: tuple
+        tuple of format described in Notes below.
+    fileobj : fileobj
+        File object to use. Make sure that the current position is at the
```
matthew-brett: Object should implement `tell, seek, read`.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+        beginning of the header (e.g. at 0).
+    parent_hdr_dict: OrderedDict
```
matthew-brett:
```
    parent_hdr_dict: None or OrderedDict, optional
        Default is None.  None results in empty `OrderedDict`.  ... 
```

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+            if def_or_name in hdr_dict:
+                n_values = hdr_dict[def_or_name]
+            else:
+                n_values = parent_hdr_dict[def_or_name]
+            for i in range(n_values):
+                value.append(parse_BV_header(format, fileobj, hdr_dict))
+        # handle conditional fields
+        elif isinstance(def_or_name, tuple):
+            if hdr_dict[def_or_name[1]] == def_or_name[2]:
+                bytes = fileobj.read(calcsize(format))
```
matthew-brett: This is a builtin for Python 3 - rename to `bytes_` or similar?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    fields (e.g. the ``prts`` field of the VTC header).
+
+    Parameters
+    ----------
+    hdr_dict_proto: tuple
+        tuple of format described in Notes of :func:`parse_BV_header`
+    hdr_dict: OrderedDict
+       hdr_dict that contains the fields and values to for the respective
+       BV file format.
+    parent_hdr_dict: OrderedDict
```
matthew-brett: Document `None`, and is optional, as for `parse_BV_header` above.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    The n_fields_name is used to indicate the name of a header field that
+    contains a number for nested header fields loops (e.g. 'NrOfSubMaps' in the
+    VMP file header).
+
+    The c_fields_name and c_fields_value parameters are used for header fields
+    that are only written depending on the value of another header field (e.g.
+    'NrOfLags' in the VMP file header).
+    """
+    hdr_dict = OrderedDict()
+    for name, format, def_or_name in hdr_dict_proto:
```
matthew-brett: `format` is a builtin function, rename to `format_` or similar?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    fields (e.g. the prts field of the VTC header).
+
+    Parameters
+    ----------
+    hdr_dict_proto: tuple
+        tuple of format described in Notes of :func:`parse_BV_header`
+    hdr_dict: OrderedDict
+       hdr_dict that contains the fields and values to for the respective
+       BV file format.
+    parent_hdr_dict: OrderedDict
```
matthew-brett: Document `None`, and that it's optional, as for `parse_BV_header` above?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+        if format == 'z':
+            hdr_size += len(value) + 1
+        # handle array fields
+        elif isinstance(format, tuple):
+            # check the length of the array to expect
+            if def_or_name in hdr_dict:
+                n_values = hdr_dict[def_or_name]
+            else:
+                n_values = parent_hdr_dict[def_or_name]
+            for i in range(n_values):
```
matthew-brett: maybe

```
hdr_size = sum(calc_BV_header_size(format_, value[i], hdr_dict) for i in range(n_values))
```

matthew-brett (commented):
Just a tiny first set of suggestions, more tomorrow.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+            sub_parts = []
+            for i in range(n_values):
+                sub_parts.append(pack_BV_header(format, value[i], hdr_dict))
+            part = b''.join(sub_parts)
+        # handle conditional fields
+        elif isinstance(def_or_name, tuple):
+            if hdr_dict[def_or_name[1]] == def_or_name[2]:
+                part = pack('<' + format, value)
+            else:
+                continue
```
matthew-brett: Would you mind putting a comment here to explain this case?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+
+    Returns
+    -------
+    hdr_dict_new : OrderedDict
+        An updated version hdr_dict correcting effects of changed nested and
+        conditional fields.
+    """
+    for name, format, def_or_name in hdr_dict_proto:
+        # handle nested loop fields
+        if isinstance(format, tuple):
```
matthew-brett: Should the `format` be ignored in this case?    Maybe a comment here?  To reduce indentation and make intention clear, mabye

```
    if not isinstance(format, tuple):
        continue
    ...
```

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+        Size of header when packed into bytes ready for writing to file.
+    """
+    hdr_size = 0
+    for name, format, def_or_name in hdr_dict_proto:
+        value = hdr_dict[name]
+        # handle zero-terminated strings
+        if format == 'z':
+            hdr_size += len(value) + 1
+        # handle array fields
+        elif isinstance(format, tuple):
```
matthew-brett: This pattern seems to repeat a lot.  Can you think of a way of making this more general, to share code between the functions with these same sets of conditionals?  I was thinking vaguely about mapping these conditionals to actions with a dictionary or something that.  Maybe a silly idea, asking in hope you can come up with something.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+
+    Parameters
+    ----------
+    hdr_dict_proto: tuple
+        tuple of format described in Notes of :func:`parse_BV_header`
+    hdr_dict_old: OrderedDict
+       hdr_dict before any changes.
+    hdr_dict_new: OrderedDict
+       hdr_dict with changed fields in n_fields_name or c_fields_name fields.
+    parent_old: OrderedDict
```
matthew-brett: Note can be None, and is optional.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+       When update_BV_header() is called recursively the not yet updated
+       (parent) hdr_dict is passed to give access to n_fields_name fields
+       outside the current scope (see below).
+    parent_new: OrderedDict
```
matthew-brett: Can be None, is optional.

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    hdr_dict_new : OrderedDict
+        An updated version hdr_dict correcting effects of changed nested and
+        conditional fields.
+    """
+    for name, format, def_or_name in hdr_dict_proto:
+        # handle nested loop fields
+        if isinstance(format, tuple):
+            # calculate the change of array length and the new array length
+            if def_or_name in hdr_dict_old:
+                delta_values = hdr_dict_new[def_or_name] - \
```
matthew-brett: Vague preference for parentheses round expression to allow line breaks, rather than continuation character `\` : https://www.python.org/dev/peps/pep-0008/#id19

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+def combine_st(st_array, inv=False):
+    """Combine spatial transformation matrices.
+
+    This recursive function returns the dot product of all spatial
+    transformation matrices given in st_array for applying them in one go.
+    The order of multiplication follow the order in the given array.
+
+    Parameters
+    ----------
+    st_array: array
```
matthew-brett: I guess this is a 3D array shape (n, 4, 4)?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+
+    inv: boolean
+        Set to true to invert the transformation matrices before
+        multiplication.
+
+    Returns
+    -------
+    combined_st : array of shape (4, 4)
+    """
+    if len(st_array) == 1:
```
matthew-brett: How about:

```
from functools import reduce
if inv:
     st_array = [np.linalg.inv(aff) for aff in st_array]
return reduce(np.dot, st_array)
```

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,952 @@
...
+    transformation matrices given in st_array for applying them in one go.
+    The order of multiplication follow the order in the given array.
+
+    Parameters
+    ----------
+    st_array: array
+        array filled with transformation matrices of shape (4, 4)
+
+    inv: boolean
+        Set to true to invert the transformation matrices before
```
matthew-brett: This is a slightly odd transformation - can you comment more?  I was expecting you to invert the result of the multiplication, not the individual elements.

Is it right that the transformation you want is st_array[-1] followed by st_array[-2] ... etc?  I mean, to do mean the order of the arrays to be the order of the transformation, or the opposite (as you get from taking the dot product from left to right).

matthew-brett (commented):
Another bite-size bit of review.

nibabel/brainvoyager/bv.py
thomastweets: I changed all other comments (from this post) but here I run into failed tests when changing to this. I will still look into it once more, however, for now I left it like it was before.

nibabel/brainvoyager/bv.py
thomastweets: I agree that this is not the most beautiful style here. However, there are a lot of weird edge cases with BV headers and I fear that while trying to make things more general one would make things really hard to understand, too. At least this repetition is confined to these module level functions in bv.py and do not repeat for different file formats.
I will give it some more thought though.

nibabel/brainvoyager/bv.py
thomastweets: OK, I am waiting for a connection flight right now that was delayed 5 hours and I am already jetlagged, so I should definitely not make up my mind about spatial transformation matrices, BUT:
What I want here is to compute a transformation matrix from the orientation of the data in the BV file to their native orientation by reversing all the transformations that were performed on the data. These transformations are stored (e.g.) in the VMR file header in an ordered fashion. So I guess if taking the dot product from left to right gets me there, this is what I want.
But again, I will better try this out with a clear mind :)

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,957 @@
...
+    def as_analyze_map(self):
+        raise NotImplementedError
+
+    def set_data_offset(self, offset):
+        """Set offset into data file to read data."""
+        self._data_offset = offset
+
+    def get_data_offset(self):
+        """Return offset into data file to read data."""
+        self.set_data_offset(calc_BV_header_size(
```
matthew-brett: Doesn't this mean that any value set in `set_data_offset` will be ignored?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,957 @@
...
+           Should implement ``write`` method
+
+        Returns
+        -------
+        None
+        """
+        binaryblock = pack_BV_header(self.hdr_dict_proto, self._hdr_dict)
+        fileobj.write(binaryblock)
+
+    def check_fix(self, logger=None, error_level=None):
```
matthew-brett: This is the same as the version in `WrapStruct` I think.  I wonder whether there is a good way of re-using that code...

nibabel/brainvoyager/bv_vmp.py
```diff
@@ -0,0 +1,195 @@
...
+        hdr = self._hdr_dict
+        # calculate dimensions
+        z = (hdr['z_end'] - hdr['z_start']) / hdr['resolution']
+        y = (hdr['y_end'] - hdr['y_start']) / hdr['resolution']
+        x = (hdr['x_end'] - hdr['x_start']) / hdr['resolution']
+        n = hdr['nr_of_submaps']
+        return tuple(int(d) for d in [n, z, y, x])
+
+    def set_data_shape(self, shape=None, zyx=None, n=None):
+        ''' Set shape of data
```
matthew-brett: blank line here

nibabel/brainvoyager/bv_vmr.py
```diff
@@ -0,0 +1,254 @@
...
+    ('vox_res_z', 'f', 1.0),
+    ('flag_vox_resolution', 'B', 0),
+    ('flag_tal_space', 'B', 0),
+    ('min_intensity', 'i', 0),
+    ('mean_intensity', 'i', 127),
+    ('max_intensity', 'i', 255),
+)
+
+
+def computeOffsetPostHDR(hdr_dict, fileobj):
```
matthew-brett: PEP8 naming -> `compute_offset_post_hdr`

nibabel/brainvoyager/bv_vmr.py
```diff
@@ -0,0 +1,254 @@
...
+    currentSeek = fileobj.tell()
```
matthew-brett: Likewise `current_seek`.

nibabel/brainvoyager/bv_vmr.py
```diff
@@ -0,0 +1,254 @@
...
+    return currentSeek + (hdr_dict['dim_x'] * hdr_dict['dim_y'] *
+                          hdr_dict['dim_z'])
+
+
+def concatePrePos(preDict, posDict):
```
matthew-brett: Consider `concat_pre_pos` or maybe `merge_pre_pos`

nibabel/brainvoyager/bv_vmr.py
```diff
@@ -0,0 +1,254 @@
...
+            STarray = np.zeros((len(self._hdr_dict['past_spatial_trans']),
+                               4, 4))
+            for st in range(len(self._hdr_dict['past_spatial_trans'])):
+                STarray[st, :, :] = \
+                    parse_st(self._hdr_dict['past_spatial_trans'][st])
+            combinedST = combine_st(STarray, inv=True)
+            M = np.dot(M, combinedST)
+        return M
+
+    get_best_affine = get_base_affine
```
matthew-brett: Maybe define these in the base class, as in:

```
def get_best_affine(self):
    return self.get_base_affine()
```

etc?

nibabel/brainvoyager/bv_vmr.py
```diff
@@ -0,0 +1,254 @@
...
+
+    get_default_affine = get_base_affine
+
+    get_affine = get_base_affine
+
+    @classmethod
+    def from_fileobj(klass, fileobj, endianness=default_endianness,
+                     check=True):
+        hdr_dictPre = parse_BV_header(VMR_PRHDR_DICT_PROTO, fileobj)
```
matthew-brett: PEP8 variable naming.

nibabel/brainvoyager/tests/test_bv.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+def test_read_c_string():
+    # sample binary block
+    binary = b'test.fmr\x00test.prt\x00'
+    with InTemporaryDirectory():
+        # create a tempfile
+        path = 'test.header'
+        fwrite = open(path, 'wb')
+
```
matthew-brett: PEP8 suggests dropping blank lines unless separating blocks of commands.

nibabel/brainvoyager/tests/test_bv.py
```diff
@@ -0,0 +1,315 @@
...
+        # test readout of one string from given position
+        fread.seek(0)
+        assert_equal([s for s in read_c_string(fread, start_pos=9)],
+                     [b'test.prt'])
+        fread.close()
+
+
+def test_combine_st():
+    vmr = BvVmrImage.from_filename(vmr_file)
+    STarray = []
```
matthew-brett: PEP8 naming -> `st_array`

nibabel/brainvoyager/tests/test_bv.py
```diff
@@ -0,0 +1,315 @@
...
+    for st in range(vmr.header._hdr_dict['nr_of_past_spatial_trans']):
+        STarray.append(parse_st(
+            vmr.header._hdr_dict['past_spatial_trans'][st]))
+    STarray = np.array(STarray)
+    combinedST = combine_st(STarray, inv=True)
+    correctCombinedST = [[1., 0., 0., 0.],
```
matthew-brett: Is this test sensitive to the order that transformations are applied?

nibabel/brainvoyager/tests/test_bv.py
```diff
@@ -0,0 +1,315 @@
...
+                         [0., 1., 0., -1.],
+                         [0., 0., 1., 1.],
+                         [0., 0., 0., 1.]]
+    assert_array_equal(combinedST, correctCombinedST)
+    combinedST = combine_st(STarray, inv=False)
```
matthew-brett: Is this test sensitive to order?

nibabel/brainvoyager/tests/test_bv.py
```diff
@@ -0,0 +1,315 @@
...
+              b'\x00sample\x00'])
+
+
+def test_read_c_string():
+    # sample binary block
+    binary = b'test.fmr\x00test.prt\x00'
+    with InTemporaryDirectory():
+        # create a tempfile
+        path = 'test.header'
+        fwrite = open(path, 'wb')
```
matthew-brett: If you can be bothered, nicer as:

```
with open(path, 'wb') as fwrite:
    fwrite.write(binary)
```

Similarly for `open` usage below.

nibabel/freesurfer/mghformat.py
```diff
@@ -281,7 +281,7 @@ def set_data_dtype(self, datatype):
         try:
             code = self._data_type_codes[datatype]
         except KeyError:
-            raise MGHError('datatype dtype "%s" not recognized' % datatype)
+            raise HeaderDataError('datatype dtype "%s" not recognized' % datatype)
```
matthew-brett: Why these error changes?

nibabel/tests/test_arrayproxy.py
```diff
@@ -50,11 +50,6 @@ def copy(self):
         return FunkyHeader(self.shape)
 
 
-class CArrayProxy(ArrayProxy):
```
matthew-brett: Did you mean to move this to the BV module?

nibabel/tests/test_image_api.py
```diff
@@ -175,12 +189,21 @@ def validate_dtype(self, imaker, params):
...
-        img.set_data_dtype(np.float32)  # assumed supported for all formats
-        assert_equal(img.get_data_dtype().type, np.float32)
+        new_dtype = np.float32
+        # some Image types accept only a few datatypes and shapes
+        # so we check and force a type change to a compatible dtype
+        try:
+            supported_dtypes = supported_np_types(img.header_class())
+            if new_dtype not in supported_dtypes:
+                new_dtype = supported_dtypes.pop()
+        except:
```
matthew-brett: Can you specify the expected errors here?  Bare `except` is a bit frightening...

matthew-brett (commented):
So sorry for long delay - was teaching this term.

Nearly there now.

Have you got any other BV files lying around we can add as more extensive tests, using http://nipy.org/nibabel/devel/add_test_data.html#adding-as-a-submodule-to-nibabel-data ?

matthew-brett:
@thomastweets - will you have any time to respond to my comments here (final stretch)?  Or would you prefer I merged this as-is, and then submitted some more PRs with edits?

thomastweets:
@matthew-brett I am very sorry for the late response: I am quite busy with my new job, however, I will try to make some time for your comments next weekend. Looking forward to having this completed :)

matthew-brett:
@thomastweets - sorry to press you - will have any time to work on this?   It's so close ...

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,957 @@
...
+                 check=True,
+                 offset=None):
+        """Initialize header from binary data block.
+
+        Parameters
+        ----------
+        binaryblock : {None, string} optional
+            binary block to set into header.  By default, None, in
+            which case we insert the default empty header block
+        endianness : {None, '<','>', other endian code} string, optional
```
thomastweets: I exchanged the binaryblock part of the docstring with a hdr_dict description. This is needed for setting parameters in the init call and also allows for easy duplication of BV files. Or maybe I do not get your intention here... :)

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,957 @@
...
+    def get_zooms(self):
+        shape = self.get_data_shape()
+        return tuple(float(self._hdr_dict['resolution'])
+                     for d in shape[0:3])
+
+    def set_zooms(self, zooms):
+        if type(zooms) == int:
+            self._hdr_dict['resolution'] = zooms
+        else:
+            if any([zooms[i] != zooms[i + 1] for i in range(len(zooms) - 1)]):
```
thomastweets: In file formats with 4 dimensions (implemented are VTC and VMP) the fourth dimension is handled quite specifically. In VTCs the time dimensions is expressed in the TR header field. In VMPs the fourth dimensions is the number of submaps. In MSK files on the other hand, only three dimensions can exist. I did not come up with a smart way of generalising this.

nibabel/brainvoyager/bv.py
thomastweets: Actually, yes. I only thought about edge cases when the user would like to actively change the data_offset. However, in most cases it would make sense to calculate it on-the-fly (like here) to capture any last minute header_dict changes that influence the offset. Should I take out functionality? I am not sure whether this violates any API dependencies though...

nibabel/freesurfer/mghformat.py
thomastweets: With the added supported_np_types check in spatialimages the missing HeaderDataError produced errors in the test_image_api and test_image_load_save tests. HeaderDataError seems to be the "correct" error to raise here.

nibabel/tests/test_arrayproxy.py
thomastweets: (I think based on your suggestion) I included the CArrayProxy class to arrayproxy.py because I need it for BV file formats. As this class was defined in test_arrayproxy.py as well I rather imported it and deleted the duplicate. But maybe I did not understand the logic here (or your question :)).

nibabel/brainvoyager/bv.py
thomastweets: Hmm! As I do not inherit from WrapStruct anymore (we started off there in before the whole hdr_dict_proto idea) I cannot think of an easy way of 'cross-importing' this single method. Can you?

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,957 @@
...
+            obj = header.copy()
+            if check:
+                obj.check_fix()
+            return obj
+        # not own type, make fresh header instance
+        obj = klass(check=check)
+        if header is None:
+            return obj
+        try:  # check if there is a specific conversion routine
+            mapping = header.as_bv_map()
```
thomastweets: That sounds good to me! As I have not worked with mixins yet: could you elaborate a bit on your suggestion? I do not fully understand your code example.

nibabel/brainvoyager/tests/test_bv.py
thomastweets: Good catch: it does not. However, as I do not have more test files right now and no access to a BrainVoyager license at the moment, I need to search for other means to solve this. I will think about it.

thomastweets:
Sorry for taking so long!
I implemented most of your comments. For some of them I would need additional feedback from you (see the comments). Also, I do not have access to a BrainVoyager license at the moment but I will try to get my hands on one soon. Therefore, creating more test files is tricky (that is, manually converting existing open datasets into BrainVoyager files as I not sure whether I can share datasets of mine here). I will look into this as soon as possible but my time is a bit restricted these days.

nibotmi:
:umbrella: The latest upstream changes (presumably #249) made this pull request unmergeable. Please resolve the merge conflicts.

nibabel/brainvoyager/bv.py
matthew-brett: Sure.   Each header defines its preferred map generation method, with default `None`.  So:


Header.map_method_name == None
AnalyzeHeader.map_method_name = 'as_analyze_map'
BvFileHeader.map_method_name = 'as_bv_map'.

Then the `Header` class has the following method (moved from the AnalyzeHeader class):

```
    @classmethod
    def from_header(klass, header=None, check=True):
        ''' Class method to create header from another header

        Parameters
        ----------
        header : ``Header`` instance or mapping
           a header of this class, or another class of header for
           conversion to this type
        check : {True, False}
           whether to check header for integrity

        Returns
        -------
        hdr : header instance
           fresh header instance of our own class
        '''
        # own type, return copy
        if type(header) == klass:
            obj = header.copy()
            if check:
                obj.check_fix()
            return obj
        # not own type, make fresh header instance
        obj = klass(check=check)
        if header is None:
            return obj
        # Check for method that returns a mapping this header understands
        map_method = (None if self.map_method_name is None
                                  else getattr(self.header, self.map_method_name, None))
        if map_method is not None:
            # header is convertible from a field mapping
            mapping = map_method()
            for key in mapping:
                try:
                    obj[key] = mapping[key]
                except (ValueError, KeyError):
                    # the presence of the mapping certifies the fields as being
                    # of the same meaning as for this class, so we can
                    # safely discard fields with names not known to this header
                    # type on the basis they are from the wrong dialect
                    pass
            # set any fields etc that are specific to this format (overriden by
            # sub-classes)
            obj._clean_after_mapping()
        # Fallback basic conversion always done.
        # More specific warning for unsupported datatypes
        orig_code = header.get_data_dtype()
        try:
            obj.set_data_dtype(orig_code)
        except HeaderDataError:
            raise HeaderDataError('Input header %s has datatype %s but '
                                  'output header %s does not support it'
                                  % (header.__class__,
                                     header.get_value_label('datatype'),
                                     klass))
        obj.set_data_dtype(header.get_data_dtype())
        obj.set_data_shape(header.get_data_shape())
        obj.set_zooms(header.get_zooms())
        if check:
            obj.check_fix()
        return obj
```

This the variant where the method goes in the Header class, but it could also be a mixin class, just containing this method, where AnalyzeHeader and BvFileHeader both inherit from both of Header and HeaderMapMixin, as in:

```
class BvFileHeader(Header, HeaderMapMixin):
```

nibabel/brainvoyager/bv.py
```diff
@@ -0,0 +1,979 @@
...
+        x = hdr['x_end']
+        y = hdr['y_end']
+        z = hdr['z_end']
+
+        # compare with possible framing cubes
+        for fc in [256, 384, 512, 768, 1024]:
+            if any([d > fc for d in (x, y, z)]):
+                continue
+            else:
+                return fc, fc, fc
```
matthew-brett: What happens if some d > 1024 ?

nibabel/brainvoyager/bv.py
matthew-brett: Your call.  You can make `set_data_offset` raise an error if you like (parrec.py does this) - or you can let people do a `set` if they want, hoping they understand the risk.

Do people tend to store stuff in BV files after the header and before the data?  That's not uncommon in Analyze, but very rare for Niftis for example.  If it's rare, maybe raising an error is better.

By the way, you're using this in the header constructor, so I guess the `offset` in the header constructor is getting thrown away, in fact.  Maybe you can remove it from the header parameter list?

nibabel/brainvoyager/bv.py
matthew-brett: Well - could use a Mixin I guess - as above.  Not sure whether it's worth it.

nibabel/freesurfer/mghformat.py
matthew-brett: OK - thanks.

matthew-brett (commented):
Just a few more comments.

effigies:
@thomastweets  Sorry to see this one fall by the wayside. Is there any interest in trying to pick it back up for the 2.5 (July) or 3.0 (October) releases? If you need help resolving conflicts, I'm happy to give it a go.

effigies:
Closing given existence of https://github.com/ofgulban/bvbabel. @ofgulban do we want to look into upstreaming your work here?

ofgulban:
@effigies , sure (sorry for the late reply). I do not have time to invest in this currently but maybe at some point in the future. I do like bvbabel to exist as a standalone package without extra overhead of maintaining similar functionality in nibabel, but I am not against if anyone else wants to pick this up.

PR #231, opened 2014-03-21
WIP: Add SpatialMeta class

moloney:
First step in integrating dcmstack functionality into nibabel. Currently this is just the  class that will store the meta data associated with an image. I am working on adding the code for extracting meta data from DICOM datasets and filtering it for PHI. Ultimately each of the DicomWrapper classes will have a "get_spatial_meta" method that will generate an appropriate SpatialMeta object.

Compared to dcmstack, this approach allows for an arbitrary number of image dimensions and a more comprehensive set of classifications. Also the meta data can now be nested into sub dictionaries. So for example we can have a "DICOM" dictionary with the (nearly) raw DICOM meta data and a "NiPy" dictionary with the most commonly used meta data extracted and normalized across vendors.

The purpose of all this is that we can 1) preserve all of the meta data from the source DICOM files even after converting to a different format and 2) we can split/merge any spatial image(s) that have associated spatial meta data and the result will preserve the appropriate set of meta data.

The API for accessing the meta data through an image still needs to be decided, but using an imaginary API the end result would provide functionality like:

    > img1 = nb.load('myimg1.nii.gz')
    > img1.spatial_meta.get_values_and_class('EchoTime')
    (40, 'varies_over()')
    > img2 = = nb.load('myimg2.nii.gz')
    > img1.spatial_meta.get_values_and_class('EchoTime')
    (80, 'varies_over()')
    > merged_img = nb.funcs.concat_images([img1, img2])
    > merged_img.get_values_and_class('EchoTime')
    ([40, 80], 'varies_over(3)')
    > merged_img.spatial_meta.get_value('EchoTime', (0, 0, 0, 0))
    40
    > merged_img.spatial_meta.get_value('EchoTime', (0, 0, 0, 1))
    80

### DIFF
--- a/nibabel/spatialmeta.py
+++ b/nibabel/spatialmeta.py
@@ -1,0 +1,670 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""
+SpatialMeta class for tracking potentially varying meta data associated
+with a spatial image. Can be merged or split along any dimension while
+maintaining the correct set of meta data.
+"""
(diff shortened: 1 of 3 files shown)


bcipolli:
@moloney any progress on this? Is this something you wanted to follow up on?

moloney:
This PR is really out of date from the approach I want to take, closing it.

PR #241, opened 2014-05-29
Track changes made to the header (NOT YET FOR MERGE)

yarikoptic:
as promised decided to give a try but as you can see -- not completed (yet).

Notes:
- `modified` in constructor of the header is probably not needed and logic to assign/reset those should go into corresponding (constructor) methods
- if to proceed this way, then all constructors for headers (and might also be for images) need to "finish" resetting the `modified` set in the header to indicate that initialization has finished
- not sure now if it is the right way to go ;)  you might cherry pick 453796e though ;-)


yarikoptic:
never finished, no point to keep it alive

PR #243, opened 2014-07-19
Streamlines API

MarcCote:
This is an attempt at providing a general API to manage different streamlines file format. This is related to the [BIAP5](https://github.com/nipy/nibabel/wiki/BIAP5).

Right now, only the TRK format is supported but I can easily add support for both TCK and VTK. One difference you might found with `nibabel.trackvis` is that it loads the streamlines as they were saved. There is no `points_space` argument but instead some functions should be used to change the space where streamlines live: `change_space(streamlines, ref=affine/nifti)`, `trackvis2mrtrix(streamlines, ref=affine/nifti)`, etc.

I can't wait to know what you guys think about this. At some point, I really see the [TractConverter](https://github.com/MarcCote/tractconverter) being part of NiBabel.


Garyfallidis:
Knock! Knock!

MrBago:
On Mon, Sep 29, 2014 at 7:50 AM, Marc-Alexandre Côté <
notifications@github.com> wrote:

> I'm not convince yet that '.trk' should always be saved as "Trackvis"
> wants them. Maybe for 'trk' and 'tck' format it is clear they are
> associated to 'Trackvis' and 'MRtrix' but some softwares save their
> streamlines using the 'vtk' format (e.g. ExploreDTI) and it is not clear
> what space their streamlines live in.
>
I'm not sure what you mean here. If we save ".trk" file that Trackvis
cannot properly interpret, presumably anyone else reading it will also
misinterpret it. When there is well defined standard, we should comply with
that standard. If there is no standard we should ask ourselves why we're
supporting that given file type and go from there. In the latter case, I
believe the answer will almost always be "so that we can pass data between
dipy and other software packages". If that's the case, we need to write
files those other packages can read properly.

> That is why I preferred loading/saving streamlines as is and was planning
> on using name of the software to create transformation functions for
> streamlines: trackvis2mrtrix, exploredti2trackvis, etc. I agree that it
> will be more user friendly to do it automatically instead of having to
> load, transform then save the streamlines manually. For example, we could
> add different loading/saving functions: load_trackvis, load_mrtrix,
> save_trackvis, save_mrtrix, etc. But in any case, those functions will
> require passing some sort of reference image from which to get the
> affine/dimension/size from.
>
I have to disagree here. This type of thing becomes very hard to support
long term. We've seen this recently with mrtrix when they changed their SH
basis between mrtrix2 and mrtrix3. If there are N software packages you
could need up to N**2 conversion functions and every time they change their
minds about how they do things, you need to go back and deal with it. I
think a much better approach is to have one consistent internal
dipy/nibabel representation and convert in and out of that representation
on read/write.


Sorry it took me so long to get back to you.

MarcCote:
Sorry, I didn't realize you were waiting for my reply.

I agree with what you said. Using a consistent internal representation is the approach I used when implementing the [TractConverter](https://github.com/MarcCote/tractconverter). It required having to specify a reference point either as an affine matrix, a nifti image or the convention's name (e.g. LPS) so we can project the streamlines in a predetermined space. Some users found it a little bit confusing because depending on the format you don't have to specify a reference point (TRK vs TCK).

We should determine and use an internal representation. What is the most convenient representation to manipulate streamlines when used in Dipy? Voxel or mm, RAS or something else, the coordinate (0,0,0) represents the center of voxel or one of its corner.

MrBago:
In dipy we've been using an affine to represent the mapping between voxel
indices and streamline points. I like this because it gives us maximum
flexibility in terms of supporting different coordinate systems and such.
It also means that we can save and load streamlines "as-is" as long as we
can encode/decode the affine on read write. For formats that don't support
encoding the affine, I think we should map the points into some space that
the format does support when saving.

Bago

Jan-Schreiber:
This sounds similar to reading and writing csv files with different encodings, delimiters, etc. 
What about using a standard interface and defining dialects for different applications as well as other parameters like coordinate systems, comments, ascii/ binary, etc.? 
nib.load_streamlines()
nib.save_streamlines()

Similar to loading volume data one could define functions like get_affine(), get_vertex_values(), get_streamline_values(), etc. whatever the file formats support. 

I guess, when writing streamline files we can specify to be compatible with some common software packages and when reading files there might be guesswork. If the file format does not tell which coordinate system it uses one has to guess or follow a given dialect. And it would be great to have a standard way to transform coordinates after loading. 

Personally, for vtk and tck files I would prefer the world coordinate system (because this is what e.g. ParaView and mrview use) but internally we should get voxel coordinates for best interaction with volumes.

MarcCote:
Following @MrBago advices, I modified the streamlines API so `Streamlines` objects use one consistent internal representation and have an affine matrix to bring them back to the world space (RAS+).

Specifically, when loading from a file, streamlines will be projected in [voxel space](http://nipy.org/nibabel/coordinate_systems.html#voxel-coordinates-are-in-voxel-space). I agree with @Jan-Schreiber that interactions with volumes are simpler that way (which is probably the main reason someone would want to load streamlines using NiBabel).

It worth mentioning that in voxel space, the X coordinate of any streamlines point will lie between `[-0.5, width-0.5)` (and similarly for the Y and Z coordinates) since a coordinate with integer value, e.g. `(0,0,0)`, represents the centre of a voxel. This means that one will have to add `0.5` then cast to `int` every coordinates before interacting with volumes. If I'm not mistaken functions in dipy take already that in count though.

On another point, I'm starting to add support for other streamlines file format (e.g. TCK) and I was wondering what metadata should be kept when converting streamlines from one format to another? Right now, a `Streamlines` object has a `header` property (a Python dict) which is filled with available metadata. I tried to standardize keys used in `header` for common metadata (you can see the current list [here](https://github.com/nipy/nibabel/pull/243/files#diff-3088f83d47b55090ba45b250ca5639af)). I guess my question is simply should we transfer metadata that is specific to a format? This would allow to not loose information if we decide to convert a streamlines file back to its original format.

MrBago:
Just to let you guys know how things work in dipy, streamlines are allowed to be represented in an arbitrary coordinate system. Dipy functions that require streamlines to interact with volumes require the streamlines and the affine which maps the volume voxels to the the streamlines coordinate system.

For example, if the streamlines are in the (RAS+) coordinate system of a nifti file, than the affine of the nifti file should be used.

Dipy tracking classes also take a data volume and an affine. They return the streamlines in the affine coordinate system. IE a point in the center of voxel `[1, 2, 3]` would be at `dot(affine, [1, 2, 3, 1])`.

I think on reading returning the streamlines in voxel-coordinates is ok, but I think returning them in any other coordinate system would also be fine as long as the affine is also returned.

Does this PR require that streamlines to be saved

MarcCote:
Yes, I agree with you, returning streamlines in any coordinate system would be fine as long you provide the affine matrix (volume voxel coordinates -> streamlines coordinates). I just think having them already in voxel coordinates would be simpler for interacting with volume (no need to apply the inverse affine matrix). Right now, `nib.streamlines.load` has a reference space parameter (either a matrix or a nifti file), that use to project the streamlines back to voxel space if needed (depends on the file format).

Also, streamlines can be created from a list of points if desired (no need to be saved). For example, creating 10 random streamlines each having 20 points in 3D would look like this
```python
>>> s = Streamlines(points=np.random.rand(10, 20, 3))
>>> s.header[Field.VOXEL_TO_WORLD]
array([[ 1.,  0.,  0.,  0.],
          [ 0.,  1.,  0.,  0.],
          [ 0.,  0.,  1.,  0.],
          [ 0.,  0.,  0.,  1.]])

>>> len(s)
10
```

So Streamlines objects are not force to be in voxel space and contain an affine matrix in their header (default is identity matrix). However, `nib.streamlines.save` function assumes Streamlines object are in voxel space and has a parameter allowing to specify a reference space (either using a matrix or a nifti file).

Whenever you have time I would appreciate some feedback about the two main files `base_format.py` and `header.py`
https://github.com/nipy/nibabel/pull/243/files#diff-555096850fc1b3fb1dfe6aa5644e3809
https://github.com/nipy/nibabel/pull/243/files#diff-3088f83d47b55090ba45b250ca5639af

Do you think it is a good approach having a `Streamlines` class regrouping all information about a streamline (scalars, properties, affine matrix, etc.)?
Once we reach consensus, I think it will be easier to support other format.

MrBago:
Will take a closer look next week. It's a pretty hefty PR, 2000+ lines but
I'll do my best :)

Bago

MarcCote:
@MrBago do you know if TrackVis (and the diffusion toolkit) considers the coordinate (0,0,0) of a streamline's point to be in the middle of the voxel or the corner? I know that MRtrix uses the former convention.

MrBago:
Trackvis uses it as the corner, ie all negative points are outside the
image volume.

Bago

MarcCote:
@MrBago I refactored a lot of code and added a lot of comments. I'm sure it will be easier to read as it is now.

Now, a `Streamlines` object `s` contains the affine matrix (`s.header.to_world_space`) needed to send the streamlines back to the world space. The convention I follow is still to have the streamlines in voxel space when loaded from a file. Afterwards, if one projects the streamlines into another space, the affine matrix will be updated as long as it uses the method made for that e.g. `s2 = s.transform(affine, lazy=False)`. So, unless modified explicitly, the affine will contain the information to bring the streamlines back to world space.

MarcCote:
@matthew-brett how can I see what coverage I'm missing exactly (which lines). When I click on the details link for "coverage/coverall", it brings me to a page where I can that the file `...ython2.7/site-packages/nibabel/testing/__init__.py` has decreased in coverage. However, when I click on it, it says source not available!

Eric89GXL:
@matthew-brett I struggled with that problem for a while trying symlinks and other solutions, the only solution seemed to be running from the package source directory. So for mne-python we run all tests except one Travis run from the source directory, and that last test we don't do a coverage report, but it checks to make sure everything still works as an installed package.

@MarcCote you should be able to do something like `nosetests --with-coverage` from the `nibabel` root directory, and it will print out a coverage report. You can also look into other options like `--cover-html` (I think?) for a more nicely formatted version.

nibabel/streamlines/trk.py
```diff
@@ -0,0 +1,520 @@
...
+                pts, scalars = read_pts_and_scalars(nb_pts)
+                properties = read_properties()
+
+                # TRK's streamlines are in 'voxelmm' space, we send them to voxel space.
+                pts = pts / self.header[Field.VOXEL_SIZES]
+                # TrackVis considers coordinate (0,0,0) to be the corner of the
+                # voxel whereas streamlines returned assume (0,0,0) to be the
+                # center of the voxel. Thus, streamlines are shifted of half
+                #a voxel.
+                pts -= np.array(self.header[Field.VOXEL_SIZES])/2.
```
MrBago: Is this right? after the points have been divided by `VOXEL_SIZE` don't you just want to do `-.5`?

MrBago:
@MarcCote I really like the changes you've made. Reading them I had an idea. How about having two versions of the `points` attribute on the streamline object, one that returns the points in voxel coordinates and one that returns them in real world coordinates.

Here is my thinking. Lets say that I read a streamline object from disk and want the real world version of the streamlines, currently I would do something like this `new_sl = sl.to_world_coordinates()` which returns a new streamline object. The new streamline object has an affine of `eye(4)`. Both sets of streamlines reference the same "real world coordinate system" but that's a pretty fuzzy concept to begin with.

Instead we can think of every streamline as having a "reference object" (you already do this in your code). The points of the streamline can be expressed using the voxel or "real world" coordinate system of the reference, but this is just a different way to express the points, not a new set of streamlines. Internally the streamlines can store the points in voxel coordinates (or something else since it's just an implementation detail in this framework).

I think transformations on the streamlines should be thought of as a change in reference and return a new streamline object with a new reference. But I think implementing this kind of transformations can be done in a separate PR. I think the reference object should have at least two things: 1) an affine and 2) a shape.

What do you think?

nibabel/streamlines/trk.py
MarcCote: Yep, you are definitively right. Clearly, it is missing unit tests!

I'll go with the `-.5` fix then.

MrBago: For IO stuff I like to include a write-read test to at least make sure that
I can at least recreate in python what I started with :).

MarcCote: Yes. I have that test but with an identity as the affine, so `np.array(self.header[Field.VOXEL_SIZES])/2 == 0.5`.

jchoude:
@MrBago I think that's a good idea. We could see the points in both spaces (voxel and world) as 2 "views" of the same streamline. I like that.

MarcCote:
@MrBago I really like that too. So, a streamline object would have a *property* named `points` and another named `points_in_world_space` (or another name, you have any suggestion?). The latter returning a generator or a list (memory usage will almost double)?

Another option would be having a *function* like `to_world_coordinates(as_generator=True)` that returns list (or generator) of points instead of `Streamlines` object.

I think the second approach is better since it allows the user to decide whether or not it is worth keeping the transformed points in memory.

@Eric89GXL thanks for the tips, I'll try that.

matthew-brett:
Guys - where are we on this one?   Bago - would you be interested in taking charge of this one, and merging when you think it's ready?

MarcCote:
I've been improving and refactoring the API a lot and I'm afraid this PR is becoming too big for a thorough  review. I propose to split it in two: 1) `Tractogram` 2) `TractogramFile`.

*Tractogram*
This part has the code to handle streamlines in memory using the Python class `Tractogram`. This class provides a mechanism to keep data along the points of every streamline. It can also keep data along each streamline. Internally, it uses what I call a `CompactList` to store efficiently the points of the streamlines. This was needed as @Garyfallidis and I realized that a list of numpy arrays has to much overhead in term of memory when handling >1M streamlines (loading a tractogram of 2.5Gb could take up to 5Gb in RAM!).

*TractogramFile*
This part has the code to save and load `Tractogram` on disk. Right now, as a proof of concept, only the TRK file format is supported.

nibabel/streamlines/trk.py
```diff
@@ -0,0 +1,739 @@
...
+
+            if cpt < trk_reader.header[Field.NB_PROPERTIES_PER_STREAMLINE]:
+                data_per_streamline_slice['properties'] = slice(cpt, trk_reader.header[Field.NB_PROPERTIES_PER_STREAMLINE])
+
+
+        if lazy_load:
+            def _read():
+                for pts, scals, props in trk_reader:
+                    data_for_points = {k: scals[:, v] for k, v in data_per_point_slice.items()}
+                    data_for_streamline = {k: props[v] for k, v in data_per_streamline_slice.items()}
```
effigies: For Python 2.6 compatibility:

    data_for_points = dict((k, scals[:, v]) for k, v in data_per_point_slice.items())
    data_for_streamline = dict((k, props[v]) for k, v in data_per_streamline_slice.items())

Garyfallidis:
@MarcCote this is an excellent work. But I think it is not a good idea to push this completely new design in this older PR because now it has changed some much and most of the previous discussions are not relevant any more. That makes reviewing this PR hard. 

I think to help the reviewers it will be better if you make a new PR (and close this one) and write in the description which are the main new features of this PR and the specific design decisions with a short explanation. For example, now we go to RAS+ by default, why?. Also we introduce the concept of a Tractogram which is a higher level of streamlines where you can save more info on the streamlines but we don't save header information there. This goes to the TractogramFile etc. But why we do such a thing? This will help a lot. Stay laconic but give the correct info please. 

This PR is absolutely crucial for the work we do in dMRI and it will be used by many people. It also solves an important memory/speed/management issue in Dipy. So, please @jchoude, @mdesco, @gabknight, @MrBago, @Jan-Schreiber,  and @arokem try this PR and give some nice reviews asap so we can finally have a stable way of reading and saving streamline-related files. We really need to get better at this ... 

Also, about splitting in two files maybe it's not a good idea because it will make it difficult to try this with real data. The PR is indeed quite large but I don't see an easy way of splitting it and trying it with real data. 

Finally, @matthew-brett we worked together with @MarcCote to make this PR and we tried to make the API look as close as what has been done for the Nifti files so that the API looks homogeneous. This PR should be merged before the 0.11 Dipy release because many new PRs depend on this new API which as I wrote above simplifies many current issues both of file formats for tractography and internal Dipy design issues. So, I would hope that it can be prioritized and be merged relatively soon. Of course with some quality reviews first. But as you can see by yourself @MarcCote has already done some excellent work here. The day were loading/saving streamlines is as easy as loading/saving nifti files is closer than ever :) :+1:

nibabel/testing/__init__.py
```diff
@@ -57,6 +59,21 @@ def assert_allclose_safely(a, b, match_nans=True):
...
 
 
+def isiterable(iterable):
+    try:
+        for _ in iterable:
+            pass
+    except:
+        return False
+
+    return True
```
effigies: This will consume generators:

    >>> def f():
    ...    for i in range(5):
    ...        yield i
    ...
    >>> x = f()
    >>> isiterable(x)
    True
    >>> list(x)
    []

What about:

    def isiterable(val):
        try:
            iter(val)
        except TypeError:
            return False
        return True

MarcCote: You are right. However, `isinstance(val, collections.Iterable)` simply checks if the object has an `__iter__` method instead of testing that you can actually iterate through it. The same goes for testing that calling `iter` is not failling.

I will agree that `isiterable` might not be the right name for this function but I lack a better one right now (if you have suggestion let me know). I use this `isiterable` function to loop through all elements of a given object, catching exception if any (could happen for my objects).

effigies: I see. I didn't look deeply into how it was being used. Maybe something like `check_iteration`?

nibabel/streamlines/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+                return format
+
+        except IOError:
+            pass
+
+    import os
+    from ..externals.six import string_types
+    if isinstance(fileobj, string_types):
+        _, ext = os.path.splitext(fileobj)
+        return FORMATS.get(ext, None)
```
martcous: Here, it won't work if the extension is in capital letters (e.g. TRACKS.TRK). Is that intended?

nibabel/streamlines/compact_list.py
```diff
@@ -0,0 +1,213 @@
...
+        elif isinstance(idx, np.ndarray) and idx.dtype == np.bool:
+            clist = CompactList()
+            clist._data = self._data
+            clist._offsets = [self._offsets[i]
+                              for i, take_it in enumerate(idx) if take_it]
+            clist._lengths = [self._lengths[i]
+                              for i, take_it in enumerate(idx) if take_it]
+            return clist
+
+        raise TypeError("Index must be a int or a slice! Not " + str(type(idx)))
```
martcous: an int* Also, general comment but you're both using "type(x) is mytype" and "isinstance(x, mytype)". And the way you use lists would work for any iterable I think, so you could make it more general which would work for tuples and so on.

nibabel/streamlines/__init__.py
MarcCote: You are right, I'll fix that. Thx.

nibabel/streamlines/compact_list.py
MarcCote: Thanks for pointing out the typo. 

Since I was trying to make the `CompactList` behaves like a `np.ndarray`, using `tuple` for indexing has a different behavior. Tuples should have the same size as the number of dimensions of the array. For example,
```python
A = np.arange(10)
A[(1, 2, 3)]  # Equivalent to A[1, 2, 3]
```
```bash
IndexError: too many indices
```
Another difference I observed is `list` of `bool` vs. `np.ndarray` of `bool` behave differently too. The former converts the boolean to integer first (so we got a list of 0's and 1's) while the latter behave like a filter (i.e. True to take the item at index `i`, False to skip it).

That said I don't mind making it more general and changing the behavior of the `tuple` (or any iterable for that matter **except** `ndarray` of `bool`) so they are behaving like `list`.

martcous: Ah, that makes sense. It's probably fine like that, then.

MarcCote: But you are right for the mix of `isinstance` and `type`. I'll change the code to use `isinstance` where I can since it is more flexible (supports subclasses too).

nibabel/testing/__init__.py
MarcCote: I like it. I replaced `isiterable` by `check_iteration`.

samuelstjean:
Ah yes, playing with different streamline format is indeed a horrible experience (and I just started playing with it), so prioritising this instead of playing with a ton of converters is a good idea.

MarcCote:
See the new PR #391

PR #251, opened 2014-08-21
NF: add version of Paul Ivanov's slice viewer

matthew-brett:
@ivanov - many thanks for this - it's very nice, and very useful.

The thing I really miss though, is having cross-hairs showing me the
corresponding current point on all three panels.

I think there's an example here:

http://marsbar.sourceforge.net/_images/roi_view.png

I experimented for a while trying to put the crosshairs on, but it was beyond
my blit fu.  Any pointers as to how to do that?  I mean, put up the
cross-hairs, different on every movement of the slices, but still be able to
blit for speed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ dist/
 .shelf
 .tox/
 .coverage
+cover/
 
 # Logs and databases #
 ######################
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,12 @@ before_install:
     - if [[ $DEPENDS == *h5py* ]]; then
       sudo apt-get install libhdf5-serial-dev;
       fi
-# command to install dependencies
+    # Create a (fake) display on Travis so that mpl tests work
+    # TODO: Could also use the `agg` backend instead
+    - if [[ $DEPENDS == "*matplotlib*" ]]; then
+      export DISPLAY=:99.0;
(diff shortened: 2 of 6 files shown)


Eric89GXL:
Should close this given `nilearn` has something similar, or you think it's still worth pursuing here?

http://nilearn.github.io/building_blocks/plotting.html

matthew-brett:
Sorry to be lazy - but at a first pass these all looked static - so you can't click around the image?

Eric89GXL:
It was dynamic last time I used it, clicking around the image should work as should linking two images together, etc. IIRC everything was working correctly when we last looked at it, we just shelved it due to the release and concerns that it should be part of the resampling PR, e.g.:

https://github.com/matthew-brett/nibabel/pull/5#issuecomment-61022377

matthew-brett:
Sorry - I mean the nilearn stuff is static - no?  If that is true, this PR still does something we need and that we don't have.

Eric89GXL:
That's true. Do you think it's worth trying to port our changes over to their code instead of partially duplicating effort? That might be a pain, though, I guess.

How's the resampling coming along? :)

matthew-brett:
Hi Eric.  @choldgraf just caught me in the corridor asking about this one.

I would like to get this merged real-soon-now (TM).    I need this stuff too.  I think we can sort out the resampling axis stuff later.   Do you have some time to rebase this?  I've got quite a bit of time in the next few days to do final review.

Eric89GXL:
That was quick, nicely done @choldgraf :)

I should be able to look tomorrow.

Eric89GXL:
Closing for #404

PR #232, opened 2014-03-21
WIP: Dicom meta data anonymization and extraction

moloney:
Adds code for anonymizing and extracting DICOM meta data.

PR #290, opened 2014-12-04
WIP: Improve meta data extraction from DICOM files

moloney:
This is basically the meta data extraction, without anonymization, stuff from PR #232 with some improvements. Added tests for most of the functionality.

Adds the meta data extraction parts from dcmstack, with some improvements.  Most notably it can also parse the Siemens XProtocol format (kind of looks like XML but isn't).

I still need to fix text encoding and filtering of non-text byte data in the `extract` module. Text in a DICOM file can basically have any encoding and there is often no specification of what that encoding is.  

@matthew-brett  how do you feel about pulling in chardet (https://github.com/chardet/chardet) as a dependency? It is less static than OrderedDict, so I guess we don't want to pull it into externals.


matthew-brett:
Brendan - any news from your researches?

matthew-brett:
Any news here?  Anything I can do to help?

moloney:
Sorry about the lack of progress. 

I think the unicode handling is reasonable now.  I use pydicom to decode the standard DICOM elements. For elements with a VR of OB, OW, or UN we will use the `chardet` package if available to try to find the encoding. We fall back to just testing if the data is ASCII (skipping elements where this is not the case).

I integrated your PLY based parser for the xprotocol, with some minor changes made on my fork. I want to do some more work on this code to produce a more user friendly result. For example it would be nice to be able to do a lookup like `result['']['Protocol0']` instead of `result[0]['blocks'][0]['value'][2]['value']`.

I am also thinking it would be nice to do the meta data "extraction" in a lazy manner. So you only pay the cost of all the extra parsing if you actually need something from one of the private sub headers. To that end, I am thinking the extraction code should really be integrated with the `DicomWrapper` objects. For example you could access the CSA Image sub header directly through a `DicomWrapper` by doing `dw['CsaImage']`. What do you think about this approach?

matthew-brett:
The lazy loading sounds like a very good idea.

Did you have a chance to look at the failing tests?

Will you have time to get to cleaning up the output from the ply parser?   It seems like a good idea to get that in, in a good state.

nibabel/nicom/utils.py
```diff
@@ -13,11 +16,11 @@ def find_private_section(dcm_data, group_no, creator):
     element number of these tags give the start of matching information, in the
     higher tag numbers.
 
-    Parameters
-    ----------
+    Paramters
```
samuelstjean: You seem to have inadvertedly removed an e.

moloney:
I am confused why the imports for 'externals.ply' are broken on Travis-CI. It works on my machine in an isolated virtual environment.  I am just running "nosetests" in the root directory, should I be doing something else?

moloney:
I just commited my first attempt at putting the extraction logic into the dicom wrapper objects and doing lazy parsing. For example you can now access the "CSA Image" sub header by just doing `dw['csa_image']`.

I made some backwards incompatible changes. In particular the dicom wrapper objects all take a list of PrivateTranslator objects as their second argument to `__init__` now.  Previously, the siemens wrappers took the CSA Image header as the second argument. I don't think a lot of user code would have used that functionality rather than just relying on `wrapper_from_data` method to do the right thing. The siemens wrappers also no longer have a `csa_header` attribute that points to the CSA Image sub header, but we could add this back in temporarily with a deprecation warning.

Also, the wrappers now require a real DICOM data set rather than a dict. Again, I don't think that is a big deal for user code.  I still need to update the tests that are broken by this change, but wanted to get some feedback first.

Finally, the dicom wrapper objects need a method to return the full JSON meta data extension described here: https://github.com/nipy/nibabel/wiki/json-header-extension

I am going to start working on improving the xpparse output as described above while I wait for any feedback on these changes.

matthew-brett:
Hi Brendan.  Sorry to be so slow to get to this.  What do you think the best way forward here is?  I'm afraid I've left a lot of changes to back up.  Maybe schedule on online or real meeting sometime to work through this stuff?

moloney:
Hi Matthew.  We just hired someone to help me out and he is working on a branch for doing the stacking part and creating the meta data extension.  I hope to spend some time in the next week cleaning up this PR, at which point it would be good to discuss further.

matthew-brett:
Hi Brendan - that's good news.  I think I've not done a good job on helping you with this PR - what do you think we should do better from now?   I should obviously give more timely feedback, I will try hard to do that, please do remind me.

effigies:
Going to close this as stale, since it's 9 years old and needs to be rebased.

PR #296, opened 2015-02-12
ENH: Add Nifti1DicomExtension + test

kastman:
Hi all,

Here's a thin wrapper class to read DICOM binary header information as encoded with the [DICOM extended header code 2](http://nifti.nimh.nih.gov/nifti-1/documentation/faq#Q21), instead of just providing the ugly & useless byte string. The extension code already written was quite helpful; thanks! 

This depends on [pydicom](https://github.com/darcymason/pydicom), but should fall back to a standard `Nifti1Extension` in the case of an `ImportError`. This is my first commit to nipy, so I'm not sure how you feel about provisional dependencies like this, but I think delegating the work to pydicom is the right approach. 

Few things still to be worked out before you merge:
* What do I have to do for CI to get this to either load or not load pydicom during testing?
* This is hard-coded for little endian encodings right now - does anyone know a heuristic? Maybe @darcymason ?
* This is written inside the Nifti1 format file, since the Nifti2 format seems to just delegate to Nifti1. Is that the right spot for it?
* I'm not a great test-writer - could I improve on what's there?
* I added `import struct` in the test to create DICOM byte-strings, but it looked out of place. Is there a better spot / method?
* Is there a requirement that Nifti extensions should be divisible by 16 bytes? I saw that in another extension test, but am not aware of the requirement and couldn't find it in the docs. If so, I'll have to add some byte-padding code.

Hope this is helpful; let me know if there are any concerns about it! Cheers,
Erik


matthew-brett:
Thanks a lot for doing this.

Absolutely fine for the pydicom optional dependency - we already have that.  Have a look at `./nibabel/nicom/tests/test_utils.py` for an example of how to deal with the optional dependency.

Did you see: http://nipy.org/nibabel/devel/add_test_data.html for adding test data?

For the tests - the trick is to go through each function / method and ask yourself 'how would I know if this method was doing the wrong thing'.  It is a rather tiring process, but it also proves very helpful in learning, at least in my experience.

kastman:
I did see the page for adding test data, but I wasn't sure if it was appropriate to add another sample image (even a small one) for a test like this. Thanks for the tip on test_utils; that will work. Looks like I'm also failing on py3; I'll fix that as well.

matthew-brett:
For a test image - if it is smaller than 50 compressed - sure - go ahead and add it to the main repo.  Otherwise, a submodule would better.    No problem for a very small submodule with only a few images.

kastman:
Ironically, my actual use-case is a PET volume that's 1.6GB, so I'll definitely be making a new image for the test. See what I can do!

matthew-brett:
Thanks for keeping up the work on this.

Are there any good docs on what the DICOM extension format should contain.  I see this : http://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/extension.html - which only says:

    2 = DICOM format (i.e., attribute tags and values)

Do you know of anyone else writing these extensions?

nibabel/nifti1.py
```diff
@@ -380,6 +380,113 @@ def write_to(self, fileobj, byteswap):
...
+            is_implicit_VR = True
+            is_little_endian = True
+        elif transfer_syntax == dicom.UID.ExplicitVRLittleEndian:
+            is_implicit_VR = False
+            is_little_endian = True
+        elif transfer_syntax == dicom.UID.ExplicitVRBigEndian:
+            is_implicit_VR = False
+            is_little_endian = False
+        elif transfer_syntax == dicom.UID.DeflatedExplicitVRLittleEndian:
+            zipped = fileobj.read()
```
matthew-brett: Have you got pyflakes or similar running in your editor?  Pyflakes tells me that `fileobj` and `zlib` are not defined - is this piece of code tested?

kastman:
I've only ever needed this once, for [PMOD](http://www.pmod.com), a closed-source tool to model PET time activity curves. They write dicom tags to indicate the start time and duration of PET frames in a 4d nifti; the closet I could find for documentation is a listing of [what they support](http://www.pmod.com/files/download/v33/doc/pbas/4854.htm) for nifti. I only noticed they were using the DICOM extension when I saw they were able to read the timing from PMOD-created files and started looking around.

The tags are written with an explicit VR syntax without any metadata. An example header looks like this:

    Nifti1Extension('dicom', 
      (0054, 1001) Units                                     CS: 'Bq/ml'
      (0055, 0010) Private Creator                     LO: 'PMOD_1'
      (0055, 1001) [Frame Start Times Vector]  FD: [0.0, 30.0, 60.0, ..., 13720.0, 14320.0]
      (0055, 1004) [Frame Durations (ms) Vector] FD: [30.0, 30.0, 30.0, ...,600.0, 600.0]')

Everyone else I know that wants DICOM info in NiFtI files uses @moloney 's [dcmstack](https://github.com/moloney/dcmstack) xml encoding; and I see there's already effort for that w/ nibabel (e.g. #232, #290). However, I figured that since I was implementing this, I might as well put in a way of storing full DICOM datasets in case someone wanted to use it.

matthew-brett:
Well - I guess we can define the standard format.  So would that be the explicit VR syntax?  Maybe with little-endian byte order?  Do the PMOD files always have little-endian order?  And then try our best to read DICOM extensions that are written with implicit VR and big-endian.

nibabel/nifti1.py
```diff
@@ -380,6 +380,113 @@ def write_to(self, fileobj, byteswap):
         # next 16 byte border
         fileobj.write(b'\x00' * (extstart + rawsize - fileobj.tell()))
```
matthew-brett: PEP8 - two lines between classes, two lines between functions one line between methods.

nibabel/nifti1.py
kastman: You're right - I took this directly from pydicom, although it wasn't abstracted in a method I could call from nibabel. The only transfer syntax tested here is ImplicitVRLittleEndian, not the zipped or big endian ones. I'll add tests and fix this.

I haven't been using Pyflakes, but just grabbed a bundle to enable it. Thanks for the suggestion!

matthew-brett: Pyflakes is huge - it really helps picking up this kind of thing.

kastman:
That sounds like a plan; little-endian seems to be preferred from everything I've seen (I'm not even sure how to guess endianness without some magic or prior, which meta-less tags don't have). And I think that being explicit about the VR is preferable to implicit - the intended use-case is archival, not transmission over the network, so clarity has the premium over compression or compactness.

I'm not sure about the endianness of PMOD files - everything I've seen has been little, but there's no doc or guarantee. I could write to the developers if you think it's worth it?

Should we toss reading full metadata dicom datasets until someone wants to use them? Seems like we should honor it if it's present...

matthew-brett:
Thinking more - I wonder if we should default to writing as the byte order of the header (`header.endianness`).   Of course this will almost invariably by little-endian these days but still.

I guess we could write explicit VR, header endianness always, and read any endianness, with or without full metadata.  I'm neutral about allowing full metadata if you have already implemented it, just because we haven't got an example of anyone using it, and it's more code maintenance - the YAGNI principle.

kastman:
There's two parts to this, the way that the header is actually written and what the header says. The dicom standard says the header is always written ExplicitVR-LittleEndian, but it can list a different method (TransferSyntax) to use for the rest of the file. However, in the case where there is no header (e.g. naked tags like PMOD) there's no TransferSyntax and no way to know the correct encoding; we can only guess it.

How about this? If there is metadata and a TransferSyntax attribute is present, write using whatever the transfer syntax says. If no TransferSyntax, always use little endian. If the dataset was inferred to be Implicit, then write with ImplicitVR, otherwise write with ExplicitVR. That way you're returning as close to what you got as possible.

I agree with you concerning YAGNI, but if I *were* going to use it, for archiving tags, I would use the whole dataset with metadata, and I would expect the writing to follow the Syntax - not doing so would definitely be a surprise. Plus, I've created example dicoms (~360Bytes) for testing different syntaxes directly from Nibabel's tests, so that reduces maintenance cost a little.

matthew-brett:
Sorry - when I was talking about the 'header' I meant the nifti header.

Do we care about what the input transfer syntax is?    I mean, endian, or implicit VR?   If we specified nifti-header-endian and explicit VR was the standard, could we persuade people that was the right way to save this stuff out?

kastman:
Ahh, I see what you mean. There are two uses here:

1. Using the nifti header as a way to store a handful of interesting tags.
2. Using the nifti header as an archive of the dicom header.

I could see the benefit of both.  The first seems to be preferred by PMOD and would certainly simplify things. I'm not sure how much benefit there would be in the second, especially given a) privacy concerns and b) dcmstack? Is there any good argument for it, besides thouroughness?

matthew-brett:
So far, all we have for the standard thing to store is:

* `2 = DICOM format (i.e., attribute tags and values)` from the standard;
* PMOD storing tags without the file metadata (is that right?)

So, adding those two things up, I would say that your option 1. is the one to go for, and we should not support (saving as) 2. unless someone pitches in with a good case.  What do you think?

kastman:
Ok, I'm getting on board with just writing the standard. 

Last question, should we remove the implicit_VR guess heuristic and *enforce* that any dicom tags in the nifti header are explicit? I feel like that will ensure people use an explicit encoding, but there are conceivably other programs outside of nipy using ecode 2 that might write with an implicit VR?

matthew-brett:
If I understand right - you are asking whether we should be able to _read_ implicit / other endian / with-metadata DICOM inputs, while always doing a _write_ with explicit VR, nifti-header-endian ?

kastman:
Yep - that's it exactly.

matthew-brett:
Yes - that seems reasonable to me - because the standard is so vague it's perfectly possible that someone thought it allowed those options...

matthew-brett:
What is the right thing to do with implicit encoding and private tags?  I guess pydicom will be able to work out VRs for most of them?

kastman:
Yeah, I’m not sure how that works but I’d defer to whatever pydicom does and not worry about it here.

kastman:
Last question - since you want to use the endianness of the header to write the DICOM extension, what's the best way for the extension to get its header parent? Doesn't look like it's referenced or passed in anywhere.

After that I think the PR should be good to pull.

matthew-brett:
Good question.  It seems like the extension should know about its parent.   You have looked at the code most recently - do you have an idea for a good solution?

matthew-brett:
Or maybe some extensions should have endianness...

kastman:
I'm leaning toward treating any BigEndian extensions as edge cases that could be handled manually with the extension's _raw_content attribute. Then the extension would always use little endian; I'm not sure why it would even be worth writing with the parent header's endianess.

Is that overly simplistic? Just trying to avoid YAGNI again.

matthew-brett:
To understand what you mean - you are suggesting raising an error if the user writes big-endian DICOM attributes into the nifti DICOM extension?   So we'd only read little-endian explicit VR format from the DICOM extension?  And we'd only write little-endian explicit VR?

I don't feel strongly either way.  I guess that is the DICOM default, so it might not be so surprising that the DICOM extension was of different endianness from the nifti header.

kastman:
The way it's set right now, the extension is always read to and written from as little-endian, which I think for almost all cases is totally fine. It's one more extra step to make the extension agree with the header (I'd pass the parent in to both the dicom extension and the regular nifti extension as an optional parent kwarg). I'm not sure if the agreement is worth even those few extra lines of code, though.

Thinking a little more though, we should probably allow the reading and writing of implicit as well as explicit tags. Some tags that are read as implicit *can't* be written as explicit, which is currently the case: if the dataset was read as implicit, it will write out as implicit as well, and I think that's ok. But that's different from the recommendation of always using only explicit/little. 

So, maybe we should write as whatever the header is, if we're not enforcing just one. I honestly think it's ok to merge right now, using little always but switching between implicit/explicit. But I want to be thorough and clean. If you think we should write with the header I'm happy to add that in quick; otherwise I think it's ready to merge. 

Sorry this is confusing, but I'm having a hard time figuring out what the "right" thing is. I just keep going back and forth, so I leave the call to you!

matthew-brett:
Just for my understanding, how does it happen that an implicit dataset can be read, but not written?   Don't we have to work out the datatype (VR) in order to read it?  Can we read an implicit dataset in big-endian and write in implicit little-endian?

kastman:
I'm not exactly sure, but there are a few union VRs - this or that. They get read in the same way but still have distinct meanings in the Dicom, so that you can use the value without knowing which VR was actually used.  I don't have an example, but I know pydicom had refused to write some datasets as explicit when I read it as implicit. 

I'll have to check on implicit big endian, but that might not exist?

kastman:
Ready to merge? Or should I add writing the endianess from the header?

doc/source/dicom/dicom_niftiheader.rst
```diff
@@ -0,0 +1,62 @@
+.. _dicom-niftiheader:
+
+##############################
+DICOM Tags in the NIfTI Header
+##############################
+
+NIfTI images include an extended header (see the `NIfTI Extensions Standard`_) to store, amongst others, DICOM tags and attributes. When NiBabel loads a NIfTI file containing DICOM information (a NIfTI extension with ``ecode == 2``), it parses it and returns a pydicom dataset as the content of the NIfTI extension. This can be read and written to in order to facilitate communication with software that uses specific DICOM codes found in the NIfTI header.
```
matthew-brett: Would you mind line wrapping these paras?  Just to avoid annoying diffs later?

doc/source/dicom/dicom_niftiheader.rst
```diff
@@ -0,0 +1,62 @@
...
+
+For example, the commercial PMOD software stores the Frame Start and Duration times of images using the DICOM tags (0055, 1001) and (0055, 1004). Here's an example of an image created in PMOD with those stored times accessed through nibabel.
+
```
matthew-brett: Maybe mention that dicom headers read as generic if pydicom not installed?

matthew-brett:
Just to confirm - this currently reads anything (LE implicit / explicit, with / without meta file header)?  And writes what it reads?

nibabel/tests/test_nifti1.py
```diff
@@ -1070,6 +1081,57 @@ def test_nifti_extensions():
...
+    dcmext = Nifti1DicomExtension(2,dcmbytes_explicit)
+    assert_equal(dcmext.__class__, Nifti1DicomExtension)
+    assert_equal(dcmext._guess_implicit_VR(),False)
+    assert_equal(dcmext.get_code(),2)
+    assert_equal(dcmext.get_content().PatientID, 'NiPy')
+    assert_equal(len(dcmext.get_content().values()), 1)
+    assert_equal(dcmext._mangle(dcmext.get_content()),dcmbytes_explicit)
+    assert_equal(dcmext.get_sizeondisk() % 16, 0)
+    
+    # create a single dicom tag (Patient ID, [0010,0020]) with Implicit VR
```
matthew-brett: Testing big and little endian explicit VR?

nibabel/tests/test_nifti1.py
```diff
@@ -40,6 +41,16 @@
 header_file = os.path.join(data_path, 'nifti1.hdr')
 image_file = os.path.join(data_path, 'example4d.nii.gz')
 
+try:
```
matthew-brett: Could also use:

```
from nibabel.optpkg import optional_package
dicom, have_dicom, _ = optional_package('dicom')
```

kastman:
Almost: this currently reads LE, implicit or explicit, *without* metadata only, and writes whatever it read.

The capability for reading with metadata was in a previous commit, but I took it out because we decided it was YAGNI and there weren't any existing use cases.

There's no BE because it's tricky to figure out endianess without the metadata. We decided this was ok because no one is using this yet and we can call it a standard; the use case doesn't justify the complexity, and if anyone wants to get at the raw bytestring and encode it as a dicom Dataset manually they can always use of hdr._raw_content.

Should I put that explanation in the rst doc too?

matthew-brett:
Ok - that seems reasonable.

Sorry - I am sure I can work this out - but what happens if the user passes in a big-endian dataset?

kastman:
Yeah, it's not a good situation - and hard to rescue from.

For example, if the dataset is a single element for PatientID, either a LE or BE reading is "valid": if it's encoded as BE and read as BE it looks right but if it's encoded as BE and read as LE it doesn't fail:

* (00**10**, 00**20**: "Patient ID")
* (**10**00, **20**00: "")

That's still a valid dicom tag, albeit the incorrect one.

On other tags that are invalid, you do get invalid reads like KeyError: "Unknown DICOM tag (694e, 7950) - can't look up VR" - those we can rescue from at least, but I wonder if it's worth it? 

Even if we do read and write according to the nifti header's endianness that's no guarantee that the extension was coded correctly.

matthew-brett:
Yes, that's a bad situation.  It would be good to somehow make sure that we were getting it right for headers that we have written.  Any ideas?  I'm really sorry for flip-flopping, but one obvious one is to write the meta-data...

kastman:
Meta data is one option, using the endianess of the nifti header is another.  

I'm feeling like your initial intuition was right and we should use the nifti header. 

Of the two I think using the header is the easiest, but it requires slightly tweaking the extension api (to pass in the parent header to the extension). 

This doesn't mean well always be right, but at least well be right for things we write ourselves.

matthew-brett:
How about always prepending the transfer syntax UID tag element ? ((0002,0010)

matthew-brett:
I guess we'd have to do a special case to check for byte swapping this tag in particular - if the first tag is (0x2000, 0x1000) then ...

matthew-brett:
Any thoughts here?  I am fine with just following the header if you think that's the best.

kastman:
I think the cleanest, most transparent thing to do it just to follow the header, and be real clear in the docs how to access the original raw extension. I'll take a crack at it next chance I get.

matthew-brett:
@kastman - how are things going with this?  It would be good to get this one in...

kastman:
Thanks for the ping - I got pulled away but will try to get to it this week! Went down a rabbit hole looking into weak references and never finished.

matthew-brett:
Hi @kastman - sorry to pester you - did you have time to get to this one?

kastman:
Hey @matthew-brett  - thanks for the poke. I looked up recommendations on weakref, and I think we don't need it. So long as we're not writing our own __del__ methods python's garbage collection should be able to handle these circular references without adding weakref complexity. 

I'm putting this together right now and should have something tested and ready in a day or two. Thanks for your patience,

matthew-brett:
Hi Erik - any progress?   I've got some more free time over the next couple of weeks, so I promise to be faster to review.

matthew-brett:
Hi @kastman - I'm really keen to get this in - do you have a moment or two to finish this one off?

kastman:
Hey @matthew-brett - I do! 

Sorry for the long delay on this - I tried to overthink it and then was pulled away. Considering how long it's been, do I have to rebase this or do anything tricky before I add the last commits?

matthew-brett:
Considering how fiercely slow I've been to get back to you, it's the least I can do to do the rebases for you - so no - go ahead with the changes, I'll rebase and merge.

kastman:
@matthew-brett Thanks for the git-assist; I'm sure your git-fu is much better than mine, but don't worry, the delay's been entirely my fault.

@bcipolli I haven't been following the CIFTI PR, but I agree there's definitely merit to encapsulation. I didn't do it because I thought that might be needlessly over complicated if it were just for the dicom extension, but I could see how it might be helpful now. It looks like you're defining the extension directly in a [parser module](https://github.com/satra/nibabel/blob/enh/cifti2/nibabel/cifti2/parse_cifti2_fast.py#L29), right? I'd be happy adding a separate file / directory for nifti extensions - both DICOM and CIFTI extensions could live in the same place.  Do we think that's worth it now?

bcipolli:
@kastman Actually, the purpose is to localize NiftiExtensions with the functional code that's using them. Anybody can define a `NiftiExtension`, and so I think new extensions belong with the code related to the format that's using the extension.

So, the Cifti `NiftiExtension` lives in the `cifti` code, and I would guess that the `DICOM` `NiftiExtension` could live within the `nicom` directory (somewhere).

matthew-brett:
Erik - can I give any help here?

kastman:
Just to make sure I'm still thinking about it right (sorry for the very long time gap) - I'm going to use the header's endianeness as authoritative, right? So even if we created a little endian Dicom Dataset, if it's written as part of a Big endian nifti header it should be written using big endian byte order. And vice versa - if we created a big endian dicom dataset but wrote it as part of a little endian nifti header, we would write it as LE. That's still the plan, right?

kastman:
Take a look at this one, @matthew-brett , and see if makes sense. Sorry again for the delay!

Nifti1DicomExtension subclasses Nifti1Extension, but defines its own `__init__` and doesn't super() to Nifti1Extensions `__init__` in order to allow dicom datasets to be passed directly (see the `if __class__ == 'Dataset'` logic at nifti1.py#L397). Because of this, I didn't need to alter the header of Nfiti1Extension to accept a parent header. This is simpler and makes fewer changes, but I wonder if I should add it to Nifti1Extension in order to make things clearer in the future?

Also, it looks like the environment is failing. Do we need to rebase now for the CI to build?

matthew-brett:
I just did a merge of the master branch (makes it simpler for you to merge into your branch).

I also did a refactor we need to do anyway to clean up importing dicom / pydicom in various places.

I made a pull request into this branch (I hope): https://github.com/kastman/nibabel/pull/1

Please do check what I did, comments welcome.

nibabel/nifti1.py
```diff
@@ -381,13 +381,89 @@ def write_to(self, fileobj, byteswap):
...
+    byte string. If pydicom is not available, it silently falls back to the
+    standard NiftiExtension class.
+    """
+    def __init__(self, code, content, parent_hdr=None):
+        self._code = code
+        if parent_hdr:
+            self._is_little_endian = parent_hdr.endianness == '<'
+        else:
+            self._is_little_endian = True
+        if content.__class__ == Dataset:
```
matthew-brett: How about `if isinstance(content, Dataset):` ?

nibabel/nifti1.py
```diff
@@ -381,13 +381,89 @@ def write_to(self, fileobj, byteswap):
...
+            self._is_implicit_VR = False
+            self._raw_content = self._mangle(content)
+            self._content = content
+        elif len(content):  # Got a byte string - unmangle it
```
matthew-brett: What can `content` be apart from a byte-string and a Dataset?  Does it make sense to ignore values that don't fit - as in the `else` block below?

nibabel/nifti1.py
```diff
@@ -381,13 +381,89 @@ def write_to(self, fileobj, byteswap):
...
 
+class Nifti1DicomExtension(Nifti1Extension):
+    """NIfTI1 DICOM header extension.
+
+    This class is a thin wrapper around pydicom to read a binary DICOM
+    byte string. If pydicom is not available, it silently falls back to the
+    standard NiftiExtension class.
+    """
+    def __init__(self, code, content, parent_hdr=None):
+        self._code = code
```
matthew-brett: Docstring here with parameters definitions would be useful.

matthew-brett:
Converting everything to the same endianness of the header sounds good to me.

matthew-brett:
Also fine to override the `__init__` of Nifti1Extension, as long as you are doing the same stuff in the init, and I think you are.

nibabel/nifti1.py
matthew-brett: Also - how about checking `elif isinstance(content, bytes):` to be more explicit?

kastman: `content` should only be a bytestring or a Dataset - I should probably raise a helpful error if its neither of those two? Also, I see that you're still pushing small commits - should I wait on the PR?

matthew-brett: Yes, I think raising an error if it's neither of those - unless you want to allow None, to mean an empty dataset?

Yes, sorry, please wait for a bit - there were some little problems, will get back to you in 20 minutes or so.

kastman: No rush - I appreciate the review, and agree with everything so far. Lemme know when it's good to go,

matthew-brett: Ok - tests passing at last : https://travis-ci.org/matthew-brett/nibabel/builds/117048622 - good to merge.

kastman:
Not sure the docstring is super-helpful - take a look? Also, fixed the value import for pydicom <1.

nibabel/nifti1.py
```diff
@@ -379,13 +381,107 @@ def write_to(self, fileobj, byteswap):
...
+
+    This class is a thin wrapper around pydicom to read a binary DICOM
+    byte string. If pydicom is not available, it silently falls back to the
+    standard NiftiExtension class.
+    """
+    def __init__(self, code, content, parent_hdr=None):
+        """
+        Parameters
+        ----------
+        code : int|str
```
matthew-brett: Detail point, but I think these should be `int or str` : see https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#id5

nibabel/nifti1.py
```diff
@@ -379,13 +381,107 @@ def write_to(self, fileobj, byteswap):
...
+          Canonical extension code as defined in the NIfTI standard, given
+          either as integer or corresponding label
+          (see :data:`~nibabel.nifti1.extension_codes`)
+        content : bytes|pydicom Dataset|None
+          Extension content - either a bytestring as read from the NIfTI file
+          header or an existing pydicom Dataset. If a bystestring, the content
+          is converted into a Dataset on initialization. If None, a new empty
+          Dataset is created.
+        parent_hdr : :data:`~nibabel.nifti1.Nifti1Header`
```
matthew-brett: Is `:data:` correct here?  I was thinking it should be `:class:`.

`optional goes at end of line, as in `parent_hdr: data:`~nibabel.nifti1.Nifti1Header`, optional`.

nibabel/nifti1.py
```diff
@@ -379,13 +381,107 @@ def write_to(self, fileobj, byteswap):
...
+            self._is_implicit_VR = False
+            self._raw_content = self._mangle(content)
+            self._content = content
+        elif isinstance(content, bytes):  # Got a byte string - unmangle it
+            self._raw_content = content
+            self._is_implicit_VR = self._guess_implicit_VR()
+            ds = self._unmangle_and_verify(content, self._is_implicit_VR,
+                                           self._is_little_endian)
+            self._content = ds
+        elif content == None:  # Otherwise, initialize a new dicom dataset
```
matthew-brett: Standard is `content is None` : PEP 8 "Comparisons to singletons like None should always be done with is or is not , never the equality operators.". 

I think 'Otherwise' doesn't apply any more in the comment.

nibabel/nifti1.py
```diff
@@ -379,13 +381,107 @@ def write_to(self, fileobj, byteswap):
...
+        else:
+            raise TypeError("content must be either a bytestring or a pydicom "
+                            "Dataset. Got %s" % content.__class__)
+
+    def _unmangle_and_verify(self, content, is_implicit_VR, is_little_endian):
+        """"Decode and verify dicom dataset"""
+        ds = self._unmangle(content, is_implicit_VR, is_little_endian)
+        for elem in ds:
+            if elem.VR not in pdcm.values.converters.keys():
+                raise StandardError  # Ensure that all VRs are valid
```
matthew-brett: Message for error?  E.g. `VR {0} not in standard`?

nibabel/tests/test_nifti1.py
```diff
@@ -1119,6 +1122,82 @@ def test_nifti_extensions():
...
+    assert_equal(exts_container.count('dicom'), 0)
+    exts_container.append(dcmext)
+    assert_equal(exts_container.count('dicom'), 1)
+    assert_equal(exts_container.get_codes(), [6, 6, 2])
+    assert_equal(dcmext._mangle(dcmext.get_content()), dcmbytes_explicit_be)
+    assert_equal(dcmext.get_sizeondisk() % 16, 0)
+
+    # creating an extension with bad content should raise
+    assert_raises(TypeError, Nifti1DicomExtension, 2, 0)
+
```
matthew-brett: Check `content = None` case?  Maybe I missed it.

nibabel/nifti1.py
kastman: I'm not sure reporting the VR itself is actually that useful - it's just a sanity check that all VRs are valid. If a VR is wrong it's less likely a problem with the VR itself and more likely a problem with the encoding. But I could be more explicit about that.

nibabel/nifti1.py
kastman: Sure, I was just following the docstring convention from Nifti1Extension (`nifti1.py#L262`). I'm fixing that here too. Thanks for the detailed numpy docstring link; it was helpful (I know it's linked at the nibabel developer help as well, but hadn't gone back to that yet).

nibabel/nifti1.py
kastman: When I built the doc `:data:` worked and created a link, but I think class may be more correct. Updated!

nibabel/tests/test_nifti1.py
kastman: Good point - I check a zero-length string (that was actually from the original `len(content) == 0` ducktyping) as the very first test but don't explicitly check None. Just added it.

nibabel/nifti1.py
kastman: Also, can you help with the ReStructuredText for linking to the pydicom Dataset class? Doesn't look like you're using the `conf.py` intersphinx mapping.

nibabel/nifti1.py
kastman: Actually, the only case I can find that will cause this is to give a malformed implicit VR, which is correctly reported by pydicom as a KeyError (`KeyError: "Unknown DICOM tag (0054, 0099) - can't look up VR"`) and not caught as a StandardError. So, I'm going to remove this validation code completely.

kastman:
@matthew-brett Anythign left that I can tweak or help you with on this one?

matthew-brett:
Sorry to be a bit slow - looks good - thanks a lot for your persistence.

PR #297, opened 2015-02-21
added ecat test data set with negative values

freec84:
I added a ecat7 file with negative values (transmission scan from ecat exac hr+ pet scanner) to the test data.
It seems that alls negative values are set to two times the maximum positive value when read  in with nibabel ecat load.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "nibabel-data/nitest-minc2"]
 	path = nibabel-data/nitest-minc2
 	url = git://github.com/matthew-brett/nitest-minc2.git
+[submodule "nipy-ecattest"]
+	path = nibabel-data/nipy-ecattest
+	url = https://github.com/freec84/nipy-ecattest


matthew-brett:
Great - thanks for this.  Do you somewhere have the minimum value in image, as it should be (if we weren't messing up the negative numbers)?  I'm thinking how to test we've got it right in due course.

matthew-brett:
@freec84 - do you have any metrics from the correctly loaded data to help with tests?  Like min, max, mean etc?    If not - let me know anyway?  Then I'll do my best to fix this without.

freec84:
Hi Matthew,

the metrics are:

min: -0.00061576
max:  0.19215
mean: 0.04933
unit: 1/cm

Bests,
Georg

matthew-brett:
Would you mind having a look at https://github.com/nipy/nibabel/pull/302 - it should fix your problem, and gives the same answers as you give above.

freec84:
Thanks a lot Matthew. You fix works. Negative values are now read in correctly.

Issue #207, opened 2013-10-11
nibabel.concat_images returns 5D images from 4D images


GaelVaroquaux:
If I concat a list of 4D images, I get a 5D image.

I was a bit surprised by this behavior: this is not what I was expecting from the docstring.

Maybe an option to avoid this behavior, and concat along the last axis, and not the last+1 axis would be useful. Another option would be to concat along a specified axis (given as a kwarg to concat_images), that could be '4' by default.

bcipolli:
Is there a recommended solution to this?  I couldn't find any elegant ways to eliminate the unary 4th dimension.  Since my images were size 1 in the 4th dimension, I could do:

```python
all_files = glob.glob('fM00223/*.img')
all_images = [nibabel.funcs.four_to_three(nibabel.load(f))[0] for f in all_files]
func_img = nibabel.funcs.concat_images(all_images)
```

matthew-brett:
I'm sorry, I just ran out of time for the 2.0 release, but I still agree that an axis keyword would be good.  Do you have any time to hack that up as a pull-request?   That would be the best solution.  Otherwise it's pretty easy to do by hand, by adapting the code in `concat_images`.

PR #298, opened 2015-02-24
Add an 'axis' parameter to concat_images, plus two tests.

bcipolli:
Trying to solve issue #207.  There are now three cases:

`axis is not None`: concat on the specified dimension
`axis is None` and all images have a final dimension of 1: concat on the final dimension (don't add a new one)
`axis is None` and any image has a final dimension != 1: concat on a new dimension.

Existing tests covered the 3rd case.  New tests were added for the first two cases.

### DIFF
--- a/nibabel/funcs.py
+++ b/nibabel/funcs.py
@@ -125,7 +125,7 @@ def concat_images(images, check_affines=True, axis=None):
             klass = img.__class__
 
             if axis is None:  # collect images in output array for efficiency
-                out_shape = (n_imgs, ) + img.shape[:3]
+                out_shape = (n_imgs, ) + img.shape
                 out_data = np.empty(out_shape)
             else:  # collect images in list for use with np.concatenate
                 out_data = [None] * n_imgs
@@ -133,23 +133,13 @@ def concat_images(images, check_affines=True, axis=None):
         elif check_affines and not np.all(img.affine == affine):
             raise ValueError('Affines do not match')
 
-        # Special case for 4D image with size[3] == 1; reshape to work!
-        if axis is None and img.get_data().ndim == 4 and img.get_data().shape[3] == 1:
-            out_data[i] = np.reshape(img.get_data(), img.get_data().shape[:-1])
-        else:
-            out_data[i] = img.get_data()
+        out_data[i] = img.get_data()
 
         del img
 
     if axis is None:
         out_data = np.rollaxis(out_data, 0, out_data.ndim)
     else:
-        # Massage the output, to allow combining 3D and 4D images.
-        is_3D = [len(d.shape) == 3 for d in out_data]
-        is_4D = [len(d.shape) == 4 for d in out_data]
-        if np.any(is_3D) and np.any(is_4D):  # Convert all to 4D
-            out_data = [data if is_4D[di] else np.reshape(data, data.shape + (1,))
-                        for di, data in enumerate(out_data)]
         out_data = np.concatenate(out_data, axis=axis)
 
     return klass(out_data, affine, header)
--- a/nibabel/funcs.py
+++ b/nibabel/funcs.py
@@ -94,13 +94,16 @@ def concat_images(images, check_affines=True, axis=None):
     Parameters
     ----------
     images : sequence
-       sequence of ``SpatialImage`` or filenames\s
+       sequence of ``SpatialImage`` or filenames of the same dimensionality\s
     check_affines : {True, False}, optional
        If True, then check that all the affines for `images` are nearly
        the same, raising a ``ValueError`` otherwise.  Default is True
     axis : None or int, optional
-        If None, concatenates on the 4th dimension.
-        If not None, concatenates on the specified dimension [-2 to 3).
+        If None, concatenates on a new dimension.  This rrequires all images
+           to be the same shape).
+        If not None, concatenates on the specified dimension.  This requires
+           all images to be the same shape, except on the specified dimension.
+           For 4D images, axis must be between -2 and 3.
     Returns
     -------
     concat_img : ``SpatialImage``
@@ -110,7 +113,7 @@ def concat_images(images, check_affines=True, axis=None):
 
     n_imgs = len(images)
     if n_imgs == 0:
-        raise ValueError('Cannot concatenate an empty list of images.')
+        raise ValueError("Cannot concatenate an empty list of images.")
 
     for i, img in enumerate(images):
         if not hasattr(img, 'get_data'):
--- a/nibabel/funcs.py
+++ b/nibabel/funcs.py
@@ -94,7 +94,7 @@ def concat_images(images, check_affines=True, axis=None):
     Parameters
     ----------
     images : sequence
-       sequence of ``SpatialImage`` or of filenames\s
+       sequence of ``SpatialImage`` or filenames\s
     check_affines : {True, False}, optional
        If True, then check that all the affines for `images` are nearly
        the same, raising a ``ValueError`` otherwise.  Default is True
@@ -109,32 +109,33 @@ def concat_images(images, check_affines=True, axis=None):
     '''
     n_imgs = len(images)
     img0 = images[0]
-    is_filename = False
     if not hasattr(img0, 'get_data'):
         img0 = load(img0)
-        is_filename = True
     affine = img0.affine
     header = img0.header
+    i0shape = img0.shape
+    del img0
 
     if axis is None:  # collect images in output array for efficiency
-        out_shape = (n_imgs, ) + img0.shape[:3]
+        out_shape = (n_imgs, ) + i0shape[:3]
         out_data = np.empty(out_shape)
     else:  # collect images in list for use with np.concatenate
         out_data = [None] * n_imgs
 
     for i, img in enumerate(images):
-        if is_filename:
+        if not hasattr(img, 'get_data'):
             img = load(img)
+
         if check_affines and not np.all(img.affine == affine):
             raise ValueError('Affines do not match')
 
+        # Special case for 4D image with size[3] == 1; reshape to work!
         if axis is None and img.get_data().ndim == 4 and img.get_data().shape[3] == 1:
             out_data[i] = np.reshape(img.get_data(), img.get_data().shape[:-1])
         else:
             out_data[i] = img.get_data()
 
-        if is_filename:
-            del img
+        del img
 
     if axis is None:
         out_data = np.rollaxis(out_data, 0, out_data.ndim)
--- a/nibabel/funcs.py
+++ b/nibabel/funcs.py
@@ -88,19 +88,6 @@ def squeeze_image(img):
                  img.extra)
 
 
-def _shape_equal_excluding(shape1, shape2, exclude_axes):
-    """ Helper function to compare two array shapes, excluding any
-    axis specified."""
-
-    if len(shape1) != len(shape2):
-        return False
-
-    idx_mask = np.ones((len(shape1),), dtype=bool)
-    idx_mask[exclude_axes] = False
-    return np.array_equal(np.asarray(shape1)[idx_mask],
-                          np.asarray(shape2)[idx_mask])
-
-
 def concat_images(images, check_affines=True, axis=None):
     ''' Concatenate images in list to single image, along specified dimension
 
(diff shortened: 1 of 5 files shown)


nibabel/funcs.py
```diff
@@ -122,8 +124,13 @@ def concat_images(images, check_affines=True):
         if check_affines:
             if not np.all(img.affine == affine):
                 raise ValueError('Affines do not match')
-        out_data[i] = img.get_data()
-    out_data = np.rollaxis(out_data, 0, len(i0shape)+1)
+        out_data[i] = img.get_data().copy()
+    if axis is not None:
+        out_data = np.concatenate(out_data, axis=axis)
+    elif np.all([d.shape[-1] == 1 for d in out_data]):
```
matthew-brett: I would prefer not to have a special case for ones on the last axis - the user can always do (when fixed) axis=-1 for that case.

nibabel/funcs.py
```diff
@@ -122,8 +124,13 @@ def concat_images(images, check_affines=True):
         if check_affines:
             if not np.all(img.affine == affine):
                 raise ValueError('Affines do not match')
-        out_data[i] = img.get_data()
-    out_data = np.rollaxis(out_data, 0, len(i0shape)+1)
+        out_data[i] = img.get_data().copy()
+    if axis is not None:
+        out_data = np.concatenate(out_data, axis=axis)
```
matthew-brett: First - I think this makes another copy of the array, which may well be large by this point.

Second - isn't this the wrong 'axis'?   For example, the most common use here would be to concatenate a (i, j, k, N1) and an (i, j, k, N2) image.   That would be axis=-1 or axis=3.  I think the code above would first - fail if N1 != N2 in the `out_data[i] = ` line and if N1 == N2, would fail for axis=3 because the last allowed axis in the call above would be 2 (it is iterating over the first axis generating 3D arrays).   For axis=-1 (corresponding to axis == 2) it would generate something of shape (i, j, k * (N1 + N2)) which is still not what you want, I don't think.

I think the solution is to use concatenate without `out_data` if `axis != None` - so have separate code paths for the previous case and the new case.

bcipolli: @matthew-brett For some reason, I had it in my head that what we actually had was a list.  I'm shocked that the few tests I did worked at all.

I'll start over on this.

nibabel/funcs.py
bcipolli: Fair enough. :)

bcipolli:
@matthew-brett Please take a look.  I know `np.asarray` copies from lists; I believe `np.concatenate` must copy as well.  I beefed up the test coverage as well.

nibabel/funcs.py
```diff
@@ -115,15 +117,19 @@ def concat_images(images, check_affines=True):
     affine = img0.affine
     header = img0.header
     out_shape = (n_imgs, ) + i0shape
```
matthew-brett: This loses the memory efficiency of the original.  How about:

```
if axis is None:  # collect images in output array for efficiency
    out_shape = (n_imgs, ) + i0shape
    out_data = np.empty(out_shape)
else:  # collect images in list for use with np.concatenate
    out_data = [None] * n_imgs
```

Then you can use `out_data[i] = img.get_data()` in the loop.

nibabel/funcs.py
```diff
@@ -115,15 +117,19 @@ def concat_images(images, check_affines=True):
...
         if check_affines:
             if not np.all(img.affine == affine):
                 raise ValueError('Affines do not match')
-        out_data[i] = img.get_data()
-    out_data = np.rollaxis(out_data, 0, len(i0shape)+1)
+        out_data.append(img.get_data())
+    if axis is not None:
+        out_data = np.concatenate(out_data, axis=axis)
+    else:
+        out_data = np.asarray(out_data)
```
matthew-brett: Restoring the empty array above, you would not need this line any more.

nibabel/tests/test_funcs.py
```diff
@@ -30,34 +30,71 @@ def _as_fname(img):
...
+                assert_raises(ValueError, concat_images, [img0, img3])
+                # except if check_affines is False
+                all_imgs = concat_images([img0, img1])
+                assert_array_equal(all_imgs.get_data(), all_data)
+                assert_array_equal(all_imgs.affine, affine)
+            # Delete images as prophylaxis for windows access errors
+            for img in imgs:
+                del(img)
+
+        # Test axis parameter and trailing unary dimension
```
matthew-brett: It would be good to check that values also match what you get from doing np.concatenate directly (as well as checking the shape).  Maybe you could do something like:

for axis in range(-3, 4):
    # some test

Maybe also check axis=None explicitly?

nibabel/funcs.py
bcipolli: Why is this less memory efficient?  It doesn't copy anywhere.  I assume that, under the covers, numpy preallocates a large array and does a memcopy.  So I believe this should be very efficient.

nibabel/funcs.py
```diff
@@ -98,7 +98,9 @@ def concat_images(images, check_affines=True):
     check_affines : {True, False}, optional
        If True, then check that all the affines for `images` are nearly
        the same, raising a ``ValueError`` otherwise.  Default is True
-
+    axis : int, optional
+        If None, concatenates on the last dimension.
```
matthew-brett: This should be:

```
    axis : None or int, optional
        If None, creates an extra final dimension, and concatenates along the new dimension.
```

nibabel/funcs.py
matthew-brett: Sorry - I should have explained - it's memory efficient because Python can delete the memory for the individual images as it goes through the loop.  Having said that, in order for that to happen, we need more logic:

```
if is_filename:
    del img
```

bcipolli: Got it.  I have an idea how to do all the cases efficiently; I'll try it out sometime today.  I'll also look into improving the test coverage and checks.

matthew-brett:
Anything I can do to help?

bcipolli:
Sorry, had a conference this weekend and have been doing some research.  I can get get back to this on Monday and try to wrap things up!

matthew-brett:
No problem - and thanks for your work on this.

bcipolli:
Hope to get to this today or tomorrow.

matthew-brett:
Great - thanks.​

bcipolli:
@matthew-brett I've got the basic code working, but had a question while testing.

How flexible do you want `concat_images` to be?  Specifically, if a mix of 3D and 4D images (with a unary 4th dimension, for example) are passed, it is possible to concat them--but it'd take an extra check to make it work.

Would you prefer simply to let that error, or to do the check to make sure those work together?  From the testing standpoint, it's no problem to test that case as I've written things now.

bcipolli:
Also, mixed image/filename is not allowed.  Easy to extend for that as well.
Finally, I think `del img` is fine without the `if` check; it just removes the reference (doesn't force garbage collection); see https://docs.python.org/2/reference/simple_stmts.html#the-del-statement

matthew-brett:
My instinct is to go for simpler code at the expense of not allowing obscurities like mixing 3D and 4D, or image objects and filenames.    We can always add those later if it looks like a significant use-case.

bcipolli:
@matthew-brett It made the code simpler to allow mixed than to disallow it.  However, the 3D/4D possibilities did make the code more complex.

The code could lose about 5-6 lines of complexity if all images must be 3D or 4D.  The testing code is going to be complex no matter what, due to the testing of all axes, None, and no argument specified.

Let me know if you'd like to trim some of the cases, but things are working as-is.

nibabel/funcs.py
```diff
@@ -88,43 +88,67 @@ def squeeze_image(img):
...
-    is_filename = False
-    if not hasattr(img0, 'get_data'):
-        img0 = load(img0)
-        is_filename = True
-    i0shape = img0.shape
-    affine = img0.affine
-    header = img0.header
-    out_shape = (n_imgs, ) + i0shape
-    out_data = np.empty(out_shape)
+    if n_imgs == 0:
```
matthew-brett: Drop this check?   I guess if they pass in an empty list they can expect an empty list back?

nibabel/funcs.py
```diff
@@ -88,43 +88,67 @@ def squeeze_image(img):
...
     ----------
     images : sequence
-       sequence of ``SpatialImage`` or of filenames\s
+       sequence of ``SpatialImage`` or filenames\s
     check_affines : {True, False}, optional
        If True, then check that all the affines for `images` are nearly
        the same, raising a ``ValueError`` otherwise.  Default is True
-
+    axis : None or int, optional
+        If None, concatenates on the 4th dimension.
```
matthew-brett: I think it concatenates on a new final dimension, technically. For example, it will concatenate 2D images to 3D.  Also, it does not by default concatenate on an existing dimension (e.g. the 4th) - right?

nibabel/funcs.py
```diff
@@ -88,43 +88,67 @@ def squeeze_image(img):
...
+        If not None, concatenates on the specified dimension [-2 to 3).
```
matthew-brett: -2 to 3 depends on the input right?   For 2D images, this will be -1 to 2 for example.

nibabel/funcs.py
```diff
@@ -88,43 +88,67 @@ def squeeze_image(img):
...
+            if axis is None:  # collect images in output array for efficiency
+                out_shape = (n_imgs, ) + img.shape[:3]
+                out_data = np.empty(out_shape)
+            else:  # collect images in list for use with np.concatenate
+                out_data = [None] * n_imgs
+
+        elif check_affines and not np.all(img.affine == affine):
+            raise ValueError('Affines do not match')
+
+        # Special case for 4D image with size[3] == 1; reshape to work!
```
matthew-brett: So - I still think this should not be special cased - because it's difficult to explain and the caller can just pass axis=-1 if they want this behavior.

nibabel/funcs.py
```diff
@@ -88,43 +88,67 @@ def squeeze_image(img):
...
+        else:
+            out_data[i] = img.get_data()
+
+        del img
+
+    if axis is None:
+        out_data = np.rollaxis(out_data, 0, out_data.ndim)
+    else:
+        # Massage the output, to allow combining 3D and 4D images.
+        is_3D = [len(d.shape) == 3 for d in out_data]
```
matthew-brett: Looking at this - I also think this is a little difficult to explain, and would prefer the simplest possible thing here (don't allow this mix).

nibabel/tests/test_funcs.py
```diff
@@ -30,34 +30,105 @@ def _as_fname(img):
...
+                        assert_raises(ValueError, concat_images, [img0, img2], **concat_imgs_kwargs)
+                        assert_raises(ValueError, concat_images, [img0, img3], **concat_imgs_kwargs)
+
+                        # except if check_affines is False
+                        try:
+                            all_imgs = concat_images([img0, img1], **concat_imgs_kwargs)
+                            assert_array_equal(all_imgs.get_data(), all_data)
+                            assert_array_equal(all_imgs.affine, affine)
+                            assert_false(expect_error, "Expected a concatenation error, but got none.")
+                        except ValueError as ve:
```
matthew-brett: Use `else` to put the `except` line below the line that will cause the error?

```
try:
    all_imgs = concat_images([img0, img1], **concat_imgs_kwargs)
except ValueError as ve:
    assert_true(expect_error, ve.message)
else:
    assert_array_equal(all_imgs.get_data(), all_data)
```

matthew-brett:
I might have missed it - but did you test concatenating 2D images?

nibabel/funcs.py
bcipolli: OK, I think I got it now.  Here's how the function will work:
Input: any mixed dimensionality, Output: error.
Input: ND dimension images, axis=None, output: an ND+1 dimension image.
Input: ND dimension images, axis=int, output: an ND dimension image.
I'll just clean away the special cases and update the tests.

bcipolli:
>I might have missed it - but did you test concatenating 2D images?

No... just 3D and 4D.  What about 1D?  5D?  I'm not sure of the limitations.

nibabel/funcs.py
bcipolli: In the past, and currently, this throws an error.  I added the check because the error did not indicate the issue clearly.

matthew-brett:
I think 2D is a reasonable case to test.

If it's easy to test 5D the user might reasonably expect that to work.

1D - that's a bit obscure :)

bcipolli:
@matthew-brett I cut out the 3D/4D special cases and added testing for 2D and 5D.

I did have to add some complexity.  I found a lot of edge cases where numpy's broadcasting, or details of `np.concatenate`, led to unexpected results.  These are unlikely edge cases, but to do things correctly (and test generically), I have to explicitly test the shape.

The upside is that the error message will be extremely clear.  The downside is that the code to check the shape properly in both axis cases was ~7 lines.

Let me know what you think!  I think we're pretty close now.

nibabel/funcs.py
```diff
@@ -88,43 +88,81 @@ def squeeze_image(img):
...
     ----------
     images : sequence
-       sequence of ``SpatialImage`` or of filenames\s
+       sequence of ``SpatialImage`` or filenames of the same dimensionality\s
     check_affines : {True, False}, optional
        If True, then check that all the affines for `images` are nearly
        the same, raising a ``ValueError`` otherwise.  Default is True
-
+    axis : None or int, optional
+        If None, concatenates on a new dimension.  This rrequires all images
```
matthew-brett: typo rrequires

nibabel/funcs.py
matthew-brett: Fair enough.

nibabel/funcs.py
```diff
@@ -88,43 +88,81 @@ def squeeze_image(img):
...
+                out_shape = (n_imgs, ) + shape
+                out_data = np.empty(out_shape)
+            else:  # collect images in list for use with np.concatenate
+                out_data = [None] * n_imgs
+
+        elif check_affines and not np.all(img.affine == affine):
+            raise ValueError('Affines do not match')
+
+        elif ((axis is None and not np.array_equal(shape, img.shape)) or
+              (axis is not None and not _shape_equal_excluding(shape, img.shape,
```
matthew-brett: Sorry - this is probably a dumb question, but won't np.concatenate error in this second situation anyway?

nibabel/tests/test_funcs.py
```diff
@@ -30,34 +30,96 @@ def _as_fname(img):
...
+                        concat_imgs_kwargs = dict(axis=axis)
+
+                    # Create expected output
+                    try:
+                        # Error will be thrown if the np.concatenate fails.
+                        #   However, when axis=None, the concatenate is possible
+                        #   but our efficient logic (where all images are
+                        #   3D and the same size) fails, so we also
+                        #   have to expect errors for those.
+                        expect_error = data0.ndim != data1.ndim
```
matthew-brett: Put this outside `try ... except` ?  Do you need the `try .. except` if `expect_error` is True?

nibabel/funcs.py
```diff
@@ -88,43 +88,81 @@ def squeeze_image(img):
...
     check_affines : {True, False}, optional
        If True, then check that all the affines for `images` are nearly
        the same, raising a ``ValueError`` otherwise.  Default is True
-
+    axis : None or int, optional
+        If None, concatenates on a new dimension.  This rrequires all images
+           to be the same shape).
+        If not None, concatenates on the specified dimension.  This requires
+           all images to be the same shape, except on the specified dimension.
+           For 4D images, axis must be between -2 and 3.
```
matthew-brett: Is this last sentence necessary - I mean - won't the concatenate raise an informative error in cases outside this range?

nibabel/tests/test_funcs.py
bcipolli: I personally try to keep indent levels to be a logical block, so sometimes I include things inside a `try...except` that may not throw, so that the logical block is maintained.  Here, the computation of `expect_error` is all at the same indent level, which I find to be semantically helpful.

With that said, I'm happy to move it out; doing that now!

nibabel/funcs.py
bcipolli: It will error, but the error message is pretty cryptic.  I added this logic in after seeing the error.

bcipolli:
@matthew-brett made the requested changes, sans the question about the error check.

matthew-brett:
OK - thanks very much for all your work on this.

The last remaining thing is that we are deleting the image even if it is passed in as an image object, in the list, and that could be bad.  We either want to delete the image only if we created it by loading the filename, or (probably better) we could use ``img.get_data(caching='unchanged')`` so we don't fill up the cache in the image if it is already empty, and the image objects will stay really small.

Also you kindly removed the expect_error from the `try` branch, but it is undefined in some tests now.

I put in a pull request to your branch just now with fixes for these, and some optional refactoring.  If you prefer my version, please go ahead and merge, otherwise - fix these last things and we'll merge asap.

Thanks again...

bcipolli:
@matthew-brett I pushed a change for `expect_error` issue.  

As for the `del`, I believe it is proper--we're not telling Python to delete the image; we're simply deleting the (local) reference to the image.  If the image was passed in as an object, the calling code will still have a reference and the image data will remain in memory; if it was loaded locally, then that reference is the only one, and the object will be marked for garbage collection.  If that's right, then there's no need for any special logic.

Thanks for the pull request / refactor.  I prefer my version with the initialization within the loop, as it potentially avoids loading the first image's data twice.  Since we worked hard to improve efficiency in this version, I think that small optimization is worthwhile.  But as with you, my preference isn't strong; happy to defer if you feel any obstacle.

Thanks also for your work, sorry I didn't get this done more smoothly!  Let me know if, after reading these comments, you think there are still some changes to make, and I'll be glad to take care of them ASAP.

matthew-brett:
For the del - yes, thinking about it, the loop iteration will make new reference to the object that may have its only other reference in a passed list.   On the other hand, this function will fill the image array cache of the passed image objects, which may be desirable or not desirable.    I think probably not. 

For the load data twice if doing initialization outside the loop - could you explain?  I think there's only one 'get_data' call in the code.

I prefer my code because - well who doesn't prefer their own code? - but also I have an allergy to doing i == 1 initialization in the loop that is instinctive, but this is your work, so this is your call.

bcipolli:
> On the other hand, this function will fill the image array cache of the passed image objects, which may be desirable or not desirable. I think probably not.

I believe I understand.  I also believe this issue existed in the old version of the code.  If so, can we open a new issue to address it?

> For the load data twice if doing initialization outside the loop - could you explain? I think there's only one 'get_data' call in the code.

I probably simply don't know the nibabel code well enough.  I assumed the data was loaded when `load(img)` was called; sounds like it's always deferred to `get_data`.  In that case, the in-loop initialization is totally unnecessary and just more confusing.

bcipolli:
Just merged your code; given what was discussed above, I think it solves the remaining issues.  Let's be done with this! :)

matthew-brett:
Thank you for being so patient with my nit-picks and slowness, and thanks very much for taking the time to do this.  I think the test fails are spurious and it's time to merge...

PR #302, opened 2015-03-29
BF+TST: test /fix for signed int data type

matthew-brett:
We were incorrectly loading signed ints as unsigned; fix and test.

See : https://github.com/nipy/nibabel/pull/297 for motivation.

### DIFF
--- a/nibabel/ecat.py
+++ b/nibabel/ecat.py
@@ -152,14 +152,19 @@ subheader_dtd = [
 subhdr_dtype = np.dtype(subheader_dtd)
 
 # Ecat Data Types
+# See:
+# http://www.turkupetcentre.net/software/libdoc/libtpcimgio/ecat7_8h_source.html#l00060
+# and:
+# http://www.turkupetcentre.net/software/libdoc/libtpcimgio/ecat7r_8c_source.html#l00717
 _dtdefs = ( # code, name, equivalent dtype
-    (1, 'ECAT7_BYTE', np.uint8),
-    (2, 'ECAT7_VAXI2', np.int16),
-    (3, 'ECAT7_VAXI4', np.int32),
-    (4, 'ECAT7_VAXR4', np.float32),
-    (5, 'ECAT7_IEEER4', np.float32),
-    (6, 'ECAT7_SUNI2', np.int16),
-    (7, 'ECAT7_SUNI4', np.int32))
+           (1, 'ECAT7_BYTE', np.uint8),
+           # Byte signed? https://github.com/nipy/nibabel/pull/302/files#r28275780
+           (2, 'ECAT7_VAXI2', np.int16),
+           (3, 'ECAT7_VAXI4', np.int32),
+           (4, 'ECAT7_VAXR4', np.float32),
+           (5, 'ECAT7_IEEER4', np.float32),
+           (6, 'ECAT7_SUNI2', np.int16),
+           (7, 'ECAT7_SUNI4', np.int32))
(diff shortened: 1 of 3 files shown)


matthew-brett:
Any reviews here?  I think this is a bug and I will merge soon unless I hear otherwise.

nibabel/ecat.py
```diff
@@ -155,10 +155,10 @@
 _dtdefs = ( # code, name, equivalent dtype
     (1, 'ECAT7_BYTE', np.uint8),
     (2, 'ECAT7_VAXI2', np.int16),
-    (3, 'ECAT7_VAXI4', np.float32),
+    (3, 'ECAT7_VAXI4', np.int32),
     (4, 'ECAT7_VAXR4', np.float32),
     (5, 'ECAT7_IEEER4', np.float32),
-    (6, 'ECAT7_SUNI2', np.uint16),
+    (6, 'ECAT7_SUNI2', np.int16),
```
Eric89GXL: These do indeed look like they were wrong. Any idea why the original author would have head `float32` for a datatype that has `I4` at the end and `uint16` when everything else is `int*`?

Eric89GXL:
LGTM, +1 for merge. My comment above is just a question for my own edification.

nibabel/ecat.py
matthew-brett: No idea I'm afraid.  I don't think I've got any real data to test with, for ECAT_VAXI4 either, so it's only a guess.

Eric89GXL: http://www.turkupetcentre.net/software/libdoc/libtpcimgio/ecat7_8h.html#a3256675b42164c36df6e3810ac8c82fa

Eric89GXL: Don't know if it's right, but it's at least another reference

Eric89GXL: ...and it seems to match what you've done here in any case

matthew-brett: I saw that file - but it wasn't immediately obvious what data type it was - did I miss something obvious?

Eric89GXL: if you go to the functions it references, you can see what data types the actually get read into

Eric89GXL: For example, ECAT7_SUNI2 does get read in as short int:

http://www.turkupetcentre.net/software/libdoc/libtpcimgio/ecat7r_8c_source.html#l00663

matthew-brett: Aha - yes - thanks for tracking that down.

matthew-brett: Hmm - this:

http://www.turkupetcentre.net/software/libdoc/libtpcimgio/ecat7r_8c_source.html#l00720

suggests to me that the BYTE type is signed - do you agree?

Eric89GXL: The pointer starts out as a `char *` and isn't re-cast as anything else, which makes me think it is unsigned. What made you think it was signed?

matthew-brett: Just some trivial testing with gcc:

```
#include <limits.h>
int main() {
    printf("Char min is %d\n", CHAR_MIN);
}
```

But I see in fact it is undefined ...

http://stackoverflow.com/questions/2054939/is-char-signed-or-unsigned-by-default

matthew-brett: The only ECAT files I can find on our system have ECAT7_SUNI2 type, so very hard to know what the intention was.

matthew-brett:
Thanks - I added a few lines pointing to the docs you found, merging.

PR #304, opened 2015-03-31
BF: work around np.allclose float128 bug

matthew-brett:
Numpy 1.9.2 allclose thinks that np.inf in float128 precision is not allclose
to itself.  Work round with expanded assert_allclose_safely function.

See: http://nipy.bic.berkeley.edu/builders/nibabel-py2.7-osx-10.6/builds/178/steps/shell_4/logs/stdio for resulting error, maybe fixed by this PR.


matthew-brett:
Any comments here?

yarikoptic:
fwiw  with such extensive test I do not have a concern...  send to be doing the right thing  :-)

matthew-brett:
:) - OK - I'll merge then.  Thanks for taking a look.

PR #306, opened 2015-04-10
RF Simplify concat_images

effigies:
Reduces code paths for `axis = None` case to a final check before re-encapsulating data into a SpatialImage. Also separates image compatibility checks from data aggregation, which incidentally allows the comparison of the first image with itself to be skipped. These make the function more readable, to me at least.

At least from timing `test_concat`, this gives a very slight speed-up (~0.5s out of a 10s test).

It does come at the cost of it not being immediately obvious that `np.concatenate(axis=None).reshape((n_imgs, ) + shape0)` should necessarily work out, but the tests show us it does.

### DIFF
--- a/nibabel/funcs.py
+++ b/nibabel/funcs.py
@@ -120,19 +120,13 @@ def concat_images(images, check_affines=True, axis=None):
     klass = img0.__class__
     shape0 = img0.shape
     n_dim = len(shape0)
-    if axis is None:
-        # collect images in output array for efficiency
-        out_shape = (n_imgs, ) + shape0
-        out_data = np.empty(out_shape)
-    else:
-        # collect images in list for use with np.concatenate
-        out_data = [None] * n_imgs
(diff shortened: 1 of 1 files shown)


nibabel/funcs.py
```diff
@@ -120,19 +120,13 @@ def concat_images(images, check_affines=True, axis=None):
     klass = img0.__class__
     shape0 = img0.shape
     n_dim = len(shape0)
-    if axis is None:
-        # collect images in output array for efficiency
-        out_shape = (n_imgs, ) + shape0
```
matthew-brett: Ah no - this was different deliberately in order to save memory.  With this path, there are not large temporary arrays for the axis == None case, we directly fill the output array.

effigies: Oh, I see. I assumed time optimization from the comment. Nevermind then.

Issue #311, opened 2015-05-12
'NoneType' object has no attribute 'groups'


cccbauer:
Hi,

I am trying to convert par/rec to nii to feed them into a real-time system, but somehow the par/rec files that Philips scanner dumps into the folder are not converted and 'NoneType' object has no attribute 'groups' appears.

thanks so much

clemens

matthew-brett:
Thanks for the report.

I guess you are using `parrec2nii` from the command line?

Can you isolate the error to a particular call to `parrec2nii`?

cccbauer:
Yes, I'm using it form the command line, I just typed in parrec2nii Dump-0000.par with no selected options.
it gives:

Caught 1 exceptions. Dump follows:

 Dump-0000.par: 'NoneType' object has no attribute 'groups'

matthew-brett:
Can you share `Dump-0000.par` publicly or privately so I can take a look?

cccbauer:
yes, here you have:

https://www.dropbox.com/sh/afvly1lu04c9ms0/AAAYDoIlq-H7EnZpfLVt4jq3a?dl=0

matthew-brett:
Thanks - I see the problem - will fix later on today.

cccbauer:
Fantastic, I appreciate it very much!

cccbauer:
Just for additional information, It also seems that the par/rec files that come from the scanner include phase images and here is a matlab code which can help figure out how the files are structured.
 thanks


x_res = 80;
y_res = 80;
dynamics = 50; 
mixes = 2; % is 2 for rt dumps
slices = 44; % is 1 for rt dumps
a = y_res * dynamics * mixes * slices;  % Number of columns of raw data
 
filenamestruct=dir('*.rec');
for dump=1:length(filenamestruct)
    storedname=filenamestruct(dump).name;
    
    fid = fopen(storedname, 'r', 'l');
    Data(:,:) = fread(fid, [x_res,a],'uint16');
    fclose(fid);
    
    % if mixes = 2, the first is the phase image, 
    % i.e. Data4D(:,:,:,1,:) is the phase matrix
    if mixes == 1
        Data4D = reshape(Data,x_res,y_res,slices,[]); % Matrix has structure [x_voxels, y_voxels, slices, dynamics]  
    elseif mixes == 2
        Data4D = reshape(Data,x_res,y_res,slices,mixes,[]); % Matrix has structure [x_voxels, y_voxels, slices, phase, dynamics]
    else
[...]

PR #312, opened 2015-05-13
MRG: test for bug in PAR header reading

matthew-brett:
Fix for reading empty parrec fields, floating point window values.

Add test.

Closes https://github.com/nipy/nibabel/issues/311


matthew-brett:
Ready for review...

nibabel/parrec.py
```diff
@@ -185,8 +185,8 @@
     ('rescale intercept', float),
     ('rescale slope', float),
     ('scale slope', float),
-    ('window center', int,),
-    ('window width', int,),
+    ('window center', float,),
+    ('window width', float,),
```
Eric89GXL: Was this the fix?

nibabel/tests/data/umass_anonymized.PAR
```diff
@@ -0,0 +1,176 @@
...
+#  scanning sequence                        (integer)
+#  index in REC file (in images)            (integer)
+#  image pixel size (in bits)               (integer)
+#  scan percentage                          (integer)
+#  recon resolution (x y)                   (2*integer)
+#  rescale intercept                        (float)
+#  rescale slope                            (float)
+#  scale slope                              (float)
+#  window center                            (integer)
+#  window width                             (integer)
```
Eric89GXL: ...if so, I'm confused. Silly neuroimaging formats.

nibabel/parrec.py
Eric89GXL: If it was, could you add a comment noting that there is a discrepancy between what is observed in PAR files (integer) and what they actually store (float)? Otherwise the next person might come across this and errantly revert your changes.

nibabel/parrec.py
```diff
@@ -244,6 +245,9 @@ class PARRECError(Exception):
     pass
 
 
+# Value after colon may be absent
+GEN_RE = re.compile(r".\s+(.*?)\s*:(.*)")
```
matthew-brett: Eric - the fix is here - this regex allows the field after the colon to be whitespace only - in the example file the Patient name was empty, which caused the previous code to barf.

nibabel/tests/data/umass_anonymized.PAR
matthew-brett: Yes - says integer here, but below, actually floating point.

nibabel/parrec.py
effigies: Instead of `value.strip()` later, why not the following regex, which simply changes the `\s+` to a `\s*`?

    GEN_RE = re.compile(r".\s+(.*?)\s*:\s*(.*)")

matthew-brett: Thanks, yes, that is better.

matthew-brett:
Any more comments?  OK to merge?

Eric89GXL:
LGTM, +1 for merge

matthew-brett:
Thanks for the review

PR #314, opened 2015-06-07
DOC Use HTTPS for outside scripts/forms

effigies:
Since `*.github.io` pages support HTTPS, pulling in content from HTTP addresses causes browsers to complain about unsafe scripts. This eliminates any warnings, on a test I ran on one of our lab web servers. Loading scripts over HTTPS from an HTTP connection shouldn't cause problems.

HTTPS might not be a priority for nibabel, so I limited this initial commit to making the nibabel site eliminate browser warnings. If it would be welcome, I can also go through other URLs throughout the documentation and upgrade them to HTTPS when they exist and have a valid certificate.

### DIFF
--- a/doc/source/_templates/indexsidebar.html
+++ b/doc/source/_templates/indexsidebar.html
@@ -6,7 +6,7 @@
     <li><a href="http://mail.scipy.org/pipermail/nipy-devel">Mailing list archive</a></li>
 </ul>
 
-<script type="text/javascript" src="http://www.ohloh.net/projects/480908/widgets/project_partner_badge"></script>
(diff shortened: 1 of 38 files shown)


matthew-brett:
Yes, sure - go wild on https :)​

effigies:
Well, a little less important now I see that https://nipy.github.io will be redirecting to the HTTP-only http://nipy.org, but I can go ahead and do this.

How should I treat dead links? If I can find a new location, I'll change to that, but if not, should I leave them as is, mark as dead somehow, or remove them entirely?

matthew-brett:
Why not remove them, and we'll all have a go at trying to find replacements
while we're reviewing the PR.​

matthew-brett:
Nice - thanks very much for doing that.

PR #315, opened 2015-06-10
WIP: Automatically reshape FreeSurfer ico7 niftis

### LINKED ISSUE #309: Nifti surface files misshapen when loaded
As noted in #100, FreeSurfer can put surface files into niftis. For example, on a machine that has FreeSurfer installed:

    $ mri_convert $SUBJECTS_DIR/fsaverage/surf/lh.orig.avg.area.{mgh,nii}

To examine the files:

    >>> fmt = os.path.join(os.environ['SUBJECTS_DIR'], 'fsaverage/surf/lh.orig.avg.area.{}').format
    >>> mgh = nib.load(fmt('mgh'))
    >>> nii = nib.load(fmt('nii'))

We see:

    >>> mgh.shape
    (163842, 1, 1)
    >>> nii.shape
    (27307, 1, 6)
    >>> np.all(np.squeeze(mgh.get_data()) == nii.get_data().transpose().reshape(-1))
    True

So the data's the same, but reshaped and has an axis rolled. I think the expected behavior would probably be a `Nifti1Image` with the same shape as the `MGHImage`.

On the plus side, `nib.load(x).to_filename(x)` seems to produce files that FreeSurfer tools accept without complaint (and look the same at first glance).

I could probably write up the above in test form in the next couple days, but I'm really not sure where to begin digging into Nifti1Image to fix this. It's also possible that the problem is that `mri_convert` is producing `nii`s with bad headers.

effigies:
Closes #309.

Tried to follow in the footsteps of f53c168. Presents programmer with data object equivalent to an MGH, leaving header intact for writing to file.

Tests planned:

1. `mri_convert` a `.mgh` file in `nibabel-data/nitest-freesurfer/fsaverage/surf` to '.nii' and verify `img.get_data()` is equal
2. Create new image with ico7 shape, verify `img.header._structarr['dim'][1:4] == (27307, 1, 6)`

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -698,16 +698,26 @@ class Nifti1Header(SpmAnalyzeHeader):
         Allows for freesurfer hack for large vectors described in
         https://github.com/nipy/nibabel/issues/100 and
         https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?spec=svn5022&r=5022#77
+
+        Allows for freesurfer hack for 7th order icosahedron surface described
+        in
+        https://github.com/nipy/nibabel/issues/309
+        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/load_nifti.m?r=8776#86
+        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?r=8776#50
         '''
         shape = super(Nifti1Header, self).get_data_shape()
         # Apply freesurfer hack for vector
-        if shape != (-1, 1, 1): # Normal case
+        if shape == (-1, 1, 1):
+            vec_len = int(self._structarr['glmin'])
+            if vec_len == 0:
+                raise HeaderDataError('-1 in dim[1] but 0 in glmin; '
+                                      'inconsistent freesurfer type header?')
+            return (vec_len, 1, 1)
+        # Apply freesurfer hack for ico7 surface
+        elif shape == (27307, 1, 6):
+            return (163842, 1, 1)
+        else:  # Normal case
             return shape
-        vec_len = int(self._structarr['glmin'])
-        if vec_len == 0:
-            raise HeaderDataError('-1 in dim[1] but 0 in glmin; inconsistent '
-                                  'freesurfer type header?')
-        return (vec_len, 1, 1)
 
     def set_data_shape(self, shape):
         ''' Set shape of data
@@ -725,12 +735,22 @@ class Nifti1Header(SpmAnalyzeHeader):
         Applies freesurfer hack for large vectors described in
         https://github.com/nipy/nibabel/issues/100 and
         https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?spec=svn5022&r=5022#77
+
+        Allows for freesurfer hack for 7th order icosahedron surface described
+        in
+        https://github.com/nipy/nibabel/issues/309
+        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/load_nifti.m?r=8776#86
+        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?r=8776#50
         '''
-        # Apply freesurfer hack for vector
         hdr = self._structarr
         shape = tuple(shape)
-        if (len(shape) == 3 and shape[1:] == (1, 1) and
-            shape[0] > np.iinfo(hdr['dim'].dtype.base).max): # Freesurfer case
+
+        # Apply freesurfer hack for ico7 surface
+        if shape == (163842, 1, 1):
+            shape = (27307, 1, 6)
+        # Apply freesurfer hack for vector
+        elif (len(shape) == 3 and shape[1:] == (1, 1) and
+                shape[0] > np.iinfo(hdr['dim'].dtype.base).max):
             try:
                 hdr['glmin'] = shape[0]
             except OverflowError:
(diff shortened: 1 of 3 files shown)


effigies:
@matthew-brett How does one actually run the `test_nifti1` tests?

    $ python -c 'import nibabel.tests.test_nifti1 as t; t.TestNifti1Image()' 
      File "/usr/lib/python2.7/unittest/case.py", line 191, in __init__
        (self.__class__, methodName))
    ValueError: no such test method in <class 'nibabel.tests.test_nifti1.TestNifti1Image'>: runTest

matthew-brett:
Run the tests locally you mean?

You could start with `git submodule update --init` to get the test data and then `nosetests nibabel\tests\test_nifti1.py` from the nibabel root directory (containing `setup.py`).

effigies:
Thanks. Wrote tests that pass locally, with change in https://bitbucket.org/nipy/nitest-freesurfer/pull-request/1/. If that PR goes through, I'll push the tests.

nibabel/nifti1.py
```diff
@@ -698,16 +698,26 @@ def get_data_shape(self):
...
         # Apply freesurfer hack for vector
-        if shape != (-1, 1, 1): # Normal case
+        if shape == (-1, 1, 1):
+            vec_len = int(self._structarr['glmin'])
+            if vec_len == 0:
+                raise HeaderDataError('-1 in dim[1] but 0 in glmin; '
+                                      'inconsistent freesurfer type header?')
+            return (vec_len, 1, 1)
+        # Apply freesurfer hack for ico7 surface
+        elif shape == (27307, 1, 6):
```
matthew-brett: Do I understand right - that this is one particular image ('ico7')?   So the hack only applies to that image?

effigies: To surfaces with that number of vertices. I can't find a good reference to the arguments as to why they do this[0], but a reasonably common FreeSurfer surface is a warped icosahedron, and in particular they [recommend](https://surfer.nmr.mgh.harvard.edu/fswiki/mri_surf2surf#line-31) using the 7th-order icosahedron. This is the number of vertices used in their `fsaverage` group template for comparisons across subjects.

Their MATLAB code indicates that this specific surface shape is hard-coded:
https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?r=8776#50
https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/load_nifti.m?r=8776#86

[0] I believe the use of the icosahedron is to allow for near constant neighbor-neighbor distances.

matthew-brett: I noticed this:

```
By default, mri_surf2surf will save the output as multiple 'slices'; has no effect for paint/w output format. For ico, the output will appear to be a 'volume' with Nv/R colums, 1 row, R slices and Nf frames, where Nv is the number of vertices on the surface. For icosahedrons, R=6. For others, R will be the prime factor of Nv closest to 6. Reshaping is for logistical purposes (eg, in the analyze format the size of a dimension cannot exceed 2^15). Use this flag to prevent this behavior. This has no effect when the output type is paint.
```

at https://surfer.nmr.mgh.harvard.edu/fswiki/mri_surf2surf - comment for `--noreshape`.

effigies: The easiest way to clarify the situation is to simply try all of the icosahedra:

    $ for ORDER in {1..7}; do
        mri_surf2surf --hemi rh --srcsubject fsaverage \
            --srcsurfval $SUBJECTS_DIR/fsaverage/surf/rh.orig.avg.area.mgh \
            --trgsubject ico --trgsurfval rh.ico$ORDER.nii \
            --trgicoorder $ORDER &> /dev/null \
        && python -c "from nibabel import load;\
              print ($ORDER, load('rh.ico$ORDER.nii').header._structarr['dim'][1:4])" \
        && rm rh.ico$ORDER.nii
      done 
    (1, array([42,  1,  1], dtype=int16))
    (2, array([162,   1,   1], dtype=int16))
    (3, array([642,   1,   1], dtype=int16))
    (4, array([2562,    1,    1], dtype=int16))
    (5, array([10242,     1,     1], dtype=int16))
    (6, array([-1,  1,  1], dtype=int16))
    (7, array([27307,     1,     6], dtype=int16))

So whatever their ideal of what ought to be done for icosahedra with >2^15 vertices, in practice, they use the (27307, 1, 6) shape for ico7, the large vector hack for ico6 and just (x, 1, 1) for lesser icosahedra. (ico8 does not work.)

matthew-brett: Is there any possibility that someone will try to use this with a volume with some other number of vertices?  Would they expect it to work?

matthew-brett:
Ben - I noticed your email to the freesurfer list, and I think it got no reply.  What do you think we should do here?

effigies:
I figured I'd at least wait until Monday, since I sent this email out on a Friday in the summer.

If we don't hear anything, my feeling is we should follow the lead of `load/save_nifti.m` and only handle these two special cases until there's evidence of other special cases in the wild.

Though the `ico6` case is indicated in the documentation, being unable to produce it with the tool the documentation was written about makes me skeptical that they've bothered implementing it. (They recommend everybody use ico7, anyway.) We could still add the case, as it's an unlikely shape to appear by accident, but then that introduces the problem of saving files: Do we use the (-1, 1, 1) or the (6827, 1, 6)?

Assuming the decision is to restrict ourselves to the existing cases, is there anything you need from me regarding the [nitest-freesurfer PR](https://bitbucket.org/nipy/nitest-freesurfer/pull-request/1/)?

matthew-brett:
Just needs tests - I guess these are waiting on the nitest-freesurfer PR

effigies:
Yup. In case BitBucket didn't alert you, I updated the PR over there.

matthew-brett:
Merged the data commit, thanks for that.

effigies:
Seems not to be fetching the new nitest-freesurfer. Is that just a thing that requires time to sync? https://travis-ci.org/nipy/nibabel/jobs/66921373

matthew-brett:
Ah no - you need to update the submodule.  The submodule points to one exact commit, not a branch - https://matthew-brett.github.io/pydagogue/git_submodules.html

matthew-brett:
Great - thanks very much for this - in it goes.

PR #318, opened 2015-06-25
BF: Use deprecate_with_doc as decorator

effigies:


matthew-brett:
Oops - I guess I should have checked - thanks.

Issue #317, opened 2015-06-24
Push loadsave logic into individual classes, remove logic duplication


bcipolli:
In modifying `loadsave.py` for `CIFTI` file format, I found three design choices that made it more challenging for me to understand and modify code:

1. The logic for whether a file is of each Image type is contained in a single function, not modularized into each image (yet the load/save functionality is).
2. Image type / file extension associations is duplicated between `load` and `save` functions.
3. The list of available file formats is duplicated (in imports, then again in the explicit logic of `load` and `save`, each).

I propose the following redesign (which I think is relatively simple):
* Each Image class should declare class variable `valid_ext` (valid filename extensions) and class function `is_image(filename)` (which returns `True` if the given filename is of the Image class.
* `loadsave.py` should have a global tuple of all image classes. `load` and `save` should simply loop through this list, should not have any image-specific code.

Pros:

* This should improve code modularity, readability, maintainability, and avoid errors in adding new formats.
* It gives a clear interface for determining file Image type.

Cons:

* This design will have a small time cost in opening files when a file extension is ambiguous about the file type, and the file type isn't the most common for that file extension.
* It gives more structure to determining image type, which could be detrimental in the future.

Thoughts?

effigies:
Not writing image classes, myself, I can only really give an aesthetic opinion. I think having each image type directly perform the work of establishing whether a file is that type is much neater than a global function that figures out the type.

To go further, perhaps `is_image` should instead be a function that returns a class or `None`? e.g.

    Minc1Image.test_image('test.nii') == None
    Minc1Image.test_image('minc1.mnc') == Minc1Image
    Minc1Image.test_image('minc2.mnc') == Minc2Image

Since `Minc2Image` is a subclass of `Minc1Image`, the `test_image` code would only need to be written once and only entered once regardless of which class is tested first. Though if it were to go that far, why not a `try_image` function that returns either a loaded image or `None`?

I think those options would address con one. Regarding the second, is it really overly structured to give a class the ability to inspect a file to determine if it is that type? (Admittedly I'm thinking of this in terms of `load`, not `save`.

Finally, a looping master tuple seems a little crude when you've got a string that should be able to index a dictionary. Why not a global type map from extensions to images that use that extension? Either classes add themselves to it as they are loaded such as with a decorator:

    @valid_exts('.ex1', '.ex')
    class Ex1Image

[...]

PR #319, opened 2015-06-25
WIP: loadsave refactoring

effigies:
A first pass at #317, because apparently I don't have real work to do.

This so far only addresses the load half of load/save, but puts into practice some of the ideas we talked about, e.g. the decorator. (At first I tried just using `klass.files_types`, but that broke `.mgz`.)

One thing this demonstrates is that the checks in `guessed_image_type` were contextual; moving these checks into the `klass.is_image` files doesn't produce standalone functions that you'd probably expect.

A possibility is to make `@valid_exts` more forceful and add an assertion around `is_image`.

I killed off some tests along with their functions, so those'll need replacing.


bcipolli:
@effigies - one design concern that I came across while looking at implementing this myself:

`valid_exts` is actually already defined in the `files_types` within a class. We have a master list of classes within `.imageclasses`. So, we can use that to create the `IMAGE_MAP` variable without the decorator.

I prefer that design, as it reuses the existing data rather than duplicates the extension information.

What do you think about migrating to that design?

matthew-brett:
Actually, `files_types` doesn't define all valid extensions, it only defines the 'characteristic' extensions for the files that make up the image.    That's why the MGHFormat is a bit confusing, because the characteristic extension is '.mgh', present in `files_types`, but `.mgz` is also acceptable as an extension.

effigies:
Tried that first, but the `.mgz` convinced me `valid_exts` was easier as a first pass. To use `files_types`, you'll need to account for that quirk.

Frankly not married to it, so do what you will.

On July 9, 2015 12:20:15 PM EDT, Ben Cipollini <notifications@github.com> wrote:
>@effigies - one design concern that I came across while looking at
>implementing this myself:
>
>`valid_exts` is actually already defined in the `files_types` within a
>class. We have a master list of classes within `.imageclasses`. So, we
>can use that to create the `IMAGE_MAP` variable without the decorator.
>
>I prefer that design, as it reuses the existing data rather than
>duplicates the extension information.
>
>What do you think about migrating to that design?
>
>---
>https://github.com/nipy/nibabel/pull/319#issuecomment-120056092

-- 
Chris Markiewicz

bcipolli:
Is there any reason not to enumerate all valid file types in `files_types`? "Characteristic" can just mean... the first entry of each type (image, header) in the list.

matthew-brett:
`files_types` is also used to construct list of files that comprise the image.   For example, when using the `to_filename` method of the image, we use `files_types` to generate the `img` and `hdr` filenames from the original passed filename.   If you add another entry, you'll get another file saved with the given extension, when saving an image of that type.

effigies:
One possibility is reducing valid_exts to additional extensions (so just .mgz) and constructing most of IMAGE_MAP from files_types.

The current construction is more explicit and eliminates the need for each image type to be imported into loadsave, but we could remove the redundancy which is really for just the one case.

On July 9, 2015 12:33:57 PM EDT, Matthew Brett <notifications@github.com> wrote:
>`files_types` is also used to construct list of files that comprise the
>image.   For example, when using the `to_filename` method of the image,
>we use `files_types` to generate the `img` and `hdr` filenames from the
>original passed filename.   If you add another entry, you'll get
>another file saved with the given extension, when saving an image of
>that type.
>
>---
>https://github.com/nipy/nibabel/pull/319#issuecomment-120061226

-- 
Chris Markiewicz

matthew-brett:
The special-casing for `.mgz` is horrible, it would good to find a better way.  For example, at the moment, we need a special hack in the `BinOpener` class to recognize the `.mgz` extension as being a compressed image, and this is not even in the `mghformat` module.

bcipolli:
We could just have `files_types` enumerate all, and have a function on SpatialImage that returns the "characteristic type" (finds all unique types in `files_types`, and returns the first extension for each).

Then just tweak `to_filename` to use that "characteristic" function.

>For example, at the moment, we need a special hack in the BinOpener class to recognize the .mgz extension as being a compressed image, and this is not even in the mghformat module.

I also see that compression extensions are defined in quite a few places (in `loadsave.py` and on a few of the image classes, and in the `Opener` class). Perhaps a design motivated by the `mgz` issue could apply more broadly to deal with that as well.

Do image types that have separate header files allow compressed files? Perhaps they compress the image only, eh...

I will think about this.

matthew-brett:
Image types that allow compressed files compress both header and image.

matthew-brett:
Summary of `files_types` for my own benefit.

`files_types` is a tuple of (key, value) string pairs, where 

* key is the an image file type (just 'image' or 'header' or far);
* value is the characteristic extension for the image type.

Used for:

* Characteristic extension for image type via `SpatialImage.get_filename()`;
* Sequence of filenames corresponding to a single image, given input single file name, via `SpatialImage.filespec_to_filemap`;
* Sequence of filenames or file objects corresponding to a single image, given optional mapping from file type to file-name / file object.

matthew-brett:
By the way, you may know that '.mgz' files can also have extension '.mgh.gz' - nibabel will open either: 

https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat

bcipolli:
>at the moment, we need a special hack in the BinOpener class to recognize the .mgz extension as being a compressed image, and this is not even in the mghformat module.

I am looking at this now. @matthew-brett Why is `BinOpener` separate from `Opener`... and not in `openers.py`?

Also, any concerns if we (secretly) allow reading of `.mgz.gz` extensions, rather than explicitly rejecting it?

matthew-brett:
It's OK to read 'mgz.gz' - if a little weird.

BinOpener, as you see, is just a hack allowing '.mgz' as an extension indicating that the file is zip compressed - so it's in volumeutils only because the hack is not intrinsic to openers, but to images.  It should probably go in the freesurfer modules, now I think about it (and the other classes use Opener)

bcipolli:
OK that's good to hear. I agree, and I think not too much work will clean things up nicely. I'll try it out now.

nibabel/minc1.py
```diff
@@ -307,6 +311,26 @@ def from_file_map(klass, file_map):
...
+    def is_image(klass, filename, sniff=None):
+        ftypes = dict(klass.files_types)
+        froot, ext, trailing = splitext_addext(filename, klass._compressed_exts)
+        lext = ext.lower()
+
+        if lext not in ftypes.values():
+            return False, sniff
+
+        fname = froot + ftypes['header'] if 'header' in ftypes else filename
+        if not sniff:
```
bcipolli: Should also check the size of the sniff

nibabel/minc1.py
```diff
@@ -307,6 +311,26 @@ def from_file_map(klass, file_map):
...
+            with BinOpener(fname, 'rb') as fobj:
+                sniff = fobj.read(4)
+
+        return klass._minctest(sniff), sniff
```
bcipolli: Is it not possible to do `header_class.is_header` here? This would eliminate the need for overriding the `SpatialImage` version of this function.

nibabel/spatialimages.py
```diff
@@ -866,6 +867,22 @@ def from_image(klass, img):
                      klass.header_class.from_header(img.header),
                      extra=img.extra.copy())
 
+    @classmethod
+    def is_image(klass, filename, sniff=None):
+        ftypes = dict(klass.files_types)
+        froot, ext, trailing = splitext_addext(filename, ('.gz', '.bz2'))
```
bcipolli: Use `klass._compressed_ext` instead of hard-coded gz/bz2?

nibabel/spatialimages.py
```diff
@@ -866,6 +867,22 @@ def from_image(klass, img):
...
+        lext = ext.lower()
+
+        if lext not in ftypes.values():
+            return False, sniff
+
+        fname = froot + ftypes['header'] if 'header' in ftypes else filename
+        if not sniff:
+            with BinOpener(fname, 'rb') as fobj:
+                sniff = fobj.read(1024)
```
bcipolli: Perhaps allow header classes to define a `sniff_size`. In that case, only sniff if `header_class.sniff_size` is defined, `header_class.is_header` is defined. 

That would also simplify the `loadsave` logic (no special-casing for len(possible_classes) == 0) and avoid a potential problem if `header_class.is_header` is not defined.

nibabel/nifti2.py
```diff
@@ -221,6 +222,17 @@ def _chk_eol_check(hdr, fix=False):
             rep.fix_msg = 'setting EOL check to 13, 10, 26, 10'
         return hdr, rep
 
+    @classmethod
+    def is_header(klass, binaryblock):
+        if len(binaryblock) < 540:
```
bcipolli: I don't believe this is the correct logic; the sniff size doesn't have to be 540; the sniffed info has to indicate that the header size is 540 exactly.

bcipolli:
@effigies I went through this today, seeing how things work if taking out the `IMAGE_MAP` and instead looping over the image classes / `files_types` (shortens the code, and avoids duplicating extension associations). 

Found some things I left comments on here; please let me know if you think I'm off on any of the comments. I plan to try and work through those, and the other changes discussed, so no need to change them yourself. But ideas / feedback welcome!

effigies:
Glad to see you say that. Your comments made me worry you were treating this as the canonical PR for #317, while I've been operating under the assumption you have your own fork going on the issue. I've done a little more work trying to make `is_image()` stand on its own as a function, but that's not really in a great place.

I'll have a look and make comments.

nibabel/nifti2.py
effigies: Nifti2 `.hdr` files are 540 bytes. Analyze/Nifti1 `.hdr` files are 348 bytes. If we end up testing a Nifti1 header with Nifti2's  `is_header`, we either need to test with `hdr = np.ndarray(shape=(), dtype=nifti1.header_dtype, buffer=binaryblock[:348])` or we need to assert that there be at least 540 bytes.

In the old style, `load` called `which_analyze_type`, which used `nifti1.header_dtype`.

nibabel/spatialimages.py
effigies: I'm guessing I made this decision when copying `guessed_image_type` logic in here. I think `klass._compressed_ext` would be fine.

nibabel/minc1.py
effigies: The reason this was an issue is because minc1 and minc2 both use the `MincHeader`, which then needs to give a different response based on whether a `Minc1Image` or `Minc2Image` called it.

There's probably a better way to do it.

nibabel/spatialimages.py
effigies: What I wanted to avoid was grabbing data from the filesystem and then finding out we need more. Hard-coding was easiest, and 1kb is small enough to avoid burdening even a slow network connection, but large enough to cover all possible headers. But if there's a cleverer way of doing it, by all means.

bcipolli:
Thanks @effigies . I have a branch that should be pretty close today (though beefing up the tests would be good). Do you prefer I do a PR against your branch, or submit an independent PR here? The design is different enough that I think doing an independent PR makes sense, but thought I'd check with you since you did significant work on it too!

In case it helps:
https://github.com/nipy/nibabel/compare/master...bcipolli:reload?expand=1

effigies:
A new PR is fine with me.

PR #325, opened 2015-07-08
BUG: work round int overflow in size calculation

matthew-brett:
Omar pointed out the integer overflow in calculating the image size of an
image of size > 4GB, on a 32-bit platform.

### DIFF
--- a/nibabel/volumeutils.py
+++ b/nibabel/volumeutils.py
@@ -14,6 +14,8 @@ import warnings
 import gzip
 import bz2
 from os.path import exists, splitext
+from operator import mul
+from functools import reduce
 
 import numpy as np
 
@@ -504,7 +506,8 @@ def array_from_file(shape, in_dtype, infile, offset=0, order='F', mmap=True):
             pass
     if len(shape) == 0:
         return np.array([])
-    n_bytes = int(np.prod(shape) * in_dtype.itemsize)
+    # Use reduce and mul to work around numpy integer overflow
+    n_bytes = reduce(mul, shape) * in_dtype.itemsize
     if n_bytes == 0:
         return np.array([])
     # Read data from file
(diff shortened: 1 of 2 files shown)


nibabel/volumeutils.py
```diff
@@ -504,7 +507,8 @@ def array_from_file(shape, in_dtype, infile, offset=0, order='F', mmap=True):
             pass
     if len(shape) == 0:
         return np.array([])
-    n_bytes = int(np.prod(shape) * in_dtype.itemsize)
+    # Use reduce and mul to work round integer overflow on 32-bit platforms
```
Eric89GXL: s/round/around/g

nibabel/volumeutils.py
```diff
@@ -504,7 +507,8 @@ def array_from_file(shape, in_dtype, infile, offset=0, order='F', mmap=True):
             pass
     if len(shape) == 0:
         return np.array([])
-    n_bytes = int(np.prod(shape) * in_dtype.itemsize)
```
effigies: You could also specify the `dtype` in `np.prod`, rather than introduce `operator.mul`:

    n_bytes = int(np.prod(shape, dtype=np.int64) * in_dtype.itemsize)

matthew-brett: The advantage of using `operator.mul` is that is uses Python integers internally, and so completely prevents overflow for any possible value of the shape.  In fact the test tests overflow for values above the int64 range, although of course you cannot make an array that large.

effigies: Fair enough. Might change the comment to reflect that it's not just avoiding int32 overflows, but that's really nit-picky.

matthew-brett:
OK to merge?

effigies:
Not sure who are valid signers-off, but it looks reasonable to me.

Eric89GXL:
Thanks @matthew-brett for the fix

PR #326, opened 2015-07-12
Use "optional_package" for scipy checks

bcipolli:
In a few places, scipy was still checked for in `try.. catch` blocks, rather than a call to `optional_package`. Switch those over.

Made another related changes as well, just because I was there:
* Turn `nibabel/tests/data/check_parrec_reslice.py` into an importable module, and fail more gracefully if `scipy` is not installed.

### DIFF
--- a/nibabel/tests/test_helpers.py
+++ b/nibabel/tests/test_helpers.py
@@ -9,7 +9,7 @@ from numpy.testing import assert_array_equal
 from ..openers import Opener
 from ..optpkg import optional_package
 from ..tmpdirs import InTemporaryDirectory
-have_scipy = optional_package('scipy')[1]
+_, have_scipy, _ = optional_package('scipy')
 
 
 def bytesio_filemap(klass):
--- a/nibabel/tests/test_helpers.py
+++ b/nibabel/tests/test_helpers.py
@@ -3,13 +3,14 @@
 from io import BytesIO
 
 import numpy as np
-from nose.tools import assert_true
-from numpy.testing import assert_array_equal
 
 from ..openers import Opener
-from ..optpkg import optional_package
 from ..tmpdirs import InTemporaryDirectory
-_, have_scipy, _ = optional_package('scipy')
+from ..optpkg import optional_package
+_, have_scipy, _ = optional_package('scipy.io')
+
+from nose.tools import assert_true
+from numpy.testing import assert_array_equal
 
 
 def bytesio_filemap(klass):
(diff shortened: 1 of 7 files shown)


nibabel/tests/test_helpers.py
```diff
@@ -42,15 +41,12 @@ def bz2_mio_error():
...
-        import scipy.io
-    except ImportError:
-        return True
     with InTemporaryDirectory():
         with Opener('test.mat.bz2', 'wb') as fobj:
             try:
+                import scipy.io
                 scipy.io.savemat(fobj, {'a': 1}, format='4')
-            except ValueError:
+            except ImportError, ValueError:
```
effigies: except (ImportError, ValueError):

nibabel/tests/test_helpers.py
```diff
@@ -1,15 +1,14 @@
 """ Helper functions for tests
 """
+from io import BytesIO
 
 import numpy as np
+from nose.tools import assert_true
```
matthew-brett: I personally prefer test imports after other imports.

nibabel/tests/test_helpers.py
```diff
@@ -42,15 +41,12 @@ def bz2_mio_error():
...
     commit 98ef522d99 (in scipy)
     """
-    try:
-        import scipy.io
-    except ImportError:
-        return True
     with InTemporaryDirectory():
         with Opener('test.mat.bz2', 'wb') as fobj:
             try:
+                import scipy.io
```
matthew-brett: On principle of isolating the line causing the error in the try block:

```
try:
    import scipy.io
except ImportError:
    return True
try:
    scipy.io.savemat(fobj, {'a': 1}, format='4')
except ValueError:
    return True
```

matthew-brett:
Otherwise, looks good to me, thanks Ben.

nibabel/tests/test_helpers.py
bcipolli: OK. In this case, I would just do the `optionalpackage` check. I thought the combined `try...catch` was more compact and easy to read, but glad to change.

nibabel/tests/test_helpers.py
bcipolli: OK, glad to follow. Just to note, I was trying to follow the PEP8 convention for grouping imports into three: https://www.python.org/dev/peps/pep-0008/#imports . I'm not arguing for following that convention here, just to share where I was coming from.

bcipolli:
Thanks @matthew-brett ; just made the requested changes (using `optional_package` instead of `try...catch` for scipy import check)

matthew-brett:
Thanks for going through those changes.

Some travis test errors...

bcipolli:
Fixed up & passing.

matthew-brett:
Thanks for your patience ...

bcipolli:
You too. Plus, this was way easier than last time! :)

Issue #323, opened 2015-07-07
Push .imageclass.class_map logic into Image classes


bcipolli:
Adding new images is easier if the information about the image is encapsulated by the Image class. Even better if the `SpatialImage` super class can define a common interface.

Currently, a number of properties about an image are defined in the `imageclasses.py:class_map` variable, including (pulling from `imageclasses.py`): 

```python
                'ext': '.img', # characteristic image extension
                'has_affine': False, # class can store an affine
                'makeable': True, # empty image can be easily made in memory
                'rw': True}, # image can be written
```

I suggest that these all get defined as properties on `SpatialImage`, and overridden with appropriate values by each subclass. Note that we're already discussing pushing the extension metadata to the image class in #317. 

If we think `class_map` (and `ext_map`) are being used by external libraries, we can simply build these maps with a simple dictionary comprehension (or list comprehension if Python 2.6 compatibility matters).

effigies:
+1

`class_map` and `ext_map` don't seem particularly necessary once the logic of load/save is pushed into the classes. Since neither of these shows up in [the documentation](http://nipy.org/nibabel/reference/nibabel.imageclasses.html), I lean toward removal rather than reconstruction.

matthew-brett:
Yes, sure, it would be better to store the information in the classes if we can.  I would prefer to reconstruct and deprecate, especially if we want to put this in before the next major release (I mean 3.0).

bcipolli:
keys to `class_map` seemed to indicate a 1-to-1 mapping between key and class and class to extension, but that's not the case--`MGHImage` maps to both `mgh` and `mgz`. Thus, the keys to `class_map` do not indicate image classes, but labels to file extensions.

To keep those keys somewhere, I need to stamp them on the actual class. But it's not a class property--it's a property per extension. Those extensions are defined within the `klass.files_types` tuple.. which I assume adding a third value to the tuple could be problematic.

It also seems silly to add a new property that duplicates the extensions, just to assign a key/nickname to each.

A bit stuck on the best way to go. *almost* everything else was easy...

bcipolli:
Another thing I don't understand: for the `PARREC` format, the class defines:
```
    files_types = (('image', '.rec'), ('header', '.par'))
```
This seems to indicate that extension `.rec` is the primary one (it is both the first in the tuple, and associated with the 'image' keyword).

However, `class_map` and `ext_map` associate this class with the .`par` extension. I'm clearly confused about the intended design here.

[...]

Issue #316, opened 2015-06-13
ENH: Add GIFTI to the list of known imageclasses to ease loading with .load()


bcipolli:
GIFTI and CIFTI are going to be more important as more people use the Human Connectome Project (HCP) data (and as those data continue to evolve).

It would be great if GIFTI were integrated into the load/save framework. It was mentioned in 
https://github.com/nipy/nibabel/pull/95 that this be done.

I have time today at the OHBM hackathon to do this; @matthew-brett what do you think? I'm new to GIFTI, so I may not be realizing some issue with doing this.

This may also make some PySurfer changes I am piloting easier; see https://github.com/nipy/PySurfer/issues/134

matthew-brett:
I don't know GIFTI either I'm afraid, and as you may have seen, that code hasn't had much care since Stefan Gerhard stopped working on it.

That makes me a bad person to design the API for this stuff.    I will do my best to give sensible comments though.

Would you be interested in taking charge of the GIFTI / CIFTI part of nibabel?

bcipolli:
@matthew-brett Thanks; let me talk with @satra today about GIFTI/CIFTI formats. When I have a better idea of what they are and how we're going to work with them, I'd be glad to consider it.

In the meantime, I'll put something together for the OHBM hackathon, and we can figure out afterwards if it's in the right direction.

Thanks as always!

bcipolli:
@matthew-brett I think I can manage this. Here's what I'm thinking:
* For CIFTI: satra suggested at #OHBM2015 that we implement the API exposed by the C++ library here: https://github.com/Washington-University/workbench/tree/master/src/Cifti
* For CIFTI and GIFTI: integrate into the load/save pathway as cleanly as possible (hints here: https://github.com/nipy/nibabel/pull/249#discussion_r32465795 )
* For CIFTI and GIFTI: alias some common metadata values that are quite embedded in the header, to be properties directly off of the Image object (need to determine good ones)

matthew-brett:
Thanks - that would be a big contribution.

Maybe we could kick off with a little sketch of what the Python API would look like, with some examples of (proposed) use?

Are there a set of standard CIFTI files we could test against?  

http://nipy.org/nibabel/devel/add_test_data.html

They could maybe go in a data repository if the license is nasty (I think it is for the HCP data):

http://nipy.org/nibabel/devel/add_test_data.html#adding-as-a-submodule-to-nibabel-data

[...]

PR #329, opened 2015-07-20
WIP: Migrate load, save, class_map, ext_map to class properties and methods

bcipolli:
This is to address #317 and #323 (and aims to supercede PR #319), both issues around image-specific code being distributed across files, rather than localized in the class definition.

Changes here include:
* Removing image-specific code from `nib.load` and `nib.save` functions, moving the logic into `ImageKlass.is_image` and `HeaderKlass.is_header` class functions.
* Deprecating `class_map` and `ext_map` objects, and moving necessary info to properties defined on the image klasses themselves.

Things needing to be done in this PR:
* Migrate some of the old tests that were deleted into similar tests in the new coding

Things that need to be done in a future PR:
* The inclusion of `.mgz` in `files_types` as an `'image'` is a bit funky, but will take thought and work to do cleanly. I have an idea of what to do, but would prefer to do a subsequent PR on it as this one is already quite sizable.


nibabel/spatialimages.py
```diff
@@ -867,6 +872,51 @@ def from_image(klass, img):
...
+        return klass.is_valid_extension(ext)
+
+    @classmethod
+    def is_image(klass, filename, sniff=None):
+        froot, ext, trailing = splitext_addext(filename, klass._compressed_exts)
+
+        if not klass.is_valid_extension(ext):
+            return False, sniff
+        elif (getattr(klass.header_class, 'sniff_size') is None or
+              getattr(klass.header_class, 'is_header') is None):
```
bcipolli: Check if `None` rather than do a `hasattr`, so that properties can be either unset, or reset to `None` (which is easier when subclassing).

nibabel/tests/test_image_load_save.py
```diff
@@ -265,62 +263,3 @@ def test_filename_save():
             del rt_img
         finally:
             shutil.rmtree(pth)
-
-
-def test_analyze_detection():
```
bcipolli: These tests (and the rest of the removed ones in this file) need to get similar tests for the current code structure.

effigies:
Small cleanups in ab2081b. Given that all `is_header` methods except Minc do their own slicing, it makes sense to me to pass `sniff` directly and slice `binaryblock` in the Minc headers as well. Also `_minctest` seems to be deprecated, so I dropped it.

nibabel/loadsave.py
```diff
@@ -106,25 +58,38 @@ def save(img, filename):
...
-        img_type = ext_map[ext]
-        klass = class_map[img_type]['class']
+    else:  # arbitrary conversion
+        valid_klasses = filter(lambda klass: klass.is_valid_extension(ext),
+                               all_image_classes)
+        try:
+            klass = next(iter(valid_klasses))
+        except StopIteration:  # if iterator is empty
+            raise ImageFileError('Cannot work out file type of "%s"' %
+                                 filename)
```
effigies: The `next(iter(filter(...)))` is awkward. Alternatives:

    valid_klasses = list(filter(lambda klass: klass.is_valid_extension(ext), all_image_classes))
    valid_klasses = [klass for klass in all_image_classes if klass.is_valid_extension(ext)]

matthew-brett:
@effigies - just to check that you are happy for this PR to supersede your #319 ?

effigies:
Yes. I used that branch to make a suggested commit, but no need to have a PR for it.

nibabel/loadsave.py
bcipolli: I'll go for the second, since we want a list anyway...

nibabel/loadsave.py
```diff
@@ -11,17 +11,10 @@
...
-from .openers import ImageOpener
-from .analyze import AnalyzeImage
-from .spm2analyze import Spm2AnalyzeImage
-from .nifti1 import Nifti1Image, Nifti1Pair, header_dtype as ni1_hdr_dtype
-from .nifti2 import Nifti2Image, Nifti2Pair
-from .minc1 import Minc1Image
-from .minc2 import Minc2Image
-from .freesurfer import MGHImage
+from .filename_parser import splitext_addext
+from .volumeutils import BinOpener
```
effigies: Should be:

    from .openers import ImageOpener

nibabel/minc1.py
```diff
@@ -14,8 +14,10 @@
 
 from .externals.netcdf import netcdf_file
 
+from .filename_parser import splitext_addext
 from .spatialimages import Header, SpatialImage
 from .fileslice import canonical_slicers
+from .volumeutils import BinOpener
```
effigies: These two added imports from 8c40ef0 are no longer used.

nibabel/spatialimages.py
```diff
@@ -866,6 +871,51 @@ def from_image(klass, img):
...
+                for tr_ext in np.unique([trailing, ''] + list(klass._compressed_exts)):
+                    metadata_filename = froot + ext + tr_ext
+                    if os.path.exists(metadata_filename):
+                        break
+
+        try:
+            if not sniff or len(sniff) < klass.header_class.sniff_size:
+                # 1024 == large size, for efficiency (could iterate over imageclasses).
+                sniff_size = np.max([1024, klass.header_class.sniff_size])
+                with BinOpener(metadata_filename, 'rb') as fobj:
```
effigies: I think this can be switched to `ImageOpener`?

bcipolli:
Thanks @effigies -- I can see that I needed to be more careful about integrating my other PR into this one. I'll use your notes and go through things a bit myself to make sure!

bcipolli:
Getting there. Added back some tests. Want to clean things up a bit further (e.g. two of the properties I put on the individual classes are really test-only, and should live in the associated tests). I also need to review the test coverage, to see why it's dropped so significantly.

effigies:
@bcipolli A couple small fixes in a863949. Mostly the same as earlier, but rebased for convenience, and changed the MINC1 test to reduce the space of false positives to other NetCDF images.

Relatedly, I think we may want to add documentation to `is_header` indicating the level of confidence one should have in the result, in particular that one should not assume that a `True` result implies that a header may be successfully read.

bcipolli:
@effigies That's great--thanks for the tweaks!

An easier workflow for me would be if you make PRs directly against my branch, and use your comments as the PR message. Then I could just iterate with you there and merge, rather than cherry picking manually (and potentially losing track of things).

Would that work for you?

effigies:
Sure, that works. Was assuming we'd want to keep conversation in one place, but I'm fine with anything.

bcipolli:
OK, got some updates from @effigies ; I think this is in a relatively good state now. With this code, I believe getting GIFTI into the load/save pipeline (see #316) should be cleaner. 

@effigies anything further you'd like to see happen here?

effigies:
I'm pretty happy with this. To reiterate this issue with `is_header`, though, the name suggests that it will return True if and only if the binary string *will* successfully load as a header, and we might want to indicate that it's a heuristic.

Also a couple style issues I'll PR on your branch in a minute.

nibabel/tests/test_image_types.py
```diff
@@ -0,0 +1,181 @@
...
+
+        sizeof_hdr = getattr(expected_img_klass.header_class, 'sizeof_hdr', 0)
+
+        for sniff_mode, sniff in dict(
+                vanilla=None,  # use the sniff of the previous item
+                no_sniff=None,  # Don't pass a sniff
+                none=None,  # pass None as the sniff, should query in fn
+                empty='',  # pass an empty sniff, should query in fn
+                irrelevant='a' * (sizeof_hdr - 1),  # A too-small sniff, query
+                bad_sniff='a' * sizeof_hdr).items():  # Bad sniff, should fail.
```
effigies: Shouldn't these be `bytes` objects (b'a')? They'll either fail or trigger a new sniff anyway, so maybe not a big deal, but if you want them to fail for the right reasons on Python 3...

bcipolli: Just to note, @effigies added a change for this (and another test issue) that I've merged in.

effigies:
@matthew-brett How do you feel about tests on deprecation warnings? Seem like they might be excessive, but I think that's part of the test coverage decrease.

nibabel/analyze.py
```diff
@@ -892,6 +892,17 @@ def _chk_pixdims(hdr, fix=False):
             rep.fix_msg = ' and '.join(fmsgs)
         return hdr, rep
 
+    @classmethod
+    def is_header(klass, binaryblock):
```
matthew-brett: How about `may_contain_header` ?

nibabel/analyze.py
```diff
@@ -892,6 +892,17 @@ def _chk_pixdims(hdr, fix=False):
...
+        if len(binaryblock) < klass.sizeof_hdr:
+            raise ValueError('Must pass a binary block >= %d bytes' %
+                             klass.sizeof_hdr)
+
+        hdr = np.ndarray(shape=(), dtype=header_dtype,
```
matthew-brett: Rename `hdr` to `hdr_struct` to indicate it's not a header object at this point?

nibabel/loadsave.py
```diff
@@ -40,55 +33,15 @@ def load(filename, **kwargs):
...
     '''
-    return guessed_image_type(filename).from_filename(filename, **kwargs)
 
+    sniff = None
+    for image_klass in all_image_classes:
+        is_valid, sniff = image_klass.is_image(filename, sniff)
+        if is_valid:
+            return image_klass.from_filename(filename, **kwargs)
 
-def guessed_image_type(filename):
```
matthew-brett: Can we leave a stub here?  I guess someone may have used this function, given it's got a public API name.

effigies: Do you think a deprecation warning or return an image class like before, but using the new mechanism? (Or both?)

matthew-brett: Both, I think.  Or do you think someone would want this function independent of the `load` mechanism?

effigies: Doesn't seem obvious that they would. If you say deprecate, that works for me.

matthew-brett: Could just be a stub that tries `load` and return the class of the result.

effigies: `load` is only 7 lines. My inclination is to just replicate it without the `.from_filename` bit. If we weren't deprecating, I'd have `load` call it like before, but might as well make the process of removing it easy.

effigies:
@matthew-brett Incorporated your suggestions. Let me know if my approach to `guessed_image_type` doesn't work.

Looks like we're back up to positive test coverage, so that's good.

nibabel/loadsave.py
```diff
@@ -212,43 +203,3 @@ def read_img_data(img, prefer='scaled'):
         if prefer == 'scaled':
             return hdr.data_from_fileobj(fileobj)
         return hdr.raw_data_from_fileobj(fileobj)
-
-
-def which_analyze_type(binaryblock):
```
matthew-brett: Sorry man - I know this is anal - but this one's part of the API too, technically.

nibabel/nifti2.py
```diff
@@ -221,6 +221,17 @@ def _chk_eol_check(hdr, fix=False):
             rep.fix_msg = 'setting EOL check to 13, 10, 26, 10'
         return hdr, rep
 
+    @classmethod
+    def may_contain_header(klass, binaryblock):
```
matthew-brett: Maybe have `_sizeof_hdr_val` as class attribute, and share this code between Nifti1 and Nifti2?

nibabel/spatialimages.py
```diff
@@ -866,6 +872,52 @@ def from_image(klass, img):
                      klass.header_class.from_header(img.header),
                      extra=img.extra.copy())
 
+    @classmethod
+    def is_valid_extension(klass, ext):
+        return np.any([ft[1] == ext.lower() for ft in klass.files_types])
+
+    @classmethod
+    def is_image(klass, filename, sniff=None):
```
matthew-brett: Consider `path_maybe_image` or similar?   I'd expect an image object as input to an `is_image` test method.

nibabel/tests/test_image_types.py
```diff
@@ -0,0 +1,186 @@
...
+                MGHImage, all_image_classes)
+
+from nose.tools import assert_true, assert_equal, assert_raises
+
+DATA_PATH = pjoin(dirname(__file__), 'data')
+
+
+def test_analyze_detection():
+    # Test detection of Analyze, Nifti1 and Nifti2
+    # Algorithm is as described in loadsave:which_analyze_type
```
matthew-brett: `which_analyze_type` now deleted no?

nibabel/tests/test_image_types.py
```diff
@@ -0,0 +1,186 @@
...
+            try:
+                if klass.may_contain_header(hdr.binaryblock):
+                    return klass
+                else:
+                    print('checked completed, but failed.')
+            except ValueError as ve:
+                print(ve)
+                continue
+        return None
+        # return nils.which_analyze_type(hdr.binaryblock)
```
matthew-brett: Remove comment

nibabel/tests/test_image_types.py
```diff
@@ -0,0 +1,186 @@
...
+
+    n1_hdr['magic'] = 'n+1'
+    assert_equal(wat(n1_hdr), Nifti1Header)
+
+    n1_hdr['magic'] = 'ni1'
+    assert_equal(wat(n1_hdr), Nifti1Header)
+
+
+def test_sniff_and_guessed_image_type(img_klasses=all_image_classes):
+    """
```
matthew-brett: I prefer comments at the top of test functions, otherwise nosetest prints out the docstring first line, making it difficult to see which test is running.  Maybe one-line comment first, then docstring?

nibabel/freesurfer/mghformat.py
```diff
@@ -459,9 +459,13 @@ class MGHImage(SpatialImage):
     """ Class for MGH format image
     """
     header_class = MGHHeader
-    files_types = (('image', '.mgh'),)
+    files_types = (('image', '.mgh'),
+                   ('image', '.mgz'))
```
matthew-brett: Is this second line necessary?  It may cause problems with the `make_file_map` function.

matthew-brett: Ah - maybe this relates to line 134 of `filename_parser.py`.  Ouch. 

The problem here is that I intended `files_types` to be a list of the types of file that make up a single image.  Here you are re-using it to be a list of types of extensions to try when loading a file.   Did I understand right?

nibabel/nifti2.py
effigies: These actually aren't the same test with different numbers, so reusing won't work. Nifti1 returns `hdr_struct['magic'] in (b'ni1', b'n+1')`.

nibabel/tests/test_image_types.py
effigies: I think this function was added in hopes of getting test coverage back up, but I don't think it explores paths that the other tests don't, so I'm for dropping it. That said, I don't want to abuse write access to @bcipolli's repo to remove code he added before he has a chance to defend it, so I'll leave that discussion there.

nibabel/spatialimages.py
effigies: So you'd use this like:

    Nifti1Image.XXXX(path)

Maybe `file_is_type`, or `file_is_image` or something?

matthew-brett: Yes, reasonable names.  Do we want to convey the 'maybe' aspect as we did for the header?

nibabel/tests/test_image_types.py
bcipolli: Agreed. drop it.

nibabel/spatialimages.py
effigies: Fair point. `path_maybe_image` or `file_maybe_image` makes sense to me.

nibabel/freesurfer/mghformat.py
effigies: That looks to me like how it's working. I feel like a conversation was had on this subject a while back, but I'm guessing it got lost from the PR after a rebase. I believe this allowed us to avoid a decorator or new class variable just to accommodate `.mgz`, but perhaps there is a fairly easy solution:

Add a default `SpatialImage` class variable `alternate_extensions = ()`, update it in `MGHImage` (possibly through the `ImageOpener.register_ext_from_image` decorator) and rewrite `SpatialImage.is_valid_extensions`:

    @classmethod
    def is_valid_extension(klass, ext):
        valid_exts = tuple(ft[1] for ft in klass.files_types) + klass.alternate_extensions
        return ext.lower() in valid_exts

bcipolli: If we go the `alternate_exts` route, would `Spm99AnalyzeImage`'s use of the `mat` extension fit in here?

```
    files_types = (('image', '.img'),
                   ('header', '.hdr'),
                   ('mat', '.mat'))
```

I'm not familiar with this, but I don't think this fits the definition for `files_types` that @matthew-brett outlined above. It may fit the `alternate_exts` idea?

effigies:
Intend to squash 2d15549 into 5eebce1 and ac92aea into d0f680d before this gets merged.

effigies:
@matthew-brett This is ready for your review. The coverage decrease was in 5eebce1, I think, with the addition of a new class variable. Not sure how to test that.

nibabel/loadsave.py
```diff
@@ -105,25 +83,38 @@ def save(img, filename):
...
     try:
         img.to_filename(filename)
     except ImageFileError:
         pass
     else:
         return
-    froot, ext, trailing = splitext_addext(filename, ('.gz', '.bz2'))
+
+    # Be nice to users by making common implicit conversions
+    froot, ext, trailing = splitext_addext(filename, img._compressed_exts)
```
matthew-brett: Use of private `img._compressed_exts`?  Anyway - is that a good idea?  What if we are trying to save a `.mnc` image as `.nii.gz` for example?

nibabel/loadsave.py
```diff
@@ -56,39 +58,15 @@ def guessed_image_type(filename):
...
-            binaryblock = fobj.read(348)
-        ft = which_analyze_type(binaryblock)
-        if ft == 'nifti2':
-            klass = Nifti2Pair
-        elif ft == 'nifti1':
-            klass = Nifti1Pair
-        else:
-            klass = Spm2AnalyzeImage
-    return klass
+    warnings.warn('guessed_image_type is deprecated', DeprecationWarning)
```
matthew-brett: Might want `stacklevel=2` for deprecation warning.

nibabel/minc1.py
```diff
@@ -269,6 +269,7 @@ class MincHeader(Header):
     """
     # We don't use the data layout - this just in case we do later
     data_layout = 'C'
+    sizeof_hdr = 4
```
matthew-brett: `sizeof_hdr` seems a bit confusing.  Means size of needed sniff, but not the size of the header.

matthew-brett:
Forgive me for rambling here but:

The task is to have a load command that asks each image type whether the filename can plausibly match the image type.   If that check passes, then the image type has to determine whether the contents of the metadata (header) file corresponding to the passed filename contains data compatible with the image type.   Logic is something like:

```
if not klass.is_valid_filename(fname): 
    return False, sniff
header_fname = klass.header_for_filename(fname)  # doesn't exist
if not os.path.exists(header_fname):
    return False, sniff
if sniff is None:
    with open(header_fname, 'rb') as fobj:
        sniff = fobj.read(max(self.sniff_size, MAX_SNIFF_SIZE))
if len(sniff) < self.sniff_size:
    return False, None
try:
    return klass.may_contain_header(sniff), sniff
except AttributeError:
    return True, sniff
```

I guess one smelly problem here is that the `header_fname` may not be the same for all image types in the load.  It is at the moment true that trying a load `.img` or `.hdr` will always end up with `header_fname == '.hdr'`, but it's a slightly weird contract that that must be the case.  For example, if there turns out to be another file type with extension `.img`, which doesn't have a `.hdr` header, but where this the `.hdr` file does exist for some reason, then the new `.img` image type will want to sniff the `.img` file, but will get a sniff from the `.hdr` file.

nibabel/minc1.py
bcipolli: We previously had a `sniff_size` property, but `sniff_size` was equal to `sizeof_hdr` in all cases where they were both defined.

I see three options:
* change `sizeof_hdr` in all cases to be the size of the header (if not 4 above, then increase it).
* read `sniff_size` if it exists, otherwise, look for a `sizeof_hdr`, otherwise don't do the sniff check.
* revert back to the `sniff_size`, stop reading `sizeof_hdr` at all.

Any preferred way to go?

bcipolli:
>I guess one smelly problem here is that the header_fname may not be the same for all image types in the load. 
>```header_fname = klass.header_for_filename(fname)  # doesn't exist```

I believe this check is done in `path_maybe_image`, for each image class:
https://github.com/bcipolli/nibabel/blob/reload/nibabel/spatialimages.py#L892
```python
        header_exts = [ft[1] for ft in klass.files_types if ft[0] == 'header']
        if len(header_exts) == 0:
            metadata_filename = filename
        else:
            # Search for an acceptable existing header;
            #   could be compressed or not...
            for ext in header_exts:
                for tr_ext in np.unique([trailing, ''] +
                                        list(klass._compressed_exts)):
                    metadata_filename = froot + ext + tr_ext
                    if os.path.exists(metadata_filename):
                        break
```

It's just not encapsulated in a separate function. The code would be the same for all classes, it could be defined on `SpatialImage` if desired.

Do you prefer that that code be moved into a separate class method?

nibabel/loadsave.py
effigies: Actually, does it make sense to have these at all? At least on my system, DeprecationWarnings are silently ignored, so they won't be raised. If they won't be raised, the `np.deprecate` is a better alert.

I'm not sure it's good manners to change the filters in a library like nibabel, but an alternative is to change the ignore rule for DeprecationWarnings.

nibabel/loadsave.py
effigies: Yeah, I think this was a mistake that assumed there wouldn't be a conversion. We'll just have to update this list with any new compression extensions that get used, which seems unlikely enough.

nibabel/minc1.py
effigies: I think I'll vote for reverting to `sniff_size`, unless there's a good reason to note the full header size.

matthew-brett:
Sorry for going quiet on you guys, I got a bit swamped by Python 3.5 coming out and lots of packages breaking.

I thought about it for a while, and I realized that I was still uncomfortable about the different ideas of valid extensions and typical file types getting fused.   So, I had a go at some changes on top of this branch to see if I could make these ideas more explicit.  would you mind taking a look?

https://github.com/bcipolli/nibabel/pull/6

effigies:
Merged bcipolli#6 and rebased to squash a couple commits. To summarize the outstanding issues:

1. `DeprecationWarning`: Worth raising, given that the `warnings` module ignores these by default? If so, fix the `stacklevel`. Possibly another PR to enable DeprecationWarnings in nibabel?
2. `Minc1Header` and `Minc2Header` should have `sizeof_hdr` class variable switched out for something more accurately named.

As an aside, I really don't understand Coveralls. I verified that the rebase left the code in an identical state, and we went from +0.003% to -0.01% on the push.

matthew-brett:
I think it's good to raise the deprecationwarnings just in case someone wants to enable them.   I think the deprecation warnings here are for fairly unusual cases, so probably we can get away without pushing up to a more visible form (like https://github.com/numpy/numpy/blob/master/numpy/__init__.py#L124).

How about deleting the 'sizeof_hdr' check in the `may_contain_header` checks, and let letting them return False when the byte strings are not long enough?

Coveralls can be a bit weird...

effigies:
Pushed `stacklevel` updates for all of the DeprecationWarnings introduced in this PR. bcipolli#4 explicitly added a `ValueError` for too-short `binaryblock` in those functions, so I'll wait for @bcipolli's thoughts before pushing that latter suggestion.

On the one hand, it would be nice to have a single location that gives a sniff size we can call by name. On the other hand, it is just the number 4 occurring twice each in two functions. If a more complicated check becomes needed, it won't be too hard to see what needs changing.

bcipolli:
All sounds fine to me. I appreciate you two working to make this happen! I've focused my efforts on downstream code at the moment, when I can find it!

effigies:
No problem. I'm mostly writing English right now, so an excuse to procrastinate with code is great.

@matthew-brett Didn't get rid of the `ValueError` test, mainly because I'd need to figure out how to make an exception for it in [this test](https://github.com/nipy/nibabel/blob/ecee5617af6be2a48b82fbfd04f5611d0efbcfe1/nibabel/tests/test_image_types.py#L51). Can do it if it's a hang-up.

effigies:
@matthew-brett Just a bump, in case you've got a bit of time to look this over.

matthew-brett:
Yes, sorry, thanks for the bump.

Sorry to be ignorant, but won't the `may_contain_header` function(s) still raise some sort of predictable error if the byte strings are too short, even if you remove the explicit check.

The image API has now expanded to include `path_maybe_image` - I believe.  It would be very nice to have some tests for the in `test_image_api.py`.

What do you think about the case where the sniff can't make any sense transferred from on image type to the next?   At the moment, this can't happen, but sort of by accident.    For example, if we are trying to load a `.mnc` file, then, because the previous files in the list don't have a `.mnc` extension as being valid, the file will not get sniffed.  But, let's imagine that there are two image types with extension `.foo`., `Image1` and `Image2`.  `Image1` has a `.bar` header, which, according to the `Image1.path_maybe_image` method, gets sniffed.  But, `Image1.path_maybe_image` correctly rejects the sniff as not being valid for `Image1`.  On we go to `Image2`.  This doesn't have a `.bar` header, but we still pass in the sniff from the `.bar` header, which `Image2` rejects.  But we should not have passed in the sniff from the `.bar` header, because `Image2` does not have `.bar` headers - we should have started again, and this time sniffed the `.foo` file.

Admittedly, this can't easily happen, but it does seem like an unstated assumption in the code, that it can't happen.

I was thinking of something like this, to make that more explicit

```
class InvalidateSniff(object):
    " Signals that previous sniff is no longer valid "


IMG_CLASS_LOAD_SEQ = [Nifti1Image, AnalyzeImage, InvalidateSniff, MINC1]
IMG_CLASS_SAVE_SEQ = [k for k in IMG_CLASS_LOAD_SEQ if not k is InvalidateSniff]
```

bcipolli:
@matthew-brett Thanks for thinking so carefully. I understand the issue you're wishing to solve.

I'd prefer a solution that uses the class metadata to invalidate sniffs. Two ways to do this:
1. Pass sniff as a tuple: `sniff=(filename, binaryblob)`. Then classes check where the sniff is coming from.
2. We could explicitly track the filename of the sniff in the `load` function, and automatically invalidate the sniff if it's from the wrong filename.

matthew-brett:
Passing the filename with the sniff seems a bit ugly.  How would you keep track of the filename in the load function?  I guess you'd have to return it from the `path_maybe_image` ?

effigies:
> Sorry to be ignorant, but won't the may_contain_header function(s) still raise some sort of predictable error if the byte strings are too short, even if you remove the explicit check.

No. It will return False because slicing beyond the end of a `bytes` simply returns something smaller than the slice.

    >>> b''[:4]
    b''
    >>> x = b''
    >>> x[:4] == b'ABCD'
    False

I'll think a bit about your other concern as I can.

matthew-brett:
Ah - yes - but isn't False the right answer in this case?

effigies:
It is. I think it was a desire for consistent behavior across image classes that led here. There's a commit in the stack that removed the `.mnc` exception, so re-adding it won't be a huge deal.

As to your other concern, I apparently have a couple thoughts already:

If passing the path back and forth is too ugly, what about switching to `(sniff, ext)`, and checking whether `ext` is a valid header extension? If so, use the sniff, if not, drop it and replace:

    is_valid, sniff, ext = image_klass.path_maybe_image(filename, sniff, ext)

On the off chance you have the sequence checks-foo, checks-bar, checks-foo, `sniff` could be a `{ext: binaryblock}` dictionary, so that sniffs aren't thrown away until they're known not to be needed. And we can add back path into this and make `sniff` be `{path: binaryblock}`.

I believe it's already pretty much strongly assumed that a header and image file are assumed to have the same basename, so the `ext` options seem safe until it becomes necessary to modify that assumption throughout the code.

matthew-brett:
Sorry - keep on getting swamped in nipy stuff.

Thinking about it, maybe the (sniff, filename) tuple isn't too bad.  I guess then you'd have to check whether the filename is the correct metadata filename for each image type, but that seems fine.   So it's just redfining a sniff as as (bytes, filename) pair, which seems OK.

I have a vague preference for not raising errors in the `maybe_header` checks, but it's vague, so - your call.

I'll be more attentive here now - but please feel free to remind me at short (one day) intervals if I'm slacking.

effigies:
I don't have a strong opinion here, but I would say if we're going to get rid of the exception in Minc1, then [checks like this](https://github.com/bcipolli/nibabel/blob/reload/nibabel/nifti1.py#L1616-L1618) should return `False` as well, for consistency.

matthew-brett:
Agree about the consistency of checks.

effigies:
Okay, just updated the doc string and killed the exceptions.

matthew-brett:
I think this is looking good now.

Last thing - would you mind adding the tests for the expanded Image API in ``test_image_api.py``?    Specifically, I think the image API has extended to include the `path_maybe_image` class method. 

I realize that test function is a bit hairy, so let me know if I can help - I should have time today and tomorrow.

effigies:
How about adding something like the following to `LoadImageAPI`?

    def validate_path_maybe_image(self, imaker, params):
        if hasattr(imaker, 'path_maybe_image'):
            for img_params in self.example_images:
                test, sniff = imaker.path_maybe_image(img_params['fname'])[0]
                assert test == True
                if sniff is not None:
                    assert isinstance(sniff[0], bytes)
                    assert isinstance(sniff[1], basestring)

matthew-brett:
Aren't we insisting that every image class have `path_maybe_image`?

Should be `assert_true` rather than assert, and `string_types` rather than `basestring` but otherwise, that looks about right.

effigies:
The PAR/REC test didn't have an `image_maker` attribute, so I added a separate attribute `klass`, which for the most part doubles up on `image_maker`. I didn't want to assert things to be makeable to test this method.

matthew-brett:
Ah yes, good point.

I think the PR is ready, if the tests pass.  Ben - any further comments before I merge?

bcipolli:
:+1

matthew-brett:
Thanks a lot for your patience - sorry it took so long.

PR #332, opened 2015-07-21
BF FreeSurfer nifti surfaces can have >3 dimensions

effigies:
Quick workaround for https://mail.python.org/pipermail/neuroimaging/2015-July/000132.html.

This uses the FreeSurfer hacks on any shapes beginning with (-1, 1, 1) or (27307, 1, 6), which seems fairly safe, but we can restrict down to 4-dimensional, or possibly only allow (-1, 1, 1, 1) or (27307, 1, 6, 1) in the 4D case.

I guess we should also add tests for these cases.

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -693,27 +693,23 @@ class Nifti1Header(SpmAnalyzeHeader):
 
         Notes
         -----
-        Allows for freesurfer hack for large vectors described in
-        https://github.com/nipy/nibabel/issues/100 and
-        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?spec=svn5022&r=5022#77
+        Applies freesurfer hack for large vectors described in `issue 100`_ and
+        `save_nifti.m <save77_>`_.
 
         Allows for freesurfer hack for 7th order icosahedron surface described
-        in
-        https://github.com/nipy/nibabel/issues/309
-        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/load_nifti.m?r=8776#86
-        https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?r=8776#50
+        in `issue 309`_, load_nifti.m_, and `save_nifti.m <save50_>`_.
         '''
         shape = super(Nifti1Header, self).get_data_shape()
-        # Apply freesurfer hack for vector
-        if shape == (-1, 1, 1):
+        # Apply freesurfer hack for large vectors
+        if shape[:3] == (-1, 1, 1):
             vec_len = int(self._structarr['glmin'])
             if vec_len == 0:
                 raise HeaderDataError('-1 in dim[1] but 0 in glmin; '
                                       'inconsistent freesurfer type header?')
-            return (vec_len, 1, 1)
+            return (vec_len, 1, 1) + shape[3:]
         # Apply freesurfer hack for ico7 surface
-        elif shape == (27307, 1, 6):
-            return (163842, 1, 1)
+        elif shape[:3] == (27307, 1, 6):
+            return (163842, 1, 1) + shape[3:]
         else:  # Normal case
             return shape
 
@@ -723,6 +719,12 @@ class Nifti1Header(SpmAnalyzeHeader):
         If ``ndims == len(shape)`` then we set zooms for dimensions higher than
         ``ndims`` to 1.0
 
+        Nifti1 images can have up to seven dimensions. For FreeSurfer-variant
+        Nifti surface files, the first dimension is assumed to correspond to
+        vertices/nodes on a surface, and dimensions two and three are
+        constrained to have depth of 1. Dimensions 4-7 are constrained only by
+        type bounds.
+
         Parameters
         ----------
         shape : sequence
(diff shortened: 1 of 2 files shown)


matthew-brett:
Seems reasonable.  I'm afraid we do need some tests :)

effigies:
Okay. Looking at https://code.google.com/p/fieldtrip/source/browse/trunk/external/freesurfer/save_nifti.m?r=8776#52 again, it looks like this MATLAB code at least only ever works with 4D volumes, but does allow the 4th dimension to vary. The tests I uploaded permit up to 7 dimensions, which is the max of `hdr['dim']` capabilities, and verify the first three dimensions are either (27307, 1, 6) or (-1, 1, 1).

bcipolli:
Looks reasonable to me as well. I would, however, add some comments to the `get` and `set` shape functions about the current limits of shape.

effigies:
How about adding the following to `set_data_shape`?

        # For Nifti surface files, the first dimension is assumed to correspond
        # to vertices/nodes on a surface, and dimensions two and three are
        # assumed to have depth of 1. Dimensions 4-7 are not subject to
        # constraints beyond type bounds.

Or should I describe the hacks? I think the code in `set_data_shape` is about as readable as any comments I could add. I'm not sure what to add to `get_data_shape` that adds any clarity, except maybe `(see set_data_shape for details)`.

bcipolli:
@effigies that's just what I was looking for. I would put that on `set_data_shape` in the docstring, or in the `Nifti1Image` docstring if it's more widely applicable.

I found the code to be very readable, I doubt inline comments could help.

effigies:
Added a variation on that comment. One possible quibble is that surface files that have <65536 nodes will not trigger constraints on dimensions 2 and 3, since they won't hit either of the large surface hacks. I'm not sure this is a problem, but I'd be open to suggestions of rewording.

bcipolli:
@effigies I appreciate the note, and I'll follow up on this (to see if it is an issue) as I get closer to working on CIFTI/GIFTI stuff.

nibabel/nifti1.py
```diff
@@ -723,6 +723,11 @@ def set_data_shape(self, shape):
         If ``ndims == len(shape)`` then we set zooms for dimensions higher than
         ``ndims`` to 1.0
 
+        Nifti1 images can have up to seven dimensions. For Nifti surface files,
```
matthew-brett: When you say "Nifti surface files" - do you mean that this is part of the NIfTI standard somewhere?  Or you just mean NIfTI as creatively reinterpreted by Freesurfer?

bcipolli: It's my understanding that CIFTI, which contains both surface and volume data, follows the NIFTI2 spec.

nibabel/tests/test_nifti1.py
```diff
@@ -261,12 +261,20 @@ def test_freesurfer_large_vector_hack(self):
         hdr.set_data_shape((too_big-1, 1, 1))
         assert_equal(hdr.get_data_shape(), (too_big-1, 1, 1))
         # The freesurfer case
+        full_shape = (too_big, 1, 1, 1, 1, 1, 1)
+        for dim in range(3, 8):
+            expected_dim = np.array([dim, -1, 1, 1, 1, 1, 1, 1])
```
matthew-brett: Maybe add comment like:

```
# First element in 'dim' field is number of dimensions
```

I got a bit confused reading this first time.

nibabel/tests/test_nifti1.py
```diff
@@ -261,12 +261,20 @@ def test_freesurfer_large_vector_hack(self):
         hdr.set_data_shape((too_big-1, 1, 1))
         assert_equal(hdr.get_data_shape(), (too_big-1, 1, 1))
         # The freesurfer case
```
matthew-brett: Maybe comment:

```
# First Freesurfer large vector hack, where large first dim goes into the glmax field
```

nibabel/nifti1.py
matthew-brett: Maybe add something to that effect here?

effigies: So, there are Nifti1 surface files: http://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h:

    #define NIFTI_INTENT_VECTOR    1007   /* for any other type of vector */

     /*! To signify that the vector value at each voxel is really a
         spatial coordinate (e.g., the vertices or nodes of a surface mesh):
           - dataset must have a 5th dimension
           - intent_code must be NIFTI_INTENT_POINTSET
           - dim[0] = 5
           - dim[1] = number of points
           - dim[2] = dim[3] = dim[4] = 1
           - dim[5] must be the dimensionality of space (e.g., 3 => 3D space).
           - intent_name may describe the object these points come from
             (e.g., "pial", "gray/white" , "EEG", "MEG").                   */

Annoyingly, this isn't how FreeSurfer actually implements its surfaces them, presumably because of the need for large vector hacks.

matthew-brett: Excellent precision of the thing - could these go in the docstring, for posterity?

nibabel/tests/test_nifti1.py
```diff
@@ -261,12 +261,20 @@ def test_freesurfer_large_vector_hack(self):
...
-        assert_equal(hdr.get_data_shape(), (too_big, 1, 1))
-        assert_array_equal(hdr['dim'][:4], [3, -1, 1, 1])
-        assert_equal(hdr['glmin'], too_big)
-        # This only works for the case of a 3D with -1, 1, 1
+            hdr.set_data_shape((too_big, 1, 1, 4))
+        assert_equal(hdr.get_data_shape(), (too_big, 1, 1, 4))
+        assert_array_equal(hdr['dim'][:5], np.array([4, -1, 1, 1, 4]))
+        # This only works when the first 3 dimensions are -1, 1, 1
         assert_raises(HeaderDataError, hdr.set_data_shape, (too_big,))
         assert_raises(HeaderDataError, hdr.set_data_shape, (too_big,1))
```
matthew-brett: I think PEP8 is going to complain about the lack of spaces between values in the tuple.

nibabel/tests/test_nifti1.py
matthew-brett: Sorry - scratch that - I see it's at the top of the function already.

matthew-brett:
Looks good apart from my line comments.

nibabel/tests/test_nifti1.py
effigies: Should I go ahead and PEP8 clean the whole file? It's got a ton of complaints, but they're not all really relevant to the PR, so I can go either way.

matthew-brett: Ah - sorry - I was reading color-blind, didn't notice this was white text not green - so ignore me, let's not extend PEP8 editing beyond your current changes.

nibabel/nifti1.py
effigies: How about the following documentation output?

https://qnl-linux1.bu.edu/~effigies/nibabel/reference/nibabel.nifti1.html#nibabel.nifti1.Nifti1Header.set_data_shape

matthew-brett: Nice - but - can you clean up the appearance of the links in the output by using ReST links, defined at the bottom of the docstring?

effigies: Learning Sphinx, today. Have another look?

matthew-brett: Nice ! Does it look OK in the source code as well?

effigies:
Pushed new commit. If source needs cleanup, let me know.

matthew-brett:
Thanks for that - looks good now.  Any more comments?  Otherwise I will merge tomorrow.

bcipolli:
@matthew-brett @effigies this seems ready to go?

matthew-brett:
Thanks for reminding me.

PR #334, opened 2015-07-28
NF: add routine to read zero-terminal strings

matthew-brett:
Routine to read zero-terminal strings of arbitrary length from binary file
stream.  Inspired by Thomas Emmerling's readCstring function in the 
BrainVoyager PR.

### DIFF
--- a/nibabel/tests/test_fileutils.py
+++ b/nibabel/tests/test_fileutils.py
@@ -1,0 +1,56 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+# ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
+# ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+""" Testing fileutils module
+"""
+
+import numpy as np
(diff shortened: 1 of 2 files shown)


nibabel/tests/test_fileutils.py
```diff
@@ -0,0 +1,57 @@
...
+        # create a tempfile
+        path = 'test.bin'
+        fwrite = open(path, 'wb')
+        # write the binary block to it
+        fwrite.write(binary)
+        fwrite.close()
+        # open it again
+        fread = open(path, 'rb')
+        # test readout of one string
+        assert_equal([s for s in read_zt_byte_strings(fread)],
```
thomastweets: This should be possible without a list comprehension as the return value is no generator anymore, right?

thomastweets:
Nice!
This would exactly fit my use case of parsing zt strings from BV file headers.
I wonder whether one should include a 'rewind' kind of parameter as the changed position of the fileobj could be a not-so-obvious side effect. Then again, maybe we can assume well informed users.

nibabel/tests/test_fileutils.py
matthew-brett: Yes, thanks for finding that one.

matthew-brett:
I left out `rewind` because I thought the most common case is that you expect the file pointer to be after the stuff you have just read.  Also, it's easy to wrap this function with a few lines to do that if you need to.

I added a note about the file position in the docstring.

thomastweets:
Perfect!

matthew-brett:
Thanks for the review - merging...

PR #335, opened 2015-07-28
Update parrec.py

yanlend:
Split up operations in `get_data()` to avoid creation of unnecessary temporary arrays that consume too much memory.


matthew-brett:
Seems reasonable to me.

Eric89GXL:
Agreed, LGTM +1 for merge

matthew-brett:
Thanks for the change - merged.

PR #336, opened 2015-08-05
BF: Fix benchmark range bounds

effigies:
These look like errors.


matthew-brett:
They do look like errors !  Thanks, merging.

PR #340, opened 2015-08-25
TEST: Nifti test bug; STY: PEP8 test_nifti1

effigies:
While fixing tests on #332 PEP8, I cleaned the file and ran into a bug.

One outstanding PEP8 issue because I couldn't parse a comment enough to reword.

### DIFF
--- a/nibabel/tests/test_nifti1.py
+++ b/nibabel/tests/test_nifti1.py
@@ -349,8 +349,9 @@ class TestNifti1PairHeader(tana.TestAnalyzeHeader, tspm.HeaderScalingMixin):
         assert_array_equal(hdr.get_sform(), empty_sform)
         assert_equal(hdr.get_qform(coded=True), (None, 0))
         assert_equal(hdr.get_sform(coded=True), (None, 0))
-        # Affine with no shears
+        # Affines with no shears
         nice_aff = np.diag([2, 3, 4, 1])
+        another_aff = np.diag([3, 4, 5, 1])
         # Affine with shears
         nasty_aff = from_matvec(np.arange(9).reshape((3, 3)), [9, 10, 11])
         fixed_aff = unshear_44(nasty_aff)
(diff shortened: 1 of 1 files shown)


nibabel/tests/test_nifti1.py
```diff
@@ -132,40 +132,40 @@ def test_slope_inter(self):
...
-            ((minf, inf), HDE, (None, None), (minf, inf)),
-            ((minf, minf), HDE, (None, None), (minf, minf)),
-            # Good slope and bad intercept generates error for get_slope_inter
-            ((2, None), HDE, HDE, (2, nan)),
-            ((2, nan), HDE, HDE, (2, nan)),
-            ((2, inf), HDE, HDE, (2, inf)),
-            ((2, minf), HDE, HDE, (2, minf))):
-            # Good slope and inter - you guessed it
-            ((2, 0), None, (None, None), (2, 0)),
-            ((2, 1), None, (None, None), (2, 1)),
```
effigies: The colon/comma swap was the bug. Last two conditions weren't iterated over but were unassigned values in the loop body.

Updated the values in the unrun tests to not fail, but this should be checked for validity.

nibabel/tests/test_nifti1.py
```diff
@@ -359,7 +360,7 @@ def test_qform_sform(self):
             aff, code = out_meth(coded=True)
             assert_array_equal(aff, nice_aff)
             assert_equal(code, 2)
-            assert_array_equal(out_meth(), nice_aff) # non coded
+            assert_array_equal(out_meth(), nice_aff)  # non coded
             # Affine can also be passed if code == 0, affine will be suitably set
```
effigies: This is the too-long line.

matthew-brett:
LGTM

nibabel/tests/test_nifti1.py
effigies: Took the time to understand the test. How about replacing with:

    # Affine may be passed if code == 0, but will be set to None

matthew-brett: Actually, the affine gets set to the passed affine, even if the code == 0.   The test doesn't do a good job of testing that, because the affine could have remained unchanged.  How about:

```
             another_aff = np.diag([3, 4, 5, 1])
             # Affine may be passed if code == 0, and will get set into header,
             # but the returned affine with 'coded=True' will be None.
             in_meth(another_aff, 0)
             assert_equal(out_meth(coded=True), (None, 0))  # coded -> None
             assert_array_almost_equal(out_meth(), another_aff)  # not coded -> input affine
```

effigies:
Okay. Slightly modified your inline comment to fit in 79 characters, and moved the `another_aff` definition up with `nice_aff` but otherwise used your fix exactly.

effigies:
Anything left to do on this one?

matthew-brett:
Thanks for the reminder - and the rewrite.

PR #342, opened 2015-08-28
TST: Call current python executable in test_back_tick

### LINKED ISSUE #341: TST: test_back_tick calls python directly
In the test `test_back_tick()` in `nibabel/nisext/tests/test_testers.py` python is called. This test will fail on systems where the binary is named differently. E.g., on systems with only Python 3 where the binary is called python3.

In the testers package the interpreter is chosen using `sys.executable`. Could that be done here as well?

effigies:
closes #341


matthew-brett:
Thanks looks good to me.

Issue #343, opened 2015-09-01
Simplify the testing infrastructure for checking warnings


bcipolli:
In implementing #329, I've hit the need to check warnings. I believe `nibabel` has two different codepaths for manipulating warnings:

* `nibabel.checkwarns` - has classes `ErrorWarnings` and `IgnoreWarnings`
* `nibabel.testing` - has classes `suppress_warnings` and `catch_warn_reset`

I'd like to combine these into a single place, with a single methodology. I believe creating a `CheckWarnings` class, that mirrors `catch_warn_reset`, and eliminating the warning classes from `nibabel.testing`, may be the best?

@matthew-brett any advice? I'd like to resolve this one before completing #329 , if that's cool. I have time to do it today.

matthew-brett:
Sorry - I was on a plane flying across the Atlantic yesterday...

Yes, it certainly seems like a good idea to clean these up.

I did an extensive refactor of these things for numpy - here : https://github.com/numpy/numpy/pull/5682 - maybe best to use a copy of that?  It had a heavy review.

We should be a little careful not to break anyone else's code with the API change.

bcipolli:
:+1: thx

bcipolli:
OK, so given this, I am thinking to:
* Deprecate `ErrorWarnings`, `IgnoreWarnings`, and `catch_warn_reset`; also deprecate `nibabel.checkwarns`.
* New interface will be `suppress_warnings` (essentially `IgnoreWarnings`), `clear_and_catch_warnings` (copy from `numpy`), and `error_warnings` (essentially `ErrorWarnings`).

Any comments on this plan are welcome.

matthew-brett:
Sounds like a good plan.   `suppress_warnings` and `error_warnings` could inherit from `clear_and_catch_warnings` too, to allow for modules where the warning has already been raised.

bcipolli:
This was fixed in #345

PR #345, opened 2015-09-02
Simplify interface for testing of warnings.

bcipolli:
This is a fix for #343 

* New interface: suppress_warnings, clear_and_catch_warnings, and error_warnings.
* Deprecate ErrorWarnings, IgnoreWarnings, and catch_warn_reset; nibabel.checkwarns.

I haven't added tests for the deprecation, but I can do that. I am just feeling lazy :sweat_smile:


nibabel/testing/__init__.py
```diff
@@ -7,23 +7,19 @@
...
 
-# Allow failed import of nose if not now running tests
-try:
-    import nose.tools as nt
-except ImportError:
-    pass
-else:
-    from nose.tools import (assert_equal, assert_not_equal,
-                            assert_true, assert_false, assert_raises)
-
```
bcipolli: @matthew-brett Any idea if `try...catch` is still necessary? I took the liberty to blow it away. I couldn't tell from the commits (5+ years ago) where this was being used. I didn't see `nt` being used anywhere in my cursory search, and tests pass...

nibabel/testing/__init__.py
```diff
@@ -56,22 +52,50 @@ def assert_allclose_safely(a, b, match_nans=True):
...
 
 
-class suppress_warnings(catch_warnings):
-    """ Version of ``catch_warnings`` class that suppresses warnings
-    """
-    def __enter__(self):
-        res = super(suppress_warnings, self).__enter__()
-        simplefilter('ignore')
-        return res
+class clear_and_catch_warnings(warnings.catch_warnings):
```
bcipolli: This is a copy of code from https://github.com/numpy/numpy/pull/5682/files . I did NOT copy over the tests.

nibabel/testing/__init__.py
```diff
@@ -79,12 +103,44 @@ def __enter__(self):
...
-        super(catch_warn_reset, self).__exit__(*exc_info)
+        super(clear_and_catch_warnings, self).__exit__(*exc_info)
         for mod in self.modules:
             if hasattr(mod, '__warningregistry__'):
                 mod.__warningregistry__.clear()
             if mod in self._warnreg_copies:
                 mod.__warningregistry__.update(self._warnreg_copies[mod])
+
+
+class error_warnings(clear_and_catch_warnings):
```
bcipolli: Inherits from `clear_and_catch_warnings`.

nibabel/testing/__init__.py
```diff
@@ -79,12 +103,44 @@ def __enter__(self):
...
+    """
+    filter = 'error'
+
+    def __enter__(self):
+        mgr = super(error_warnings, self).__enter__()
+        warnings.simplefilter(self.filter)
+        return mgr
+
+
+class suppress_warnings(error_warnings):
```
bcipolli: Lost the docstring. I will recover it shortly.

nibabel/testing/__init__.py
```diff
@@ -79,12 +103,44 @@ def __enter__(self):
...
+    filter = 'ignore'
+
+
+class catch_warn_reset(clear_and_catch_warnings):
```
bcipolli: Removed docstring, and all functions take `*args`, `**kwargs` for all deprecated classes.

nibabel/tests/test_arraywriters.py
```diff
@@ -11,24 +11,19 @@
 import numpy as np
 
 from ..externals.six import BytesIO
-
 from ..arraywriters import (SlopeInterArrayWriter, SlopeArrayWriter,
                             WriterError, ScalingError, ArrayWriter,
                             make_array_writer, get_slope_inter)
-
 from ..casting import int_abs, type_info, shared_range, on_powerpc
-
```
bcipolli: drive-by linting. I couldn't resist!

nibabel/tests/test_checkwarns.py
```diff
@@ -1,42 +1,9 @@
...
-        warn('Here is a warning, you will not see it')
-        warn('Nor this one', DeprecationWarning)
-    assert_equal(n_warns, len(filters))
-    # Check other errors are propagated
-    def f():
-        with IgnoreWarnings():
-            raise ValueError('An error')
-    assert_raises(ValueError, f)
+from ..testing import (error_warnings, suppress_warnings,
+                       clear_and_catch_warnings)
```
bcipolli: Need to add deprecation tests.

nibabel/testing/__init__.py
bcipolli: Note: I just changed the default to `record=True` for the nibabel version of this code. This is what was used in all of these helper classes in the `nibabel` codebase previously, and I think that makes a lot of sense.

nibabel/testing/__init__.py
matthew-brett: This is so that people can still import nibabel if they don't have nose installed.

nibabel/testing/__init__.py
matthew-brett: Would you mind copying over the tests as well?

nibabel/tests/test_arraywriters.py
matthew-brett: :)

matthew-brett:
Oops - mixed in with the array class refactoring?

bcipolli:
Everything's set, except one test failure (checking warnings from the `checkwarns` module import). Strangely, the tests succeed locally. Any ideas?

nibabel/tests/test_checkwarns.py
```diff
@@ -1,42 +1,26 @@
...
-        with IgnoreWarnings():
-            raise ValueError('An error')
-    assert_raises(ValueError, f)
+from ..testing import (error_warnings, suppress_warnings,
+                       clear_and_catch_warnings, get_fresh_mod)
+
+
+def test_ignore_and_error_warnings():
+    with clear_and_catch_warnings() as w:
+        from .. import checkwarns
```
matthew-brett: I'm guessing what is happening here is that nose has previously imported `checkwarns` as part of test discovery for the while test-suite, and therefore there is no Future warning, because the import has already happened.

nibabel/tests/test_checkwarns.py
```diff
@@ -1,42 +1,26 @@
...
+        assert_equal(len(w), 1)
+        assert_equal(w[0].category, FutureWarning)
+
+    with clear_and_catch_warnings() as w:
+        with checkwarns.IgnoreWarnings():
```
matthew-brett: What are you testing here?   Sorry for my confusion.

bcipolli: deprecation warning in object creation

matthew-brett: I see.  I suppose you don't need the `with` in that case?

nibabel/tests/test_checkwarns.py
matthew-brett: By the way - I do see the error if I run the full test-suite locally, but not if I run the individual test.

bcipolli:
I changed the offending test to an `ignore_warnings` block, as I'm stumped and would rather use the time on more important things. Hope that's cool!

I looked over the coverage change, and I didn't see any obvious gaps (as well as I could from the reports available).

I believe this is ready to go, and to focus on #329 , then back to GIFTI / CIFTI (which is what this is all really about!)

matthew-brett:
Yes, that's fine, thanks for doing that.

PR #347, opened 2015-09-06
BF: Misplaced parenthesis

effigies:
Always returned `False`.


matthew-brett:
Good spot.  Fancy adding a test?

effigies:
Tests for negative values in each slot added.

matthew-brett:
Nice, thanks.

PR #348, opened 2015-09-11
NF: -H for nib-ls to list desired header fields

yarikoptic:
saw https://neurostars.org/p/3361/ which triggered to do this tiny hack (time for tests for the beast will come!)

Now it should be possible to specify some header fields desired to be output, which will get listed as well, e.g.:

```
$> $PWD/bin/nib-ls -H dim,bitpix /usr/share/data/fsl-mni152-templates/MNI152*brain.nii.gz
/usr/share/data/fsl-mni152-templates/MNI152lin_T1_1mm_brain.nii.gz uint8 [182, 218, 182] 1.00x1.00x1.00   [  3 182 218 182   1   1   1   1]  8
/usr/share/data/fsl-mni152-templates/MNI152lin_T1_2mm_brain.nii.gz uint8 [ 91, 109,  91] 2.00x2.00x2.00   [  3  91 109  91   1   1   1   1]  8
/usr/share/data/fsl-mni152-templates/MNI152_T1_1mm_brain.nii.gz    int16 [182, 218, 182] 1.00x1.00x1.00   [  3 182 218 182   1   1   1   1] 16
/usr/share/data/fsl-mni152-templates/MNI152_T1_2mm_brain.nii.gz    int16 [ 91, 109,  91] 2.00x2.00x2.00   [  3  91 109  91   1   1   1   1] 16
```

using '*' as argument would spit out all the header fields

### DIFF
--- a/bin/nib-ls
+++ b/bin/nib-ls
@@ -13,7 +13,7 @@ Output a summary table for neuroimaging files (resolution, dimensionality, etc.)
 from __future__ import division, print_function, absolute_import
 
 __author__ = 'Yaroslav Halchenko'
-__copyright__ = 'Copyright (c) 2011-2012 Yaroslav Halchenko ' \
+__copyright__ = 'Copyright (c) 2011-2015 Yaroslav Halchenko ' \
                 'and NiBabel contributors'
 __license__ = 'MIT'
 
@@ -141,6 +141,10 @@ def get_opt_parser():
                dest="verbose", default=0,
                help="Make more noise.  Could be specified multiple times"),
 
+        Option("-H", "--header-fields",
+               dest="header_fields", default='',
+               help="Header fields (comma separated) to be printed as well (if present)"),
+
         Option("-s", "--stats",
                action="store_true", dest='stats', default=False,
                help="Output basic data statistics"),
@@ -179,6 +183,22 @@ def proc_file(f, opts):
         row += ['@l#exts: %d' % len(h.extensions)]
     else:
         row += [ '' ]
+
+    if opts.header_fields:
+        # signals "all fields"
+        if opts.header_fields == '*':
+            # TODO: might vary across file types, thus prior sensing would be needed
+            header_fields = h.keys()
+        else:
+            header_fields = opts.header_fields.split(',')
+
+        for f in header_fields:
+            if not f: # skip empty
+                continue
+            try:
+                row += [str(h[f])]
+            except (KeyError, ValueError):
+                row += [ '' ]
 
     try:
         if (hasattr(h, 'get_qform') and hasattr(h, 'get_sform')
--- a/bin/nib-ls
+++ b/bin/nib-ls
@@ -186,7 +186,7 @@ def proc_file(f, opts):
 
     if opts.header_fields:
         # signals "all fields"
-        if opts.header_fields == '*':
+        if opts.header_fields == 'all':
             # TODO: might vary across file types, thus prior sensing would be needed
             header_fields = h.keys()
         else:
--- a/bin/nib-ls
+++ b/bin/nib-ls
@@ -198,7 +198,7 @@ def proc_file(f, opts):
             try:
                 row += [str(h[f])]
             except (KeyError, ValueError):
-                row += [ '' ]
+                row += [ 'error' ]
 
     try:
         if (hasattr(h, 'get_qform') and hasattr(h, 'get_sform')
--- a/bin/nib-ls
+++ b/bin/nib-ls
@@ -187,7 +187,8 @@ def proc_file(f, opts):
     if opts.header_fields:
         # signals "all fields"
         if opts.header_fields == 'all':
-            # TODO: might vary across file types, thus prior sensing would be needed
+            # TODO: might vary across file types, thus prior sensing
+            # would be needed
             header_fields = h.keys()
         else:
             header_fields = opts.header_fields.split(',')
--- a/nibabel/testing/__init__.py
+++ b/nibabel/testing/__init__.py
@@ -9,6 +9,7 @@
 ''' Utilities for testing '''
 from __future__ import division, print_function
 
+import re
 import os
 import sys
 import warnings
@@ -57,6 +58,17 @@ def assert_allclose_safely(a, b, match_nans=True):
     if b.dtype.kind in 'ui':
         b = b.astype(float)
     assert_true(np.allclose(a, b))
+
+
+def assert_re_in(regex, c, flags=0):
+    """Assert that container (list, str, etc) contains entry matching the regex
+    """
+    if not isinstance(c, (list, tuple)):
+        c = [c]
+    for e in c:
+        if re.match(regex, e, flags=flags):
+            return
+    raise AssertionError("Not a single entry matched %r in %r" % (regex, c))
 
 
 def get_fresh_mod(mod_name=__name__):
--- a/nibabel/tests/test_scripts.py
+++ b/nibabel/tests/test_scripts.py
@@ -82,7 +82,9 @@ def test_nib_ls_multiple():
     assert_equal(len(stdout_lines), 4)
     # they should be indented correctly.  Since all files are int type -
     ln = max(len(f) for f in fnames)
-    assert_equal([l[ln:ln+2] for l in stdout_lines], [' i']*4)
+    assert_equal([l[ln:ln+2] for l in stdout_lines], [' i']*4,
+                 msg="Type sub-string didn't start with 'i'. "
+                     "Full output was: %s" % stdout_lines)
     # and if disregard type indicator which might vary
     assert_equal(
         [l[l.index('['):] for l in stdout_lines],
--- a/nibabel/tests/test_scripts.py
+++ b/nibabel/tests/test_scripts.py
@@ -20,6 +20,7 @@ from ..orientations import flip_axis, aff2axcodes, inv_ornt_aff
 
 from nose.tools import (assert_true, assert_false, assert_not_equal,
                         assert_equal)
+from nose import SkipTest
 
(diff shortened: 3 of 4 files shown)


bin/nib-ls
```diff
@@ -180,6 +184,22 @@ def proc_file(f, opts):
     else:
         row += [ '' ]
 
+    if opts.header_fields:
+        # signals "all fields"
+        if opts.header_fields == '*':
```
matthew-brett: Is this the right character to use?   Maybe 'all' string instead to avoid shell?

bin/nib-ls
```diff
@@ -180,6 +184,22 @@ def proc_file(f, opts):
...
+            # TODO: might vary across file types, thus prior sensing would be needed
+            header_fields = h.keys()
+        else:
+            header_fields = opts.header_fields.split(',')
```
matthew-brett: Maybe filter up here for invalid or empty keys?  Raise error instead of passing silently?

bin/nib-ls
yarikoptic: Isn't there a higher chance that some format would define "all" header field?

bin/nib-ls
yarikoptic: raising error -- I would prefer to avoid it... at times I just run nib-ls on a bunch of files of different types, so if in some type some header field is missing -- oh well.  Similar logic here is for other types (e.g. no qform/sform in [sq]form in AnalyzeHeader, but otherwise I print 'error'. Ok -- can print 'error' here as well for consistency I guess

bin/nib-ls
matthew-brett: There is.  Tradeoff between ease of confusion when shell takes over the `*` and likelihood of some future format defining a field called 'all'.

bin/nib-ls
matthew-brett: OK

TheChymera:
I support this pull request :)

matthew-brett:
Yarik - time for some tests?

yarikoptic:
Not ATM... Busy pushing kids through the Disneyland

matthew-brett:
Back from Disneyland ? :)

matthew-brett:
Yarick - time for tests?   You know you love them.

yarikoptic:
I love when they exist indeed... would need to recall how you test those cmdline utils... ideally I think all that functionality should be coded within e.g. nibabel/cmdline and then only interfaced through tiny scripts calling into them.
do you have already some helpers to mock stdout/print statements?  otherwise -- we do have one in datalad (https://github.com/datalad/datalad/blob/master/datalad/utils.py#L225) but IIRC it doesn't play exactly nice with nosetests when that one is ran withough '-s'...

matthew-brett:
Did you already see : https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_scripts.py#L57 ?

nibabel/testing/__init__.py
```diff
@@ -59,6 +60,17 @@ def assert_allclose_safely(a, b, match_nans=True):
     assert_true(np.allclose(a, b))
 
 
+def assert_re_in(regex, c, flags=0):
+    """Assert that container (list, str, etc) contains entry matching the regex
+    """
```
yarikoptic: this one came from datalad (my copyright/license, so no problem there), so I decided just to keep it as is ;)

nibabel/tests/test_testing.py
```diff
@@ -142,3 +142,24 @@ def f():
...
+
+    # Sufficient to have one entry matching
+    assert_re_in("ab", ["", "abc", "laskdjf"])
+    assert_raises(AssertionError, assert_re_in, "ab$", ["ddd", ""])
+
+    # Tuples should be ok too
+    assert_re_in("ab", ("", "abc", "laskdjf"))
+    assert_raises(AssertionError, assert_re_in, "ab$", ("ddd", ""))
+
+    # shouldn't "match" the emty list
```
matthew-brett: emty typo.  Capitalize first word of comments?

Worth having an option for search instead of match?

nibabel/tests/test_scripts.py
```diff
@@ -54,15 +54,62 @@ def script_test(func):
 
 
 @script_test
-def test_nib_ls():
+def _test_nib_ls_example4d(opts=[], hdrs_str=""):
```
matthew-brett: Elsewhere I've been using `check_` prefix for things that do asserts, that aren't meant to be tests directly.

Does it need the decorator?

nibabel/tests/test_scripts.py
```diff
@@ -54,15 +54,62 @@ def script_test(func):
...
+                   % hdrs_str)
+    cmd = ['nib-ls'] + opts + [fname]
     code, stdout, stderr = run_command(cmd)
     assert_equal(fname, stdout[:len(fname)])
-    assert_not_equal(re.match(expected_re, stdout[len(fname):]), None)
+    assert_re_in(expected_re, stdout[len(fname):])
+
+@script_test
+def test_nib_ls():
+    yield _test_nib_ls_example4d
```
matthew-brett: Don't feel too strongly, but I prefer to avoid yield because it makes the test failures harder to debug.

nibabel/tests/test_scripts.py
```diff
@@ -54,15 +54,62 @@ def script_test(func):
...
+    fnames = [
+        pjoin(DATA_PATH, f)
+        for f in ('example4d.nii.gz', 'example_nifti2.nii.gz', 'small.mnc', 'nifti2.hdr')
+    ]
+    code, stdout, stderr = run_command(['nib-ls'] + fnames)
+    stdout_lines = stdout.split('\n')
+    assert_equal(len(stdout_lines), 4)
+    # they should be indented correctly.  Since all files are int type -
+    ln = max(len(f) for f in fnames)
+    assert_equal([l[ln:ln+2] for l in stdout_lines], [' i']*4,
```
matthew-brett: Would you mind running a PEP8 checker over the changes?

nibabel/tests/test_testing.py
yarikoptic: thanks, will do/fix

someone might want/will do the option whenever necessity comes ;-)

nibabel/tests/test_scripts.py
yarikoptic: do you have other constructs for parametric tests?  I hate duplication and debugging loops within the test is no better ;)

nibabel/tests/test_scripts.py
yarikoptic: is there a convenient way to run check only on changes?  otherwise -- there is more to fix up:
```
$> flake8 nibabel/testing/__init__.py
nibabel/testing/__init__.py:3:1: E266 too many leading '#' for block comment
nibabel/testing/__init__.py:8:1: E266 too many leading '#' for block comment
nibabel/testing/__init__.py:23:5: F401 'assert_false' imported but unused
nibabel/testing/__init__.py:23:5: F401 'assert_not_equal' imported but unused
nibabel/testing/__init__.py:23:5: F401 'assert_raises' imported but unused
nibabel/testing/__init__.py:179:80: E501 line too long (84 > 79 characters)
2 10297 ->1.....................................:Mon 30 Nov 2015 01:34:44 PM EST:.
(git)hopa:~nipy-suite/nibabel[nibls-header]git
$> flake8 nibabel/tests/test_scripts.py
nibabel/tests/test_scripts.py:12:1: F401 're' imported but unused
nibabel/tests/test_scripts.py:21:1: F401 'assert_not_equal' imported but unused
nibabel/tests/test_scripts.py:24:1: F401 'assert_array_equal' imported but unused
nibabel/tests/test_scripts.py:41:16: E251 unexpected spaces around keyword / parameter equals
nibabel/tests/test_scripts.py:41:18: E251 unexpected spaces around keyword / parameter equals
nibabel/tests/test_scripts.py:42:20: E251 unexpected spaces around keyword / parameter equals
nibabel/tests/test_scripts.py:42:22: E251 unexpected spaces around keyword / parameter equals
nibabel/tests/test_scripts.py:51:29: E261 at least two spaces before inline comment
nibabel/tests/test_scripts.py:67:1: E302 expected 2 blank lines, found 1
nibabel/tests/test_scripts.py:70:80: E501 line too long (97 > 79 characters)
nibabel/tests/test_scripts.py:72:1: E302 expected 2 blank lines, found 1
nibabel/tests/test_scripts.py:77:80: E501 line too long (89 > 79 characters)
nibabel/tests/test_scripts.py:105:80: E501 line too long (95 > 79 characters)
nibabel/tests/test_scripts.py:106:80: E501 line too long (95 > 79 characters)
nibabel/tests/test_scripts.py:107:80: E501 line too long (94 > 79 characters)
nibabel/tests/test_scripts.py:108:80: E501 line too long (81 > 79 characters)
nibabel/tests/test_scripts.py:114:1: E303 too many blank lines (3)
```
meanwhile will trim the lines I have introduced

yarikoptic: actually trimming some would only bring disturbance into the field for no benefit (e.g. matching those output lines with stats in them), so I will let those stay as is

yarikoptic:
not that there is still a failed test to deal with:  with numpy 1.5.1 for some reason .mnc output is just `/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/tests/data/small.mnc             failed`... I guess it must not be numpy version specific... will look into it after lunch

yarikoptic:
gy, adding more tests decreased the coverage %... sorry about that (can't see which are those two lines, probably try/except clause).  Otherwise should be ready for consumption I guess.

matthew-brett:
OK - sorry to be slow - thanks - in it goes.

yarikoptic:
hip hip hoorray! thanks @matthew-brett !

PR #352, opened 2015-10-06
RF: getArraysFromIntent => get_arrays_from_intent

bcipolli:
GIFTI has a function in CamelCase, not snake_case. Deprecate that.


matthew-brett:
Yup, fair enough.

PR #353, opened 2015-10-07
RF: Use python properties, rather than set/get methods

bcipolli:
`numDA`, `version`, and `filename` properties on GIFTI have `set_` and `get_` methods. Seems cleaner (and more pythonic) to use Python properties instead.

Question: is GIFTI support officially published? I.e. do I need to add deprecation warnings for the old methods, and deprecation tests?

Also renamed tests to be clear that they're testing the loading of these properties, not necessarily of their setting or saving.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -133,9 +133,30 @@ class GiftiLabel(object):
         self.blue = blue
         self.alpha = alpha
 
+    @np.deprecate_with_doc("Use the rgba property instead.")
     def get_rgba(self):
+        return self.rgba
+
+    @property
+    def rgba(self):
         """ Returns RGBA as tuple """
         return (self.red, self.green, self.blue, self.alpha)
+
+    @rgba.setter
+    def rgba(self, rgba):
+        """ Set RGBA via tuple
+
+        Parameters
+        ----------
+        rgba : tuple (red, green, blue, alpha)
+
+        """
+        if len(rgba) != 4:
+            raise ValueError('rgba must be length 4.')
+        self.red = rgba[0]
+        self.green = rgba[1]
+        self.blue = rgba[2]
+        self.alpha = rgba[3]
 
 
 def _arr2txt(arr, elem_fmt):
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -380,6 +380,14 @@ class GiftiImage(object):
             raise ValueError("Not a valid GiftiLabelTable instance")
         self._labeltable = labeltable
 
+    @np.deprecate_with_doc("Use the gifti_img.labeltable property instead.")
+    def set_labeltable(self, labeltable):
+        self.labeltable = labeltable
+
+    @np.deprecate_with_doc("Use the gifti_img.labeltable property instead.")
+    def get_labeltable(self):
+        return self.labeltable
+
     @property
     def meta(self):
         return self._meta
@@ -400,6 +408,14 @@ class GiftiImage(object):
             raise ValueError("Not a valid GiftiMetaData instance")
         self._meta = meta
  
+    @np.deprecate_with_doc("Use the gifti_img.labeltable property instead.")
+    def set_metadata(self, meta):
+        self.meta = meta
+
+    @np.deprecate_with_doc("Use the gifti_img.labeltable property instead.")
+    def get_meta(self):
+        return self.meta
+
     def add_gifti_data_array(self, dataarr):
         """ Adds a data array to the GiftiImage
 
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -153,11 +153,7 @@ class GiftiLabel(object):
         """
         if len(rgba) != 4:
             raise ValueError('rgba must be length 4.')
-        self.red = rgba[0]
-        self.green = rgba[1]
-        self.blue = rgba[2]
-        self.alpha = rgba[3]
-
+        self.red, self.green, self.blue, self.alpha = rgba
 
 def _arr2txt(arr, elem_fmt):
     arr = np.asarray(arr)
@@ -438,7 +434,7 @@ class GiftiImage(object):
         if not isinstance(meta, GiftiMetaData):
             raise ValueError("Not a valid GiftiMetaData instance")
         self._meta = meta
- 
+
     @np.deprecate_with_doc("Use the gifti_img.labeltable property instead.")
     def set_metadata(self, meta):
         self.meta = meta
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -344,47 +344,31 @@ class GiftiDataArray(object):
 
 
 class GiftiImage(object):
-
-    numDA = int
-    version = str
-    filename = str
-
     def __init__(self, meta=None, labeltable=None, darrays=None,
                  version="1.0"):
         if darrays is None:
             darrays = []
-        self.darrays = darrays
         if meta is None:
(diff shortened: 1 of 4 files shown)


matthew-brett:
Sorry to say, but yes, we have advertised the GIFTI support, and people have used it, so I think we do have to deprecate the get_ / set_ methods...

bcipolli:
OK. Is it worth it to make this change, in this case? Happy to deprecate & test, but also OK to toss. I'm just working through the code on my way to load/save GIFTI.

matthew-brett:
If you have the heart to do the deprecations, I think it's an improvement...

bcipolli:
I sure do! Will get to that tonight or tomorrow, then :)

bcipolli:
Added the get/set deprecations and deprecation tests.

nibabel/gifti/gifti.py
```diff
@@ -427,14 +425,12 @@ def add_gifti_data_array(self, dataarr):
         """
         if isinstance(dataarr, GiftiDataArray):
             self.darrays.append(dataarr)
-            self.numDA += 1
```
matthew-brett: numDA being tested properly already?

bcipolli: Not sure about tests, will check.
This line was removed as I made `numDA` a computed read-only property.

matthew-brett: Yes, sorry to be confusing - I saw you didn't need the line, but I was wondering if there were tests to make sure the computation was working right.

bcipolli: There were not; I just added some.

bcipolli:
I updated some other functions that are better as properties and added basic tests. Please take another look when possible!

nibabel/gifti/gifti.py
```diff
@@ -392,15 +403,24 @@ def set_labeltable(self, labeltable):
         labeltable : GiftiLabelTable
 
         """
-        if isinstance(labeltable, GiftiLabelTable):
-            self.labeltable = labeltable
-        else:
-            print("Not a valid GiftiLabelTable instance")
+        if not isinstance(labeltable, GiftiLabelTable):
+            raise ValueError("Not a valid GiftiLabelTable instance")
```
bcipolli: Note the change from `print` to `ValueError` here.

nibabel/gifti/gifti.py
```diff
@@ -128,10 +133,31 @@ def __init__(self, key=0, label='', red=None, green=None, blue=None,
...
+        ----------
+        rgba : tuple (red, green, blue, alpha)
+
+        """
+        if len(rgba) != 4:
+            raise ValueError('rgba must be length 4.')
+        self.red = rgba[0]
+        self.green = rgba[1]
+        self.blue = rgba[2]
+        self.alpha = rgba[3]
```
effigies: Personal preference, feel free to reject:

    self.red, self.green, self.blue, self.alpha = rgba

bcipolli: Good call!

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -108,9 +108,6 @@ def StartElementHandler(self, name, attrs):
             self.img = GiftiImage()
             if 'Version' in attrs:
                 self.img.version = attrs['Version']
-            if 'NumberOfDataArrays' in attrs:
-                self.img.numDA = int(attrs['NumberOfDataArrays'])
-                self.count_da = False
```
effigies: I think `attrs['NumberOfDataArrays']` should at least be validated and throw a warning if the count doesn't match the actual number of data arrays. Something like:

    if 'NumberOfDataArrays' in attrs:
        self.expected_numda = int(attrs['NumberOfDataArrays'])

nibabel/gifti/tests/test_gifti.py
```diff
@@ -24,6 +26,35 @@ def test_gifti_image():
...
+    da = GiftiDataArray(data='data')
+    gi.add_gifti_data_array(da)
+
+    gi.remove_gifti_data_array_by_intent(0)
+    assert_equal(gi.numDA, 1)
+
+    gi.darrays[0].intent = 0
+    gi.remove_gifti_data_array_by_intent(0)
+    assert_equal(gi.numDA, 0)
+
```
effigies: Maybe a check that trying to remove a data array when `numDA == 0` produces an `IndexError` and maintains a `numDA == 0`? At least with the old version where the number was incremented and decremented manually, I'd want to be sure it can't go negative.

nibabel/gifti/tests/test_gifti.py
```diff
@@ -36,3 +67,58 @@ def test_dataarray():
...
+def test_gifti_label_rgba():
+    rgba = np.random.rand(4)
+    kwargs = dict(zip(['red', 'green', 'blue', 'alpha'], rgba))
+
+    gl = GiftiLabel(**kwargs)
+    assert_equal(kwargs['red'], gl.rgba[0])
+    assert_equal(kwargs['green'], gl.rgba[1])
+    assert_equal(kwargs['blue'], gl.rgba[2])
+    assert_equal(kwargs['alpha'], gl.rgba[3])
```
effigies: Why not just:

     assert_equal(rgba, gl.rgba)

effigies:
Read-through done. Logic looks good.

nibabel/gifti/parse_gifti_fast.py
bcipolli: Yes. I will also add a test for this. In fact, we lack malformed GIFTI file tests; see https://github.com/nipy/nibabel/issues/357

nibabel/gifti/tests/test_gifti.py
bcipolli: I believe that test is here: https://github.com/nipy/nibabel/pull/353#discussion-diff-42019172R42

effigies: Wow. Yup. Now I'm questioning my ability to read. Should probably have another pass through in a bit.

nibabel/gifti/tests/test_gifti.py
bcipolli: Doesn't work for me; I get:
```ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()```

But `numpy.assert_all_equal` works :)

**Comment by bcipolli** [comment]
Thanks @effigies ; I just updated the `rgba` logic and tests, added requested check for `NumberOfDataArrays` (and added a test), and opened #357 so we can add more 'malformed' GIFTI tests.

**Comment by bcipolli** [comment]
Updated based on last comments.

File: nibabel/gifti/tests/test_gifti.py [hunk 7]
```diff
@@ -36,3 +67,63 @@ def test_dataarray():
...
+    assert_raises(ValueError, assign_rgba, gl3, rgba.tolist() + rgba.tolist())
+
+    # Test deprecation
+    with clear_and_catch_warnings() as w:
+        warnings.filterwarnings('once', category=DeprecationWarning)
+        assert_equal(kwargs['red'], gl3.get_rgba()[0])
+
+    # Test default value
+    gl4 = GiftiLabel()
+    assert_is_none(gl4.rgba[0])
```
effigies: 2.6 seems upset.

    assert_true(gl4.rgba[0] is None)

bcipolli: That's what I get for trying to be too fancy :)

effigies:
Okay, this LGTM.

@matthew-brett The Travis build seems to be failing on unrelated things. (Happening here and on #354.)

PR #354, opened 2015-10-07
BF: Make sure xml is encoded as utf-8

bcipolli:
Valid XML has to be utf-8--in bytes. Current code doesn't explicitly encode or decode, and so different versions of Python return bytes (Python 2) or Unicode (Python 3). 

This caused test errors in the `cifti` PR (and was fixed there); a similar issue needs to be fixed for GIFTI.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -185,6 +185,17 @@ class GiftiCoordSystem(xml.XmlSerializable):
         print('Affine Transformation Matrix: \n', self.xform)
 
 
+@np.deprecate_with_doc("This is an internal API that will be discontinued.")
+def data_tag(dataarray, encoding, datatype, ordering):
+    class DataTag(xml.XmlSerializable):
+        def __init__(self, *args):
+            self.args = args
+        def _to_xml_element(self):
+            return _data_tag_element(*self.args)
+
+    return DataTag(dataarray, encoding, datatype, ordering).to_xml()
+
+
 def _data_tag_element(dataarray, encoding, datatype, ordering):
     """ Creates the data tag depending on the required encoding,
     returns as XML element"""
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -9,10 +9,10 @@
 from __future__ import division, print_function, absolute_import
 
 import sys
-import xml.etree.ElementTree as xml
 
 import numpy as np
 
+from .. import xmlutils as xml
 from ..nifti1 import data_type_codes, xform_codes, intent_codes
 from .util import (array_index_order_codes, gifti_encoding_codes,
                    gifti_endian_codes, KIND2FMT)
@@ -23,19 +23,7 @@ from .util import (array_index_order_codes, gifti_encoding_codes,
 import base64
 
 
-class XmlSerializable(object):
-    """ Basic interface for serializing an object to xml"""
-    def _to_xml_element(self):
-        """ Output should be a xml.etree.ElementTree.Element"""
-        raise NotImplementedError()
-
-    def to_xml(self, enc='utf-8'):
-        """ Output should be an xml string with the given encoding.
-        (default: utf-8)"""
-        return xml.tostring(self._to_xml_element(), enc)
-
-
-class GiftiMetaData(XmlSerializable):
+class GiftiMetaData(xml.XmlSerializable):
     """ A list of GiftiNVPairs in stored in
     the list self.data """
     def __init__(self, nvpair=None):
@@ -87,7 +75,7 @@ class GiftiNVPairs(object):
         self.value = value
 
 
-class GiftiLabelTable(XmlSerializable):
+class GiftiLabelTable(xml.XmlSerializable):
 
     def __init__(self):
         self.labels = []
@@ -113,7 +101,7 @@ class GiftiLabelTable(XmlSerializable):
         print(self.get_labels_as_dict())
 
 
-class GiftiLabel(XmlSerializable):
+class GiftiLabel(xml.XmlSerializable):
     key = int
     label = str
     # rgba
@@ -166,7 +154,7 @@ def _arr2txt(arr, elem_fmt):
     return '\n'.join(fmt % tuple(row) for row in arr)
 
 
-class GiftiCoordSystem(XmlSerializable):
+class GiftiCoordSystem(xml.XmlSerializable):
     dataspace = int
     xformspace = int
     xform = np.ndarray  # 4x4 numpy array
@@ -199,7 +187,7 @@ class GiftiCoordSystem(XmlSerializable):
 
 def _data_tag_element(dataarray, encoding, datatype, ordering):
     """ Creates the data tag depending on the required encoding,
-    returns as bytes"""
+    returns as XML element"""
     import zlib
     ord = array_index_order_codes.npcode[ordering]
     enclabel = gifti_encoding_codes.label[encoding]
@@ -225,7 +213,7 @@ def data_tag(dataarray, encoding, datatype, ordering):
     return xml.tostring(data, 'utf-8')
 
 
-class GiftiDataArray(XmlSerializable):
+class GiftiDataArray(xml.XmlSerializable):
 
     # These are for documentation only; we don't use these class variables
     intent = int
@@ -362,7 +350,7 @@ class GiftiDataArray(XmlSerializable):
         return self.meta.metadata
 
 
-class GiftiImage(XmlSerializable):
+class GiftiImage(xml.XmlSerializable):
 
     def __init__(self, meta=None, labeltable=None, darrays=None,
                  version="1.0"):
@@ -506,4 +494,4 @@ class GiftiImage(XmlSerializable):
         """ Return XML corresponding to image content """
         return b"""<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE GIFTI SYSTEM "http://www.nitrc.org/frs/download.php/115/gifti.dtd">
-""" + XmlSerializable.to_xml(self, enc)
+""" + xml.XmlSerializable.to_xml(self, enc)
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -9,6 +9,7 @@
 from __future__ import division, print_function, absolute_import
 
 import sys
+import xml.etree.ElementTree as xml
 
 import numpy as np
 
@@ -51,17 +52,14 @@ class GiftiMetaData(object):
         return self.data_as_dict
 
     def to_xml(self):
-        if len(self.data) == 0:
-            return b"<MetaData/>\n"
-        res = "<MetaData>\n"
+        metadata = xml.Element('MetaData')
         for ele in self.data:
-            nvpair = """<MD>
-\t<Name><![CDATA[%s]]></Name>
-\t<Value><![CDATA[%s]]></Value>
-</MD>\n""" % (ele.name, ele.value)
-            res = res + nvpair
-        res = res + "</MetaData>\n"
-        return res.encode('utf-8')
+            md = xml.SubElement(metadata, 'MD')
+            name = xml.SubElement(md, 'Name')
+            value = xml.SubElement(md, 'Value')
+            name.text = ele.name
+            value.text = ele.value
+        return xml.tostring(metadata, 'utf-8')
 
     def print_summary(self):
         print(self.metadata)
@@ -89,24 +87,15 @@ class GiftiLabelTable(object):
         return self.labels_as_dict
 
     def to_xml(self):
-        if len(self.labels) == 0:
-            return b"<LabelTable/>\n"
-        res = "<LabelTable>\n"
+        labeltable = xml.Element('LabelTable')
         for ele in self.labels:
-            col = ''
-            if not ele.red is None:
-                col += ' Red="%s"' % str(ele.red)
-            if not ele.green is None:
-                col += ' Green="%s"' % str(ele.green)
-            if not ele.blue is None:
-                col += ' Blue="%s"' % str(ele.blue)
-            if not ele.alpha is None:
-                col += ' Alpha="%s"' % str(ele.alpha)
-            lab = """\t<Label Key="%s"%s><![CDATA[%s]]></Label>\n""" % \
-                (str(ele.key), col, ele.label)
-            res = res + lab
-        res = res + "</LabelTable>\n"
-        return res.encode('utf-8')
+            label = xml.SubElement(labeltable, 'Label')
+            label.attrib['Key'] = str(ele.key)
+            label.text = ele.label
+            for attr in ['Red', 'Green', 'Blue', 'Alpha']:
+                if getattr(ele, attr.lower(), None) is not None:
+                    label.attrib[attr] = str(getattr(ele, attr.lower()))
+        return xml.tostring(labeltable, 'utf-8')
 
     def print_summary(self):
         print(self.get_labels_as_dict())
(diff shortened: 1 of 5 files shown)


nibabel/gifti/gifti.py
```diff
@@ -192,7 +193,7 @@ def data_tag(dataarray, encoding, datatype, ordering):
         raise NotImplementedError("In what format are the external files?")
     else:
         da = ''
-    return "<Data>" + da + "</Data>\n"
+    return ("<Data>" + da + "</Data>\n").encode('utf-8')
```
matthew-brett: I guess `da` is always a `str` (for Pythons 2 and 3)?

bcipolli: Yes; it is computed above, not passed it, so we can be confident.

matthew-brett:
Sorry - I have a feeling of exhaustion thinking about this, but I know that's not very helpful.

I guess the rule is that stuff output from the code in XML should always be bytes encoded in UTF-8, and stuff output as text from methods (not XML) should always be unicode / Python 3 str decoded from UTF-8?

bcipolli:
>I guess the rule is that stuff output from the code in XML should always be bytes encoded in UTF-8, and stuff output as text from methods (not XML) should always be unicode / Python 3 str decoded from UTF-8?

If you call `to_xml()`, you should get proper XML, which is bytes-encoded UTF-8. For other methods, I think you get bytes in Python2, unicode in Python 3. I think that's inline with the rest of the code--we don't force into unicode whenever we return a string type.

matthew-brett:
You've put quite a few `.encode` calls in, right?   These are always going to return unicode I think.   So the return type will sometimes (depending on the method) be Python 2 str, sometimes Python 2 unicode?

bcipolli:
`.encode` should return bytes; `.decode` should return unicode. I `.decode` to combine strings, then `.encode` on the way out of `.to_xml` to make sure I get `utf-8` bytes. I'm pretty confident that's right.

matthew-brett:
Sorry, got my encode / decodes the wrong way round.

I think it's clear what to do for the stuff that is going out to XML, that should always be bytes.

I'm worrying about the consistency of stuff that is coming in from XML as return arguments from methods - so I should have asked about the 'decode' calls - where you are returning unicode in Python 2.

Sorry if I'm not thinking clearly though.

effigies:
@matthew-brett I don't see any methods calling `.to_xml()` except other `.to_xml()` methods and the one situation in `giftiio.py` where the result is written to a binary-mode filehandle. So mixing `.to_xml()` bytes with unicode/Py3 str doesn't seem to be a risk.

Or am I misunderstanding the concern?

nibabel/gifti/gifti.py
```diff
@@ -280,21 +281,21 @@ def to_xml(self):
...
         dt_kind = data_type_codes.dtype[self.datatype].kind
         result += data_tag(self.data,
                            gifti_encoding_codes.specs[self.encoding],
                            KIND2FMT[dt_kind],
-                           self.ind_ord)
-        result = result + self.to_xml_close()
-        return result
+                           self.ind_ord).decode('utf-8')
+        result = result + self.to_xml_close().decode('utf-8')
+        return result.encode('utf-8')
```
effigies: This entire function is an encoded concatenation of decoded bytestrings. Can't all of these changes be dropped?

If you simply want to explicitly label this function as dealing in `bytes` and not `unicode`, you could use a `BytesIO` to accumulate:

    def to_xml(self):
        # fix endianness to machine endianness
        self.endian = gifti_endian_codes.code[sys.byteorder]
        result = BytesIO(self.to_xml_open())
        # write metadata
        if not self.meta is None:
            result.write(self.meta.to_xml())
        # write coord sys
        if not self.coordsys is None:
            result.write(self.coordsys.to_xml())
        # write data array depending on the encoding
        dt_kind = data_type_codes.dtype[self.datatype].kind
        result.write(data_tag(self.data,
                              gifti_encoding_codes.specs[self.encoding],
                              KIND2FMT[dt_kind],
                              self.ind_ord))
        result.write(self.to_xml_close())
        return result.getvalue()

While decoding and encoding will be slightly more expensive, writing to a BytesIO object will be slightly less.

effigies:
Does it make sense in the `.to_xml` functions (and helpers like `_arr2txt` and `data_tag`) to explicitly label Unicode strings with `u''` notation? In Python 2, you'll coerce up to `unicode` when needed, and in Python 3 it's just a string, but it might be a good tag, for hygiene.

bcipolli:
How about using the `xml` library to dump the XML? I think it cleans up the code nicely; @effigies @matthew-brett ?

If this looks good, we can use something similar for CIFTI.

effigies:
I think the idea of using an XML library rather than hand-generating it is a solid one, especially if it doesn't introduce a new dependency.

I haven't looked deeply at this diff, though. Is there a way to verify that the output of this and the previous output are the same (at least to the extent that parsing the files leads to identical data structures)?

bcipolli:
I believe the `xml` library is a standard one, and the interface I used wraps whatever parser is available (Expat by default) https://docs.python.org/2/library/xml.html

I counted on the GIFTI tests. I could beef those up if they're not sufficient. There were definitely errors when I screwed things up, but no guarantee they'd fully cover all cases.

effigies:
It looks like `tests/test_giftiio.py` do save-load loops and check consistency. I think that's probably fine. It doesn't test the XML structure itself, but if it produces the same numpy arrays, that seems like as good a verification.

nibabel/gifti/gifti.py
```diff
@@ -46,17 +47,14 @@ def get_metadata(self):
...
-\t<Value><![CDATA[%s]]></Value>
-</MD>\n""" % (ele.name, ele.value)
-            res = res + nvpair
-        res = res + "</MetaData>\n"
-        return res
+            md = xml.SubElement(metadata, 'MD')
+            name = xml.SubElement(md, 'Name')
+            value = xml.SubElement(md, 'Value')
+            name.text = ele.name
+            value.text = ele.value
```
effigies: Did we lose the `CDATA` bit here? Maybe:

    name.text = '<![CDATA[{0}]]>'.format(ele.name)
    value.text = '<![CDATA[{0}]]>'.format(ele.value)

bcipolli: I believe the `CDATA` bit is a way to avoid encoding strings as XML-safe. That trick isn't necessary; the library will XML encode.

In addition, if you do as you suggest, I believe the `<` and `>` characters will be escaped, and you'll get something you don't want.

effigies: You're right. Just did a little reading and playing in ipython.

effigies:
Okay, I've looked through and all of the changes make sense to me, and I do agree it looks much cleaner this way.

I will note that this does again run into encode-decode-encode loops, but in XML instead of Unicode. My impression of these functions is that this is plumbing, not API (but I could be wrong and the `bytes` output is non-negotiable), so would it make sense to reorganize so that `.to_xml()` returns XML `Element` objects? If so, that leaves three options that I see:

1) `GiftiImage.to_xml()` is an exception to the rule, and continues returning a `bytes` object.
2) Rename `GiftiImage.to_xml()` to `GiftiImage.to_xml_string()` (or `.to_bytes`?)
3) Replace the write command in `giftiio` like so: `f.write(xml.tostring(img.to_xml(), 'utf-8'))`.

bcipolli:
Thanks @effigies ! I agree the current encode/decode isn't the best. I considered the three solutions you gave, but did it this way as the current interface has been published.

It would be great to migrate the interface (since `CIFTI` support should keep a similar nomenclature). I'm just not sure how to migrate the return type...

I suppose a fourth option would be to have an internal `_to_xml_element` object that `to_xml` calls into.

effigies:
Fair. That last option also makes sense, but these are presumably rare enough operations that it's more a decision about aesthetics than efficiency, so I'd be inclined to leave it as is, unless it seems like a good idea to deprecate the public `.to_xml` interface anyway.

bcipolli:
I don't think `GIFTI.save()` should be too rare, so efficiency would be nice. Especially because I'd like to reuse the logic here for CIFTI, which could have great use (HCP data).

I also wouldn't want a public function to return an object type like `xml.etree.ElementTree.Element` when others could work just as well; having public functions be string makes sense too.

I'm leaning more towards having a `_to_xml_element` private function, which is documented to use the specific xml object model. The implementation was simple (I have it on my machine).

bcipolli:
@effigies I pushed two changes:
* Add `_to_xml_element` and an `XmlSerializable` interface
* Move code to `xml.py`, so that CIFTI support can follow and so that all references to `xml.etree.ElementTree` will be contained there. It would make changing the XML parser potentially easier.

Maybe it's overkill; I can back out those commits if y'all don't like.

nibabel/gifti/gifti.py
```diff
@@ -176,8 +162,9 @@ def print_summary(self):
         print('Affine Transformation Matrix: \n', self.xform)
 
 
-def data_tag(dataarray, encoding, datatype, ordering):
-    """ Creates the data tag depending on the required encoding """
+def _data_tag_element(dataarray, encoding, datatype, ordering):
+    """ Creates the data tag depending on the required encoding,
+    returns as bytes"""
```
effigies: returns as *XML Element

nibabel/gifti/gifti.py
```diff
@@ -192,10 +179,18 @@ def data_tag(dataarray, encoding, datatype, ordering):
         raise NotImplementedError("In what format are the external files?")
     else:
         da = ''
-    return "<Data>" + da + "</Data>\n"
+
+    data = xml.Element('Data')
+    data.text = da
+    return data
```
effigies: return xml.Element('Data', text=da)

bcipolli: Yeah, I read the docs and had tried this. On my machine it set an attribute: `<Data text='blah' />`.

nibabel/gifti/gifti.py
bcipolli: Fixed!

effigies:
To be clear, I meant rare as in "probably no more than one per second". But that interface doesn't strike me as overkill, especially if it's usable in CIFTI.

bcipolli:
@effigies Cool. Yes, this will be directly applicable to CIFTI, which is also XML-based.

bcipolli:
@matthew-brett Any ETA on when you'd be able to look over this PR? I believe @effigies looked overthe current design and didn't have any major objections.

This is one of two PRs I need in for a GIFTI `load`/`save` PR, and also for the CIFTI PR (#353 is the other).

Thanks!

matthew-brett:
Ben - so sorry - I have been completely swamped these last few days with various things.

I see you're in mid-refactor, and that I'm slowing you up.

Chris - would you consider being the the main reviewer for Ben's set of PRs on the XML-related code?   I mean, can you take responsibility for the PRs, and merge them when all comments are in?   In order not to hold Ben up?

effigies:
@matthew-brett I can, if you don't mind occasional pinging about specific questions (especially API issues). My current model of API is: if it's not prefaced with a `_`, it can't be removed without deprecation.

As to this PR, I have two questions that I was assuming would get answered when you reviewed:

1) The [removal of `GiftiCoordSystem.data_tag`](https://github.com/nipy/nibabel/pull/354/files#diff-efdef6e61829a924d939cdd10fe1748dL179) for `GiftiCoordSystem._data_tag_element`: Should this be a deprecation?

2) Are you comfortable with the [xmlutils](https://github.com/bcipolli/nibabel/blob/gifti-fix3/nibabel/xmlutils.py) going into `nibabel`? That's a big decision for me to make for not-my-project.

@bcipolli Conservatively (absent a contrary opinion from Matthew), I'd say we should address my first question by deprecating `data_tag` with a stub that just runs `xml.to_string(self._data_tag_element(...), 'utf-8')`. It can get removed in a major release.

bcipolli:
@effigies I believe it's `gifti.data_tag` - I don't believe that function belongs to any object. But regardless, it's so easy to deprecate and I think you're right we should do so... so I've updated! :)

bcipolli:
As for `xmlutils.py`, it could also be named as `xmlimages.py`; in the next iteration I refactor `SpatialImage` to extract out the file-based bits that are common across spatial and non-spatial images (which I call `FileBasedImage`). CIFTI and GIFTI could inherit `XmlImage`, which inherits from `FileBasedImage`. 

These utilities could live in an `xmlimages.py`, if that better fits the nibabel file schema.

matthew-brett:
I agree about the deprecation stuff - no `_` - then some kind of deprecation warning - as a guideline - exceptions allowed with a good argument.

By `xmlutils` - you mean the wrapper that is only a couple of methods at the moment?   No problem in general, your call.  Whatever leads to the greatest degree of simplicity and readability.

I'll try and pitch in when I can, but I'm getting ready to head out for a trip, so I may be a bit slow to reply.  Please do ping me though, past the point where you think it's polite, I'll do my best.

matthew-brett:
Chris, Ben - I've added you as maintainers - so you can both merge if you want to.

Chris - can I leave you to be the responsible reviewer on this set of PRs from Ben?    Ben, maybe you can be the responsible reviewer for Chris' PRs if he's also working on this stuff?

effigies:
@matthew-brett Yup, I'll be responsible reviewer on these. (Just realized you'd asked, not assigned.)

effigies:
@bcipolli Do you want to rebase these PRs?

bcipolli:
Will do, thanks!

effigies:
Overlap with #353. Another rebase needed.

bcipolli:
@effigies rebased, Travis tests pass.

nibabel/gifti/gifti.py
```diff
@@ -298,51 +299,35 @@ def from_array(klass,
...
-                      str(self.num_dim),
-                      str(di),
-                      gifti_encoding_codes.specs[self.encoding],
-                      gifti_endian_codes.specs[self.endian],
-                      self.ext_fname,
-                      self.ext_offset,
-                      )
-
-    def to_xml_close(self):
-        return "</DataArray>\n"
```
effigies: Stupid, but do you think we should have deprecated `to_xml_open` and `to_xml_close`? They were public, even if they were kind of weird. If so, I think the easiest way is probably just to restore them verbatim, since neither corresponds to a full XML element.

bcipolli: It's painful. But you're right, good call! Will work on this.

nibabel/gifti/tests/test_gifti.py
```diff
@@ -67,6 +66,17 @@ def test_dataarray():
         da = GiftiDataArray.from_array(bs_arr, 'triangle')
         assert_equal(da.datatype, data_type_codes[arr.dtype])
 
+    # Smoke test on deprecated functions
+    da = GiftiDataArray.from_array(np.ones((1,)))
```
effigies: Missing an intent code.

bcipolli: Thanks, fixed.

nibabel/gifti/tests/test_gifti.py
```diff
@@ -67,6 +66,17 @@ def test_dataarray():
         da = GiftiDataArray.from_array(bs_arr, 'triangle')
         assert_equal(da.datatype, data_type_codes[arr.dtype])
 
+    # Smoke test on deprecated functions
+    da = GiftiDataArray.from_array(np.ones((1,)), 'triangle')
+    with clear_and_catch_warnings() as w:
+        warnings.filterwarnings('always', category=DeprecationWarning)
+        assert_true(isinstance(da.to_xml_open(), string_types))
```
effigies: `string_types` needs to be imported from externals.six.

Rebase and I think we're almost done.

bcipolli: Thanks. Sorry for all the errors; I'm juggling changes three branches (trying to keep things simple/clean #fail) and I'm doing a poor job :)

Fixed, rebased, and pushed up. Looking forward to getting this one in; the next PR is the most interesting one! Looking forward to discussing!

effigies: No worries. Thanks for your patience. Heading out of town in a couple hours, so the next PR might have to wait for Monday, but I do want to get this in so you don't have to juggle too much more.

nibabel/gifti/gifti.py
```diff
@@ -299,26 +300,62 @@ def from_array(klass,
...
+            print(self.coordsys.print_summary())
+
+    @np.deprecate_with_doc("Use the metadata property instead.")
+    def get_metadata(self):
+        return self.meta.metadata
+
+    @property
+    def metadata(self):
+        """ Returns metadata as dictionary """
+        return self.meta.metadata
```
effigies: Is this reordering intentional? PR would be a little simpler if `print_summary` and `{get_,}metadata` went back below `to_xml_*`.

nibabel/gifti/tests/test_gifti.py
bcipolli: Thanks for all your work. No worries--that one is a WIP and will take some time. I've just been looking forward to beginning that discussion!

I made the change requested. Just more sloppiness from my side. Reordered & diff looks cleaner now.

effigies:
There we go. Thanks for all your work on these. Good luck with the next stage!

PR #358, opened 2015-10-15
BF: stride_tricks produces unbroadcastable array

effigies:
Hopefully this will fix the Travis errors.

### DIFF
--- a/nibabel/fileslice.py
+++ b/nibabel/fileslice.py
@@ -767,9 +767,15 @@ def strided_scalar(shape, scalar=0.):
     strided_arr : array
         Array of shape `shape` for which all values == `scalar`, built by
         setting all strides of `strided_arr` to 0, so the scalar is broadcast
-        out to the full array `shape`.
+        out to the full array `shape`. `strided_arr` is flagged as not
+        `writeable`.
+
+        The array is set read-only to avoid a numpy error when broadcasting -
+        see https://github.com/numpy/numpy/issues/6491
     """
     shape = tuple(shape)
     scalar = np.array(scalar)
     strides = [0] * len(shape)
-    return np.lib.stride_tricks.as_strided(scalar, shape, strides)
+    strided_scalar = np.lib.stride_tricks.as_strided(scalar, shape, strides)
+    strided_scalar.flags.writeable = False
+    return strided_scalar
--- a/nibabel/tests/test_fileslice.py
+++ b/nibabel/tests/test_fileslice.py
@@ -639,8 +639,14 @@ def test_strided_scalar():
         assert_equal(observed.shape, shape)
         assert_equal(observed.dtype, expected.dtype)
         assert_array_equal(observed.strides, 0)
-        observed[..., 0] = 99
-        assert_array_equal(observed, expected * 0 + 99)
+        # Strided scalars are set as not writeable
+        # This addresses a numpy 1.10 breakage of broadcasting a strided
+        # array without resizing (see GitHub PR #358)
+        assert_false(observed.flags.writeable)
+        def setval(x):
+            x[..., 0] = 99
+        # RuntimeError for numpy < 1.10
+        assert_raises((RuntimeError, ValueError), setval, observed)
     # Default scalar value is 0
     assert_array_equal(strided_scalar((2, 3, 4)), np.zeros((2, 3, 4)))
 


nibabel/parrec.py
```diff
@@ -609,7 +609,7 @@ def __getitem__(self, slicer):
                                  'F')
         # Broadcast scaling to shape of original data
         slopes, inters = self._slice_scaling
-        fake_data = strided_scalar(self._shape)
```
matthew-brett: Ah - no - that was deliberate to save memory - what is the error?

effigies: I figured that was the case. The error is here: https://travis-ci.org/nipy/nibabel/jobs/85392218#L977

Seems to be an immutability issue, but hard to diagnose, so this PR was me trying to use Travis to figure out if my guess was right.

matthew-brett: Ouch.  I guess that's numpy 1.10.

effigies: Yeah. Does it make sense to try to find another low-memory option, here, or is this something that needs to be addressed in numpy? And in the meantime, should we skip this test on `numpy >= 1.10`, use `np.zeros`, or put up with failing builds?

bcipolli: I think you just need to create a new array, as `self._shape` is probably read-only. Perhaps 
```
        fake_data = strided_scalar(self._shape.copy())
```
 will work.

effigies: Just tried. `self._shape` is a tuple, so that doesn't work. (Failed to try the right tests, locally. Want `test_proxy_api`, if you're playing along at home.)

bcipolli: This works:
`_, slopes, inters = np.broadcast_arrays(fake_data.copy(), slopes.copy(),
inters.copy())`

Just gotta narrow down from there.

effigies:
Of course Travis would pick now to start failing to even get to the tests...

Anyway, what we're looking at is [https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/numpy/lib/stride_tricks.py#L70](this line) which seems like it can be skipped by setting the `fake_data.flags.writeable` flag to `False`. We can either do this in `PARRECArrayProxy.__getitem__` as I show here or in `strided_scalar`. It seems safe to do the latter, since the array should never be writable, but figured I'd (a) check this works before digging around farther afield (b) get your thoughts.

nibabel/parrec.py
effigies: Yup, but that creates a writeable array. Unsetting the writeable flag was the needed trick.

effigies:
The [remaining AppVeyor error](https://ci.appveyor.com/project/Eric89GXL/nibabel/build/1.0.270/job/koby3voyqie0x5d8#L793) is completely different, though possibly numpy 1.10 related. Seems to be 32-bit Windows specific, and out of scope.

If the Travis builds succeed, I plan to just ignore that failure for Ben's PRs.

nibabel/parrec.py
```diff
@@ -610,6 +610,7 @@ def __getitem__(self, slicer):
         # Broadcast scaling to shape of original data
         slopes, inters = self._slice_scaling
         fake_data = strided_scalar(self._shape)
+        fake_data.flags.writeable = False
```
matthew-brett: How about

```
# Create broadcast array using writeable memory for numpy 1.10
fake_data = strided_scalar(np.array(self._shape))
```

?  Only because the writeable line looks a bit frightening.

effigies: I can try, but I really don't think the mutability of the shape tuple is relevant. 

On October 15, 2015 12:35:57 AM EDT, Matthew Brett <notifications@github.com> wrote:
>> @@ -610,6 +610,7 @@ def __getitem__(self, slicer):
>>          # Broadcast scaling to shape of original data
>>          slopes, inters = self._slice_scaling
>>          fake_data = strided_scalar(self._shape)
>> +        fake_data.flags.writeable = False
>
>How about
>
>```
># Create broadcast array using writeable memory for numpy 1.10
>fake_data = strided_scalar(np.array(self._shape))
>```
>
>?  Only because the writeable line looks a bit frightening.
>
>---
>https://github.com/nipy/nibabel/pull/358/files#r42082838

-- 
Chris Markiewicz

effigies:
Tried a few strategies to create writeable arrays, and none of those work in 1.10. Now trying to make the shape tuple writeable, but since it's immediately [cast to tuple](https://github.com/effigies/nibabel/blob/e2873a01ebca2120b42900033c4b5ee7dce918b3/nibabel/fileslice.py#L772), I don't anticipate improvement.

I think I have misstated the problem, so let me try my new understanding:

[`as_strided`](https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/numpy/lib/stride_tricks.py#L38) returns a writeable array. [`broadcast_arrays`](https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/numpy/lib/stride_tricks.py#L137) [calls](https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/numpy/lib/stride_tricks.py#L199) [_broadcast_to](https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/numpy/lib/stride_tricks.py#L57) on each array, and the `fake_data` array is the one it's getting hung up on, as demonstrated by 5576c23 (on the test that actually ran, at any rate). So when [this check](https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/numpy/lib/stride_tricks.py#L69) is called, it's because there is no opportunity to set `readonly` in `broadcast_arrays`, and `fake_data.writeable == True`, but for some reason `result` cannot be set `writeable`. (I suspect it's because `np.nditer(..., op_flags['readonly'], ...)` was used to construct `result`.)

It may be worthwhile to submit a PR, and we can use this as a test case, but I don't see anything on the `nibabel` end that will resolve this except by unsetting the `writeable` flag. I would be inclined to do this in `strided_array` itself, because it is simply a bunch of views on a scalar, so any changes to a single location change the entire array, which is almost certainly unwanted: if a new array with one memory address is wanted, then call `strided_array` again with the new scalar; if only one location is to be changed, a copy needs to be made anyway.

effigies:
@matthew-brett Just a ping, since I want to get this in before moving on Ben's PRs.

Since I don't see much alternative, the current version just follows my above reasoning and modifies `strided_array`, adds immutability to its documentation, and updates tests (which specifically tested mutability). Let me know if you disagree with that reasoning and what you think the best way forward is.

Left commits so you can see the changes and their effects on tests. Will squash to a single commit if you give the go-ahead.

nibabel/tests/test_fileslice.py
```diff
@@ -639,8 +639,10 @@ def test_strided_scalar():
         assert_equal(observed.shape, shape)
         assert_equal(observed.dtype, expected.dtype)
         assert_array_equal(observed.strides, 0)
-        observed[..., 0] = 99
-        assert_array_equal(observed, expected * 0 + 99)
+        def setval(x):
```
matthew-brett: Add comment explaining reason for test?

matthew-brett:
I think you're right, setting the flag is the only way.   I put in a numpy bug report : https://github.com/numpy/numpy/issues/6491

effigies:
Added a comment, one more test, and squashed to a single commit.

nibabel/tests/test_fileslice.py
```diff
@@ -639,8 +639,14 @@ def test_strided_scalar():
         assert_equal(observed.shape, shape)
         assert_equal(observed.dtype, expected.dtype)
         assert_array_equal(observed.strides, 0)
-        observed[..., 0] = 99
-        assert_array_equal(observed, expected * 0 + 99)
+        # Strided scalars are set as not writeable
+        # This addresses a numpy 1.10 breakage of broadcasting a strided
+        # array without resizing (see GitHub PR #358)
+        assert_false(x.flags.writeable)
```
bcipolli: @effigies this line is causing a test failure; `x` is not defined. Perhaps this should be inside `setval`?

effigies: Woops. `s/x/observed/`

effigies:
Fixed that new test and Travis is passing.

nibabel/fileslice.py
```diff
@@ -767,9 +767,12 @@ def strided_scalar(shape, scalar=0.):
     strided_arr : array
         Array of shape `shape` for which all values == `scalar`, built by
         setting all strides of `strided_arr` to 0, so the scalar is broadcast
-        out to the full array `shape`.
+        out to the full array `shape`. `strided_arr` is flagged as not
+        `writeable`.
```
matthew-brett: Still feel that we need a comment here - otherwise it's not obvious why this is true.

effigies: What kind of comment are you thinking? Noting the numpy bug (with some level of detail, or reference to this PR) or the reasoning that setting one element of the array would set all (and thus is probably not the intended behavior)?

matthew-brett: Just something like `We set the array read-only to avoid a numpy error when broadcasting - see https://github.com/numpy/numpy/issues/6491`.

effigies: Updated.

matthew-brett:
Thanks for your patience.

effigies:
Thanks!

PR #361, opened 2015-10-20
RF: remove imageopener decorator

effigies:
Popping this up from @matthew-brett's fork off #329.

Pros for the decorator:

* Adding allowed `valid_exts` and specifying how it should be opened happen in
  the same place (decorator at top of class);
* Using a decorator allows change of internal storage of the ImageOpener
  class.

Cons:

* Decorators can be hard to read, especially when the decorator is a function
  call (so the decorator returns a decorator);
* It's hard to see at a glance which extensions are valid, because the
  decorator is at the top of the class, and adds later to `valid_exts`,
  whereas the apparent definition of `valid_exts` is after the docstring;
* The decorator has to know about the internals of the class (in this case,
  the `valid_exts` attribute);
* The role of the decorator is to signal that a certain extension has to be
  opened in a certain way, but it seems a bit odd that this signal also adds
  the extension to the class.  To me it seems more natural to make this signal
  a line of its own, next to the `valid_exts` definition;
* Modifying the ImageOpener dictionary directly doesn't stop us from making
  the dictionary a property or another object type, and changing the
  underlying storage;
* No decorator means less code.

### DIFF
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -149,42 +149,10 @@ class Opener(object):
 
 
 class ImageOpener(Opener):
-    """ Opener-type class passed to image classes to collect compressed extensions
+    """ Opener-type class to collect extra compressed extensions
 
-    This class allows itself to have image extensions added to its class
-    attributes, via the `register_ex_from_images`.  The class can therefore
-    change state when image classes are defined.
+    A trivial sub-class of opener to which image classes can add extra
+    extensions with custom openers, such as compressed openers.
     """
+    # Add new extensions to this dictionary
     compress_ext_map = Opener.compress_ext_map.copy()
-
-    @classmethod
-    def register_ext_from_image(opener_klass, ext, func_def):
-        """Decorator for adding extension / opener_function associations.
-
-        Should be used to decorate classes. Updates ImageOpener class with
-        desired extension / opener association. Updates decorated class by
-        adding ```ext``` to ```klass.alternate_exts```.
-
-        Parameters
-        ----------
-        opener_klass : decorated class
-        ext : file extension to associate `func_def` with.
-          should start with '.'
-        func_def : opener function/parameter tuple
-          Should be a `(function, (args,))` tuple, where `function` accepts
-          a filename as the first parameter, and `args` defines the
-          other arguments that `function` accepts. These arguments must
-          be any (unordered) subset of `mode`, `compresslevel`,
-          and `buffering`.
-
-        Returns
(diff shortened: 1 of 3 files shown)


bcipolli:
Looks OK to me. My only concern is exposing `ImageOpener` details to images. 

What about just calling `ImageOpener.register_ext_from_image('.mgz', ImageOpener.gz_def)` as a function where you do the assignment now?

effigies:
I don't feel very strongly either way, so I think it's good that you popped in.

That does seem to be a decent compromise between either `ImageOpener` needing to know about image class internals, or image classes needing to know about `ImageOpener` internals. That said, dictionaries are simple. It's a question of an indirect `f(x, y)` vs `d.__setitem__(x, y)`.

To my mind a function makes sense if you want a constraint. For instance, we currently disallow setting an entry twice, which this PR does not preserve. If that's worth keeping, then I'd say we go ahead with a lightweight wrapper around `compressed_ext_map.__setitem__`, named something like `register_compressed_ext`.

matthew-brett:
I think we aren't exposing much, just that the ImageOpener has a dict
interface...​

bcipolli:
@matthew-brett After further review, I think you're right. I'm still not used to overriding operators in Python, it seems!

LGTM thanks!

effigies:
Will note again that we're losing the [non-overwrite constraint](https://github.com/nipy/nibabel/pull/361/files#diff-6ea16d18716da15eb1a0a16836a98644L185), but if you're both okay with that, I'm okay with that.

bcipolli:
Fine with me!

effigies:
Removed an unused line and adapted the usage docs from the decorator to the class itself.

nibabel/openers.py
```diff
@@ -149,42 +149,22 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        ImageOpener.compress_ext_map[ext] = func_def
+
+    ``ext`` is a file extension beginning with '.' and should be included in
+    the image class's ``valid_exts`` tuple.
+
+    ``func_def`` is a `(function, (args,))` tuple, where `function accepts a
+    filename as the first parameter, and `args` defines the other arguments
+    that `function` accepts. These arguments must be any (unordered) subset of
+    `mode`, `compresslevel`, and `buffering`.
```
bcipolli: Nice docstring.

bcipolli:
Re-reviewed, still LGTM!

PR #364, opened 2015-10-21
RF: move load/save logic out of SpatialImage

bcipolli:
This PR precedes #360, which builds off of this work. I'll copy some text from that PR; there's quite a bit of discussion about how this fits into forward-looking design of surface vs. spatial, binary vs. xml object models.

**Issue:**
The final issue blocking streamlined GIFTI/CIFTI in nibabel is: image file I/O functions are tied directly to the `SpatialImage` class; non-volumetric images (GIFTI/CIFTI) can't extend `SpatialImage` and so can't easily get in the `load`/`save` patheway.

**Proposed Solution:**
The good news is, the file I/O code actually is quite general. We can easily abstract the file I/O into a higher-level class (which I've called `FileBasedHeader`/`FileBasedImage`, as the more generic `Header` was already taken) that `SpatialImage` and `GiftiImage` inherit from.

**To do:**
- [X] Move file I/O from `SpatialImage` into `FileBasedHeader`/`FileBasedImage` 
- [X] Update tests that assume all image types are of type `SpatialImage`

**Notes:**
* Given the above design, I believe this is mostly done. Rather than fine-tuning the code now, I'd rather get ideas about the overall design first. @matthew-brett and @effigies , your input is most welcome when you have the time!
* Whatever we decide here, I'll be using to push forward GIFTI (#360) and CIFTI (#249) support

### DIFF
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -417,6 +417,22 @@ class FileBasedImage(object):
         img.to_filename(filename)
 
     @classmethod
+    def from_image(klass, img):
+        ''' Class method to create new instance of own class from `img`
+
+        Parameters
+        ----------
+        img : ``spatialimage`` instance
+           In fact, an object with the API of ``FileBasedImage``.
+
+        Returns
+        -------
+        cimg : ``spatialimage`` instance
+           Image, of our own class
+        '''
+        raise NotImplementedError()
+
+    @classmethod
     def is_valid_extension(klass, ext):
         valid = tuple(ft[1] for ft in klass.files_types) + klass.alternate_exts
         return ext.lower() in valid
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -6,7 +6,7 @@
 #   copyright and license terms.
 #
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
-''' Simple interfaces for images'''
+''' Common interface for any image format--volume or surface, binary or xml.'''
 
 try:
     basestring
@@ -68,11 +68,10 @@ class FileBasedHeader(object):
 
 class FileBasedImage(object):
     '''
-    It also has a ``header`` - some standard set of meta-data that is specific to
-    the image format, and ``extra`` - a dictionary container for any other
-    metadata.
+    This abstract image class defines an interface for loading/saving images
+    from disk. It doesn't define any image properties.
 
-    It has
+    It has:
 
     attributes:
 
@@ -84,18 +83,6 @@ class FileBasedImage(object):
        * affine
        * header
        * dataobj
-
-    You cannot slice an image, and trying to slice an image generates an
-    informative TypeError.
-    '''
-
-
-    '''
-    This abstract image class defines an interface for loading/saving images
-    from disk. It doesn't define any image properties.
-
-    It has
-
 
     methods:
        * .get_header() (deprecated, use header property instead)
@@ -110,6 +97,13 @@ class FileBasedImage(object):
        * from_file_map(fmap) - make instance from file map
        * instance_to_filename(img, fname) - save ``img`` instance to
          filename ``fname``.
+
+    It also has a ``header`` - some standard set of meta-data that is specific to
+    the image format, and ``extra`` - a dictionary container for any other
+    metadata.
+
+    You cannot slice an image, and trying to slice an image generates an
+    informative TypeError.
 
 
     There are several ways of writing data.
@@ -177,24 +171,6 @@ class FileBasedImage(object):
     carry the position at which a write (with ``to_files``) should place the
     data.  The ``file_map`` contents should therefore be such, that this will
     work:
-
-       >>> # write an image to files
-       >>> from io import BytesIO
-       >>> file_map = nib.AnalyzeImage.make_file_map()
-       >>> file_map['image'].fileobj = BytesIO()
-       >>> file_map['header'].fileobj = BytesIO()
-       >>> img = nib.AnalyzeImage(data, np.eye(4))
-       >>> img.file_map = file_map
-       >>> img.to_file_map()
-       >>> # read it back again from the written files
-       >>> img2 = nib.AnalyzeImage.from_file_map(file_map)
-       >>> np.all(img2.get_data() == data)
-       True
-       >>> # write, read it again
-       >>> img2.to_file_map()
-       >>> img3 = nib.AnalyzeImage.from_file_map(file_map)
-       >>> np.all(img3.get_data() == data)
-       True
     '''
     files_types = (('image', None),)
     alternate_exts = ()  # Modified by @ImageOpener.register_ext_from_image
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -76,9 +76,7 @@ class FileBasedImage(object):
     properties:
 
        * shape
-       * affine
        * header
-       * dataobj
 
     methods:
        * .get_header() (deprecated, use header property instead)
@@ -213,9 +211,7 @@ class FileBasedImage(object):
     def __getitem__(self):
         ''' No slicing or dictionary interface for images
         '''
-        raise TypeError("Cannot slice image objects; consider slicing image "
-                        "array data with `img.dataobj[slice]` or "
-                        "`img.get_data()[slice]`")
+        raise TypeError("Cannot slice image objects.")
 
     def get_header(self):
         """ Get header from image
@@ -412,34 +408,13 @@ class FileBasedImage(object):
 
         Parameters
         ----------
-        img : ``spatialimage`` instance
-           In fact, an object with the API of ``spatialimage`` - specifically
-           ``dataobj``, ``affine``, ``header`` and ``extra``.
+        img : ``any FileBasedImage`` instance
+
         filename : str
            Filename, implying name to which to save image.
         '''
         img = klass.from_image(img)
         img.to_filename(filename)
-
-    @classmethod
-    def from_image(klass, img):
-        ''' Class method to create new instance of own class from `img`
-
-        Parameters
-        ----------
-        img : ``spatialimage`` instance
-           In fact, an object with the API of ``spatialimage`` -
-           specifically ``dataobj``, ``affine``, ``header`` and ``extra``.
-
-        Returns
-        -------
-        cimg : ``spatialimage`` instance
-           Image, of our own class
-        '''
-        return klass(img.dataobj,
-                     img.affine,
-                     klass.header_class.from_header(img.header),
-                     extra=img.extra.copy())
 
     @classmethod
     def is_valid_extension(klass, ext):
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -8,16 +8,12 @@
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 ''' Common interface for any image format--volume or surface, binary or xml.'''
 
-try:
-    basestring
-except NameError:  # python 3
-    basestring = str
-
 import os.path
 import warnings
 
 import numpy as np
 
+from .externals.six import string_types
 from .fileholders import FileHolder
 from .filename_parser import (types_filenames, TypesFilenamesError,
                               splitext_addext)
@@ -400,7 +396,7 @@ class FileBasedImage(object):
         for key, ext in klass.files_types:
             file_map[key] = FileHolder()
             mapval = mapping.get(key, None)
-            if isinstance(mapval, basestring):
+            if isinstance(mapval, string_types):
                 file_map[key].filename = mapval
             elif hasattr(mapval, 'tell'):
                 file_map[key].fileobj = mapval
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -8,10 +8,7 @@
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 ''' Common interface for any image format--volume or surface, binary or xml.'''
 
-import os.path
 import warnings
-
-import numpy as np
 
 from .externals.six import string_types
 from .fileholders import FileHolder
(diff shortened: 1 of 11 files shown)


nibabel/spatialimages.py
```diff
@@ -312,24 +308,18 @@ def supported_np_types(obj):
     return set(supported)
 
 
-class ImageDataError(Exception):
-    pass
+class Header(SpatialHeader):
+  '''Alias for SpatialHeader; kept for backwards compatibility.'''
```
effigies: Should this be marked deprecated? e.g.

    def __init__(self, *args, **kwargs):
        warnings.warn('Header is deprecated, use SpatialHeader', DeprecationWarning)
        SpatialHeader.__init__(self, *args, **kwargs)

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,569 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+''' Simple interfaces for images'''
```
effigies: Maybe 'Common interface for binary and XML image formats'?

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,569 @@
...
+
+try:
+    basestring
+except NameError:  # python 3
+    basestring = str
```
effigies: from .externals.six import string_types

And replace `basestring` with `string_types` throughout.

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,569 @@
...
+       >>> # read it back again from the written files
+       >>> img2 = nib.AnalyzeImage.from_file_map(file_map)
+       >>> np.all(img2.get_data() == data)
+       True
+       >>> # write, read it again
+       >>> img2.to_file_map()
+       >>> img3 = nib.AnalyzeImage.from_file_map(file_map)
+       >>> np.all(img3.get_data() == data)
+       True
+    '''
```
effigies: Much of these docs seem to be specific to `SpatialImage`.

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,569 @@
...
+        return self._header
+
+    def __getitem__(self):
+        ''' No slicing or dictionary interface for images
+        '''
+        raise TypeError("Cannot slice image objects; consider slicing image "
+                        "array data with `img.dataobj[slice]` or "
+                        "`img.get_data()[slice]`")
+
+    def get_header(self):
```
effigies: Mark deprecated?

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,569 @@
...
+        """ Get header from image
+
+        Please use the `header` property instead of `get_header`; we will
+        deprecate this method in future versions of nibabel.
+        """
+        return self.header
+
+    def get_filename(self):
```
effigies: Worth propertizing `filename`? Or keep this?

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,569 @@
...
+
+        Returns
+        -------
+        cimg : ``spatialimage`` instance
+           Image, of our own class
+        '''
+        return klass(img.dataobj,
+                     img.affine,
+                     klass.header_class.from_header(img.header),
+                     extra=img.extra.copy())
```
effigies: Does this work for non-`SpatialImage`s?

nibabel/spatialimages.py
```diff
@@ -136,15 +136,11 @@
     basestring
 except NameError:  # python 3
     basestring = str
```
effigies: `basestring` unused in this file.

nibabel/spatialimages.py
```diff
@@ -312,24 +308,18 @@ def supported_np_types(obj):
     return set(supported)
 
 
-class ImageDataError(Exception):
-    pass
+class Header(SpatialHeader):
+  '''Alias for SpatialHeader; kept for backwards compatibility.'''
+  pass
```
effigies: 4-space indentation.

bcipolli:
Thanks @effigies! 

A question: for some comments, it's about old code. Most of the code is copy-pasted from `spatialimage.py`; it seemed dangerous to move a bunch of code, and make changes at the same time (as git doesn't really track those changes as different).

I'm happy to deal with those things here, but I wanted to double-check first. Thanks!

effigies:
Could be a separate PR, but my thought is that while people are looking at code is a good time to make changes. And the changes will be tracked in the commits, if not the full PR diff.

nibabel/spatialimages.py
bcipolli: :+1:

nibabel/filebasedimages.py
bcipolli: I put: `''' Common interface for any image format--volume or surface, binary or xml.'''`

nibabel/spatialimages.py
effigies: I should probably note that you'll want `stacklevel=2` in that warning.

nibabel/filebasedimages.py
bcipolli: :+1:

nibabel/filebasedimages.py
bcipolli: Moved this to the `SpatialImage` class docstring. Also fixed up the docs; they were a bit hacked up.

nibabel/filebasedimages.py
bcipolli: :+1:

bcipolli: There are a number of uses of `get_header` in the code that I'll change over to `header` as well.

effigies: Think it's just ecat? The `test_image_api` bit should stay until `get_header` is removed entirely.

nibabel/filebasedimages.py
bcipolli: We can propertize it and document that we'll deprecate it in the future.

effigies: This one I want @matthew-brett's comment on. Since other get/sets in the vicinity seem to have been deprecated, there might be a good reason this one hasn't been.

nibabel/filebasedimages.py
bcipolli: No. Anything with `dataobj` or `affine` should not appear in `filebasedimage`. There are also some dirty docs; fixing this now.

nibabel/spatialimages.py
bcipolli: :+1:

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,538 @@
...
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+''' Common interface for any image format--volume or surface, binary or xml.'''
+
+from ..externals.six import string_types
```
effigies: `.externals`, not `..externals`

bcipolli: Yep. I pushed before running tests; give me a min..

bcipolli:
OK @effigies ; cleaned up and ran tests; pls take a look!

bcipolli:
doctests are failing; will work on those now.

nibabel/filebasedimages.py
matthew-brett: Sorry - might be moot now - but `get_filename` is different because it returns the canonical filename for the image - for example will return `image.img` when input filename might have been `image.hdr`, for a Nifti Pair.

effigies: For clarity, you think it should stay as `{get,set}_filename`? Or does the entire `filename` interface seem irrelevant, now?

matthew-brett: On Wed, Oct 21, 2015 at 4:01 PM, Chris Markiewicz <notifications@github.com>
wrote:

> In nibabel/filebasedimages.py
> <https://github.com/nipy/nibabel/pull/364#discussion_r42674196>:
>
> > +    def __getitem__(self):
> > +        ''' No slicing or dictionary interface for images
> > +        '''
> > +        raise TypeError("Cannot slice image objects; consider slicing image "
> > +                        "array data with `img.dataobj[slice]` or "
> > +                        "`img.get_data()[slice]`")
> > +
> > +    def get_header(self):
> > +        """ Get header from image
> > +
> > +        Please use the `header` property instead of `get_header`; we will
> > +        deprecate this method in future versions of nibabel.
> > +        """
> > +        return self.header
> > +
> > +    def get_filename(self):
>
> For clarity, you think it should stay as {get,set}_filename? Or does the
> entire filename interface seem irrelevant, now?
>

Sorry not to be clear - I think, while the FileBasedImage can have more
than one filename used to identify the file (e.g '.img', '.hdr') it should
stay as `get/set_filename`...

bcipolli:
Made the code updates. A bit worried about deprecating `Header` now; should we simply document that they're equivalent and it will be deprecated (a la `get_header()`)?

Also a bit concerned about changing the base class of images from `Header` to `SpatialHeader`. Would this break `pickle`d headers?

effigies:
@bcipolli Regarding 9cc27ea: For images where there's a `filename` interface (basically, if there CAN be multiple files, then return the canonical file), then `{get,set}_` are [preferred](https://github.com/nipy/nibabel/pull/364#discussion_r42674598). So that commit can be stripped out.

Is there any possibility that XML-based datasets will have multiple files (so a canonical filename is needed)? If not, then `{get,set}_filename` could be moved back into `SpatialImage`. But I don't think there's anything wrong with leaving them in `FileBasedImage`, either.

effigies:
> A bit worried about deprecating `Header` now; should we simply document that they're equivalent and it will be deprecated (a la `get_header`())?

To be honest, I'm not quite sure what the difference is between posting an intent to deprecate and actually deprecating but not removing. Documentation should be updated ASAP, and `DeprecationWarning`s are silent by default.

> Also a bit concerned about changing the base class of images from `Header` to `SpatialHeader`. Would this break pickled headers?

Good question. Is pickling really a thing we care about? The only case I can see is in `tests/test_analyze.py`. I guess I'm confused why you'd pickle an image rather than save the image. But we can try pickling under `Header` and loading under `SpatialHeader`, to be safe.

In any event, I'm also okay with leaving things as subclassed from `Header` and not `SpatialHeader`.

bcipolli:
>that commit can be stripped out.

Done

>Is there any possibility that XML-based datasets will have multiple files

I don't see any reason to keep it off of `FileBasedImage`; there's nothing to suggest surface images couldn't use a similar approach at some point. CIFTI does something close (using multiple sub-extensions to specify different data types related to a study), but not quite.

>Is pickling really a thing we care about?

I'm not sure; perhaps some people or tools use `pickle` or `shelve` to store working sessions.

>In any event, I'm also okay with leaving things as subclassed from Header and not SpatialHeader.

I'd prefer not to do that; I think that introduces complexity without much of the benefits. I suggest we either use `SpatialHeader` and deprecate `Header`, or just have `Header` without `SpatialHeader.

I'll try this out.

bcipolli:
@effigies I removed the commit and tested `shelve` (which uses `pickle`) manually (create a `Minc1Header` in `master`, load up in this branch. No issue.

effigies:
Okay. So no barrier to deprecating `Header` for `SpatialHeader`?

bcipolli:
Not that I see. I agree that deprecating directly is a fine user experience, from my POV

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,537 @@
...
+    contain enough information so that an existing image instance can save
+    itself back to the files pointed to in ``file_map``.  When a file holder
+    holds active file-like objects, then these may be affected by the
+    initial file read; in this case, the contains file-like objects need to
+    carry the position at which a write (with ``to_files``) should place the
+    data.  The ``file_map`` contents should therefore be such, that this will
+    work:
+    '''
+    files_types = (('image', None),)
+    alternate_exts = ()  # Modified by @ImageOpener.register_ext_from_image
```
effigies: `valid_exts`

nibabel/filebasedimages.py
```diff
@@ -0,0 +1,537 @@
...
+        -------
+        cimg : ``spatialimage`` instance
+           Image, of our own class
+        '''
+        raise NotImplementedError()
+
+    @classmethod
+    def is_valid_extension(klass, ext):
+        valid = tuple(ft[1] for ft in klass.files_types) + klass.alternate_exts
+        return ext.lower() in valid
```
effigies: `is_valid_extension` was removed from `SpatialImage` in 0c06464.

nibabel/filebasedimages.py
bcipolli: `_meta_sniff_len` is optional, and on the header.

`valid_exts` is required, but is also required to be defined by each subclass. Defining it here makes it easier to forget to do that, so I excluded it.

effigies: `sizeof_hdr` is on the header, [`_meta_sniff_len`](https://github.com/nipy/nibabel/pull/364/files#diff-1053478051df4b99b0c69a5365f5f0beL326) is on the image and used in [`path_maybe_image`](https://github.com/nipy/nibabel/pull/364/files#diff-496e321b697d515cfea0150d3c97dfb5R530) without being checked to exist, first. (It does check whether the header class defines `may_contain_header`, but that's not tied to presence of `_meta_sniff_len`, AFAICT.)

`files_types` must also be defined by subclasses, correct? I think its purpose here is documentary. But the point of my comment was that `alternate_exts` has been removed from `SpatialImage`.

effigies:
Had another read through, and once we agree on the above issues, I think this can go in.

nibabel/filebasedimages.py
bcipolli: :+1:

bcipolli:
@effigies Made the tweaks requested, tests pass... I think we're good to go! Excited! This is the last refactoring needed to really jump into #249 for CIFTI :)

effigies:
Pushed a PEP8 cleanup for `filebasedimages.py`, as that was easier than pointing them out. Have a suspicion removing the `numpy` import will break the doctests. If so, I'll put it into the doctests themselves.

bcipolli:
>Pushed a PEP8 cleanup for filebasedimages.py, as that was easier than pointing them out.

Agree, thanks for doing that!

effigies:
Alright. If you're happy with that commit (and the tests pass), in it goes.

bcipolli:
:+1:

PR #365, opened 2015-10-23
RF: Use XmlSerializable, XmlParser for Gifti load/save

bcipolli:
`CiftiImage` (#249) and `GiftiImage` share a lot of logic. Both have functions for parsing, both will use similar logic for loading and saving. `XmlBasedImage` and `XmlImageParser` encapsulates the shared code, and defines a (hopefully) clean object model for any future images that are defined in XML.

* `XmlImageParser` - a new abstract class that does most of the work here. Defines a structure for parsing XML files; every xml-based image will have to define one, specific to that image format. This essentially generalizes the object model used in `parse_gifti_fast.py`.
* `XmlBasedImage` - a light wrapper around `FileBasedImage` that calls into the `XmlImageParser` for the image class on read, and takes care of small details on `write` (call into `to_xml()`). New required class property: `parser` (similar usage to `header`)

Things done:
* [x] Define `XmlImageParser` and `XmlBasedImage`
* [x] Migrate `GiftiImage` to use `XmlBasedImage`, and `parse_gifti_fast` to use `XmlImageParser`
* [x] Deprecate `parse_gifti_file` and related objects.

Things to do:
* [x] Determine the file structure; there is an unresolvable dependency between `gifti.py` and `parse_gifti_fast.py`; each refers to objects in the other. Perhaps we can simply combine them?
* [x] Write tests
* [x] Migrate `parse_gifti_fast` tests to new object model.
* [x] Add deprecation tests for `parse_gifti_file`
* [x] Add documentation
* [x] Scrub GIFTI parser code
* [x] Add test to test `buffer_size` passing via `nibabel.load()`

I'm pushing this PR out there now as (1) it seems to pass tests and (2) to get feedback on the object model / design as early in the development process as possible.

I'm not in a big rush on this; just wanted to get it out there while it was on my mind.


bcipolli:
Sorry, this depends on #364 ; the diff will not be useful until that's merged; here's an easier diff for now: https://github.com/bcipolli/nibabel/compare/gifti-loadsave...bcipolli:xmlbasedimages

nibabel/gifti/gifti.py
```diff
@@ -386,14 +385,44 @@ def metadata(self):
         return self.meta.metadata
 
 
-class GiftiImage(FileBasedImage, xml.XmlSerializable):
+class GiftiImage(xml.XmlBasedImage):
+    """
+    The Gifti spec suggests using the following suffixes to your
```
bcipolli: Here, I moved the docstring from `GiftiImage.to_file_map` to `GiftiImage`, so that `to_file_map` could be moved to `XmlBasedImage`.

nibabel/gifti/gifti.py
```diff
@@ -386,14 +385,44 @@ def metadata(self):
...
 
     def __init__(self, header=None, extra=None, file_map=None, meta=None,
                  labeltable=None, darrays=None, version="1.0"):
-        FileBasedImage.__init__(self, header=header, extra=extra,
-                                file_map=file_map)
+        super(GiftiImage, self).__init__(header=header, extra=extra,
+                                         file_map=file_map)
+        # placed here temporarily for git diff purposes
+        from .parse_gifti_fast import GiftiImageParser
+        GiftiImage.parser = GiftiImageParser
```
bcipolli: Temporary trickery to avoid recursive imports; it's in the "todo" checklist.

nibabel/gifti/gifti.py
```diff
@@ -501,7 +530,7 @@ def getArraysFromIntent(self, intent):
 
     def print_summary(self):
         print('----start----')
-        print('Source filename: ', self.filename)
+        print('Source filename: ', self.get_filename())
```
bcipolli: `GiftiImage` was storing `filename` on it's own. It should use the mechanisms in place for `FileBasedImage`.

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -71,14 +70,12 @@ def read_data_block(encoding, endian, ordering, datatype, shape, data):
     return newarr
 
 
-class Outputter(object):
+class GiftiImageParser(XmlImageParser):
```
bcipolli: I used the structure of this `Outputter` class, which is duplicated in CIFTI (#249), as the structure for an `XmlImageParser` class.

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -95,7 +92,6 @@ def initialize(self):
 
         # where to write CDATA:
         self.write_to = None
-        self.img = None
```
bcipolli: super class has this.

nibabel/xmlbasedimages.py
```diff
@@ -0,0 +1,146 @@
...
+"""
+Thin layer around xml.etree.ElementTree, to abstract nibabel xml support.
+"""
+from xml.etree.ElementTree import Element, SubElement, tostring
+from xml.parsers.expat import ParserCreate
+
+from .filebasedimages import FileBasedHeader, FileBasedImage
+
+
+class XmlSerializable(object):
```
bcipolli: Moved here from `xmlutils.py`

nibabel/xmlbasedimages.py
```diff
@@ -0,0 +1,146 @@
...
+        (default: utf-8)"""
+        return tostring(self._to_xml_element(), enc)
+
+
+class XmlBasedHeader(FileBasedHeader, XmlSerializable):
+    """ Basic wrapper around FileBasedHeader and XmlSerializable."""
+    pass
+
+
+class XmlImageParser(object):
```
bcipolli: New class, based on `parse_gifti_fast.py:Outputter`

nibabel/xmlbasedimages.py
```diff
@@ -0,0 +1,146 @@
...
+        self.buffer_size = buffer_size
+        self.img = None
+
+    def _create_parser(self):
+        """Internal function that allows subclasses to mess
+        with the underlying parser, if desired."""
+
+        parser = ParserCreate(encoding=self.encoding)  # from xml package
+        parser.buffer_text = True
+        parser.buffer_size = self.buffer_size
```
bcipolli: Old code checked before setting, but that's for pre-2.6 compat. Simplify the code...

bcipolli: `buffer_size` makes a significant difference in performance, so it is important to keep.

nibabel/xmlbasedimages.py
```diff
@@ -0,0 +1,146 @@
...
+        raise NotImplementedError
+
+    def EndElementHandler(self, name):
+        raise NotImplementedError
+
+    def CharacterDataHandler(self, data):
+        raise NotImplementedError
+
+
+class XmlBasedImage(FileBasedImage, XmlSerializable):
```
bcipolli: Moved here from `xmlutils.py`

nibabel/xmlbasedimages.py
```diff
@@ -0,0 +1,146 @@
...
+        -------
+        None
+        """
+        if file_map is None:
+            file_map = self.file_map
+        f = file_map['image'].get_prepare_fileobj('wb')
+        f.write(self.to_xml())
+
+    @classmethod
+    def from_file_map(klass, file_map, buffer_size=35000000):
```
bcipolli: Moved here from `GiftiImage`. Added `buffer_size` parameter so that `buffer_size` can be specified from `nibabel.load`. Should add a test to make sure this works.

nibabel/xmlbasedimages.py
```diff
@@ -0,0 +1,146 @@
...
+        -------
+        img : GiftiImage
+            Returns a GiftiImage
+         """
+        img = klass.parser(buffer_size=buffer_size).parse(
+            fptr=file_map['image'].get_prepare_fileobj('rb'))
+        return img
+
+    @classmethod
+    def from_filename(klass, filename, buffer_size=35000000):
```
bcipolli: Added this function so that `buffer_size` can be passed from `load`. This is similar to how other classes deal with optional args--override the base class.

bcipolli:
Checked off all items but one: not sure the best way to avoid the recursive imports here. Should be ready for a looksie anytime--I even tried scrubbing some of the old code :)

bcipolli:
@effigies I put `CiftiImage` on top of this code this morning. Turns out that `CiftiImage` is a `Nifti2Image` with an xml-based *extension*; it is not an Xml-based image. This changed two things:

* `CiftiImage` will not subclass `XmlBasedImage`; only `GiftiImage` will.
* The `CiftiImage` extension will need an xml parser. Using the object model here was great, but `XmlImageParser` is too specific a name now. I am thinking to change to `XmlParser`.

I think the object model and abstractions here are good; I'd like to keep them. But, since `GiftiImage` is the only one using `XmlBasedImage`, it could be considered overkill. Alternately, I could remove `XmlBasedImage`, move logic back to `GiftiImage, keep `xmlutils.py`, and simply add the parser there.

effigies:
I suppose I'm inclined to not bother with `XMLBasedImage` if `GiftiImage` is its only subclass, but I should make it clear that I have not even started to look at this PR. (Busy week. I don't expect to be useful on here until Monday at the earliest.)

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -309,64 +319,23 @@ def flush_chardata(self):
...
-    if fptr is None:
-        with open(fname, 'rb') as datasource:
-            return parse_gifti_file(fptr=datasource, buffer_size=buffer_size)
-    else:
-        datasource = fptr
+class Outputter(GiftiImageParser):
+    @np.deprecate_with_doc("Use GiftiImageParser instead.")
+    def __init__(self, *args, **kwargs):
+        super(Outputter, self).__init__(*args, **kwargs)
+        self.img = None
```
effigies: `Outputter` had an `initialize` method that should be preserved.

effigies:
Sorry for the delay. This looks reasonable to me.

The main design issue I see is still whether or not we want an `XMLBasedImage` with only one subclass. In favor of `XMLBasedImage`, the decoupling work has already been done, and the alternative is to stick generic code into `GiftiImage`. In opposition, now you have to read another class to understand `GiftiImage`.

I think I can go either way. If anybody else has strong opinions, I can probably be pushed to one side or the other.

bcipolli:
I would prefer to move the code back into `GiftiImage`, rename `XmlImageParser` to `XmlParser` (as it's the CIFTI header that's XML-based), and move `xmlimages.py` back to `xmlutils.py`.

effigies:
That sounds good to me.

bcipolli:
@effigies I did as outlined above, and deprecated `Outputter.initialize`. Tests pass as well.

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -309,64 +321,25 @@ def flush_chardata(self):
...
-    except ExpatError:
-        print('An expat error occured while parsing the  Gifti file.')
-
-    # Reality check for pending data
-    assert out.pending_data is False
-    # update filename
-    out.img.filename = fname
-    return out.img
+    def initialize(self, *args, **kwargs):
+        self.__init__(*args, **kwargs)
```
effigies: `Outputter.__init__` doesn't quite do the same thing as `Outputter.initialize`, since it also calls `super(GiftiImageParser, self).__init__`, which could change the buffer size and all that. I would just do the [original contents](https://github.com/bcipolli/nibabel/blob/xmlbasedimages/nibabel/gifti/parse_gifti_fast.py#L84-L102), but short, like:

    def initialize(self):
        """ Initialize outputter
        """
        self.fsm_state = []
        self.nvpair = self.da = self.coordsys = self.lata = self.label = None
        self.meta_global = self.meta_da = self.write_to = None
        self._char_blocks = None
        self.count_da = True

bcipolli: Setting `encoding` and `buffer_size` is new. I figured since `initialize` is deprecated, nobody will call it with those args, nor would those args be non-defaults. I'd prefer to keep the # of lines for deprecated functions minimal.

As a side note, I could remove the arguments from `initialize` and `__init__` to match the original definitions.

LMK what you think.

effigies: I would keep the original prototypes and docstring on `initialize`. But point taken on only defaults being used in any case, so revert the prototypes and I'm okay here.

bcipolli:
Done.

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -309,64 +321,27 @@ def flush_chardata(self):
...
-    """
-    assert (fname is not None) + (fptr is not None) == 1, "Specify only fname or fptr, not both"
-
-    if fptr is None:
-        with open(fname, 'rb') as datasource:
-            return parse_gifti_file(fptr=datasource, buffer_size=buffer_size)
-    else:
-        datasource = fptr
+class Outputter(GiftiImageParser):
+    @np.deprecate_with_doc("Use GiftiImageParser instead.")
```
effigies: Should this deprecation be on the class instead of `__init__`?

bcipolli: I probably don't understand what deprecating the class would do; could you help make that clear to me?

In previous cases I remember deprecating `__init__` so that the deprecation is lazy--it only triggers if someone instantiates the class, and not when the file is simply imported. That's the only goal here.

effigies: I thought the `@np.deprecate` functions modify documentation. Do they also raise warnings? If so, this makes sense.

bcipolli: yep, they raise warnings--I add a deprecation warning test case each time we add this decorator, to make sure.

effigies:
Okay, LGTM then.

Anybody else have comments?

bcipolli:
Any chance we can merge this one?

nibabel/gifti/gifti.py
yarikoptic: Ask to-do s are checked... So decided just to stick to this implementation?

bcipolli: Yep. the image and parsing code refer to each other. I think it helps to keep them as separate files, to keep files short. So this is necessary.

The same design is used in the Cifti code. There, it's even more helpful to keep the object defs and parsing code separate, as both files are relatively long.

effigies:
I think people have had a chance to voice objections. In it goes.

matthew-brett:
Yes, sorry I haven't had a chance to review.   Next week is better for me, by Tuesday afternoon I will have more time.

nibabel/gifti/gifti.py
```diff
@@ -386,14 +386,44 @@ def metadata(self):
...
     valid_exts = ('.gii',)
     files_types = (('image', '.gii'),)
 
     def __init__(self, header=None, extra=None, file_map=None, meta=None,
                  labeltable=None, darrays=None, version="1.0"):
-        FileBasedImage.__init__(self, header=header, extra=extra,
-                                file_map=file_map)
+        super(GiftiImage, self).__init__(header=header, extra=extra,
+                                         file_map=file_map)
+        # placed here temporarily for git diff purposes
```
matthew-brett: Rewrite comment now you've decided in this implementation?

nibabel/gifti/gifti.py
```diff
@@ -562,40 +576,31 @@ def to_file_map(self, file_map=None):
...
         if file_map is None:
             file_map = self.file_map
         f = file_map['image'].get_prepare_fileobj('wb')
         f.write(self.to_xml())
+
+    @classmethod
+    def from_file_map(klass, file_map, buffer_size=35000000):
+        """ Load a Gifti image from a file_map
+
+        Parameters
```
matthew-brett: Not numpy-docstring compliant.

Also - file_map is not a string, but a dictionary, containing fileholders.

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -58,27 +56,31 @@ def read_data_block(encoding, endian, ordering, datatype, shape, data):
...
-class Outputter(object):
+class GiftiImageParser(XmlParser):
 
-    def __init__(self):
-        self.initialize()
+    def __init__(self, encoding=None, buffer_size=35000000, verbose=0):
+        super(GiftiImageParser, self).__init__(encoding=encoding,
+                                               buffer_size=buffer_size,
+                                               verbose=verbose)
+        # output
```
matthew-brett: What does 'output' mean here?

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -309,64 +321,27 @@ def flush_chardata(self):
...
-    out = Outputter()
-    for name in HANDLER_NAMES:
-        setattr(parser, name, getattr(out, name))
-    try:
-        parser.ParseFile(datasource)
-    except ExpatError:
-        print('An expat error occured while parsing the  Gifti file.')
-
-    # Reality check for pending data
-    assert out.pending_data is False
```
matthew-brett: Worth preserving something like this check?

effigies:
@matthew-brett Should I revert this merge?

matthew-brett:
Oh - no - sorry - you did good to merge, thanks for doing it.  I was only adding post-merge suggestions.  Happy to do that as a PR if it's easier.

PR #368, opened 2015-10-28
[WIP] BF: Check for attribute existence before value

### LINKED ISSUE #366: Tests: failure of `nisext.tests.test_doctest_markup.test_byter` with Python 3.5
The test `nisext.tests.test_doctest_markup.test_byter` fails with nibabel 2.0.1. Only fails with Python 3.5 only.

http://hydra.nixos.org/build/27003289/nixlog/1/raw

effigies:
Indentation seemed to be erroneous, as the presence of keyword arguments should not be prerequisite for `*args` and `**kwargs`.

Closes #366.

### DIFF
--- a/nisext/codegen.py
+++ b/nisext/codegen.py
@@ -174,14 +174,14 @@ class SourceGenerator(NodeVisitor):
                 paren_or_comma()
                 self.write(keyword.arg + '=')
                 self.visit(keyword.value)
-            if node.starargs is not None:
-                paren_or_comma()
-                self.write('*')
-                self.visit(node.starargs)
-            if node.kwargs is not None:
-                paren_or_comma()
-                self.write('**')
-                self.visit(node.kwargs)
+        if hasattr(node, 'starargs') and node.starargs is not None:
+            paren_or_comma()
+            self.write('*')
+            self.visit(node.starargs)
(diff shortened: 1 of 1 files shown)


nisext/codegen.py
```diff
@@ -174,14 +174,14 @@ def paren_or_comma():
...
                 self.visit(keyword.value)
-            if node.starargs is not None:
-                paren_or_comma()
-                self.write('*')
-                self.visit(node.starargs)
-            if node.kwargs is not None:
-                paren_or_comma()
-                self.write('**')
-                self.visit(node.kwargs)
+        if hasattr(node, 'starargs'):
```
bcipolli: Will this cause downstream problems if `node.starargs` is `None`?

bcipolli:
Is there test coverage of this codepath? It is explicitly omitted from `.coveragerc`.

nisext/codegen.py
effigies: Woops, fixed.

effigies:
Added the check in the wrong place, but if there's a possibility of failure in `visit_Call`, might there not also be a possibility in `visit_ClassDef`?

Someone with more experience fiddling with the AST should probably review this.

PR #369, opened 2015-10-28
RF: remove doctest string markup

matthew-brett:
The doctest string markup was for allowing doctests running on Python 3
and Python < 2.6.   We haven't used it for a few years.  This commit
removes that code in the hope that no-one else is using it.


matthew-brett:
A google and OpenHub search for uses of the `nisext` package and `py3builder` doesn't come up with anything outside our little world, so I think this is safe to merge.

effigies:
No objections.

matthew-brett:
Last call for comments?

PR #371, opened 2015-10-31
RF: Move extension_codes to within Nifti1Extensions.

bcipolli:
Note: this is taken from https://github.com/satra/nibabel/pull/2 ; an implementation extending `extension_codes` for `CIFTI` can be found there.

`nifti1.extension_codes` is fixed (for the `Nifti1Image1` format, the list is fixed) and extensible (new `Nifti1Extension`s can be declared for image formats that extend `Nifti1Image`).

To support extensibility via encapsulation, I am suggesting `extension_codes` be a property of the `*Extensions` object. That way, `Nifti1Extensions.extension_codes` can remain fixed, while new image formats can extend this (e.g. `CiftiExtensions.extension_codes` extends `Nifti1Extensions` and adds extension codes). 

This should avoid potential side-effects of adding new `extension_codes` for new image formats.

***Notes:***

I found the implementation of `Nifti1Extension` and `Nifti1Extensions` to be a bit challenging to work with. `Nifti1Extension` seems like it should stand alone, as each extension is independent. But it doesn't; it refers to `extension_codes`--the entire list of `Nifti1Extension`s. `Nifti1Extensions` is a wrapper around that list of codes, and where those `extension_codes` logically belong. 

Since both objects use the list, they now depend on each other circularly, so both objects must point to the list. It's definitely unhappy, but I didn't see another way to go (without changing `Nifti1Extension` drastically, to avoid the need to refer to `extension_codes`).

### DIFF
--- a/nibabel/tests/test_nifti1.py
+++ b/nibabel/tests/test_nifti1.py
@@ -20,8 +20,7 @@ from nibabel.eulerangles import euler2mat
 from nibabel.externals.six import BytesIO
 from nibabel.nifti1 import (load, Nifti1Header, Nifti1PairHeader, Nifti1Image,
                             Nifti1Pair, Nifti1Extension, Nifti1Extensions,
-                            data_type_codes, extension_codes,
-                            slice_order_codes)
+                            data_type_codes, slice_order_codes)
 from nibabel.openers import ImageOpener
 from nibabel.spatialimages import HeaderDataError
 from nibabel.tmpdirs import InTemporaryDirectory
@@ -35,11 +34,13 @@ from numpy.testing import (assert_array_equal, assert_array_almost_equal,
                            assert_almost_equal)
 from nose.tools import (assert_true, assert_false, assert_equal,
                         assert_raises)
-
 from ..testing import data_path, suppress_warnings, runif_extra_has
 
 from . import test_analyze as tana
 from . import test_spm99analyze as tspm
+from .nibabel_data import get_nibabel_data, needs_nibabel_data
+from .test_arraywriters import rt_err_estimate, IUINT_TYPES
+from .test_helpers import bytesio_filemap, bytesio_round_trip
 
 header_file = os.path.join(data_path, 'nifti1.hdr')
 image_file = os.path.join(data_path, 'example4d.nii.gz')
@@ -1034,9 +1035,8 @@ def test_ext_eq():
 
 
 def test_extension_codes():
-    for k in extension_codes.keys():
+    for k in Nifti1Extensions.extension_codes.keys():
         Nifti1Extension(k, 'somevalue')
-
 
 def test_extension_list():
     ext_c0 = Nifti1Extensions()
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -252,6 +252,7 @@ class Nifti1Extension(object):
     as `comment`. More sophisticated extensions should/will be supported by
     dedicated subclasses.
     """
+
     def __init__(self, code, content):
         """
         Parameters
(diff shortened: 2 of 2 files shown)


effigies:
Does it make sense to use `Nifti1Extensions.codes` instead of `Nifti1Extensions.extension_codes`? Or could that lead to ambiguity? (I don't ever use this, so I defer to the judgment of people who do.)

bcipolli:
@matthew-brett Any comment on this refactoring? The proposed downstream use is here:
https://github.com/satra/nibabel/pull/2/files#diff-3e6083e43c037b3efa25b74f45038a9cR52

nibabel/tests/test_nifti1.py
```diff
@@ -1034,10 +1035,9 @@ def test_ext_eq():
 
 
 def test_extension_codes():
-    for k in extension_codes.keys():
+    for k in Nifti1Extensions.extension_codes.keys():
```
matthew-brett: Maybe add test for `Nifti1Extension.extension_codes` as well?

matthew-brett:
I hate to ask, but how about just adding the Cifti code to the list of known extension codes, and leave the code as was otherwise?

It feels strange to put the codes in the `Nifti1Extensions` object somehow.

Just to remind myself, the extensions recoder object records the numerical codes, label and class that handles the matching code.  It's a sort of dispatch handler, so it does seem like a separate object to me.

bcipolli:
@matthew-brett In my view that breaks encapsulation. Nifti is well-defined irrespective of Cifti. It shouldn't need to know a thing about Cifti. The framework *almost* works great; Cifti could simply register itself as an extension.

To me, that seems more powerful and correct: Nifti is an extensible framework, and the Nifti objects can be defined such that they can be extended by downstream image types--whether known (such as Cifti) or unknown (some user-defined image type).

> Just to remind myself, the extensions recoder object records the numerical codes, label and class that handles the matching code. It's a sort of dispatch handler, so it does seem like a separate object to me.

As far as I understood, the recoder is just a list of associations. It's the `Nifti1Extensions` class that does the dispatching. I don't see one as really meaningful without the other--you can't do dispatching without the list of codes, and the list of codes aren't meaningful without the dispatching code.

That's why separating the recoder and dispatching code causes problems--`Nifti1Extensions` makes assumptions about where it finds codes for dispatching. This refactor makes the assumption overrideable / extendable.

Another way would be for the Cifti class to directly manipulate `extension_codes`. I think that's a worse option, and we avoided such side-effect methods when we worked on other code.

matthew-brett:
I see what you mean, but ... feels painful.

So I guess after this you'd have to do something like:

```
class CiftiExtensions(Nifti1Extensions):
    extension_codes = Nifti1Extensions.extension_codes.copy()
    extension_codes.add_codes((
    (CiftiExtension.code, "cifti", CiftiExtension),))

class CiftiExtension(Nifti1Extension):
    extension_codes = CiftiExtensions.extension_codes
```

?

bcipolli:
Something along those lines. I agree it's not ideal. I can improve that interface, perhaps by giving `Nifti1Extensions` an interface for adding codes that hides the need to copy. I'll take a look.

bcipolli:
I'm having trouble with this one, to make a clean interface that allows encapsulation. The major issue is that things would be class-level properties, not instance-level properties. If instance-level properties, I could  copy the data inside `__init__`. In classes, to do it automatically, I'd have to use a metaclass or some number of awkward tricks.

I'll revisit this in a couple of weeks. It's important to me that different image classes be encapsulated, so that adding new image types is as clean and self-contained as possible.

bcipolli:
@matthew-brett very similar to your suggestion, to simply adding the `Cifti` code into the list defined in `nifti1.py`, I could just add the `Cifti` extension from `cifti.py` to the `nifti1.extension_codes` list.

I was working hard to avoid side-effects like that, but that essentially fits the current design while allowing cifti information to be encapsulated in cifti code.

In that case, I would simply close this PR.

matthew-brett:
Ben - do you prefer to close this PR then?   Would you do the honors if so?

bcipolli:
My pleasure!

PR #379, opened 2015-11-06
MRG: Prepare for pydicom 1.0

### LINKED ISSUE #351: New import names in pydicom 1.0
See: https://github.com/darcymason/pydicom/wiki/Porting-to-pydicom-1.0

Need wrapper something like:

```
try:
    import dicom as pydicom
except ImportError:
    import pydicom

try:
    read_file = pydicom.dicomio.read_file
except AttributeError:
    read_file = pydicom.read_file
```

effigies:
pydicom 1.0 will change the import name from `dicom` to `pydicom`, and move `dicom.read_file` to `pydicom.dicomio.read_file`. Adjust imports accordingly, and rename `dicom` to `pydicom`, when imported directly.

Closes #351

### DIFF
--- a/nibabel/nicom/tests/test_dicomreaders.py
+++ b/nibabel/nicom/tests/test_dicomreaders.py
@@ -40,7 +40,10 @@ def test_passing_kwds():
     # Check that we correctly pass keywords to dicom
     dwi_glob = 'siemens_dwi_*.dcm.gz'
     csa_glob = 'csa*.bin'
-    import dicom
+    try:
+        from dicom.filereader import InvalidDicomError
+    except ImportError:
+        from pydicom.filereader import InvalidDicomError
     for func in (didr.read_mosaic_dwi_dir, didr.read_mosaic_dir):
         data, aff, bs, gs = func(IO_DATA_PATH, dwi_glob)
         # This should not raise an error
@@ -49,14 +52,14 @@ def test_passing_kwds():
             dwi_glob,
             dicom_kwargs=dict(force=True))
         assert_array_equal(data, data2)
-        # This should raise an error in dicom.read_file
+        # This should raise an error in pydicom.dicomio.read_file
         assert_raises(TypeError,
                       func,
                       IO_DATA_PATH,
                       dwi_glob,
                       dicom_kwargs=dict(not_a_parameter=True))
         # These are invalid dicoms, so will raise an error unless force=True
-        assert_raises(dicom.filereader.InvalidDicomError,
+        assert_raises(InvalidDicomError,
                       func,
                       IO_DATA_PATH,
                       csa_glob)
--- a/nibabel/nicom/tests/test_dicomwrappers.py
+++ b/nibabel/nicom/tests/test_dicomwrappers.py
@@ -9,12 +9,17 @@ from copy import copy
 
 import numpy as np
 
+have_dicom = True
 try:
-    import dicom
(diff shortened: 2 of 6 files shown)


effigies:
@ignatenkobrain Do you want to try this branch out?

bcipolli:
This looks ok to me. My concern is: how are we going to test the two codepaths?

effigies:
Until pydicom 1.0 is actually released, I'm not sure. I don't know if we can set up a Travis test to use a commit off Github. But there is a Travis test using the latest pydicom release, so we'll at least make sure we're not breaking anybody who's not bleeding edge.

I'm also okay with waiting to merge this until the 1.0 release of pydicom, in which case @ignatenkobrain can test this patch (which gives us some assurance that it's correct) and include it in his package. That way the patch is at least the one we expect to go in, so he's not stuck with an orphaned patch.

arokem:
Would pip installing from github work for this?

http://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip

matthew-brett:
Some test failures...

effigies:
Thanks. Scanning today, so intermittent ability to check.

nibabel/nicom/tests/test_dicomwrappers.py
```diff
@@ -9,12 +9,17 @@
...
 except ImportError:
-    have_dicom = False
-else:
-    have_dicom = True
+    try:
+        import pydicom
+    except ImportError:
+        have_dicom = False
+    else:
+        from pydicom.dicomio import read_file
```
effigies: This is a little ugly, reflecting the need to discover two different pydicom versions. The try/except/else allows us to detect a failure to import `pydicom` separately from a failure to find `pydicom.dicomio.read_file`.

effigies:
Tests passing. Code review appreciated.

nibabel/nicom/tests/test_dicomreaders.py
```diff
@@ -40,7 +40,10 @@ def test_passing_kwds():
     # Check that we correctly pass keywords to dicom
     dwi_glob = 'siemens_dwi_*.dcm.gz'
     csa_glob = 'csa*.bin'
-    import dicom
+    try:
+        import dicom as pydicom
+    except ImportError:
+        import pydicom
```
bcipolli: Is it worth importing `InvalidDicomError` here, like you did with `read_file`? I know there's no change in the package path, but would be more consistent with the others (including `Dataset`)

bcipolli:
LGTM. Essentially all of the `try..catch` blocks are in test code, and I don't think it's worth any extra effort to try and make things "prettier". I also think there's no need for further code comments; the code is pretty clear about what changed, and the commits/ PR document the time & reason for the change well. 

All that to, again, say: LGTM

ignatenkobrain:
Hi folks, I want. I will test it during weekend.

ignatenkobrain:
I have checked this with PR for 2.0.2. Fixes my issue and works perfectly! Would be great if this also will be part of 2.0.2 (but not required, I will backport this patch myself)

effigies:
I'd like to put this into 2.0.2.

Squashed down to two commits. Ready to merge? @matthew-brett @bcipolli

bcipolli:
:+1: Took a final look, happy from my side

matthew-brett:
Thanks for this - in it goes.

Issue #209, opened 2013-11-14
ENH: Improve read pattern in gzip


GaelVaroquaux:
**Problems**

_Memory usage_

When opening big gzip files, as for instance those shipped by the human connectome project, nibabel takes too much memory (see for instance #208 ).

The breakage happens in gzip.open and can be explained as follow (see the traceback in #208 to understand better):
- gzip does not know how much memory it needs to allocate, so it uses 1) for decompression a growing chunk size, with powers of two and capped at gzip.GzipFile.max_read_chunk, 2) for storing the output a buffer that it grows progressively by appending the chunk read.
- for memory problems, that buffer is the worst problem as its memory footprint gets doubled when the append happens (it is not an append: a new buffer is created, and the old one removed).
- if the final data has 4Gb + epsilon, the gzip opener will have at a given instant 2 4Gb buffers, and, in @rphlypo's case blow the memory

_Read is slow_

In addition to the memory usage, this pattern leads to very inefficient code, as the temporary buffer is grown again and again, which leads to costly memory copies. These are especially costly since the size of the memory copy isn't small compared to the total size of the memory, and thus the system has to move other memory pages belonging to other programs to allocate this memory. This is typically the behavior that renders systems unresponsive when the memory usage gets close 100%. To witness how bad the I/O speed is with data of a size on the same order of magnitude than the total memory, an easy experiment is to compare the time a 'nibabel.get_data()' takes to the time required to:
- gzip.GzipFile('file.nii.gz').read()    (slow)
- gunzip file.nii.gz -c > /dev/null      (almost twice as fast)
- open in fslview                              (almost twice as fast)

**Proposed improvement**

The only way that I can think of to improve the memory usage would be to preallocate a buffer big enough. We roughly know the size. However, the gzip module does not allow us to do this, and it would require fairly deep monkey patching that I think is not reasonable. I give up on this aspect of the problem.

For the speed, the simple option is to increase the 'max_read_chunk' of the GzipFile object. With a file that is 400Mb big, here is an experiment that shows the benefit:

<pre>
import gzip
gzip_file = gzip.open('0010042_rest_tshift_RPI_voreg_mni.nii.gz')
gzip_file.max_read_chunk = max_read_chunk = 100 * 1024 * 1024 # 100Mb
data = gzip_file.read()
</pre>

The extra line (third line) makes a factor of 2 of speed in the above example.

To see the impact on nibabel, a little bit of monkey-patching enables experimentation:

<pre>
import gzip
[...]

PR #383, opened 2015-11-15
Fix for issue 362: nibabel fails to stream gzipped files > 4GB (uncompressed) in Python 3.5

bcipolli:
This not a `nibabel` bug, but works around a Python bug (https://bugs.python.org/issue25626).

The workaround is: we wrap `gzip.GzipFile` with buffering, so that files > 4GB require multiple calls to `GzipFile.readinto`.

I've also added test functionality: if `NIPY_EXTRA_TESTS` contains `'slow'`, slowly running tests can be run. I've used this to include a test for creation of a large file.

### DIFF
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -17,11 +17,42 @@ import bz2
 GZIP_MAX_READ_CHUNK = 100 * 1024 * 1024  # 100Mb
 
 
+class BufferedGzipFile(gzip.GzipFile):
+    """GzipFile capable to readinto buffer >= 2**32 bytes."""
+        # Speedup for #209; breaks in Python 3.5
+    def __init__(self, fileish, mode='rb', compresslevel=9, buffer_size=2**32-1):
+        super(BufferedGzipFile, self).__init__(fileish, mode=mode, compresslevel=compresslevel)
+        if hasattr(self, 'max_chunk_read'):
+            gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
+        self.buffer_size = buffer_size
+
+    def readinto(self, buf):
+        """Uses self.buffer_size to do a buffered read."""
+        n_bytes = len(buf)
+        try:
+            # This works around a known issue in Python 3.5.
+            # See https://bugs.python.org/issue25626
+            mv = memoryview(buf)
+            n_read = 0
+            max_read = 2 ** 32 - 1  # Max for unsigned 32-bit integer
+            while (n_read < n_bytes):
+                n_wanted = min(n_bytes - n_read, max_read)
+                n_got = super(BufferedGzipFile, self).readinto(
+                    mv[n_read:n_read + n_wanted])
+                n_read += n_got
+                if n_got != n_wanted:
+                    break
+        except NameError:  # Python 2.6 or old 2.7: memoryview does not exist.
+            raise
+            n_read = super(BufferedGzipFile, self).readinto(buf)
+        return n_read
+
+
 def _gzip_open(fileish, *args, **kwargs):
     # open gzip files with faster reads on large files using larger chunks
     # See https://github.com/nipy/nibabel/pull/210 for discussion
-    gzip_file = gzip.open(fileish, *args, **kwargs)
-    gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
+
+    gzip_file = BufferedGzipFile(fileish, *args, **kwargs)
     return gzip_file
 
 
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -13,7 +13,6 @@ import bz2
 import gzip
 import sys
 from os.path import splitext
-from distutils.version import LooseVersion
 
 
 # The largest memory chunk that gzip can use for reads
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -9,33 +9,42 @@
 """ Context manager openers for various fileobject types
 """
 
-from os.path import splitext
-import gzip
 import bz2
+import gzip
+import sys
+from os.path import splitext
+from distutils.version import LooseVersion
+
 
 # The largest memory chunk that gzip can use for reads
 GZIP_MAX_READ_CHUNK = 100 * 1024 * 1024  # 100Mb
 
 
 class BufferedGzipFile(gzip.GzipFile):
-    """GzipFile able to readinto buffer >= 2**32 bytes."""
-    def __init__(self, fileish, mode='rb', compresslevel=9, buffer_size=2**32-1):
-        super(BufferedGzipFile, self).__init__(fileish, mode=mode, compresslevel=compresslevel)
-        self.buffer_size = buffer_size
+    """GzipFile able to readinto buffer >= 2**32 bytes.
 
-        # Speedup for #209; attribute not present in in Python 3.5
-        # open gzip files with faster reads on large files using larger chunks
-        # See https://github.com/nipy/nibabel/pull/210 for discussion
-        if hasattr(self, 'max_chunk_read'):
-            gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
+    This class only differs from gzip.GzipFile
+    in Python 3.5.0.
 
-    def readinto(self, buf):
-        """Uses self.buffer_size to do a buffered read.
+    This works around a known issue in Python 3.5.
+    See https://bugs.python.org/issue25626"""
 
-        This works around a known issue in Python 3.5.
-        See https://bugs.python.org/issue25626"""
-        n_bytes = len(buf)
-        try:
+    # This helps avoid defining readinto in Python 2.6,
+    #   where it is undefined on gzip.GzipFile.
+    # It also helps limit the exposure to this code.
+    if sys.version_info[:3] == (3, 5, 0):
+        def __init__(self, fileish, mode='rb', compresslevel=9,
+                     buffer_size=2**32-1):
+            super(BufferedGzipFile, self).__init__(fileish, mode=mode,
+                                                   compresslevel=compresslevel)
+            self.buffer_size = buffer_size
+
+        def readinto(self, buf):
+            """Uses self.buffer_size to do a buffered read."""
+            n_bytes = len(buf)
+            if n_bytes < 2 ** 32:
+                return super(BufferedGzipFile, self).readinto(buf)
+
             # This works around a known issue in Python 3.5.
             # See https://bugs.python.org/issue25626
             mv = memoryview(buf)
@@ -48,14 +57,18 @@ class BufferedGzipFile(gzip.GzipFile):
                 n_read += n_got
                 if n_got != n_wanted:
                     break
-        except NameError:  # Python 2.6 or old 2.7: memoryview does not exist.
-            raise
-            n_read = super(BufferedGzipFile, self).readinto(buf)
-        return n_read
+            return n_read
 
 
 def _gzip_open(fileish, *args, **kwargs):
     gzip_file = BufferedGzipFile(fileish, *args, **kwargs)
+
+    # Speedup for #209; attribute not present in in Python 3.5
+    # open gzip files with faster reads on large files using larger
+    # See https://github.com/nipy/nibabel/pull/210 for discussion
+    if hasattr(gzip_file, 'max_chunk_read'):
+        gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
+
     return gzip_file
 
 
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -18,16 +18,22 @@ GZIP_MAX_READ_CHUNK = 100 * 1024 * 1024  # 100Mb
 
 
 class BufferedGzipFile(gzip.GzipFile):
-    """GzipFile capable to readinto buffer >= 2**32 bytes."""
-        # Speedup for #209; breaks in Python 3.5
+    """GzipFile able to readinto buffer >= 2**32 bytes."""
     def __init__(self, fileish, mode='rb', compresslevel=9, buffer_size=2**32-1):
         super(BufferedGzipFile, self).__init__(fileish, mode=mode, compresslevel=compresslevel)
-        if hasattr(self, 'max_chunk_read'):
-            gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
         self.buffer_size = buffer_size
 
+        # Speedup for #209; attribute not present in in Python 3.5
+        # open gzip files with faster reads on large files using larger chunks
+        # See https://github.com/nipy/nibabel/pull/210 for discussion
+        if hasattr(self, 'max_chunk_read'):
+            gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
+
     def readinto(self, buf):
-        """Uses self.buffer_size to do a buffered read."""
+        """Uses self.buffer_size to do a buffered read.
+
+        This works around a known issue in Python 3.5.
+        See https://bugs.python.org/issue25626"""
         n_bytes = len(buf)
         try:
             # This works around a known issue in Python 3.5.
@@ -49,9 +55,6 @@ class BufferedGzipFile(gzip.GzipFile):
 
 
 def _gzip_open(fileish, *args, **kwargs):
-    # open gzip files with faster reads on large files using larger chunks
-    # See https://github.com/nipy/nibabel/pull/210 for discussion
-
     gzip_file = BufferedGzipFile(fileish, *args, **kwargs)
     return gzip_file
 
--- a/nibabel/testing/__init__.py
+++ b/nibabel/testing/__init__.py
@@ -9,12 +9,14 @@
 ''' Utilities for testing '''
 from __future__ import division, print_function
 
+import os
 import sys
 import warnings
 from os.path import dirname, abspath, join as pjoin
 
 import numpy as np
 
+from numpy.testing.decorators import skipif
 # Allow failed import of nose if not now running tests
 try:
     from nose.tools import (assert_equal, assert_not_equal,
@@ -164,3 +166,11 @@ class catch_warn_reset(clear_and_catch_warnings):
         warnings.warn('catch_warn_reset is deprecated and will be removed in '
                       'nibabel v3.0; use nibabel.testing.clear_and_catch_warnings.',
                       FutureWarning)
+
+
+def runif_extra_has(test_str):
+    """Decorator checks to see if NIPY_EXTRA_TESTS env var contains test_str"""
+    def decorator(func):
+        skip_set = os.environ.get('NIPY_EXTRA_TESTS', '').split(',')
+        return skipif(test_str not in skip_set, "Skip slow tests.")(func)
+    return decorator
--- a/nibabel/testing/__init__.py
+++ b/nibabel/testing/__init__.py
@@ -168,9 +168,10 @@ class catch_warn_reset(clear_and_catch_warnings):
                       FutureWarning)
 
 
+EXTRA_SET = os.environ.get('NIPY_EXTRA_TESTS', '').split(',')
+
+
 def runif_extra_has(test_str):
     """Decorator checks to see if NIPY_EXTRA_TESTS env var contains test_str"""
-    def decorator(func):
-        skip_set = os.environ.get('NIPY_EXTRA_TESTS', '').split(',')
-        return skipif(test_str not in skip_set, "Skip slow tests.")(func)
-    return decorator
+    return skipif(test_str not in EXTRA_SET,
+                  "Skip {0} tests.".format(test_str))
--- a/nibabel/tests/test_nifti1.py
+++ b/nibabel/tests/test_nifti1.py
@@ -1248,11 +1248,14 @@ class TestNifti1General(object):
 @runif_extra_has('slow')
 def test_large_nifti1():
     image_shape = (91, 109, 91, 1200)
-    img = Nifti1Image(np.zeros(image_shape, dtype=np.float32),
+    img = Nifti1Image(np.ones(image_shape, dtype=np.float32),
                       affine=np.eye(4))
+    # Dump and load the large image.
     with InTemporaryDirectory():
         img.to_filename('test.nii.gz')
         del img
         data = load('test.nii.gz').get_data()
-    assert_array_equal(np.asarray(image_shape), data.shape)
-    assert_true(np.all(data == 0.))
+    # Check that te data are all ones
+    assert_equal(image_shape, data.shape)
+    n_ones = np.sum((data == 1.))
+    assert_equal(np.prod(image_shape), n_ones)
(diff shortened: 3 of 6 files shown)


nibabel/openers.py
```diff
@@ -17,11 +17,36 @@
 GZIP_MAX_READ_CHUNK = 100 * 1024 * 1024  # 100Mb
 
 
+class BufferedGzipFile(gzip.GzipFile):
+    """GzipFile capable to readinto buffer >= 2**32 bytes."""
```
matthew-brett: "able to" rather than "capable to"

nibabel/openers.py
```diff
@@ -17,11 +17,36 @@
...
+    def __init__(self, fileish, buffer_size=2**32-1, *args, **kwargs):
+        super(BufferedGzipFile, self).__init__(fileish, *args, **kwargs)
+        # Speedup for #209; breaks in Python 3.5
```
matthew-brett: Maybe `# Speedup for #209; attribute not present for Python 3.5`

nibabel/openers.py
```diff
@@ -17,11 +17,36 @@
...
+        if hasattr(self, 'max_chunk_read'):
+            gzip_file.max_read_chunk = GZIP_MAX_READ_CHUNK
+        self.buffer_size = buffer_size
+
+    def readinto(self, buf):
+        """Uses self.buffer_size to do a buffered read."""
+        n_bytes = len(buf)
+        n_read = 0
+        max_read = 2 ** 32 - 1  # Max for unsigned 32-bit integer
```
matthew-brett: Maybe comment to point to python bug report?

nibabel/openers.py
```diff
@@ -17,11 +17,36 @@
...
+            n_got = super(BufferedGzipFile, self).readinto(
+                buf[n_read:n_read + n_wanted])
+            n_read += n_got
+            if n_got != n_wanted:
+                break
+        return n_read
+
+
 def _gzip_open(fileish, *args, **kwargs):
     # open gzip files with faster reads on large files using larger chunks
```
matthew-brett: I guess comment now needs to go in the wrapper?

nibabel/testing/__init__.py
```diff
@@ -164,3 +166,9 @@ def __init__(self, *args, **kwargs):
         warnings.warn('catch_warn_reset is deprecated and will be removed in '
                       'nibabel v3.0; use nibabel.testing.clear_and_catch_warnings.',
                       FutureWarning)
+
+
+def skipif_no_slow_tests(func):
```
matthew-brett: How about something like:

```
def runif_extras_has(test_type):
    extra_tests = os.environ.get('NIPY_EXTRA_TESTS', '').split(',')
    return skipif(test_type not in extra_tests, 'Skipping {0} tests'.format(test_type))
```

nibabel/tests/test_nifti1.py
```diff
@@ -13,17 +13,18 @@
...
-from ..tmpdirs import InTemporaryDirectory
-from ..spatialimages import HeaderDataError
-from ..eulerangles import euler2mat
-from ..affines import from_matvec
-from .. import nifti1 as nifti1
-from ..nifti1 import (load, Nifti1Header, Nifti1PairHeader, Nifti1Image,
-                      Nifti1Pair, Nifti1Extension, Nifti1Extensions,
-                      data_type_codes, extension_codes, slice_order_codes)
-
+from nibabel import nifti1 as nifti1
```
matthew-brett: Reason to avoid relative imports?

nibabel/testing/__init__.py
bcipolli: My thought was: it's good to have the very clear headers with no parameters, so the options stay very limited and no mistakes can be made. The generic function above could be used as the base for each, and could be written if/when needed...

That was my thought at least.

nibabel/tests/test_nifti1.py
bcipolli: I thought the standard is to use absolute imports in tests. At least, @effigies and I have been migrating to that standard as we touch test files.

nibabel/openers.py
```diff
@@ -17,11 +17,39 @@
...
+        """Uses self.buffer_size to do a buffered read.
+
+        This works around a known issue in Python 3.5.
+        See https://bugs.python.org/issue25626"""
+        n_bytes = len(buf)
+        n_read = 0
+        max_read = 2 ** 32 - 1  # Max for unsigned 32-bit integer
+        while (n_read < n_bytes):
+            n_wanted = min(n_bytes - n_read, max_read)
+            print(n_read, n_wanted, n_bytes)
```
bcipolli: need to remove.

nibabel/openers.py
```diff
@@ -17,11 +17,39 @@
...
+            n_got = super(BufferedGzipFile, self).readinto(
+                buf[n_read:n_read + n_wanted])
```
bcipolli: @matthew-brett this is not working; this slice apparently creates a different memory location to write to. Any idea about how to manage this?

nibabel/tests/test_nifti1.py
matthew-brett: Any reason for that that you know?

nibabel/testing/__init__.py
matthew-brett: In the interests of DRY, maybe:

``` 
skipif_no_slow_tests = runif_extras_has('slow')
```
?

nibabel/tests/test_nifti1.py
bcipolli: I got it from nilearn: https://github.com/nilearn/nilearn/issues/384
Who got it from scikit-learn: http://scikit-learn.org/stable/developers/contributing.html#coding-guidelines

> Unit tests are an exception to the previous rule; they should use absolute imports, exactly as client code would. A corollary is that, if sklearn.foo exports a class or function that is implemented in sklearn.foo.bar.baz, the test should import it from sklearn.foo.

nibabel/testing/__init__.py
bcipolli: Right. There's no DRY concern here; there's no repetition. I will just do it now, but this adds code that's (very slightly) more complex than currently needed... which I've been well-trained to avoid :)

matthew-brett: Sure.  It feels slightly cleaner to me to do:

```
@runif_extras_has('slow')
def test_one():
    pass

@runif_extras_has('bigmem')
def test_two():
    pass
```

but it's true that it's possible to get the wrong string.  I guess that could fixed with ``if test_type not in ('slow', 'bigmem')`` check in the decorator.  But I don't feel strongly.

nibabel/tests/test_nifti1.py
matthew-brett: OK - I don't feel strongly either way.

nibabel/openers.py
matthew-brett: Hmm - no.  I suppose a numpy array doesn't expose the right interface?

I'm surprised though:

```
In [1]: a = bytearray(10)

In [2]: a
Out[2]: bytearray(b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')

In [3]: a[1:3] = b'abc'

In [4]: a
Out[4]: bytearray(b'\x00abc\x00\x00\x00\x00\x00\x00\x00')
```

bcipolli: I'm surprised too; the numpy docs say simple slicing returns a view. But the tests are failing, and if I remove the slicing, they pass.

Oh, probably there's a bug in the indexing code, duh. Will fix.

matthew-brett: S'not numpy though, it's a Python bytearray.

effigies:
I haven't been following this PR, but I'm blocking 2.0.2 on this. Let me know if I shouldn't be.

bcipolli:
@effigies This "fix" didn't actually work. I will submit a temporary workaround we discussed previously, this afternoon. Job interview this morning! :)

bcipolli:
`read` suffers from the same `OverflowError`. However, this webpage covers how to solve the buffering issue I hit here.
http://eli.thegreenplace.net/2011/11/28/less-copies-in-python-with-the-buffer-protocol-and-memoryviews

I will work on an update...

bcipolli:
OK, fixed this by using `memoryview` if available (3.0+ for sure, often in 2.7), falling back to vanilla `readinto` if not (up to 3.4). Also created a `runif_extra_has` header and a large file test (off by default). I ran the test locally in Python 3.5, and works for me...

If someone could pull this branch and run the Python 3.5 test, would be great. Will have to set `NIPY_EXTRA_TESTS=slow` in the environment.

Last question: where to document the use of `NIPY_EXTRA_TESTS`?

nibabel/openers.py
```diff
@@ -17,11 +17,45 @@
 GZIP_MAX_READ_CHUNK = 100 * 1024 * 1024  # 100Mb
 
 
+class BufferedGzipFile(gzip.GzipFile):
+    """GzipFile able to readinto buffer >= 2**32 bytes."""
+    def __init__(self, fileish, mode='rb', compresslevel=9, buffer_size=2**32-1):
+        super(BufferedGzipFile, self).__init__(fileish, mode=mode, compresslevel=compresslevel)
```
effigies: Looks like `gzip.GzipFile` is [not a new-style class](https://travis-ci.org/nipy/nibabel/jobs/91709435#L1043) in 2.6.

I think this should work:

    gzip.GzipFile.__init__(self, fileish, mode=mode, compresslevel=compresslevel)

nibabel/openers.py
```diff
@@ -17,11 +17,45 @@
...
+            while (n_read < n_bytes):
+                n_wanted = min(n_bytes - n_read, max_read)
+                n_got = super(BufferedGzipFile, self).readinto(
+                    mv[n_read:n_read + n_wanted])
+                n_read += n_got
+                if n_got != n_wanted:
+                    break
+        except NameError:  # Python 2.6 or old 2.7: memoryview does not exist.
+            raise
+            n_read = super(BufferedGzipFile, self).readinto(buf)
```
effigies: Going to guess this will have issues, too.

nibabel/tests/test_nifti1.py
```diff
@@ -1242,3 +1243,16 @@ def test_rt_bias(self):
...
+@runif_extra_has('slow')
+def test_large_nifti1():
+    image_shape = (91, 109, 91, 1200)
+    img = Nifti1Image(np.zeros(image_shape, dtype=np.float32),
+                      affine=np.eye(4))
+    with InTemporaryDirectory():
+        img.to_filename('test.nii.gz')
+        del img
+        data = load('test.nii.gz').get_data()
+    assert_array_equal(np.asarray(image_shape), data.shape)
```
effigies: Shouldn't these both be tuples? If one doesn't need `np.asarray`, I imagine the other doesn't, either. At which point, you could probably just do `assert_equal`.

nibabel/openers.py
```diff
@@ -17,11 +17,45 @@
...
+            max_read = 2 ** 32 - 1  # Max for unsigned 32-bit integer
+            while (n_read < n_bytes):
+                n_wanted = min(n_bytes - n_read, max_read)
+                n_got = super(BufferedGzipFile, self).readinto(
+                    mv[n_read:n_read + n_wanted])
+                n_read += n_got
+                if n_got != n_wanted:
+                    break
+        except NameError:  # Python 2.6 or old 2.7: memoryview does not exist.
+            raise
```
effigies: Am I misunderstanding, or are you reraising `NameError` here? I think this will always be run in 2.6.

nibabel/openers.py
```diff
@@ -17,11 +17,45 @@
...
+    def readinto(self, buf):
+        """Uses self.buffer_size to do a buffered read.
+
+        This works around a known issue in Python 3.5.
+        See https://bugs.python.org/issue25626"""
+        n_bytes = len(buf)
+        try:
+            # This works around a known issue in Python 3.5.
+            # See https://bugs.python.org/issue25626
+            mv = memoryview(buf)
```
matthew-brett: I prefer keeping `try` .. `except` blocks as short as possible to make sure that it's clear what is being trapped and why.  So here, something like:

```
try:
    mv = memoryview(buf)
except NameError:
    n_read = super(BufferedGzipFile, self).readinto(buf)
else:
   n_read = 0
   ...
```

effigies: Does it make sense to invert?

    try:
        n_read = super(BufferedGzipFile, self).readinto(buf)
    except OverflowError:
        # Workaround for known issue in Python 3.5
        ...

Or is the `memoryview` strategy preferred except in Python 2.6?

nibabel/openers.py
bcipolli: Yeah, my bad--that was code to make sure I was NOT getting in here...

matthew-brett:
Sorry to ask - but would you consider adding a small "Running the tests" section to `doc/source/installation.rst`, maybe including `NIPY_EXTRA_TESTS`?

bcipolli:
>Sorry to ask - but would you consider adding a small "Running the tests" section to doc/source/installation.rst, maybe including NIPY_EXTRA_TESTS?

Great; exactly what I was looking for. Will do.

nibabel/openers.py
bcipolli: :+1:

nibabel/openers.py
bcipolli: I am fine on the `try...catch...else`. I wrote it the way I did because separating as above breaks into slightly weird chunks.

I believe that the `gzip` failure is not that fast, which is why I avoided the `gzip` failure by default. But it's hard to tell. I don't believe either of these methods is "preferred"; I believe `readinto` is deprecated.

bcipolli:
@matthew-brett a bit puzzled about pushing documentation to `installation.rst`. Isn't running tests something for either development or deployment? Or, could you help me understand what you had in mind for adding to `installation.rst`?

nibabel/openers.py
effigies: Point taken about waiting on an OverflowError. What about making the check explicitly for Python 3.5.0, instead of testing for `memoryview`? 3.5.1 should fix the issue, and I'm not super comfortable with using a hand-rolled buffering loop by default on 2.7+ because of a 3.5.0-specific bug.

And my understanding is that `readinto` hasn't been deprecated; it's just been pushed [up the hierarchy](https://docs.python.org/3/library/io.html#io.BufferedIOBase.readinto).

bcipolli:
`readinto` isn't defined in Python 2.6, so this code is invalid for that version. Taking @effigies hesitancy above as well:
* `BufferedGZipFile` only differs from `gzip.GZipFile` for Python 3.5.0
* The buffering code is only hit if the `len(buf) >= 2**32`

I tested this locally on my machine, it worked well. I will temporarily push a change to `.travis.yml` to test this on other architectures.

nibabel/openers.py
```diff
@@ -9,19 +9,62 @@
...
+            self.buffer_size = buffer_size
+
+        def readinto(self, buf):
+            """Uses self.buffer_size to do a buffered read."""
+            n_bytes = len(buf)
+            if n_bytes < 2 ** 32:
+                return super(BufferedGzipFile, self).readinto(buf)
+
+            # This works around a known issue in Python 3.5.
+            # See https://bugs.python.org/issue25626"""
```
effigies: No `"""` needed.

nibabel/openers.py
```diff
@@ -9,19 +9,62 @@
...
 
 
+class BufferedGzipFile(gzip.GzipFile):
+    """GzipFile able to readinto buffer >= 2**32 bytes."""
+    # Python 2.6 doesn't have a 'readinto' attribute,
+    # nothing is broken in anything but Python 3.5.0, so
+    # Only try this for Python 3.5.0.
+    #
+    # This works around a known issue in Python 3.5.
+    # See https://bugs.python.org/issue25626"""
```
effigies: `"""` only needed if this comment becomes part of docstring.

bcipolli:
AppVeyor fails as well. So now all we have is local testing by me on the new code.

effigies:
@surchs @arthurmensch Can you test this branch?

effigies:
@matthew-brett Are slow tests something we can use the buildbots for?

surchs:
Hey,
sorry I didn't get back on this but I'll test the branch!

bcipolli:
Great! @surchs please set environment variable before running tests: `NIPY_EXTRA_TESTS=slow`

nibabel/openers.py
```diff
@@ -9,19 +9,66 @@
 """ Context manager openers for various fileobject types
 """
 
-from os.path import splitext
-import gzip
 import bz2
+import gzip
+import sys
+from os.path import splitext
+from distutils.version import LooseVersion
```
effigies: Is this `LooseVersion` used for anything?

nibabel/openers.py
```diff
@@ -9,19 +9,66 @@
 """ Context manager openers for various fileobject types
 """
 
-from os.path import splitext
-import gzip
 import bz2
```
effigies: Have we checked to make sure large `.bz2` files aren't affected by the same upstream bug?

nibabel/testing/__init__.py
```diff
@@ -164,3 +166,11 @@ def __init__(self, *args, **kwargs):
...
                       'nibabel v3.0; use nibabel.testing.clear_and_catch_warnings.',
                       FutureWarning)
+
+
+def runif_extra_has(test_str):
+    """Decorator checks to see if NIPY_EXTRA_TESTS env var contains test_str"""
+    def decorator(func):
+        skip_set = os.environ.get('NIPY_EXTRA_TESTS', '').split(',')
+        return skipif(test_str not in skip_set, "Skip slow tests.")(func)
+    return decorator
```
effigies: This calls all extra tests "slow tests", and re-parses `NIPY_EXTRA_TESTS` each time. You could remove a nested function and gain a little brevity doing something like this:

At top:

    EXTRA_TESTS = set(os.environ.get('NIPY_EXTRA_TESTS', '').split(','))

New decorator:

    def runif_extra_has(test_str):
        """Decorator checks to see if NIPY_EXTRA_TESTS env var contains test_str"""
        return skipif(test_str not in EXTRA_TESTS, "Skip {0} tests.".format(test_str))

At the very least, I'd fix the `skipif` string.

nibabel/openers.py
bcipolli: Nope. What file types allow `.bz`? `.nii.bz` didn't seem to work.

bcipolli: Duh, `.bz2`. Trying it now.

matthew-brett:
We can certainly run the memory hungry tests on the larger buildbot machines.  That reminds me, I was going to give you access ... will do that now.

nibabel/openers.py
bcipolli: Yes, fails :(
image

effigies: That looks like 2.6, and so is probably a different issue from this PR's concern. Possibly worth looking into in a new PR, but low priority in my mind. (Nobody seems to be using large .bz2 files, and who's working on HCP data with seriously old Python?)

What about 3.5?

bcipolli: That was Python 2.7, but good point--I didn't use Python3. Python 3 is hanging, using a ton of resources.

effigies: Okay. `bz2` sounds like its own issue, not something we can quickly take care of while we're here.

bcipolli: OK, it finally returned successfully in Python 3.5

bcipolli:
ok, updated code again per comments.

nibabel/testing/__init__.py
```diff
@@ -164,3 +166,14 @@ def __init__(self, *args, **kwargs):
...
+
+EXTRA_SET = os.environ.get('NIPY_EXTRA_TESTS', '').split(',')
+
+
+def runif_extra_has(test_str):
+    """Decorator checks to see if NIPY_EXTRA_TESTS env var contains test_str"""
+    def decorator(func):
+        return skipif(test_str not in EXTRA_SET,
+                      "Skip {0} tests.".format(test_str))(func)
+    return decorator
```
effigies: Sorry, I'm going to be annoying and insist on removing the `decorator` function. What you have is:

    def f(x):
        return g(x)
    return f

When you could just have `return g`. Just use:

    def runif_extra_has(test_str):
        return skipif(test_str not in EXTRA_SET, "Skip {0} tests.".format(test_str))

effigies:
Thanks for your patience. LGTM.

nibabel/tests/test_nifti1.py
```diff
@@ -1242,3 +1243,19 @@ def test_rt_bias(self):
...
+def test_large_nifti1():
+    image_shape = (91, 109, 91, 1200)
+    img = Nifti1Image(np.ones(image_shape, dtype=np.float32),
+                      affine=np.eye(4))
+    # Dump and load the large image.
+    with InTemporaryDirectory():
+        img.to_filename('test.nii.gz')
+        del img
+        data = load('test.nii.gz').get_data()
+    # Check that te data are all ones
```
matthew-brett: type 'te'

matthew-brett: Wow' that's meta.  I meant "typo 'te'"

arthurmensch:
I can confirm that HCP is working again on Python 3.5. Cheers !

bcipolli:
:+1: @matthew-brett still looking for some details on best place for docs; I think it's the last thing to do.
https://github.com/nipy/nibabel/pull/383#issuecomment-157594704

nibabel/tests/test_nifti1.py
```diff
@@ -13,17 +13,18 @@
...
+from nibabel import nifti1 as nifti1
+from nibabel.affines import from_matvec
+from nibabel.casting import type_info, have_binary128
+from nibabel.eulerangles import euler2mat
+from nibabel.externals.six import BytesIO
+from nibabel.nifti1 import (load, Nifti1Header, Nifti1PairHeader, Nifti1Image,
+                            Nifti1Pair, Nifti1Extension, Nifti1Extensions,
+                            data_type_codes, extension_codes,
+                            slice_order_codes)
+from nibabel.openers import ImageOpener
```
effigies: Unused import.

matthew-brett:
Ben - sorry - are you asking me where to put the docs?

bcipolli:
Yes, I'm not sure where to add the documentation about the new test env variable. "Installation" felt odd to me, as it seems like a developer / release issue, so I wanted to double-check. In "installation", can you give me an idea how you envisioned it fitting in?

matthew-brett:
How about a little section on 'testing' in the installation docs, pointing to a document `advanced_testing` in `doc/source/devel` ?

bcipolli:
@matthew-brett Not sure if this is what you had in mind... 

image

image

matthew-brett:
I was thinking of something like 'to run tests, run `nosetests nibabel` or `python -c "import nibabel; nibabel.test()"`  See also advanced testing. 

Then advanced testing would be using the git submodules, as in:

``git submodule update --init``

and the slow tests stuff.

bcipolli:
OK. Here's what it looks like now:
image
image

matthew-brett:
Looks good - maybe add that `nosetests` etc should be run from the terminal rather than python / IPython.

bcipolli:
Done. I think we're ready to go!

matthew-brett:
Great - thanks again for wading through.

matthew-brett:
Running the slow tests on a big Mac buildbot : http://nipy.bic.berkeley.edu/builders/nibabel-py2.7-osx-10.10/builds/27/steps/shell_5/logs/stdio

PR #386, opened 2015-11-22
DOC: update release docs to use `try_branch.py`

matthew-brett:
Also, remove some emphasis on outdated check scripts.

### DIFF
--- a/doc/source/devel/make_release.rst
+++ b/doc/source/devel/make_release.rst
@@ -6,6 +6,24 @@ A guide to making a nibabel release
 
 This is a guide for developers who are doing a nibabel release.
 
+The general idea of these instructions is to go through the following steps:
+
+* Make sure that the code is in the right state for release;
+* update release-related docs such as the Changelog;
+* update various documents giving dependencies, dates and so on;
+* check all standard and release-specific tests pass;
+* make the *release commit* and release tag;
+* check Windows binary builds and slow / big memory tests;
+* push source and windows builds to pypi;
+* push docs;
+* push release commit and tag to github;
+* announce.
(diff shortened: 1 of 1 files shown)


doc/source/devel/make_release.rst
```diff
@@ -142,6 +121,44 @@ Release checklist
...
+  Then get the built binaries in:
+
+  * https://nipy.bic.berkeley.edu/nibabel-dist
+
+  When you've done the release to pypi, you can upload them to pypi with the
+  admin files interface.
+
+  If you are already on a Windows machine, you could have done the manual
+  command to build instead: ``python setup.py bdist_wininst``.
+
```
effigies: I feel like this makes sense to move to below the release commit, otherwise you end up with `x.y.z.dev` binaries. I ended up doing this sequence:

* Update docs and `nibabel/info.py` on a PR branch
* Merge PR in Github
* Pull `maint/2.0.x`, tag and push tag
* Push to pypi
* Push to buildbots

matthew-brett: Ah yes, you're right. But I personally like to avoid pushing the tag for as long as possible because it can't be reversed after.  So I suggest try_branch pushing to the buildbots after the release commit, but before pushing the release commit or the release tag.

effigies: I checked the builds on the branch that had been merged in before tagging, but that's a good policy.

matthew-brett:
OK to merge this one?

effigies:
LGTM

matthew-brett:
Thanks - merging.

PR #389, opened 2015-12-02
MRG: fix for Freesurfer mmap bug

matthew-brett:
We were accidentally passing an ImageOpener instance as the file object to the array proxy, and Pyton 3.5 pointed this out by failing:

http://nipy.bic.berkeley.edu/builders/nibabel-bdist32-35/builds/9/steps/shell_6/logs/stdio

This branch fixes the Python 3.5 tests:

http://nipy.bic.berkeley.edu/builders/nibabel-bdist32-35/builds/11/steps/shell_6/logs/stdio

### DIFF
--- a/nibabel/fileholders.py
+++ b/nibabel/fileholders.py
@@ -90,6 +90,12 @@ class FileHolder(object):
         return ((self.filename == other.filename) and
                 (self.fileobj == other.fileobj))
 
+    @property
+    def file_like(self):
+        """ Return ``self.fileobj`` if not None, otherwise ``self.filename``
+        """
+        return self.fileobj if self.fileobj is not None else self.filename
+
 
 def copy_file_map(file_map):
     ''' Copy mapping of fileholders given by `file_map`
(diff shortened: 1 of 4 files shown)


effigies:
Okay, if I'm understanding you, the bug only manifests if Windows tries to treat a zipping `ImageOpener` like an `mmap`able file? Does it make sense to try to catch the Windows error?

I checked the `ArrayProxy` interface, and your story checks out, so LGTM. Just want to grok the situation a little better.

matthew-brett:
I did think of trying to make the memmap testing working correctly with ImageOpener instances, but that seemed like overkill.  In this case, the fact that we were passing ImageOpener instances to the ArrayProxy was just a bug - we normally pass the filename so the image does not need to keep a file-handle open for the object's lifetime.

matthew-brett:
Chris - OK to merge this one?

effigies:
Oh, yes.

On December 4, 2015 8:57:35 PM EST, Matthew Brett <notifications@github.com> wrote:
>Chris - OK to merge this one?
>
>---
>https://github.com/nipy/nibabel/pull/389#issuecomment-162125979

-- 
Chris Markiewicz

PR #393, opened 2015-12-18
BF+TST: Skip over missing CSA header elem instead of raising.

moloney:
Allows us to load data where the PrivateCreator is present but
the element containing the header info is not. Ran into this with
anonymized data.

### DIFF
--- a/nibabel/nicom/csareader.py
+++ b/nibabel/nicom/csareader.py
@@ -66,6 +66,7 @@ def get_csa_header(dcm_data, csa_type='image'):
     try:
         tag = dcm_data[(0x29, element_no)]
     except KeyError:
+        # The element could be missing due to anonymization
         return None
     return read(tag.value)
 
--- a/nibabel/nicom/csareader.py
+++ b/nibabel/nicom/csareader.py
@@ -63,8 +63,10 @@ def get_csa_header(dcm_data, csa_type='image'):
     if section_start is None:
         return None
     element_no = section_start + element_offset
-    # Assume tag exists
-    tag = dcm_data[(0x29, element_no)]
+    try:
+        tag = dcm_data[(0x29, element_no)]
+    except KeyError:
+        return None
     return read(tag.value)
 
 
--- a/nibabel/nicom/tests/test_csareader.py
+++ b/nibabel/nicom/tests/test_csareader.py
@@ -1,6 +1,7 @@
 """ Testing Siemens CSA header reader
 """
 from os.path import join as pjoin
+from copy import deepcopy
 import gzip
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


moloney:
The failure on 2.6 is from [this](https://github.com/darcymason/pydicom/issues/236) pydicom bug which is now fixed in master. Should I mark it as a known failure?

matthew-brett:
Marking as knownfailure just for Python 2.6 seems good.

We can drop Python 2.6 soon, if it makes it easier, we could drop it now?

nibabel/nicom/csareader.py
```diff
@@ -63,8 +63,10 @@ def get_csa_header(dcm_data, csa_type='image'):
     if section_start is None:
         return None
     element_no = section_start + element_offset
-    # Assume tag exists
-    tag = dcm_data[(0x29, element_no)]
+    try:
+        tag = dcm_data[(0x29, element_no)]
```
bcipolli: @moloney Can you add a comment about when this would fail, and why `None` is an OK value to return?

nibabel/nicom/tests/test_csareader.py
```diff
@@ -14,6 +15,8 @@
 
 from .test_dicomwrappers import (have_dicom, dicom_test,
                                  IO_DATA_PATH, DATA, DATA_FILE)
+if have_dicom:
+    from .test_dicomwrappers import pydicom
```
bcipolli: Another test imports this directly in the relevant test. Seems better to follow the same style; do you mind to move?

bcipolli: Additionally, better to use `skipIf` decorator on the test: `skipif(not have_dicom)`

matthew-brett:
Brendan - can I help with this one?

matthew-brett:
Brendan - what news?  Do you have time to take a look?

nibabel/nicom/csareader.py
moloney: I added a note about why it may fail. The docstring includes info about returning None.

nibabel/nicom/tests/test_csareader.py
moloney: I ended up keeping the top level import and switching the other function to use it. This way the import logic doesn't get replicated and I can use the pydicom version in the `skipif` decorator.

matthew-brett:
Great - thanks for finishing this one up.   Ben - any more comments before I merge?

matthew-brett:
Thanks - Brendan - merging as is, Ben feel free to follow up here with comments if you have them, we'll address them with a new PR.

PR #396, opened 2016-01-07
DOC: Fixed typo

samuelstjean:


matthew-brett:
Thanks for taking the time to do that.

samuelstjean:
Web edits are easy, so it's not a problem to PR it when looking at the doc anyway.

PR #401, opened 2016-02-05
Test for warnings emitted by nibabel.

bcipolli:
When we raise warnings, we should validate them with tests. Here's a couple that were missed.


bcipolli:
I can't get this to work and it's so small, I've lost interest :)

matthew-brett:
I think you just have to import the test function from nibabel.testing ...

effigies:
@matthew-brett Isn't it? Or does it need to be non-relative? https://github.com/bcipolli/nibabel/blob/cfa3ec2bc8da0fb6d1f9d250a39b024ef3102acc/nibabel/tests/test_parrec.py#L26

PR #405, opened 2016-02-12
BF: Load gifti parser iff parsing gifti files

### LINKED ISSUE #392: gifti read fails
I do not know what has been changes in last commits but it seems that parser is added to GiftiImage class only if called using ```__init__```
if using nibabel.load of nibabel.gifti.read before ```GiftiImage.__init__``` has been called once, the following occurs

```
curv=nb.gifti.read('................/lh.white.32k.gii')
AttributeError                            Traceback (most recent call last)
<ipython-input-25-d0fe17dcf62e> in <module>()
----> 1 curv=nb.gifti.read('............./lh.white.32k.gii')

/home/bpinsard/.virtualenvs/prod/lib/python2.7/site-packages/numpy/lib/utils.pyc in newfunc(*args, **kwds)
     98             """`arrayrange` is deprecated, use `arange` instead!"""
     99             warnings.warn(depdoc, DeprecationWarning)
--> 100             return func(*args, **kwds)
    101 
    102         newfunc = _set_function_name(newfunc, old_name)

/home/bpinsard/.virtualenvs/prod/lib/python2.7/site-packages/nibabel/gifti/giftiio.pyc in read(filename)
     29      """
     30     from ..loadsave import load
---> 31     return load(filename)
     32 
     33 

/home/bpinsard/.virtualenvs/prod/lib/python2.7/site-packages/nibabel/loadsave.pyc in load(filename, **kwargs)
     39         is_valid, sniff = image_klass.path_maybe_image(filename, sniff)
     40         if is_valid:
---> 41             return image_klass.from_filename(filename, **kwargs)
     42 
     43     raise ImageFileError('Cannot work out file type of "%s"' %

/home/bpinsard/.virtualenvs/prod/lib/python2.7/site-packages/nibabel/gifti/gifti.pyc in from_filename(klass, filename, buffer_size)
    603     def from_filename(klass, filename, buffer_size=35000000):
    604         file_map = klass.filespec_to_file_map(filename)
--> 605         img = klass.from_file_map(file_map, buffer_size=buffer_size)
    606         return img

/home/bpinsard/.virtualenvs/prod/lib/python2.7/site-packages/nibabel/gifti/gifti.pyc in from_file_map(klass, file_map, buffer_size)
    595             Returns a GiftiImage
    596          """
--> 597         parser = klass.parser(buffer_size=buffer_size)
    598         parser.parse(fptr=file_map['image'].get_prepare_fileobj('rb'))
    599         img = parser.img

AttributeError: type object 'GiftiImage' has no attribute 'parser'
```

effigies:
Setting `parser` as a class attribute turns out not to be very necessary. Avoid the circular reference by calling the parser directly when needed.

Fixes #392.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -423,10 +423,6 @@ class GiftiImage(xml.XmlSerializable, FileBasedImage):
                  labeltable=None, darrays=None, version="1.0"):
         super(GiftiImage, self).__init__(header=header, extra=extra,
                                          file_map=file_map)
-        # placed here temporarily for git diff purposes
-        from .parse_gifti_fast import GiftiImageParser
-        GiftiImage.parser = GiftiImageParser
-
         if darrays is None:
             darrays = []
         if meta is None:
@@ -596,7 +592,8 @@ class GiftiImage(xml.XmlSerializable, FileBasedImage):
         img : GiftiImage
             Returns a GiftiImage
          """
-        parser = klass.parser(buffer_size=buffer_size)
+        from .parse_gifti_fast import GiftiImageParser
+        parser = GiftiImageParser(buffer_size=buffer_size)
         parser.parse(fptr=file_map['image'].get_prepare_fileobj('rb'))
         img = parser.img
         return img
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -592,6 +592,12 @@ class GiftiImage(xml.XmlSerializable, FileBasedImage):
         img : GiftiImage
             Returns a GiftiImage
          """
+        # Note for refactorers: We were aiming to make ``parser`` a class
+        # attribute such that klass.parser().parse(...) would return an
+        # instance of klass.
+        #
+        # In the current setup of parse_gifti_fast.GiftiImageParser, this
+        # results in a circular dependency, so using a local import only.
         from .parse_gifti_fast import GiftiImageParser
         parser = GiftiImageParser(buffer_size=buffer_size)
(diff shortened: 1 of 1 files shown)


effigies:
Couldn't figure out a new test to write, since loading files is already tested. Could `del GiftiImage.parser` before trying to load, but this change removes the attribute altogether, so the test would be obsolete as soon as it's written.

bcipolli:
The intention here was to set a class-level property, similar to how `header_klass` is set on images. So now the corresponding parser class isn't easily discoverable for an image requiring one.

With that said, LGTM! This fixes a bug, and I don't see that having the parser class set explicitly on the class will matter anytime soon. Thanks @effigies !

effigies:
Definitely open to alternatives, but my only other idea was to just merge the two files together and eliminate the import problem.

I could add a comment, in case somebody decides to come back to this, if that would be helpful.

bcipolli:
Right. It's confusing. My solution wasn't the best either.

I think what you've done, with an informative comment, is the best choice available.

bcipolli:
("it's confusing" === "the best solution is hard to figure out")

matthew-brett:
Ouch - nasty - the Image class needs a Parser class and the Parser class needs the Image class.

I stared at it for a while - what do y'all think of this alternative? 

https://github.com/nipy/nibabel/compare/master...matthew-brett:alternative-gifti-fix

effigies:
@matthew-brett :+1: Looks like a good alternative. No suggestions from my end. Feel free to close this PR if you go ahead with it.

matthew-brett:
Closed by #406

PR #406, opened 2016-02-13
MRG: delay set of gifti parser for circular import

matthew-brett:
The GiftiImage class sets GiftiImageParser as a class attribute, but the GiftiImageParser needs to import and use the GiftiImage class.  So, delay the setting of the parser class attribute until GiftiImage is fully defined, and therefore usable from `parse_fast_gifti.py`, which defines GiftiImageParser.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -419,14 +419,15 @@ class GiftiImage(xml.XmlSerializable, FileBasedImage):
     valid_exts = ('.gii',)
     files_types = (('image', '.gii'),)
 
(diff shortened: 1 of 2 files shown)


matthew-brett:
Closes #405 - if accepted ...

effigies:
Test fails before and succeeds after, so seems good as long as nose keeps doing things lexically.

LGTM.

matthew-brett:
Chris - OK to merge this one?

effigies:
:+1:

matthew-brett:
Thanks Chris.

PR #403, opened 2016-02-07
More GIFTI fixes

bcipolli:
This is an old branch that I never got to submitting. More commits to make `GIFTI` more pythonic and test the functionality a bit more thoroughly.

PR #407, opened 2016-02-13
MRG: rewrite gifti docstrings; refactor init

matthew-brett:
Responding to comments on Ben's previous PR #403.

Expand and fix docstrings and gifti attribute descriptions from gifti
spec.

Rewrite GiftiDataArray init to allow full initialization of object, that
can be used from GiftiDataArray.from_array.  Add tests.

Fix up the tests - there were some duplicate test names so some tests were not being run.

Remove unused `label` attribute of `GiftiLabel` object.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -14,6 +14,7 @@ from http://www.nitrc.org/projects/gifti/
 from __future__ import division, print_function, absolute_import
 
 import sys
+import warnings
 
 import numpy as np
 
@@ -407,6 +408,10 @@ class GiftiDataArray(xml.XmlSerializable):
         -------
         da : instance of our own class
         """
+        warnings.warn(
+            "Please use GiftiDataArray constructor instead of from_array "
+            "class method",
+            DeprecationWarning, stacklevel=2)
         return klass(data=darray,
                      intent=intent,
                      datatype=datatype,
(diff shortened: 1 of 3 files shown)


nibabel/gifti/gifti.py
```diff
@@ -104,23 +119,38 @@ def print_summary(self):
...
+    Notes
+    -----
+    freesurfer examples seem not to conform to datatype "NIFTI_TYPE_RGBA32"
+    because they are floats, not 4 8-bit integers.
     """
 
     def __init__(self, key=0, label='', red=None, green=None, blue=None,
                  alpha=None):
         self.key = key
-        self.label = label
```
effigies: Guessing you're removing this because `GiftiLabel().label` isn't referenced anywhere? If so, should it still be in the parameter list?

effigies:
Didn't get to the tests, but the docstrings look good.

Does `GiftiDataArray.from_array` make sense to deprecate, now that it simply passes through to `__init__`?

matthew-brett:
Yes, I guess we could deprecate `from_array`...

effigies:
I read through the tests. Looks reasonable.

I think this refactor of `__init__` makes sense, but somebody who has actually used Gifti might have other opinions.

matthew-brett:
Ben - any comments here?

matthew-brett:
OK - will merge this one tomorrow unless I hear otherwise.

Issue #408, opened 2016-02-13
PARREC:  bug in sorting for some scan types


grlee77:
The current .PAR/.REC implementation uses a fairly simplistic sorting scheme.  This has worked for the majority of the data I have tested it with, but fails for some scan types.  The main problem is that sorting is done via a call to `np.lexsort`, but only a few parameters are used as keys.  In particular, the method `vol_numbers` assigns slices to volumes based on the order they are encountered in the .PAR file.  However, there is no guarantee that this corresponds to the actual acquisition order.

Here are the first several frames of a .PAR file corresponding to a T1-weighted structural, but using two echos rather than one: 

```
# === IMAGE INFORMATION ==========================================================
#  sl ec  dyn ph ty    idx pix scan% rec size                (re)scale              window        angulation              offcentre        thick   gap   info      spacing     echo     dtime   ttime    diff  avg  flip    freq   RR-int  turbo delay b grad cont anis         diffusion       L.ty

 54   1    1  1 0 2     0  16   100  256  256     0.00000   9.97143 3.65651e-003  1070  1860   0.00  -0.00  -0.00   -5.06   12.64   34.10  1.000  0.000 0 2 0 2  1.000  1.000   2.30    0.00     0.00    0.00   1    8.00     0    0    0   448   0.0  1   1    7    0   0.000    0.000    0.000  1
 91   1    1  1 0 2     1  16   100  256  256     0.00000   9.97143 3.65651e-003  1070  1860   0.00  -0.00  -0.00   -5.06   12.64   -2.90  1.000  0.000 0 2 0 2  1.000  1.000   2.30    0.00     0.00    0.00   1    8.00     0    0    0   448   0.0  1   1    7    0   0.000    0.000    0.000  1
 90   1    1  1 0 2     2  16   100  256  256     0.00000   9.97143 3.65651e-003  1070  1860   0.00  -0.00  -0.00   -5.06   12.64   -1.90  1.000  0.000 0 2 0 2  1.000  1.000   2.30    0.00     0.00    0.00   1    8.00     0    0    0   448   0.0  1   1    7    0   0.000    0.000    0.000  1
[...]

Issue #331, opened 2015-07-21
Coveralls config doesn't allow viewing source code coverage changes


bcipolli:
So, I submitted a PR and am getting a failure check because coveralls reports decreased code coverage. I want to fix this:
![image](https://cloud.githubusercontent.com/assets/4072455/8790934/99be1b28-2f08-11e5-8762-7e952f4ba2f5.png)

So I click on the source file to see the coverage changes. But, the repo root directory is not set properly, so I see:
![image](https://cloud.githubusercontent.com/assets/4072455/8790939/b2fb8a08-2f08-11e5-88a9-c866e91d42ec.png)

Not sure how to fix this. Spent about 20 mins trying to find an answer, but only found other people with the same question. Perhaps someone else has solved this before?

Eric89GXL:
I've had this problem on other repos. Tried a lot of things (symlinks, etc.) and the only solution was to ensure `nosetests` (or `py.test`) was run from the `git` root directory, so it imported from the `src` path instead of the install path. This lead to splitting some builds into `install` version and some to `src` version testing.

In the meantime you can get coverage locally. Not as convenient, but looking at the files that have been touched can work okay.

Eric89GXL:
It looks like there are some other workarounds I didn't try (https://github.com/lemurheavy/coveralls-public/issues/471), but I'm not sure it would work.

bcipolli:
@Eric89GXL The link you found is one of the ones I found as well.. but couldn't figure out how to change the settings, even for repos that I own. Note that this is essentially the same as https://github.com/nipy/nipy/issues/311

In the meantime, how would I go about getting coverage locally?

Eric89GXL:
`nosetests --with-coverage --cover-package=nibabel` should work. FYI you can look at what `.travis.yml` says the call should be is since it generates coverage reports (it's the same as what I wrote).

If you want it nicely annotated, I think you can do `--cov-html` as well.

You can also look at the report generated by the Travis bots, but that's a lot slower than running it on your own.

PR #409, opened 2016-02-13
Parrec sort fix

grlee77:
This PR is an attempt to address #408.

I added a new boolean option upon loading a .PAR/.REC file called `strict_sort`.  If it is set to `True` the `get_sorted_slice_indices` method of `PARRECHeader` uses several additional keys in the call to `np.lexsort`.  `strict_sort=False` corresponds to the previously existing sort mode.  This option is made available via a `--strict_sort` command line option to `parrec2nii` as well.

In general, I think it is probably preferable to use `strict_sort=True` in all cases, but I have left the default as `False` for now to ensure backwards compatibility.  Perhaps after a deprecation/testing period we could make `True` the new default.  In particular, it may not be true that for any given DTI scan the b-vals/vectors would be sorted in exactly the same order as previously (If the user is using the b-vals/vectors as exported by `parrec2nii` this should not be an issue, but could cause problems if mixed with a previously stored table from the other approach).

There could be some debate as to the preferred order of priority for the sort keys.  I chose one that seemed logical to me.  As before, slices is the lowest priority so that dimension 3 of the data is still always slices and `is_full` is the highest so that partially acquired scans get moved to the end for possible discarding.  The sorting is a bit complicated, but I tried to add comments to make the reasoning clearer.  The full set of potential sort keys implemented is:
```
slice_nos
echos
cardiac phases
b-vectors
b-values
asl_labels
dynamics
vol_numbers()
is_full
```


grlee77:
If I get a chance, I may scan a phantom to generate a couple of additional complex .PAR files to add to the tests.

matthew-brett:
That would be very good.

If you happen to get a chance to scan a phantom with (different) rotations about all three axes, that would always help a lot - we don't know what order PAR / REC uses for these at the moment.

nibabel/parrec.py
```diff
@@ -631,10 +632,14 @@ def __init__(self, info, image_defs, permit_truncated=False):
         permit_truncated : bool, optional
             If True, a warning is emitted instead of an error when a truncated
             recording is detected.
+        strict_sort : bool, optional, keyword-only
+            If True, a larger number of header fields are used while sorting
+            the REC data array.
```
matthew-brett: Perhaps mention that this can give a different order of slices.

nibabel/tests/test_parrec.py
```diff
@@ -517,9 +554,10 @@ class FakeHeader(object):
     """ Minimal API of header for PARRECArrayProxy
     """
 
-    def __init__(self, shape, dtype):
+    def __init__(self, shape, dtype, strict_sort=False):
         self._shape = shape
         self._dtype = np.dtype(dtype)
+        self.strict_sort = strict_sort
```
matthew-brett: Is this actually used?

nibabel/parrec.py
```diff
@@ -1008,13 +1018,71 @@ def get_sorted_slice_indices(self):
...
+            keys = (slice_nos, echos, phases) + \
+                diffusion_keys + asl_keys + (dynamics, )
+
+            """
+            Data sorting is done in two stages:
+                - run an initial sort using the keys defined above
+                - call vol_is_full to identify potentially missing volumes
+                - call vol_numbers to assign unique volume numbers if for some
+                  reason the keys defined above don't provide a unique sort
+                  order (e.g. this occurs for the Trace volume in DTI)
```
matthew-brett: Do we have an example in the tests where the keys don't produce a unique sort order?

nibabel/parrec.py
```diff
@@ -1008,13 +1018,71 @@ def get_sorted_slice_indices(self):
...
+        else:
+            # Sort based on a larger number of keys.  This is more complicated
+            # but works for .PAR files that get missorted by the above method
+            slice_nos = self.image_defs['slice number']
+            dynamics = self.image_defs['dynamic scan number']
+            phases = self.image_defs['cardiac phase number']
+            echos = self.image_defs['echo number']
+
+            # try adding keys only present in a subset of .PAR files
+            try:
```
matthew-brett: How about:

```
idefs = self.image_defs
asl_keys = (idefs['label_type'],) if 'label_type' in idefs else ()
```

nibabel/parrec.py
```diff
@@ -1008,13 +1018,71 @@ def get_sorted_slice_indices(self):
...
+                asl_labels = self.image_defs['label type']
+                asl_keys = (asl_labels, )
+            except:
+                asl_keys = ()
+            if not self.general_info['diffusion'] == 0:
+                try:
+                    # only present for .PAR v4.1+
+                    bvals = self.image_defs['diffusion b value number']
+                    bvecs = self.image_defs['gradient orientation number']
+                except:
```
matthew-brett: Very good to avoid bare `except` statements - can catch syntax errors, control-C, etc.   How about something like:

```
def get_def(name):
    return self.image_defs[name] if name in self.image_defs else None

bvals = get_def('diffusion b value number')
if bvals is None:
    bvals = get_def('diffusion_b_factor')
bvecs = get_def('gradient orientation number')
if bvecs is None:
    bvecs = tuple(get_def('diffusion'))
```

nibabel/parrec.py
```diff
@@ -1008,13 +1018,71 @@ def get_sorted_slice_indices(self):
...
+                    bvals = self.image_defs['diffusion_b_factor']
+                    # use hash to get a single sortable value per direction
+                    bvecs = [hash(tuple(
```
matthew-brett: I think the only guarantee you have for a hash is that two values that compare equal will have the same hash : https://docs.python.org/2/library/functions.html#hash

There's no guarantee that two values that compare unequal will have a different hash though.  I think the sort order is arbitrary too.   Can't you just use the tuple itself for the sort?

nibabel/parrec.py
```diff
@@ -1008,13 +1018,71 @@ def get_sorted_slice_indices(self):
...
-        keys = (slice_nos, vol_numbers(slice_nos), np.logical_not(is_full))
-        # Figure out how many we need to remove from the end, and trim them
-        # Based on our sorting, they should always be last
+        if not self.strict_sort:
+            slice_nos = self.image_defs['slice number']
+            is_full = vol_is_full(slice_nos, self.general_info['max_slices'])
+            keys = (slice_nos, vol_numbers(slice_nos), np.logical_not(is_full))
+            sort_order = np.lexsort(keys)
+        else:
+            # Sort based on a larger number of keys.  This is more complicated
```
matthew-brett: This function is getting long - maybe have a separate helper function for this?

matthew-brett:
Can you also add a test for `strict_sort` in the script, in `nibabe/test_scripts.py`?

I'm afraid I still haven't got my head round the sorting, but I'll play with the code soon to see if I can understand it.

nibabel/parrec.py
grlee77: Yeah, this case is a bit tricky.  I haven't tested with tuples as the keys to `lexsort`, but maybe that would work.  A simpler option would be to just not support strict_sort with V4.0 PAR files.  As is, the sorting would behave differently for V4.0 where b-values are sorted by magnitude, but in V4.1 they are sorted using a "diffusion b value number" field which, I am not certain, but may instead be related to the order the b-values are scanned.

nibabel/parrec.py
grlee77: Because `image_defs` is a labelled array and not a dictionary I had to modify `get_def` a bit:
```python
    def get_def(self, name):
        """ Return a single image definition field. """
        idef = self.image_defs
        return idef[name] if name in idef.dtype.names else None
```
otherwise the following problem occurs:
```FutureWarning: elementwise == comparison failed and returning scalar instead; this will raise an error or perform elementwise comparison in the future.
  return self.image_defs[name] if name in self.image_defs else None```

File: nibabel/parrec.py (hunk 6 above)
Comment (grlee77) [inline nibabel/parrec.py]: I didn't have luck with tuple of tuples in lexsort, so I think I will handle this via:

```python
bvecs = _direction_numbers(self.get_def('diffusion'))
```
where I define the utility function
```python
def _direction_numbers(bvecs):
    """ Enumerate directions from an (N, 3) array of direction vectors.
    """
    cnt = 0
    bvec_dict = {}
    bvec_nos = np.zeros(len(bvecs))
    for i, bvec in enumerate(bvecs):
        bv = tuple(bvec)
        if bv not in bvec_dict:
            bvec_dict[bv] = cnt
            cnt += 1
        bvec_nos[i] = bvec_dict.get(bv)
    return bvec_nos
```

nibabel/parrec.py
grlee77: After looking into it a bit more closely, I don't think a non-unique order should ever occur (unless we are missing a sort key) and vol_numbers can be removed in the `strict_sort=True` case.  The Trace volume in DTI does actually have a different bval/bvec combination than any of the other volumes, I just didn't read the columns properly before.

One case where I found non-unique sort order was that I was missing a key for `image_type_mr` so when I saved a .PAR with multiple image types (real, imaginary, magnitude, phase), these didn't get sorted properly and there were multiple copies of the same slice per volume, etc unless vol_numbers was used.  However, if the `image_type_mr` key is added then there is a unique sort order again and it is unneeded.

grlee77:
@matthew-brett 
As requested above (but unrelated to this PR), .PAR/.REC files corresponding to various oblique image orientations are made freely available here:
https://github.com/grlee77/parrec_oblique

I have not yet had the time to fully compare the affines produced by the scanner to those generated by `parrec2nii`.

grlee77:
I tried to look at the details for the coveralls failure above, but I get the following error:
```
Source Not Available

The file "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/parrec.py" isn't available on github. Either it's been removed, or the repo root directory needs to be updated.
```

any ideas why that is?

effigies:
@grlee77 See #331.

matthew-brett:
Great - thanks for the data - I make a PR testing against it at #420

nibabel/parrec.py
```diff
@@ -990,6 +1019,54 @@ def get_rec_shape(self):
...
+
+        # Define the desired sort order (last key is highest precedence)
+        keys = (slice_nos, echos, phases) + \
+            diffusion_keys + asl_keys + (dynamics, image_type)
+
+        initial_sort_order = np.lexsort(keys)
+        is_full = vol_is_full(slice_nos[initial_sort_order],
+                              self.general_info['max_slices'])
+
+        # have to "unsort" is_full to match the other sort keys
```
matthew-brett: This logic took me a bit to work out - is there a good test for it?

nibabel/parrec.py
```diff
@@ -329,6 +329,21 @@ def _process_image_lines(image_lines, version):
     return image_defs
 
 
+def _direction_numbers(bvecs):
```
matthew-brett: Maybe put in a test for this helper function?  You could copy https://github.com/yarikoptic/nitest-balls1/blob/master/PARREC/DTI.PAR into the `tests/data` directory to get some sample bvecs.

nibabel/parrec.py
grlee77: Do you mean the `is_full` case specifically?  If so, I don't think I tested that aside from verifying that it still passes the existing tests.  None of the new data sets I added have missing data, although I suppose we could truncate one of the files if desired.

I don't have an exhaustive set of tests for all available sort keys, but I did collect a few .PAR files with 4-6 dimensions to use as tests.  These are the new .PAR files I added in tests/data:

- *T1_dual_echo.PAR*:  A 4D data set:  (x, y, z, echo)
- *T1_3cho_mag_real_imag_phase.PAR*:  A 5D data set:  (x, y, z, echo, image_type)
- *ASL_3D_Multiecho.PAR*:  A 6D data set:  (x, y, z, echo, 'label type', dynamics)
The tests make sure the orders are sorted as expected for these cases.  I did not collect any data with cardiac phases, for example.

One other test I can think of would be to manually scramble the order of the entries in one of the .PAR files and make sure the sort keys that come out at the end still match the unscrambled case.

nibabel/parrec.py
grlee77: okay.  perhaps that was the source of the coveralls complaint

grlee77: Actually, no bvecs will be present in older V4 .PAR files so this `_direction_numbers` function is unnecessary and would never be called.  I will remove it.

I submitted a small PR to fix an existing bug when the 'diffusion' field isn't present in `image_defs`#426

grlee77:
@matthew-brett 
I think this is now ready for review again.  The primary changes since last time you reviewed are:

- I removed the previously untested `_direction_numbers` routine since it was based on a 'diffusion' field in `image_defs` that isn't even present in the V4 .PAR files that would have been the only case to call this routine (see #426)
- In the tests for strict sort I made sure they pass even when the image_defs array is randomly sorted via `np.shuffle`.
- Since V4 .PAR files have no `gradient orientation number` key, I added `vol_numbers` back in as a key in the strict sort routine so the diffusion directions would get assigned unique volume numbers even for V4 .PAR files.  As was already the case for `vol_is_full` this has to be done after the initial stage of sorting.  In this way the sorting is robust even to random shuffling of the image info lines from the .PAR file.

nibabel/parrec.py
```diff
@@ -725,19 +734,26 @@ def get_bvals_bvecs(self):
...
+            # All 3 values of bvecs should be same within volume
+            assert not np.any(np.diff(bvecs, axis=0))
         bvecs = bvecs[0]
         # rotate bvecs to match stored image orientation
         permute_to_psl = ACQ_TO_PSL[self.get_slice_orientation()]
         bvecs = apply_affine(np.linalg.inv(permute_to_psl), bvecs)
         return bvals, bvecs
 
+    def get_def(self, name):
+        """ Return a single image definition field. """
```
matthew-brett: Add "or None if field not present" ?

nibabel/parrec.py
```diff
@@ -725,19 +734,26 @@ def get_bvals_bvecs(self):
         n_slices, n_vols = self.get_data_shape()[-2:]
         bvals = self.image_defs['diffusion_b_factor'][reorder].reshape(
             (n_slices, n_vols), order='F')
-        # All bvals within volume should be the same
-        assert not np.any(np.diff(bvals, axis=0))
+        if not self.strict_sort:
```
matthew-brett: Can you think of a better place for these checks?  Maybe there's a way to check these after doing the sorting?

nibabel/parrec.py
```diff
@@ -990,6 +1006,62 @@ def get_rec_shape(self):
...
+        (see :func:`vol_numbers`).  This may occur for diffusion scans from
+        older V4 .PAR format, where diffusion direction info is not stored.
+
+        Data sorting is done in two stages:
+            - run an initial sort using several keys of interest
+            - call `vol_is_full` to identify potentially missing volumes
+              and add the result to the list of sort keys
+        """
+        # Sort based on a larger number of keys.  This is more complicated
+        # but works for .PAR files that get missorted by the above method
```
matthew-brett: I guess it's not "the above method" now.

Maybe put the other sorting method into a `lax_sort` method, and refer to that?

nibabel/parrec.py
```diff
@@ -990,6 +1006,62 @@ def get_rec_shape(self):
...
+        slice_nos = self.image_defs['slice number']
+        dynamics = self.image_defs['dynamic scan number']
+        phases = self.image_defs['cardiac phase number']
+        echos = self.image_defs['echo number']
+        image_type = self.image_defs['image_type_mr']
+
+        # try adding keys only present in a subset of .PAR files
+        idefs = self.image_defs
```
matthew-brett: Put this at top and use it for first set of `image_defs` references?

nibabel/parrec.py
```diff
@@ -990,6 +1006,62 @@ def get_rec_shape(self):
...
+        asl_keys = (idefs['label type'], ) if 'label type' in \
+            idefs.dtype.names else ()
+
+        if not self.general_info['diffusion'] == 0:
```
matthew-brett: Or:

```
if self.general_info['diffusion'] != 0:
```

nibabel/parrec.py
```diff
@@ -1008,13 +1085,18 @@ def get_sorted_slice_indices(self):
             array, and (equivalently) the only dimension of
             ``self.image_defs``.
         """
-        slice_nos = self.image_defs['slice number']
-        is_full = vol_is_full(slice_nos, self.general_info['max_slices'])
-        keys = (slice_nos, vol_numbers(slice_nos), np.logical_not(is_full))
-        # Figure out how many we need to remove from the end, and trim them
-        # Based on our sorting, they should always be last
+        if not self.strict_sort:
```
matthew-brett: Maybe move into `_lax_sort_keys method` and do `keys = self._strict_sort_keys() if self.strict_sort else self._lax_sort_keys()`

matthew-brett:
Great - thanks for the changes - nearly there.

It's good to avoid backslashes for line continuation if possible - https://www.python.org/dev/peps/pep-0008/#id17

nibabel/parrec.py
grlee77: The check should also be performed in the strict sort case since these lines occur after the sort, so this change was a mistake!  I think these checks should then stay in this function.

grlee77:
I made the suggested changes above.

I also simplified the final lines of `_strict_sort_order`.  I was able to remove the "unsorting" entirely.  Now it is an initial sort as before, followed by a second sort that is just:

```initial_sort_order[np.lexsort((vol_nos, is_full))]```

The docstring was expanded to (hopefully!) better describe what is going on in the function`.

nibotmi:
:umbrella: The latest upstream changes (presumably #426) made this pull request unmergeable. Please resolve the merge conflicts.

grlee77:
rebased to fix merge conflicts with #426

matthew-brett:
Thanks for your patience - it looks good - merging.

grlee77:
Great!  This will be very useful for some of the data we acquire.

PR #413, opened 2016-02-27
MRG: function to return voxel size from affine

matthew-brett:
Voxel sizes are the Euclidean lengths of the columns of the affine
(excluding the last column, and assuming zeros in the last row).


Eric89GXL:
LGTM, +1 for merge

matthew-brett:
Sorry - I completely rewrote the explanation in the hope that it is more clear.

Eric89GXL:
Still LGTM :)

matthew-brett:
OK - thanks, merging.

PR #414, opened 2016-03-01
ENH: Add write_morph_data to freesurfer.io

effigies:
The first commit was in the midst of #249 and causing PEP8 warnings. Not otherwise relevant to Cifti, so  I'm submitting as its own PR. Fixed the style and brought a little more in line with `freesurfer.io` functions.

### DIFF
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -6,6 +6,7 @@ import time
 
 
 from .. externals.six.moves import xrange
+from ..openers import Opener
 
 
 def _fread3(fobj):
@@ -160,8 +161,8 @@ def read_morph_data(filepath):
     return curv
 
 
-def write_morph_data(filepath, values, fnum=0):
-    """Write Freesurfer morphometry data `values` to file `filepath`
+def write_morph_data(file_like, values, fnum=0):
+    """Write Freesurfer morphometry data `values` to file-like `file_like`
 
     Equivalent to FreeSurfer's `write_curv.m`_
 
@@ -173,8 +174,9 @@ def write_morph_data(filepath, values, fnum=0):
 
     Parameters
     ----------
-    filepath : str
-        Path to annotation file to be written
+    file_like : file-like
+        String containing path of file to be written, or file-like object, open
+        in binary write (`'wb'` mode, implementing the `write` method)
     values : array-like
         Surface morphometry values
     fnum : int, optional
@@ -193,13 +195,13 @@ def write_morph_data(filepath, values, fnum=0):
     if len(array.shape) > 1:
         raise ValueError("Multi-dimensional values not supported")
 
-    with open(filepath, 'wb') as fobj:
-        magic_bytes.tofile(fobj)
+    with Opener(file_like, 'wb') as fobj:
+        fobj.write(magic_bytes)
 
         # vertex count, face count (unused), vals per vertex (only 1 supported)
-        np.array([len(values), fnum, 1], dtype='>i4').tofile(fobj)
+        fobj.write(np.array([len(values), fnum, 1], dtype='>i4'))
 
-        array.tofile(fobj)
+        fobj.write(array)
 
 
 def read_annot(filepath, orig_ids=False):
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -160,23 +160,27 @@ def read_morph_data(filepath):
     return curv
 
 
-def write_morph_data(filename, values):
-    '''
-    '''
-    with open(filename, 'wb') as f:
+def write_morph_data(filepath, values):
+    """Write out a Freesurfer morphometry data file.
 
-      # magic number
-      np.array([255], dtype='>u1').tofile(f)
-      np.array([255], dtype='>u1').tofile(f)
-      np.array([255], dtype='>u1').tofile(f)
+    See:
+    http://www.grahamwideman.com/gw/brain/fs/surfacefileformats.htm#CurvNew
 
-      # vertices number and two un-used int4
-      np.array([len(values)], dtype='>i4').tofile(f)
-      np.array([0], dtype='>i4').tofile(f)
-      np.array([1], dtype='>i4').tofile(f)
+    Parameters
+    ----------
+    filepath : str
+        Path to annotation file to be written
+    values : ndarray, shape (n_vertices,)
+        Surface morphometry values
+    """
+    magic_bytes = np.array([255, 255, 255], dtype=np.uint8)
+    with open(filepath, 'wb') as fobj:
+        magic_bytes.tofile(fobj)
 
-      # now the data
-      np.array(values, dtype='>f4').tofile(f)
+        # vertex count, face count (unused), vals per vertex (only 1 supported)
+        np.array([len(values), 0, 1], dtype='>i4').tofile(fobj)
+
+        np.array(values, dtype='>f4').tofile(fobj)
 
 
 def read_annot(filepath, orig_ids=False):
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -160,6 +160,25 @@ def read_morph_data(filepath):
     return curv
 
 
+def write_morph_data(filename, values):
+    '''
+    '''
+    with open(filename, 'wb') as f:
+
+      # magic number
+      np.array([255], dtype='>u1').tofile(f)
+      np.array([255], dtype='>u1').tofile(f)
+      np.array([255], dtype='>u1').tofile(f)
+
+      # vertices number and two un-used int4
+      np.array([len(values)], dtype='>i4').tofile(f)
+      np.array([0], dtype='>i4').tofile(f)
+      np.array([1], dtype='>i4').tofile(f)
+
+      # now the data
+      np.array(values, dtype='>f4').tofile(f)
+
+
 def read_annot(filepath, orig_ids=False):
     """Read in a Freesurfer annotation from a .annot file.
 
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -179,17 +179,20 @@ def write_morph_data(file_like, values, fnum=0):
         in binary write (`'wb'` mode, implementing the `write` method)
     values : array-like
         Surface morphometry values
+
+        Shape must be (N,), (N, 1), (1, N) or (N, 1, 1)
     fnum : int, optional
         Number of faces in the associated surface
     """
     magic_bytes = np.array([255, 255, 255], dtype=np.uint8)
 
-    array = np.asarray(values).astype('>f4').squeeze()
-    if len(array.shape) > 1:
-        raise ValueError("Multi-dimensional values not supported")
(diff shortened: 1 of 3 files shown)


agramfort:
LGTM

nibabel/freesurfer/io.py
```diff
@@ -160,6 +160,29 @@ def read_morph_data(filepath):
     return curv
 
 
+def write_morph_data(filepath, values):
+    """Write out a Freesurfer morphometry data file.
```
matthew-brett: Mention parameters in first line?  Something like

```
Write freesurfer morphometry data `values` to file `filepath`
```

nibabel/freesurfer/io.py
```diff
@@ -160,6 +160,29 @@ def read_morph_data(filepath):
...
+
+    See:
+    http://www.grahamwideman.com/gw/brain/fs/surfacefileformats.htm#CurvNew
+
+    Parameters
+    ----------
+    filepath : str
```
matthew-brett: Any reason this cannot be file-like?

nibabel/freesurfer/io.py
```diff
@@ -160,6 +160,29 @@ def read_morph_data(filepath):
...
+        Path to annotation file to be written
+    values : ndarray, shape (n_vertices,)
```
matthew-brett: This can be array-like (list, tuple)?

nibabel/freesurfer/io.py
```diff
@@ -160,6 +160,29 @@ def read_morph_data(filepath):
...
+        Surface morphometry values
+    """
+    magic_bytes = np.array([255, 255, 255], dtype=np.uint8)
+    with open(filepath, 'wb') as fobj:
+        magic_bytes.tofile(fobj)
+
+        # vertex count, face count (unused), vals per vertex (only 1 supported)
+        np.array([len(values), 0, 1], dtype='>i4').tofile(fobj)
```
matthew-brett: Need to check length fits in `i4`?

nibabel/freesurfer/io.py
```diff
@@ -160,6 +160,29 @@ def read_morph_data(filepath):
...
+
+        np.array(values, dtype='>f4').tofile(fobj)
```
matthew-brett: Just possible (if this has been loaded off disk for example) array is already in correct format, in which case you can avoid copy by `np.asarray(values).astype('>f4').tofile(fobj)`.

matthew-brett: Worth checking that `values` is 1D?  Or should 2D be allowed?

nibabel/freesurfer/tests/test_io.py
```diff
@@ -92,6 +92,11 @@ def test_morph_data():
     curv = read_morph_data(curv_path)
     assert_true(-1.0 < curv.min() < 0)
     assert_true(0 < curv.max() < 1.0)
+    with InTemporaryDirectory():
+        new_path = 'test'
+        write_morph_data(new_path, curv)
+        curv2 = read_morph_data(new_path)
+        assert_equal(curv2, curv)
```
matthew-brett: Maybe a test reading an existing curv data file from freesurfer examples, re-writing and reading again?  E.g. `nibabel-data/nitest-freesurfer/fsaverage/surf/lh.curv` using decorator like : https://github.com/nipy/nibabel/blob/master/nibabel/freesurfer/tests/test_io.py#L48

nibabel/freesurfer/io.py
```diff
@@ -160,6 +160,29 @@ def read_morph_data(filepath):
...
+
+    Parameters
+    ----------
+    filepath : str
+        Path to annotation file to be written
+    values : ndarray, shape (n_vertices,)
+        Surface morphometry values
+    """
+    magic_bytes = np.array([255, 255, 255], dtype=np.uint8)
+    with open(filepath, 'wb') as fobj:
```
matthew-brett: If you go with `file_like` for the input argument this could be:

```
from ..openers import Opener
with Opener(file_like, 'wb') as fobj:
```

nibabel/freesurfer/io.py
effigies: Not that I know of. Would the convention for that would be the following?

    file_like : file-like

matthew-brett: Yup, then something in the description to say parameter can be filename string or open file-like object with constraints, which in your case are that the object has to be open in binary write mode and implement the `write` method (I suppose, for the array `tofile` method).

effigies:
Looking at [`write_curv.m`](https://github.com/neurodebian/freesurfer/blob/debian-sloppy/matlab/write_curv.m), the MATLAB convention is to pass an `fnum` (number of faces) to write as the second of that triplet. Should that be an optional argument, to make converting MATLAB scripts as straightforward as possible?

nibabel/freesurfer/io.py
effigies: `np.array([255, 255, 255], dtype=np.uint8).tofile(fobj)` for such a `fobj` results in:

Python 2.7:

    IOError: first argument must be a string or open file

Python 3:

    OSError: first argument must be a string or open file

(Numpy 1.8.2, in both cases.)

nibabel/freesurfer/tests/test_io.py
effigies: That's what this modification does.

nibabel/freesurfer/io.py
effigies: Turns out `tofile` doesn't like things that aren't actual files. Switched to `fobj.write()`.

effigies:
Added `fnum`. I can drop that if it seems like a bad idea. Other comments addressed.

nibabel/freesurfer/tests/test_io.py
matthew-brett: Oh sorry - yes I see.

matthew-brett: Add tests for errors for too-large array and array > 1D?

effigies:
New tests added. Subtle bug for vectors of shape `(1, n)` fixed.

nibabel/freesurfer/io.py
```diff
@@ -160,6 +161,49 @@ def read_morph_data(filepath):
...
+        String containing path of file to be written, or file-like object, open
+        in binary write (`'wb'` mode, implementing the `write` method)
+    values : array-like
+        Surface morphometry values
+    fnum : int, optional
+        Number of faces in the associated surface
+    """
+    magic_bytes = np.array([255, 255, 255], dtype=np.uint8)
+
+    array = np.asarray(values).astype('>f4').squeeze()
```
matthew-brett: Just checking, but you want to allow arrays shape (1, 1, 20, 1, 1)?

effigies: Is there a good reason not to?

matthew-brett: You could imagine that user was making a mistake, passing in a multidimensional array, that just happened to have only one dimension not length 1.   I mean, maybe you actually want to accept any of shape (N,), (1, N), (N, 1), but not other shapes.   As in `if array.ndim > 2 or (array.ndim == 2 and not 1 in array.shape):`

Actually maybe `array` is a bit confusing as name (looks like `np.array`) ...

effigies: I suppose. Maybe it makes more sense to let people do their own squeezing, if that's what they want.

I would say allow (N, 1, 1) as well, though, since that's the default shape coming in from MGH or surface Nifti files.

matthew-brett: Sounds reasonable.

nibabel/freesurfer/io.py
```diff
@@ -160,6 +161,52 @@ def read_morph_data(filepath):
...
+
+        Shape must be (N,), (N, 1), (1, N) or (N, 1, 1)
+    fnum : int, optional
+        Number of faces in the associated surface
+    """
+    magic_bytes = np.array([255, 255, 255], dtype=np.uint8)
+
+    vector = np.asarray(values)
+    vnum = np.prod(vector.shape)
+    if vector.shape not in ((vnum,), (vnum, 1), (1, vnum), (vnum, 1, 1)):
```
matthew-brett: Nice.

matthew-brett:
Great - thanks for your patience.

matthew-brett:
Oops - Appveyor failure ...

effigies:
32bit Windows issue: https://ci.appveyor.com/project/Eric89GXL/nibabel/build/1.0.587/job/xrwxr6jdbj3s4vj9#L876

`np.lib.stride_tricks.as_strided(np.array(0), (np.iinfo('i4').max + 1,), [0])` results in `OverflowError: Python int too large to convert to C long`.

Not sure where to start with that one.

nibabel/freesurfer/tests/test_io.py
```diff
@@ -92,6 +94,32 @@ def test_morph_data():
...
+    with InTemporaryDirectory():
+        for shape in okay_shapes:
+            write_morph_data('test.curv', values.reshape(shape))
+            # Check ordering is preserved, regardless of shape
+            assert_equal(values, read_morph_data('test.curv'))
+        assert_raises(ValueError, write_morph_data, 'test.curv',
+                      np.zeros(shape), big_num)
+        if ctypes.c_long is not ctypes.c_int32:
+            assert_raises(ValueError, write_morph_data, 'test.curv',
+                          strided_scalar((big_num,)))
```
effigies: Alternatives, if this is ugly:

1) Check `np.iinfo(ctypes.c_long)`
2) `try: strided_scalar((big_num,))`

matthew-brett: How about:

```
if np.dtype(np.long) != np.dtype(np.int32):  # Windows 32-bit overflows Python int
```

effigies: Doesn't seem to capture the case for [PYTHON_VERSION=2.7, PYTHON_ARCH=32](https://ci.appveyor.com/project/Eric89GXL/nibabel/build/1.0.589).

nibabel/freesurfer/tests/test_io.py
```diff
@@ -92,6 +93,33 @@ def test_morph_data():
...
+    big_num = np.iinfo('i4').max + 1
+    with InTemporaryDirectory():
+        for shape in okay_shapes:
+            write_morph_data('test.curv', values.reshape(shape))
+            # Check ordering is preserved, regardless of shape
+            assert_equal(values, read_morph_data('test.curv'))
+        assert_raises(ValueError, write_morph_data, 'test.curv',
+                      np.zeros(shape), big_num)
+        # Windows 32-bit overflows Python int
+        if np.dtype(np.long) != np.dtype(np.int32):
```
matthew-brett: Whoops.  How about `np.dtype(int) != np.dtype(np.int32)` ?

matthew-brett:
Looks good on the tests - will merge unless I hear otherwise tomorrow.

effigies:
Just squashed the final two commits. Should be good to go. Thanks for the review.

matthew-brett:
Great - thanks a lot.

PR #416, opened 2016-03-03
ENH: Add "element" containers and make dicom wrappers compatible

moloney:
Add two main types: `ElemDict` and `ElemList`. Each can only store
objects with an attribute `value`. Standard lookups with
`__getitem__` and `get` return just this `value` attribute. To
access the underlying object, a `get_elem` method is provided.

Also updated the DICOM wrappers to provide a `get_elem` method that
returns the full DICOM element rather than just the value. Added a
`__iter__` method that returns the available DICOM keys.

### DIFF
--- a/nibabel/elemcont.py
+++ b/nibabel/elemcont.py
@@ -2,9 +2,9 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 '''Containers that provide easy access to the values of nested elements
 
-These containers are for storing "elements" which have both a core data value 
-as well as some additional meta data. When indexing into these containers it is 
-this core value that is returned, which allows for much cleaner and more 
+These containers are for storing "elements" which have both a core data value
+as well as some additional meta data. When indexing into these containers it is
+this core value that is returned, which allows for much cleaner and more
 readable access to nested structures.
 
 Each object stored in these containers must have an attribute `value` which
@@ -26,7 +26,7 @@ class MetaElem(object):
 
 class InvalidElemError(Exception):
     '''The object being added to the container doesn't have a `value` attribute
-	'''
+    '''
     def __init__(self, invalid_val):
         message = ("Provided value '%s' of type %s does not have a 'value' "
                    "attribute" % (invalid_val, type(invalid_val)))
@@ -35,10 +35,10 @@ class InvalidElemError(Exception):
 
 class ElemDict(MutableMapping):
     '''Ordered dict-like providing easy access to nested elements
- 
-    Each value added to the dict must in turn have a `value` attribute, which 
-	is what is returned by subsequent calls to `__getitem__`. To get the 
-	element itself use the `get_elem` method.
+
+    Each value added to the dict must in turn have a `value` attribute, which
+    is what is returned by subsequent calls to `__getitem__`. To get the
+    element itself use the `get_elem` method.
     '''
 
     def __init__(self, *args, **kwargs):
@@ -94,13 +94,14 @@ class ElemDict(MutableMapping):
 class ElemList(MutableSequence):
     '''A list-like container providing easy access to nested elements
 
-    Each value added to the list must in turn have a `value` attribute, which 
-	is what is returned by subsequent calls to `__getitem__`. To get the 
-	element itself use the `get_elem` method.
+    Each value added to the list must in turn have a `value` attribute, which
+    is what is returned by subsequent calls to `__getitem__`. To get the
+    element itself use the `get_elem` method.
     '''
     def __init__(self, data=None):
         self._elems = list()
-        if data is None: return
+        if data is None:
+            return
         if isinstance(data, self.__class__):
             for idx in range(len(data)):
                 self.append(data.get_elem(idx))
@@ -114,11 +115,11 @@ class ElemList(MutableSequence):
         (start, end, step) = slc.indices(len(self))
         # Replace (0, -1, 1) with (0, 0, 1) (misfeature in .indices()).
         if step == 1:
-          if end < start:
-            end = start
-          step = None
+            if end < start:
+                end = start
+            step = None
         if slc.step is None:
-          step = None
+            step = None
         return (start, end, step)
 
     def __getitem__(self, idx):
@@ -187,4 +188,3 @@ class ElemList(MutableSequence):
 
     def get_elem(self, idx):
         return self._elems[idx]
-
--- a/nibabel/elemcont.py
+++ b/nibabel/elemcont.py
@@ -28,9 +28,8 @@ class InvalidElemError(Exception):
     '''The object being added to the container doesn't have a `value` attribute
 	'''
     def __init__(self, invalid_val):
-        self.invalid_val = invalid_val
         message = ("Provided value '%s' of type %s does not have a 'value' "
-                   "attribute" % (self.invalid_val, type(invalid_val)))
+                   "attribute" % (invalid_val, type(invalid_val)))
         super(InvalidElemError, self).__init__(message)
 
 
--- a/nibabel/elemcont.py
+++ b/nibabel/elemcont.py
@@ -17,7 +17,7 @@ from .externals import OrderedDict
 from .externals.six import iteritems
 
 
-class Elem(object):
+class MetaElem(object):
     '''Basic element type has a `value` and a `meta` attribute.'''
     def __init__(self, value, meta=None):
         self.value = value
--- a/nibabel/elemcont.py
+++ b/nibabel/elemcont.py
@@ -100,13 +100,13 @@ class ElemList(MutableSequence):
     '''
     def __init__(self, data=None):
         self._elems = list()
-        if data is not None:
-            if isinstance(data, self.__class__):
-                for idx in range(len(data)):
-                    self.append(data.get_elem(idx))
-            else:
-                for elem in data:
-                    self.append(elem)
+        if data is None: return
+        if isinstance(data, self.__class__):
+            for idx in range(len(data)):
+                self.append(data.get_elem(idx))
+        else:
+            for elem in data:
+                self.append(elem)
 
     def _tuple_from_slice(self, slc):
         '''Get (start, end, step) tuple from slice object.
@@ -117,7 +117,7 @@ class ElemList(MutableSequence):
           if end < start:
             end = start
           step = None
-        if slc.step == None:
+        if slc.step is None:
           step = None
         return (start, end, step)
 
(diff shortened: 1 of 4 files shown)


moloney:
This is a bite-sized chunk from #290. The goal is to provide a nice standard way to access meta data structures, especially once they become nested. So instead of having to do something like:

```python
meta['key1'].value['key2'].value[0].value
```

we can do:

```python
meta['key1']['key2'][0]
```

If you need to access the full "element" you can use the `get_elem` method. The dicomwrappers classes were lacking this ability and so I added a `get_elem` method that returns the full pydicom `DataElement` object. I also added a `__iter__` method that produces the available DICOM keywords for the data set. It may also make sense to add some other dict-like methods, I am thinking at least `keys`.

moloney:
I think we should make meta data access on the DICOM wrappers the same as meta data access on other Nibabel image classes.  Are we okay with making the image classes dict-like for meta data access? If not, we should figure out what we want and implement it for the DICOM wrappers as well.

For the other image classes things are a bit more complicated as there is also the existing "header" attribute which contains some meta data. I guess we could make it so that `nii.header['dim_info']` is the same as `nii['header']['dim_info']`?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+'''Containers for storing "elements" which have both a core data value as well
```
matthew-brett: Single first line (docstring standard - also makes auto-generated docs look nicer).

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+
+class Elem(object):
+    '''Basic element type has a `value` and a `meta` attribute.'''
+    def __init__(self, value, meta=None):
+        self.value = value
+        self.meta = {} if meta is None else meta
+
+
+class InvalidElemError(Exception):
+    '''Raised when trying to add an object without a `value` attribute to an
```
matthew-brett: Single self-contained first line of docstring.

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+    `ElemDict`.'''
+    def __init__(self, invalid_val):
+        self.invalid_val = invalid_val
```
matthew-brett: Do you need to store the value?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+provides the core data value for the element. To get the element object itself
+the `get_elem` method must be used.
+'''
+from collections import MutableMapping, MutableSequence
+
+from .externals import OrderedDict
+from .externals.six import iteritems
+
+
+class Elem(object):
```
matthew-brett: What do you think of the name 'MetaValue'?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+    `ElemDict`.'''
+    def __init__(self, invalid_val):
+        self.invalid_val = invalid_val
+        message = ("Provided value '%s' of type %s does not have a 'value' "
+                   "attribute" % (self.invalid_val, type(invalid_val)))
+        super(InvalidElemError, self).__init__(message)
+
+
+class ElemDict(MutableMapping):
+    '''Ordered dict-like where each value is an "element", which is defined as
```
matthew-brett: Self-contained first line of docstring.

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+    any object which has a `value` attribute.
+
+    When looking up an item in the dict, it is this `value` attribute that
+    is returned. To get the element itself use the `get_elem` method.
+    '''
+
+    def __init__(self, *args, **kwargs):
```
matthew-brett: Do I understand right that `args[0]` can be any of

* ElemDict
* dict
* sequence of (key, value) sequences ?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __len__(self):
+        return len(self._elems)
+
+    def __repr__(self):
+        return ('ElemDict(%s)' %
+                ', '.join(['%r=%r' % x for x in self.items()]))
+
+    def update(self, other):
+        if hasattr(other, 'get_elem'):
```
matthew-brett: Maybe add `iter_elems` method to return key, element pairs, and use that?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+        else:
+            for key, elem in iteritems(other):
+                self[key] = elem
+
+    def get_elem(self, key):
+        return self._elems[key]
+
+
+class ElemList(MutableSequence):
+    '''A list-like container where each value is an "element", which is
```
matthew-brett: Self-contained single line.

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+    defined as any object which has a `value` attribute.
+
+    When looking up an item in the list, it is this `value` attribute that
+    is returned. To get the element itself use the `get_elem` method.
+    '''
+    def __init__(self, data=None):
+        self._elems = list()
+        if data is not None:
```
matthew-brett: Reduce indentation with `if data is None: return`.

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+    def __getitem__(self, idx):
+        if isinstance(idx, slice):
+            return ElemList(self._elems[idx])
+        else:
+            return self._elems[idx].value
+
+    def __setitem__(self, idx, val):
+        if isinstance(idx, slice):
+            (start, end, step) = self._tuple_from_slice(idx)
+            if step != None:
```
matthew-brett: I think `if step is not None` preferred.

matthew-brett: Maybe (indentation again):

```
if step is None:
    # Normal slice
    for j, assign_val in enumerate(val):
        self.insert(start + j, assign_val)
        return
# Extended slice
...
```

Sorry to keep on with this kind of thing, but I personally find that easier to read, because then, after I've read some `if something: do something; return` clause, I know I can forget about that clause.  Otherwise I've got to look further down the method to see whether something else is happening as well.

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __len__(self):
+        return len(self._elems)
+
+    def __repr__(self):
+        return ('ElemList([%s])' % ', '.join(['%r' % x for x in self]))
+
+    def __add__(self, other):
+        result = self.__class__(self)
+        if isinstance(other, self.__class__):
```
matthew-brett: How about an `iter_elems` method, use that here and in `__init__` ?

nibabel/nicom/dicomwrappers.py
```diff
@@ -21,6 +21,14 @@
 from ..openers import ImageOpener
 from ..onetime import setattr_on_read as one_time
 
+try:
```
matthew-brett: Worth a `pydicom_adapter` module for this kind of thing?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __getitem__(self, key):
+        return self._elems[key].value
+
+    def __setitem__(self, key, val):
+        if not hasattr(val, 'value'):
+            raise InvalidElemError(val)
+        self._elems[key] = val
+
+    def __delitem__(self, key):
```
matthew-brett: You could avoid these three methods by subclassing `dict` - any reason not to?

nibabel/elemcont.py
```diff
@@ -0,0 +1,188 @@
...
+                for j, assign_val in enumerate(val):
+                    self.insert(indices[j], assign_val)
+            else:
+                # Normal slice
+                for j, assign_val in enumerate(val):
+                    self.insert(start + j, assign_val)
+        else:
+            self.insert(idx, val)
+
+    def __delitem__(self, idx):
```
matthew-brett: Maybe subclass `list` to avoid these guys?

nibabel/elemcont.py
moloney: Maybe 'MetaElem'? I think it helps make the difference between 'elements' and 'values' clear.

nibabel/elemcont.py
moloney: I suppose not, the exception itself makes it pretty clear what is wrong with the value.

nibabel/elemcont.py
moloney: This is trying to replicate the behavior of the `dict` constructor. You are correct that `args[0]` can be dict-like of a list tuples. Either way the elements need to have a `value` attribute.

nibabel/elemcont.py
moloney: I believe that subclassing MutableMapping is preferred for more "custom" dict-like classes, rather than just extending the API of a standard `dict`. In particular, other standard `dict` methods won't call your custom `__getitem__` and `__setitem__` methods.

nibabel/elemcont.py
moloney: Same as above, I think MutableSequence is preferred in this situation.

nibabel/elemcont.py
moloney: I agree `iter_elems` is good for the `ElemDict` class, but this should really be using `enumerate` right?

nibabel/nicom/dicomwrappers.py
moloney: Yes, this sounds like a good idea.  Would it go in the `nicom` package, or somewhere else?

nibabel/elemcont.py
matthew-brett: At least this:

```
class MyDict(dict):
    def __getitem__(self, key):
        return 'foo'

    def my_method(self):
        return self['something']

d = MyDict()
print(d.my_method())
```

gives `foo` - or did you mean something else?

I guess `MutableMapping` doesn't have things like `__len__` etc?

nibabel/nicom/dicomwrappers.py
matthew-brett: Yes, `nicom` seems reasonable.

nibabel/elemcont.py
matthew-brett: Yup - I like `MetaElem`

nibabel/elemcont.py
matthew-brett: How about `iter_mitems` for the dict and `iter_melems` for the list?

nibabel/nicom/dicomwrappers.py
matthew-brett: I ended up writing something of the sort for pr/296 - see : https://github.com/kastman/nibabel/pull/1/commits/2065503e9566855a156c34a97d131db73975726c

nibotmi:
:umbrella: The latest upstream changes (presumably #439) made this pull request unmergeable. Please resolve the merge conflicts.

effigies:
@moloney Thanks for rebasing this against master.

@matthew-brett Do you have the time to resume reviewing this (and/or #417, #419)?

effigies:
Please feel free to reopen with a fresh PR.

matthew-brett:
I can't this week, but if you can wait until middle of next, I can review.

PR #417, opened 2016-03-03
ENH: Add writer for Siemens CSA header

moloney:
Allows us to take a parsed CSA header and convert it back into a
string. Useful for things like DICOM anonymization, or perhaps
round tripping DICOM -> Nifti -> DICOM.

### DIFF
--- a/nibabel/nicom/csareader.py
+++ b/nibabel/nicom/csareader.py
@@ -188,11 +188,11 @@ def write(csa_header):
         result.append(csa_header['unused0'])
     if not 0 < csa_header['n_tags'] <= 128:
         raise CSAWriteError('Number of tags `t` should be '
-                           '0 < t <= 128')
+                            '0 < t <= 128')
     result.append(struct.pack('2I',
                               csa_header['n_tags'],
                               csa_header['check'])
-                 )
+                  )
 
     # Build list of tags in correct order
     tags = list(csa_header['tags'].items())
@@ -212,7 +212,7 @@ def write(csa_header):
                                   tag_dict['syngodt'],
                                   n_items,
                                   tag_dict['last3'])
-                     )
+                      )
 
         # Figure out the number of values for this tag
         if vm == 0:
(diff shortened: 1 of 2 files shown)


moloney:
This is another bite-sized chunk from #290. 

I was tempted to make a `csa` module with the read/writer functionality and then setup some proxies in `csareader` for backwards compatibility, but I am not sure it is worth the hassle? 

It would also be nice to use `ElemDict` objects (as specified in #416) for the parsed representation. Again, I guess we need to provide some backwards compatibility here.

nibotmi:
:umbrella: The latest upstream changes (presumably #393) made this pull request unmergeable. Please resolve the merge conflicts.

matthew-brett:
The csa representation as we wrote it originally is a bit kludgy, I think - if you're interested - it would be worth improving that with a new csa module, and depracting the `csareaders` module.  

I don't think we have to worry as much as usual about backwards compatibility - the `nicom` module has always had a bit fat warning on it.  So, avoid breakage if possible, but don't do excessive work to keep compatibility otherwise.

matthew-brett:
By the way - thanks very much for breaking out the change from the big PR - it's extremely helpful.

moloney:
@ZviBaratz @matthew-brett @effigies 

During our recent meeting, I forgot the original reason I wrote this was to support anonymization of Siemens mosaic/DWI Dicom files (not round-tripping back to DICOM).  The standard way to anonymize DICOM removes all private headers, but then you lose the mosaic info and b-values/b-vecs.   You can whitelist the CSA sub-header with this info, but at least some versions of the scanner software will use uninitialized memory for the CSA buffer and then leak PHI in the unused portions.  So by parsing it and rewriting it with this code (which does zero out the memory) you can avoid this issue. So... it is pretty niche but still useful.  

This also got me thinking, is dicom_parser going to make any attempt at supporting anonymization? Should we start a dicom_anon project?

ZviBaratz:
I haven't really given it much thought so far, but I definitely support the idea. Do you think a separate package will be better? I feel like it could probably be added to dicom_parser fairly easily, but it's possible I'm simply not aware of some anonymization edge-cases, such as the CSA thing.

effigies:
I'm going to go ahead and merge this. Coverage is reasonable, and can be built off of. Sorry for taking so long to get back to this.

PR #418, opened 2016-03-03
ENH: Add parser for Siemens "ASCCONV" text format

moloney:
This format is included in most Siemens DICOM files as well as
other MR related files. It contains many important bits of meta
data.


nibabel/nicom/tests/data/ascconv_sample.txt
```diff
@@ -0,0 +1,919 @@
...
+asCoilSelectMeas[0].aFFT_SCALE[8].lRxChannel = 9
+asCoilSelectMeas[0].aFFT_SCALE[9].flFactor = 4.57011
+asCoilSelectMeas[0].aFFT_SCALE[9].bValid = 1
+asCoilSelectMeas[0].aFFT_SCALE[9].lRxChannel = 10
+asCoilSelectMeas[0].aFFT_SCALE[10].flFactor = 4.6211
+asCoilSelectMeas[0].aFFT_SCALE[10].bValid = 1
+asCoilSelectMeas[0].aFFT_SCALE[10].lRxChannel = 11
+asCoilSelectMeas[0].aFFT_SCALE[11].flFactor = 4.69845
+asCoilSelectMeas[0].aFFT_SCALE[11].bValid = 1
+asCoilSelectMeas[0].aFFT_SCALE[11].lRxChannel = 12
```
matthew-brett: Would be nice to add a test for one of these long list of lists type values.

moloney: Currently we don't parse these as lists, instead the key ends up being 'asCoilSelectMeas[0].aFFT_SCALE[6].lRxChannel'. 

It would be nice to convert these to lists, but then we need to deal with some potential weirdness (like it seems some lists are using 0-based indexing and others are using 1-based indexing).

matthew-brett: Yes, I remember now.   I guess we could check whether the LHS list exists and fill with None up to the required index if not?  As in:

```
 [1] = 2
```
would create an object `sGroupArray` (if it doesn't exist), create a list `sGroupArray.anMember` with contents [None, None]...    Would be a little bit of work though.

moloney: Yeah that seems reasonable and worthwhile.

matthew-brett: I had some ideas - want me to have a go for a couple of hours, or do you already know how to solve it?

moloney: By all means, please take a crack at it!  Thanks!

nibabel/nicom/ascconv.py
```diff
@@ -0,0 +1,60 @@
...
+import ast, re
+from ..externals import OrderedDict
+
+
+ASCCONV_RE = re.compile(
+    r'### ASCCONV BEGIN((?:\s*[^=\s]+=[^=\s]+)*) ###\n(.*?)\n### ASCCONV END ###',
+    flags=re.M | re.S)
+
+
+def parse_ascconv(csa_key, ascconv_str):
```
matthew-brett: Maybe it would be more obvious what is going on if we had

```
parse_ascconv(ascconv_str, str_delim):
```

Then we put the logic for which string delimiter per protocol somewhere else?   The `csa_key` is just signaling the string delimiter at the moment.

nibabel/nicom/tests/data/ascconv_sample.txt
matthew-brett: As you might have predicted, it took more than a couple of hours, but:  https://github.com/moloney/nibabel/pull/3

matthew-brett:
Brendan - did you get a chance to look at my PR to your branch?

moloney:
I screwed up the merge here by doing the rebase before I merged Matthew's work, closing in favor of #896

PR #420, opened 2016-03-11
MRG: add Gregory Lee's new PARREC data

matthew-brett:
Test parrec data with multiple rotations.

Modify translation order to minimize error between our rotations and
those from the NIfTI images in the conversion.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "nibabel-data/nitest-freesurfer"]
 	path = nibabel-data/nitest-freesurfer
 	url = https://bitbucket.org/nipy/nitest-freesurfer.git
+[submodule "nibabel-data/parrec_oblique"]
+	path = nibabel-data/parrec_oblique
+	url = https://github.com/grlee77/parrec_oblique.git
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -126,6 +126,8 @@ ACQ_TO_PSL = dict(
                       [0, 0, 0, 1]])
 )
 
+DEG2RAD = np.pi / 180.
+
 # General information dict definitions
 # assign props to PAR header entries
 # values are: (shortname[, dtype[, shape]])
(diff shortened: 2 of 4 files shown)


matthew-brett:
The slight offset in the translations is still very puzzling to me.

I see that the amount that the afffine is off, depends on the rotations.  Here is a print of the filename, the rotations in FH, AP, RL order, and the error in the translations:

```
T1W_cor_0_0_0_SENSE_13_1.PAR 0.0 0.0 0.0 [-0.0943233  -0.00043228 -0.0945    ]
T1W_cor_5_15_30_SENSE_12_1.PAR 5.0 15.0 30.0 [-0.08251608 -0.07266869 -0.07619181]
T1W_sag_0_0_0_SENSE_8_1.PAR 0.0 0.0 0.0 [ 0.00018147 -0.09492847 -0.0945    ]
T1W_sag_5_15_30_SENSE_11_1.PAR 5.0 15.0 30.0 [ 0.03278593 -0.1254854  -0.03404345]
T1W_trans_0_0_0_SENSE_4_1.PAR 0.0 0.0 0.0 [-0.0943233  -0.09492847  0.        ]
T1W_trans_5_15_30_SENSE_10_1.PAR 5.0 15.0 30.0 [-0.06638231 -0.10357589  0.05282434]
```

Has anyone got any ideas how to track this down?

matthew-brett:
Here's a script to investigate - https://gist.github.com/d251bd4f6c5aab653161 - nothing much came to me looking at the output.

matthew-brett:
Any thoughts on the incorrect translations?  I can't think of anything to try.

grlee77:
It is good to see that the rotations are a very close match now.

I played with it for awhile, but I couldn't get a better result for the translations either.  The peak translation error across the test cases is about 0.3 voxels, which I could live with.  Without access to the vendor's conversion code, it probably won't be easy to track down the source of the discrepancy.

matthew-brett:
I'm surprised by the apparent order of the rotations - and it's very difficult to check visually with the one-slice images here.

@grlee77 - should you be bored one day in front of the scanner - would you consider taking some images with a fair few slices and three biggish rotations (maybe 15 20 25 degrees or something) so we can see visually if the rotations are right?

matthew-brett:
I will merge this one soon unless I hear otherwise.

PR #421, opened 2016-03-11
BF: fix dependency on numpy >= 1.8 in viewer

matthew-brett:
Viewer was using `np.take` with integer argument for `indices`, only
allowed in numpy 1.8 or greater.

See:
http://nipy.bic.berkeley.edu/builders/nibabel-py2.6-32/builds/352/steps/shell_5/logs/stdio

Instead, use rollaxis and slicing to get the slice that we need.  This
should also be faster, because we always get a view rather than a copy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,10 @@ matrix:
     - python: 2.7
       env:
         - DEPENDS=numpy==1.5.1 PYDICOM=0
+    # Absolute minimum dependencies plus oldest MPL
+    - python: 2.7
+      env:
+        - DEPENDS=numpy==1.5.1 matplotlib==1.3.1 PYDICOM=0
     # Minimum pydicom dependency
     - python: 2.7
       env:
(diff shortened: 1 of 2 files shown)


matthew-brett:
Fixes buildbot error `try_branch.py nibabel-py2.6-32`:

http://nipy.bic.berkeley.edu/builders/nibabel-py2.6-32/builds/354/steps/shell_5/logs/stdio

matthew-brett:
Would be grateful for quick review - this one is a very small change.

grlee77:
LGTM.  

as a double check, I ran the following empirical test:
```python
import numpy as np
from numpy.testing import assert_array_equal
a = np.random.randn(4, 8, 12)
order=[1, 2, 0]
idx=[5, 9, 1]
for ii in range(3):
    data = np.rollaxis(cdata, order[ii])[idx[ii]]
    data2 = np.take(cdata, idx[ii], axis=order[ii])
    assert_array_equal(data, data2)
```

matthew-brett:
Great - thanks a lot for checking.

PR #423, opened 2016-03-12
MRG: testing on numpy 1.12 plus fixes

matthew-brett:
Numpy 1.12 (not yet released) breaks some stuff; have travis test against daily
numpy builds and fix problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,12 @@ sudo: false
 cache:
   directories:
     - $HOME/.cache/pip
-
 addons:
   apt:
     packages:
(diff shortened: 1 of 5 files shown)


matthew-brett:
This one also pretty straightforward - 5 minute review?

grlee77:
LGTM.  

I only recently learned about the --pre flag to pip when we had an issue with an pip being outdated on our Appveyor builds in PyWavelets, but it looks like you already install a current version of pip in the scripts here and the tests are passing.

matthew-brett:
Thanks for the review.

PR #426, opened 2016-03-14
FIX: properly handle V4 .PAR files with diffusion info

grlee77:
This PR fixes a bug in `get_bvals_bvecs` for V4 .PAR files (the bug is only present for V4, not the more recent V4.1 or V4.2).  In V4, the `diffusion` field does not exist and the `bvecs` cannot be determined.


grlee77:
apologies... should have tested properly before uploading.  I forgot that `image_defs` is a recarray and not a dictionary.  will upload a fix later this evening

matthew-brett:
Thanks - looks good to me.

grlee77:
I can rebase to a single commit if you prefer.

matthew-brett:
It's fine as is - was waiting to see if anyone else wanted to say something, but looks like no - merging.

PR #427, opened 2016-03-15
ENH: add a function to retrieve .PAR data labels for the 4th dimension

grlee77:
This was pulled out of the previously submitted #278 to make reviewing easier.  This complements the functionality in the recently merged #409.

The primary enhancement is a new method in the PARREC header called `get_dimension_labels` that lists which .PAR fields the data stored along the 4th dimension corresponds to.  (i.e. all data >4D is still collapsed into 4 dimensions exactly as before, but labels can be retrieved to make the data ordering unambiguous)

As a concrete example, consider the file `tests/data/T1_3echo_mag_real_imag_phase.PAR` which corresponds to a 3D scan acquired with 3 echos, but where 4 data types were also saved (real, imaginary, magnitude, phase).  In this case:
```python
t1_par = pjoin(DATA_PATH, 'T1_3echo_mag_real_imag_phase.PAR')
with open(t1_par, 'rt') as fobj:
    t1_hdr = PARRECHeader.from_fileobj(fobj, strict_sort=True)

# check volume labels
vol_labels = t1_hdr.get_dimension_labels()
print(vol_labels)
```
gives
```
 {'echo number': array([1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3]),
 'image_type_mr': array([0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3])}
```
which lets us know that the 4th dimension is a combination of echos and image types and that the echo number varies faster than image type along the 4th dimension.

The secondary enhancement is a new option to `parrec2nii` that prints out this arrays of labels as a JSON file (with suffix `.ordering.json`).  I would appreciate feedback on whether this is a reasonable choice of output format or if there is a better alternative.  Once that is decided, I can add a test case for it to `test_scripts.py`.

### DIFF
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -1170,9 +1170,8 @@ class PARRECHeader(SpatialHeader):
             dynamic_keys.remove('slice number')
 
         # remove dynamic keys that may not be present in older .PAR versions
-        for key in dynamic_keys:
-            if key not in image_defs.dtype.fields:
-                dynamic_keys.remove(key)
+        dynamic_keys = [d for d in dynamic_keys if d in
+                        image_defs.dtype.fields]
 
         non_unique_keys = []
         for key in dynamic_keys:
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -569,6 +569,21 @@ class PARRECArrayProxy(object):
         self._slice_scaling = header.get_data_scaling(scaling)
         self._rec_shape = header.get_rec_shape()
 
+        try:
+            # store additional slice & volume label information
+            self._dim_4_labels = header.sorted_labels(
+                    sorted_indices=self._slice_indices,
+                    collapse_slices=True)
+            self._dim_3_4_labels = None
+        except:
+            # if non-unique slice orientations or image angulations, need to
+            # also keep labels for the 3rd (slice) dimension
+            self._dim_4_labels = None
+            self._dim_3_4_labels = header.sorted_labels(
+                sorted_indices=self._slice_indices,
+                collapse_slices=False)
+
+
     @property
     def shape(self):
         return self._shape
@@ -1113,6 +1128,75 @@ class PARRECHeader(SpatialHeader):
         # Based on our sorting, they should always be last.
         n_used = np.prod(self.get_data_shape()[2:])
         return sort_order[:n_used]
+
+    def sorted_labels(self, sorted_indices=None, collapse_slices=True):
+        """ return a dictionary of lists of the varying values in
+        ``self.image_defs``.  This is useful for custom data sorting.  only
+        keys that have more than 1 unique value across the dataset will exist
+        in the returned `sort_info` dictionary.
+
+        Parameters
+        ----------
+        sorted_indices : array, optional
+            sorted slice indices as returned by
+            ``self.get_sorted_slice_indices``.
+        collapse_slices : bool, optional
+            if True, only return indices corresponding to the first slice
+
+        Returns
+        -------
+        sort_info : dict
+            Each key corresponds to a dynamically varying sequence dimension.
+            The ordering of values corresponds to that in sorted_indices.
+        """
+
+        # define which keys to store sorting info for
+        dynamic_keys = ['slice number',
+                        'cardiac phase number',
+                        'echo number',
+                        'label type',
+                        'image_type_mr',
+                        'dynamic scan number',
+                        'slice orientation',
+                        'image_display_orientation',  # ????
+                        'image angulation',
+                        'scanning sequence',
+                        'gradient orientation number',
+                        'diffusion b value number']
+
+        if sorted_indices is None:
+            sorted_indices = self.get_sorted_slice_indices()
+
+        if collapse_slices:
+            dynamic_keys.remove('slice number')
+
+        non_unique_keys = []
+        for key in dynamic_keys:
+            if len(np.unique(self.image_defs[key])) > 1:
+                non_unique_keys.append(key)
+
+        if collapse_slices:
+            if 'slice orientation' in non_unique_keys:
+                raise ValueError("for non-unique slice orientation, need "
+                                 "collapse_slice=False")
+            if 'image angulation' in non_unique_keys:
+                raise ValueError("for non-unique image angulation, need "
+                                 "collapse_slice=False")
+            if 'image image_display_orientation' in non_unique_keys:  # ???
+                raise ValueError("for non-unique display orientation, need "
+                                 "collapse_slice=False")
+            sl1_indices = self.image_defs['slice number'][sorted_indices] == 1
+
+        sort_info = {}
+        for key in non_unique_keys:
+            if collapse_slices:
+                sort_info[key] = list(
+                    self.image_defs[key][sorted_indices][sl1_indices])
+            else:
+                sort_info[key] = list(
+                    self.image_defs[key][sorted_indices])
+
+        return sort_info
 
 
 class PARRECImage(SpatialImage):
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -1161,7 +1161,7 @@ class PARRECHeader(SpatialHeader):
         # the value at slice 1.
         sl1_indices = image_defs['slice number'][sorted_indices] == 1
 
-        sort_info = OrderedDict(non_unique_keys)
+        sort_info = OrderedDict()
         for key in non_unique_keys:
             sort_info[key] = image_defs[key][sorted_indices][sl1_indices]
         return sort_info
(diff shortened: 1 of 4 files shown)


nibabel/parrec.py
```diff
@@ -1114,6 +1126,88 @@ def get_sorted_slice_indices(self):
         n_used = np.prod(self.get_data_shape()[2:])
         return sort_order[:n_used]
 
+    def get_dimension_labels(self, collapse_slices=True):
```
matthew-brett: Isn't this better called 'get_volume_labels`?

nibabel/parrec.py
```diff
@@ -1114,6 +1126,88 @@ def get_sorted_slice_indices(self):
...
+
+        # define which keys to store sorting info for
+        dynamic_keys = ['slice number',
+                        'cardiac phase number',
+                        'echo number',
+                        'label type',
+                        'image_type_mr',
+                        'dynamic scan number',
+                        'slice orientation',
+                        'image_display_orientation',  # ????
```
matthew-brett: More detail in the comment here?

nibabel/parrec.py
```diff
@@ -1114,6 +1126,88 @@ def get_sorted_slice_indices(self):
...
+                                 "collapse_slice=False")
+            if 'image angulation' in non_unique_keys:
+                raise ValueError("for non-unique image angulation, need "
+                                 "collapse_slice=False")
+            if 'image_display_orientation' in non_unique_keys:  # ???
+                raise ValueError("for non-unique display orientation, need "
+                                 "collapse_slice=False")
+            sl1_indices = image_defs['slice number'][sorted_indices] == 1
+
+        sort_info = {}
```
matthew-brett: Maybe consider OrderedDict here?   I guess you don't want the order of keys to vary randomly?

nibabel/parrec.py
```diff
@@ -249,6 +249,18 @@ class PARRECError(Exception):
 GEN_RE = re.compile(r".\s+(.*?)\s*:\s*(.*)")
 
 
+def _unique_rows(a):
```
matthew-brett: Would be reassuring to have some tests of this function (I know it's exercised in the tests for the dimension labels, but still).

nibabel/parrec.py
```diff
@@ -1114,6 +1126,88 @@ def get_sorted_slice_indices(self):
...
+                        'image_type_mr',
+                        'dynamic scan number',
+                        'slice orientation',
+                        'image_display_orientation',  # ????
+                        'image angulation',
+                        'scanning sequence',
+                        'gradient orientation number',
+                        'diffusion b value number']
+
+        sorted_indices = self.get_sorted_slice_indices()
```
matthew-brett: Put this in `if collapse_slices:` branch further down, to make clear where it's used?

matthew-brett: Ah no - sorry - scratch that - I see you do use it in the non collapse slices branch.   But still - maybe put this closer to the point that it is used.

nibabel/parrec.py
```diff
@@ -1114,6 +1126,88 @@ def get_sorted_slice_indices(self):
...
+            elif ndim == 2:
+                # for 2D cases, e.g. 'image angulation'
+                num_unique = len(_unique_rows(image_defs[key]))
+            else:
+                raise ValueError("unexpected image_defs shape > 2D")
+            if num_unique > 1:
+                non_unique_keys.append(key)
+
+        if collapse_slices:
+            if 'slice orientation' in non_unique_keys:
```
matthew-brett: Or:

``` 
for key in ('slice_orientation', 'image_angulation', 'image_display_orientation'):
    if key in non_unique_keys:
        raise ValueError('for values of {0} that differ over slices, use '
             '"collapse_slice=False"'.format(key))
```

nibabel/parrec.py
```diff
@@ -249,6 +249,18 @@ class PARRECError(Exception):
 GEN_RE = re.compile(r".\s+(.*?)\s*:\s*(.*)")
 
 
+def _unique_rows(a):
+    """
+    find unique rows of a 2D array.  based on discussion at:
+    http://stackoverflow.com/questions/16970982/find-unique-rows-in-numpy-array
+    """
+    if a.ndim != 2:
```
matthew-brett: Why not cover 1D case here too?

nibabel/parrec.py
matthew-brett: Ah - now I read more, I understand better.

I'm not getting why you are checking these values particularly.   I was thinking you might want to check any set of `image_defs` values to make sure that they were the same across a volume, in the `collapse_slices` case, but here you are checking whether some values vary at all (including across volumes).  Can you explain more?

matthew-brett:
How about a CSV file for the output of the volume labels, with first line being the label strings, and then one row per volume?

grlee77:
The most recent commit does not have the CSV-related changes yet, but demonstrates the simplification from removing the per-slice attributes.  The `_unique_rows` utility was no longer needed after the simplification and was removed as well.

matthew-brett:
What happens for labels of incomplete volumes?

nibabel/parrec.py
```diff
@@ -1114,6 +1114,57 @@ def get_sorted_slice_indices(self):
...
+                        'scanning sequence',
+                        'gradient orientation number',
+                        'diffusion b value number']
+
+        # remove dynamic keys that may not be present in older .PAR versions
+        dynamic_keys = [d for d in dynamic_keys if d in
+                        image_defs.dtype.fields]
+
+        non_unique_keys = []
+        for key in dynamic_keys:
```
matthew-brett: What happens for attributes that vary across slices but not volumes?

Does it matter that the attributes vary across slices?

grlee77: I removed any attributes that would vary within slices of a volume.

Do you want an explicit check that for each slice in the volume all of the attributes in `dynamic_keys` have the same value and raise an error if not?

Mainly what I want to achieve in this PR is a reasonable labelling for the 4th dimension of the data in cases where multiple of these fields may vary.  Practical examples of this are the test cases added in `test_parrec.py` by this PR.

grlee77:
No labels will be generated for incomplete volumes.  The incomplete valumes are removed by `get_sorted_slice_indices` which is called at the start of `get_volume_labels`.  The length of the label lists generated here will aways exactly match the size of the data in the final NIFTI output of `parrec2nii`.

nibabel/parrec.py
matthew-brett: Up to you about checking whether the attributes vary across slice or not.  If you think that could not happen, maybe there no need to check.

matthew-brett:
I guess then that there will be volume labels if you pass `permit_truncated=True`, but that's probably what the user would expect.

nibabel/parrec.py
```diff
@@ -1114,6 +1114,57 @@ def get_sorted_slice_indices(self):
...
+            else:
+                raise ValueError("unexpected image_defs shape > 1D")
+            if num_unique > 1:
+                non_unique_keys.append(key)
+
+        # each key in dynamic keys will be identical across slices, so use
+        # the value at slice 1.
+        sl1_indices = image_defs['slice number'][sorted_indices] == 1
+
+        sort_info = {}
```
matthew-brett: I still think an OrderedDict would be better here - otherwise you'll have random ordering of your csv file columns across different runs on the same file, or across different files.

grlee77: I agree, just forgot to address that one!

grlee77:
If you pass `permit_truncated=True`, the truncated volumes still always get dropped by `get_sorted_slice_indices`.  I believe the only difference is if `permit_truncated=False` a PARRECError would be raised if any truncation is present in the file.
An alternative implementation would be to zero-fill the missing slices in the truncated case, but that is not a question for this PR.

nibabel/tests/test_parrec.py
```diff
@@ -325,6 +340,14 @@ def test_sorting_multiple_echos_and_contrasts():
     assert_equal(np.all(sorted_types[ntotal//2:3*ntotal//4] == 2), True)
     assert_equal(np.all(sorted_types[3*ntotal//4:ntotal] == 3), True)
 
+    # check volume labels
+    vol_labels = t1_hdr.get_volume_labels()
+    assert_equal(sorted(list(vol_labels.keys())),
```
matthew-brett: With the ordered dict you can predict (and test for) the key order.  I think that means you can just drop the `sorted` here.

nibabel/tests/test_parrec.py
```diff
@@ -366,6 +389,15 @@ def test_sorting_multiecho_ASL():
     # check that slices vary fastest
     assert_array_equal(sorted_slices[:nslices], np.arange(1, nslices+1))
 
+    # check volume labels
+    vol_labels = asl_hdr.get_volume_labels()
+    assert_equal(sorted(list(vol_labels.keys())),
```
matthew-brett: Can drop `sorted` and test for order `['echo number', 'label type', 'dynamic scan number']`

matthew-brett:
Made a couple of tiny testing suggestions, but otherwise, I think this is good to go.

matthew-brett:
Thanks again - in it goes.

Issue #399, opened 2016-01-28
PARREC; B1 map double TR error


ilogue:
```
In[6]:  prov.add(parfile)
[provenance] Error inspecting file: /redshirt_array/data/phantom_qT1_slu/phantom_qT1_ ~SNIP~ _WIP_B1MAP_SENSE_24_1.PAR
Traceback (most recent call last):
Out[6]:   File "/home/mrjeffs/Software/niprov/niprov/adding.py", line 69, in add
    img.inspect()
  File "/home/mrjeffs/Software/niprov/niprov/parrec.py", line 14, in inspect
    img = self.libs.nibabel.load(self.path)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/loadsave.py", line 44, in load
    return guessed_image_type(filename).from_filename(filename, **kwargs)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/keywordonly.py", line 16, in wrapper
    return func(*args, **kwargs)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/parrec.py", line 1048, in from_filename
    scaling=scaling)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/keywordonly.py", line 16, in wrapper
    return func(*args, **kwargs)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/parrec.py", line 1013, in from_file_map
    permit_truncated=permit_truncated)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/parrec.py", line 632, in from_fileobj
    info, image_defs = parse_PAR_header(fileobj)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/parrec.py", line 461, in parse_PAR_header
    general_info = _process_gen_dict(gen_dict)
  File "/home/mrjeffs/Software/anaconda/lib/python2.7/site-packages/nibabel/parrec.py", line 295, in _process_gen_dict
    value = props[1](value)
ValueError: invalid literal for float(): 70.892  190.892
(<niprov.parrec.ParrecFile at 0x7f8d0a3b1650>, 'failed')
```

@Eric89GXL @mrjeffs
We have a B1 map in a PARREC file that lists two TR values in the header, as expected for this protocol, but are running into an error for it.

The field format for TR is defined here:
[...]

PR #429, opened 2016-03-16
ENH: support multiple TR values in PARREC headers

grlee77:
This is a tiny modification to allow reading .PAR files with more than one repetition time listed in the header as requested in #399.

I am not entirely certain if this is a standard feature of .PAR files, or due to a custom software patch.  If it is the former, I think we should support it in nibabel.

if multiple TR values are found, a warning is issued and zooms[3] is set to the first TR listed

### DIFF
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -129,6 +129,7 @@ ACQ_TO_PSL = dict(
 # General information dict definitions
 # assign props to PAR header entries
 # values are: (shortname[, dtype[, shape]])
+# if shape is None, the number of elements is to be determined on read
 _hdr_key_dict = {
     'Patient name': ('patient_name',),
     'Examination name': ('exam_name',),
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -148,7 +148,7 @@ _hdr_key_dict = {
     'Technique': ('tech',),
     'Scan resolution  (x, y)': ('scan_resolution', int, (2,)),
     'Scan mode': ('scan_mode',),
-    'Repetition time [ms]': ('repetition_time', float),
+    'Repetition time [ms]': ('repetition_time', float, None),
     'FOV (ap,fh,rl) [mm]': ('fov', float, (3,)),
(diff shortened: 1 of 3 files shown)


matthew-brett:
I think we need a test - maybe an edited version of another .PAR file - if we can't get the file that caused the problem?

grlee77:
In practice there can be more than one TR value in a pulse sequence (e.g. alternating repetition time SSFP,  MR Fingerprinting, etc)

Whether or not this is officially supported by Philips as part of the .PAR format, I don't know.
It doesn't break anything to allow multiple TR values for this field, so I am in favor of adding it.

nibabel/parrec.py
```diff
@@ -148,7 +148,7 @@
     'Technique': ('tech',),
     'Scan resolution  (x, y)': ('scan_resolution', int, (2,)),
     'Scan mode': ('scan_mode',),
-    'Repetition time [ms]': ('repetition_time', float),
+    'Repetition time [ms]': ('repetition_time', float, None),
```
matthew-brett: Would you mind adding a comment at the top of parrec above `_hdr_key_dict` noting that None means "length of vector to be determined on read"?

nibabel/tests/test_parrec.py
```diff
@@ -777,3 +779,11 @@ def test_exts2par():
                    list(nii_img.header.extensions)):
         hdrs = exts2pars(source)
         assert_equal(len(hdrs), 2)
+
+
+def test_dualTR():
+    with open(DUAL_TR_PAR, 'rt') as fobj:
+        with suppress_warnings():
```
matthew-brett: Sorry to ask, but would you mind also checking the warning is raised?   Something like:

```
    with clear_and_catch_warnings(modules=[parrec], record=True) as wlist:
        warnings.simplefilter('always')
        dualTR_hdr = PARRECHeader.from_fileobj(fobj)
        assert_equal(len(wlist), 1)
```

matthew-brett:
Great - thanks as ever for your patience dealing with these small things.

PR #430, opened 2016-03-18
BF: a couple of tiny fixes

matthew-brett:
Use externals version of OrderedDict for parrec.

Use ``next(something)`` instead of ``something.__next__``.


matthew-brett:
I think these qualify as trivial fixes - merging.

PR #431, opened 2016-03-19
MAINT: switch to codecov for coverage report

matthew-brett:
Scipy, Pandas, scikit-image use codecov.

Example output (via travis-ci):
https://codecov.io/github/matthew-brett/nibabel/nibabel?ref=c576b5d459c12ae8e052bd21fd27f5faff336e46

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,6 +91,7 @@ before_install:
     - if [ "${COVERAGE}" == "1" ]; then
       pip install coverage;
       pip install coveralls;
+      pip install codecov;
       fi
 # command to install dependencies
 install:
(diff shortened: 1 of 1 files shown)


yarikoptic:
No need to switch, you can have both afaik... At least I just did that for datalad and fail2ban.
Also it presumably (didn't with for me yet) has extensions for browsers to highlight coverage straight while browsing github PR etc... Should be cool... I like it more overall but still keep coveralls enabled since seems to not hurt

matthew-brett:
There's a browser extension for looking at coverage on github : https://codecov.io/#features

matthew-brett:
Do you then have both coveralls and codecov enabled in the github status reports?

yarikoptic:
in the status report only coveralls appears but both insert status boxes above, see e.g. https://github.com/datalad/datalad/pull/377

matthew-brett:
Any benefit to keeping both?  Has there been a case where the coveralls result was more useful?

yarikoptic:
Fwiw not yet. Redundant free service though is good for resilience against failures of one of the two

matthew-brett:
Fair enough - I left coveralls in.

matthew-brett:
Will merge tomorrow unless I hear otherwise.

PR #434, opened 2016-03-23
BF: fix shebang lines

matthew-brett:
Fix shebang lines of scripts so setuptools installs them correctly.

### DIFF
--- a/bin/nib-dicomfs
+++ b/bin/nib-dicomfs
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!python
 # emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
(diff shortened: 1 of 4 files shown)


matthew-brett:
2 minute review anyone?  With this one in we should be back to all tests passing on the buildbots.

grlee77:
This worked okay when I tried it, but what was the problem with using `#!/usr/bin/env python`?

matthew-brett:
There's some discussion here : https://github.com/nipy/nipy/pull/393  - setuptools will rewrite the shebang lines correctly for the python for which it is installing - e.g. a virtualenv - but - for some versions of setuptoools - only if the shebang line in the original script is exactly `#!python`.

matthew-brett:
Will merge tomorrow unless I hear otherwise.

PR #437, opened 2016-03-27
ENH+RF: nib-ls -c/--counts to report counts for each value (useful for ROI maps)

yarikoptic:
Also adjusted slightly output of --stats to include range in [from, to]


nibabel/tests/test_scripts.py
```diff
@@ -68,7 +68,15 @@ def check_nib_ls_example4d(opts=[], hdrs_str=""):
 def test_nib_ls():
     yield check_nib_ls_example4d
     yield check_nib_ls_example4d, \
-          ['-H', 'dim,bitpix'], " \[  4 128  96  24   2   1   1   1\] 16"
+        ['-H', 'dim,bitpix'], " \[  4 128  96  24   2   1   1   1\] 16"
+    yield check_nib_ls_example4d, ['-c'], "", " 2:3 3:2 4:1 5:1.*"
```
matthew-brett: Consider removing 'yield's here?  They make test failures much harder to debug

bin/nib-ls
```diff
@@ -22,6 +22,7 @@ import sys
 from math import ceil
 from optparse import OptionParser, Option
 from io import StringIO
+from collections import Counter
```
matthew-brett: Not compatible with Python 2.6.  We can drop Python 2.6 unless there's an obvious workaround.

effigies: Below you have:

```Python
counter = Counter()
counter.update(d.flatten())
# go through each entry and report only non-0 ones
row += ["@l" + " ".join("%g:%d" % (i, counter[i]) for i in sorted(counter))]
```

You could do something like:

```Python
row += ["@l" + " ".join("%g:%d" % (i, np.sum(d == i)) for i in np.unique(d))]
```

I suspect this would be slower than Counter, so take it for whatever it's worth.

yarikoptic: I don't mind reimplementing without Counter if necessary. Myself stopped carrying about 2.6 about a year ago

nibabel/tests/test_scripts.py
yarikoptic: What is recommended way to run parametric tests nowadays in nibabel? Just call them?

bin/nib-ls
yarikoptic: Yeap... Would be much slower happen I feed some float volume in by mistake ;-)

matthew-brett: Ok - then let's drop Python 2.6.  I'll email the list.

nibabel/tests/test_scripts.py
matthew-brett: Yes, the yields mean that you still run further tests if one fails, but given we clean up our test errors quickly, that isn't worth that much.  So, I just do multiple `assert_true(something)` etc in the same test function.

yarikoptic: On Wed, 06 Apr 2016, Matthew Brett wrote:

>    In nibabel/tests/test_scripts.py:

>  > @@ -68,7 +68,15 @@ def check_nib_ls_example4d(opts=[], hdrs_str=""):
>  >  def test_nib_ls():
>  >      yield check_nib_ls_example4d
>  >      yield check_nib_ls_example4d, \
>  > -          ['-H', 'dim,bitpix'], " \[  4 128  96  24   2   1   1   1\] 16"
>  > +        ['-H', 'dim,bitpix'], " \[  4 128  96  24   2   1   1   1\] 16"
>  > +    yield check_nib_ls_example4d, ['-c'], "", " 2:3 3:2 4:1 5:1.*"

>    Yes, the yields mean that you still run further tests if one fails, but
>    given we clean up our test errors quickly, that isn't worth that much. So,
>    I just do multiple assert_true(something) etc in the same test function.

I do regular asserts within test bodies as well not allowing to fall
through a failing one (so failures could cascade etc).

But in this case it is just a parametric test which invokes the same
body of the larger testing logic under various conditions.  A bit
different (although could be argued to be the same of cause just at a
different level).  I also like such tests to see the progress
which shows that a given test is validated against different
options/conditions.

Just my .1 cent, will later change to regular calls to the check_... but
in my experience such level of yields is imho only beneficial.  It was
always as easy to isolate or troubleshoot as if I would just be
calling those large test helpers.

bin/nib-ls
```diff
@@ -214,18 +219,24 @@ def proc_file(f, opts):
...
-            row += [len(d) and '%.2g:%.2g' % (np.min(d), np.max(d)) or '-']
+            if opts.stats:
+                # just # of elements
+                row += ["@l[%d]" % np.prod(d.shape)]
+                # stats
+                row += [len(d) and '@l[%.2g, %.2g]' % (np.min(d), np.max(d)) or '-']
+            if opts.counts:
+                counter = Counter()
+                counter.update(d.flatten())
+                # go through each entry and report only non-0 ones
```
matthew-brett: Won't this report zeros if you have `ops.stats_zeros` True?

bin/nib-ls
```diff
@@ -214,18 +219,24 @@ def proc_file(f, opts):
...
-            row += ["[%d] " % np.prod(d.shape)]
-            # stats
-            row += [len(d) and '%.2g:%.2g' % (np.min(d), np.max(d)) or '-']
+            if opts.stats:
+                # just # of elements
+                row += ["@l[%d]" % np.prod(d.shape)]
+                # stats
+                row += [len(d) and '@l[%.2g, %.2g]' % (np.min(d), np.max(d)) or '-']
+            if opts.counts:
+                counter = Counter()
```
matthew-brett: How about using the code from `scipy.stats.itemfreq`?

```
items, inv = np.unique(d, return_inverse=True)
freq = np.bincount(inv)
row += ["@l" + " ".join("%g:%d" % zip(items, freq))
```

nibabel/tests/test_scripts.py
matthew-brett: I don't feel super strongly - but I have often run into problem debugging these guys.

bin/nib-ls
yarikoptic: how many your terminal can tollerate? ;)  let's make it a 1000?

bin/nib-ls
yarikoptic: yeap -- adjusted the doc for the cmdline arg

nibabel/tests/test_scripts.py
yarikoptic: coolio, then I will keep it as is for now ;)

bin/nib-ls
yarikoptic: scipy seems to be an optional dependency.... if decision is made not to drop python 2.6 then I will look into it, but otherwise better keep it leaner ... I could look into redoing it with pure numpy or dict I guess if needed.

matthew-brett: Sounds good

matthew-brett: The code I posted is copy-pasted from scipy.stats.itemfreq - it doesn't use scipy.

For me that code is leaner than the counter algorithm, but tastes differ.

matthew-brett:
Yarik - where are we on this one?

yarikoptic:
On Wed, 20 Apr 2016, Matthew Brett wrote:

>    Yarik - where are we on this one?

just got a bit overwhelmed after coming back from the trip... either I
will get to it today/tomorrow or feel very welcome to tune it to your
liking... wanted to check if your scipy way would work fine with floats
;)
-- 
Yaroslav O. Halchenko
Center for Open Neuroscience     http://centerforopenneuroscience.org
Dartmouth College, 419 Moore Hall, Hinman Box 6207, Hanover, NH 03755
Phone: +1 (603) 646-9834                       Fax: +1 (603) 646-1419
WWW:   http://www.linkedin.com/in/yarik

matthew-brett:
It does work for floats, yes - the np.unique step does that, and the inv return values gives integer values with each one corresponding to one unique float value.

yarikoptic:
On Wed, 20 Apr 2016, Matthew Brett wrote:

>    It does work for floats, yes - the np.unique step does that, and the inv
>    return values gives integer values with each one corresponding to one
>    unique float value.

then I guess it is perfect solution ;)
-- 
Yaroslav O. Halchenko
Center for Open Neuroscience     http://centerforopenneuroscience.org
Dartmouth College, 419 Moore Hall, Hinman Box 6207, Hanover, NH 03755
Phone: +1 (603) 646-9834                       Fax: +1 (603) 646-1419
WWW:   http://www.linkedin.com/in/yarik

bin/nib-ls
```diff
@@ -214,18 +218,23 @@ def proc_file(f, opts):
...
-            row += ["[%d] " % np.prod(d.shape)]
-            # stats
-            row += [len(d) and '%.2g:%.2g' % (np.min(d), np.max(d)) or '-']
+            if opts.stats:
+                # just # of elements
+                row += ["@l[%d]" % np.prod(d.shape)]
+                # stats
+                row += [len(d) and '@l[%.2g, %.2g]' % (np.min(d), np.max(d)) or '-']
+            if opts.counts:
+                items, inv = np.unique(d, return_inverse=True)
```
matthew-brett: Check for excessive `len(items)` ?

yarikoptic: you really don't want me off the hook, don't you? ;)  Let's may be chat about something more interesting at some point ;)

ok -- will do now for len "properly" ;)

yarikoptic:
@matthew-brett now I wonder if I haven't overdone it in 480de5b ... I would agree to revert to simpler implementation truncating solely the output

bin/nib-ls
```diff
@@ -214,18 +227,43 @@ def proc_file(f, opts):
...
+                d = d.reshape(-1)
+            if opts.stats:
+                # just # of elements
+                row += ["@l[%d]" % np.prod(d.shape)]
+                # stats
+                row += [len(d) and '@l[%.2g, %.2g]' % (np.min(d), np.max(d)) or '-']
+            if opts.counts:
+                maxunique = 1000  # constant
+                # Since d might be quite large and we don't want to exhaust memory and
+                # thrush, let's do in manageable batches
```
matthew-brett: 'thrash' for 'thrush'

I guess the risk is for very large floating point images, where the number of unique values will be close to the number of values.  Maybe refuse to do counts on very large images with (on-disk) datatypes > 16 bits, with a warning?

matthew-brett: Or, I guess - go back to the counter implementation and barf out when you get to too many unique values?

yarikoptic:
ok -- redid now... satisfied now? ;)

matthew-brett:
Yup - thanks - in it goes...

PR #438, opened 2016-03-30
BF: Remove assumption about StackID index.

embaker:
In determining the shape of a multi-frame DICOM, the first value in the DimensionIndexValues was taken to be the StackID. The DICOM standard does not appear to guarantee this to always be true. Replaced the assumption by using the DimensionIndexSequence to determine the position of the StackID in the DimensionIndexValues.


matthew-brett:
Did you close this on purpose?

embaker:
Yes, I did close on purpose. I realized that I had not updated all the tests that were dependent on the changes I made. I just opened up a new pull request that should be complete.

Thanks,

Eric M. Baker
Senior Research Assistant
Advanced Imaging Research Center
Oregon Health & Science University
________________________________

PR #439, opened 2016-03-30
BF: Remove assumption about StackID index

embaker:
In determining the shape of a multi-frame DICOM, the first value in the _DimensionIndexValues_ was taken to be the _StackID_. The DICOM standard does not appear to guarantee this to always be true. Replaced the assumption by using the _DimensionIndexSequence_ to determine the position of the _StackID_ in the _DimensionIndexValues_.


matthew-brett:
Thanks a lot for this.  Is there anything in the DICOM standard docs that you could copy / paste into the docstring here to explain the algorithm?

matthew-brett:
Generally this looks good and well-written, but I don't know the part of the standard that this is implementing - can you give me some pointers to help review?

Brendan - do you have time to take a look?

embaker:
I am currently inserting into the docstring a description of the algorithm and references to the DICOM standard.  The relevant parts of [Supplement 49 of the DICOM standard](ftp://medical.nema.org/medical/dicom/final/sup49_ft.pdf) are:
- C.7.6.17 Multi-frame Dimension Module
- C.7.6.16.2.2  Frame Content Macro

matthew-brett:
Thanks - that will be very helpful.

matthew-brett:
I think this is the definitive source for the supplement : ftp://medical.nema.org/medical/dicom/final/sup49_ft.pdf - is that right?

But - looking at the standard, it seems the supplement has already been included, e.g. `C.7.6.16 Multi-frame Functional Groups Module` in http://dicom.nema.org/medical/dicom/current/output/pdf/part03.pdf . So I guess the standard is the correct reference?

nibabel/nicom/dicomwrappers.py
```diff
@@ -471,13 +494,28 @@ def image_shape(self):
...
+        if len(stack_ids) > 1:
+            raise WrapperError("File contains more than one StackID. "
+                               "Cannot handle multi-stack files")
+        # Determine if one of the Dimension indices refers to the stack id
+        dim_seq = [(dim.DimensionIndexPointer, dim.FunctionalGroupPointer)
+                   for dim in self.get('DimensionIndexSequence')]
+        # the pointer pair (StackID Dicom-tag, FrameContentSequence Dicom-tag)
+        # that indicates that the corresponding axis in the DimensionIndexValues
+        # attribute refers to the StackID
+        stack_id_dim_pointer = ((0x20, 0x9056), (0x20, 0x9111))
```
matthew-brett: Maybe get these programmatically to make it more readable?  As in something like:

```
from nibabel.pydicom_compat import pydicom as pd
stack_id_dim_pointer = pd.datadict.tag_for_name('StackID'), pd.datatict.tag_for_name('FrameContentSequence')
```

matthew-brett: Do we need to check the second "FrameContentSequence" tag?   What if the first Tag is StackID and the second is something else?  Is this an error in the DICOM?

embaker: I will update the references to point to the current standard.  I have been working off of the 2015c version of the standard that did not include the supplement. 

I agree that using _tag_for_name_ is preferable. I will rewrite like how you suggested.  This brings up another thought that occurred to me: do you think that it would be beneficial to have a method/attribute that returned/stored the DICOM attribute name that each axis of the data corresponds to?  A user could use the "DimensionIndexSequence" to determine these on their own, but doing it in such a manner could be error prone for them since we are removing the "StackID" axis.

I believe you are correct that the "FrameContentSequence" tag is not needed.  It does appear that it would be an error in the DICOM (C.7.6.17.1 Dimension Indices) if the "FunctionalGroupPointer" was something other than the "FrameContentSequence" tag of "StackID".  However, it does not appear that the DICOM standard guarantees that each axis will have a "FunctionalGroupPointer" attribute when one is not needed for that axis (i.e. the dimension attribute is not nested).  If this was the case for an axis, then how I have it coded up now will not work.  So i am going rewrite the code to just rely on the "DimensionIndexPointer".

matthew-brett: Thanks for the feedback.    It would be excellent to have a way of returning an attribute name per axis.   I'm hoping to get some time in the next few months to add named axes to nibabel image objects, and it would be very helpful to be able to get names from dicom image axes.

matthew-brett: For the reference, given that the supplement has just been included, maybe add a reference to the supplement and the most recent version of the standard.

matthew-brett:
Thanks for this - the code is well written and tested, good to read.

PR #440, opened 2016-04-06
BF: old Pydicom uses Decimal strings

matthew-brett:
Cast values to float before setting into nifti header.

Fixes error seen at
http://nipy.bic.berkeley.edu/builders/nibabel-py2.7-pyd0.9.7/builds/212


effigies:
Seems reasonable.

matthew-brett:
Thank - good enough for me !

Issue #443, opened 2016-04-10
`test_dft` errors with `Exception: fromstring() has been removed. Please call frombytes() instead.`


arokem:
With PIL 1.1.7 on Python 3.5:

```
chlosyne:nibabel (dft-fromstring) $nosetests nibabel/tests/test_dft.py 
.....E.
======================================================================
ERROR: nibabel.tests.test_dft.test_png
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/numpy-1.11.0b2-py3.5-macosx-10.5-x86_64.egg/numpy/testing/decorators.py", line 147, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/source/nibabel/nibabel/tests/test_dft.py", line 89, in test_png
    data = studies[0].series[0].as_png()
  File "/Users/arokem/source/nibabel/nibabel/dft.py", line 142, in as_png
    data.tobytes())
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/PIL/Image.py", line 2063, in fromstring
    "Please call frombytes() instead.")
Exception: fromstring() has been removed. Please call frombytes() instead.

----------------------------------------------------------------------
```

PR #444, opened 2016-04-10
BF: Use newer call to frombytes, replacing fromstring.

arokem:
This addresses #443 on my machine, but is clearly not the complete solution to this issue.

### DIFF
--- a/nibabel/dft.py
+++ b/nibabel/dft.py
@@ -138,8 +138,8 @@ class _Series(object):
             max = data.max()
             data = data * 255 / (max - min)
         data = data.astype(numpy.uint8)
-        im = PIL.Image.fromstring('L', (self.rows, self.columns),
-                                  data.tostring())
+        im = PIL.Image.frombytes('L', (self.rows, self.columns),
+                                  data.tobytes())
         s = BytesIO()
         im.save(s, 'PNG')
         return s.getvalue()
--- a/nibabel/dft.py
+++ b/nibabel/dft.py
@@ -139,7 +139,7 @@ class _Series(object):
             data = data * 255 / (max - min)
         data = data.astype(numpy.uint8)
         im = PIL.Image.frombytes('L', (self.rows, self.columns),
-                                  data.tobytes())
+                                 data.tostring())
         s = BytesIO()
         im.save(s, 'PNG')
         return s.getvalue()
--- a/nibabel/dft.py
+++ b/nibabel/dft.py
@@ -124,6 +124,13 @@ class _Series(object):
 
     def as_png(self, index=None, scale_to_slice=True):
         import PIL.Image
+        # For compatibility with older versions of PIL that did not
+        # have `frombytes`:
+        if hasattr(PIL.Image, 'frombytes'):
+            frombytes = PIL.Image.frombytes
+        else:
+            frombytes = PIL.Image.fromstring
+
         if index is None:
             index = len(self.storage_instances) // 2
         d = self.storage_instances[index].dicom()
@@ -138,8 +145,9 @@ class _Series(object):
             max = data.max()
             data = data * 255 / (max - min)
         data = data.astype(numpy.uint8)
-        im = PIL.Image.frombytes('L', (self.rows, self.columns),
-                                 data.tostring())
+        im = frombytes('L', (self.rows, self.columns),
+                       data.tostring())
+
         s = BytesIO()
         im.save(s, 'PNG')
         return s.getvalue()
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ addons:
       - libatlas-base-dev
 env:
     global:
-        - DEPENDS="numpy scipy matplotlib h5py"
+        - DEPENDS="numpy scipy matplotlib h5py pillow"
         - PYDICOM=1
         - INSTALL_TYPE="setup"
 python:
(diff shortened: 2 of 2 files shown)


effigies:
#443 seems to be introduced in [Pillow v3.0.0+](https://github.com/python-pillow/Pillow/commit/71c95c8e5f3bba1845444a246d04646825e6bab3). As far as I can tell, Pillow is the generally accepted fork of PIL, so following their deprecations seems pretty safe.

That said, it should still be [`data.tostring`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tostring.html), not [`data.tobytes`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tobytes.html). `tostring` was renamed to `tobytes` in numpy 1.9.0+, but `tostring` was left as a compatibility alias.

nibabel/dft.py
```diff
@@ -138,8 +138,8 @@ def as_png(self, index=None, scale_to_slice=True):
             max = data.max()
             data = data * 255 / (max - min)
         data = data.astype(numpy.uint8)
-        im = PIL.Image.fromstring('L', (self.rows, self.columns),
-                                  data.tostring())
+        im = PIL.Image.frombytes('L', (self.rows, self.columns),
+                                  data.tobytes())
```
effigies: PEP8: `data` should align with `'L'`.

effigies:
Not sure the best way to test this. Can't get Travis to install Pillow==3: https://travis-ci.org/effigies/nibabel/jobs/122307466

matthew-brett:
Have a look at http://travis-wheels.scikit-image.org/ for the wheels stored in the 'wheelhouse' we are using.  I just built Pillow 3.2 (using https://github.com/matthew-brett/travis-wheel-builder) - so I think `Pillow==3.2` will work.

Probably best way is to add `pillow` to DEPENDS for the full tests, and then add the earlier pillow version (say 2.6) to the matrix entry with the minimum pydicom version.

arokem:
How's this?

effigies:
LGTM

matthew-brett:
Ariel - can you rebase on master to get rid of unrelated test failures?

Maybe you'll have to use `fromstring` on Pillow < 3?

effigies:
> Maybe you'll have to use fromstring on Pillow < 3?

`fromstring` was [deprecated](https://github.com/python-pillow/Pillow/blob/2.0.0/PIL/Image.py#L1828-L1835) in 2.0.

matthew-brett:
Hum - but I suppose there will be some people still using PIL.

matthew-brett:
Maybe test against PIL on Python 2.7 instead of older Pillow.

effigies:
Travis' default install [supports](https://travis-ci.org/effigies/nibabel/builds/122322097) `PIL.Image.frombytes`, which did not exist in [PIL 1.1.7](https://bitbucket.org/effbot/pil-117/src/f356a1f64271e9d3206fcf9059492ba1c9e163d6/PIL/Image.py), but I can [break the build](https://travis-ci.org/effigies/nibabel/builds/122324272#L1754) by directly [installing](https://github.com/effigies/nibabel/commit/53e05306110e6a90465823ab09be18e6485a0528) PIL.

So, you could do something like:

```Python
import PIL.Image
if hasattr(PIL.Image, 'frombytes'):
    frombytes = PIL.Image.frombytes
else:
    frombytes = PIL.Image.fromstring
...
im = frombytes('L', (self.rows, self.columns), data.tostring())
```

Still... PIL itself hasn't been updated since Py2.6, and has zero Py3 support. Looking around, Linux and Python distributions prefer Pillow to PIL. Supporting ancient PIL only seems to make sense to me if we're going to continue supporting 2.6.

matthew-brett:
OK - I see the PIL doesn't have anything that can be got from pip, so `pip install pil` will always fail.

So, I agree, I think we can forget about PIL support now.

effigies:
Also need to revert 6b9ce76 (unless we want to support PIL without testing it?). Or just reset to 310b6d8 and force push.

Otherwise, still LGTM.

arokem:
Sorry @effigies -- I might be misunderstanding your comment. My understanding is that `Import PIL` will import pillow on systems that don't have PIL installed, so technically we are not supporting PIL here. The commit https://github.com/nipy/nibabel/commit/6b9ce763d25ed1c50cf33f2873132fb6dc535f2e did not introduce a PIL/pillow dependency -- that was there before (for a while, apparently at least since 2011, according to the blame view of the `import PIL` line).

effigies:
Sorry, for the imprecision. Unfortunately Pillow and PIL are both imported as PIL, so it may be unavoidable. When I write PIL, I'm referring to the library that ended at version 1.1.7, not the import.

The question is not whether to have the dependency, but whether to permit a dependency to support the defunct PIL proper or just stick with the actively maintained Pillow. `.fromstring` is a holdover from PIL that Pillow deprecated for `.frombytes` as of 2.0.0 (released March 2013).

If we don't mean to support PIL, which is ancient, then adding a codepath to use `.fromstring` as a backup seems extraneous. BUT, if we are going to have that codepath, it seems to me that we should go ahead and have a PIL-based Travis test. So my preference would be to merge either 6b9ce76 (support PIL 1.1.7 and Pillow >= 1.0) or 310b6d8 (only support Pillow >= 2.0).

That said, I think the code is in a fine state, so I'll leave remaining decisions to @matthew-brett.

arokem:
OK - thanks for clarifying! I will wait for @matthew-brett to weigh in.

matthew-brett:
Looks good to me - thanks.

PR #446, opened 2016-04-13
TST: Add testing on Python 3.5 on Appveyor.

arokem:
Feel free to ignore this. I am trying to understand the failures I am seeing over here: https://github.com/conda-forge/staged-recipes/pull/284

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,15 +30,18 @@ install:
   # the parent CMD process).
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
 
+  # Set up a conda environment:
+  - conda config --set always_yes yes
+  - conda update -q conda
+  - conda info -a
+  - conda create -q -n test-environment python=%PYTHON_VERSION%
+  - activate test-environment
+
   # Check that we have the expected version and architecture for Python
   - "python --version"
   - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
 
   # Install the dependencies of the project.
-  - conda update -q conda
-  - conda info -a
-  - conda create -q -n test-environment python=%PYTHON_VERSION%
-  - activate test-environment
   - "conda install --yes --quiet numpy scipy matplotlib nose h5py"
   - "pip install pydicom"
   - "python setup.py install"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,6 +35,10 @@ install:
   - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
 
   # Install the dependencies of the project.
+  - conda update -q conda
+  - conda info -a
+  - conda create -q -n test-environment python=%PYTHON_VERSION%
+  - activate test-environment
   - "conda install --yes --quiet numpy scipy matplotlib nose h5py"
   - "pip install pydicom"
   - "python setup.py install"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,10 @@ environment:
       PYTHON_VERSION: "3.4"
       PYTHON_ARCH: "64"
 
+    - PYTHON: "C:\\Python35-conda64"
+      PYTHON_VERSION: "3.5"
+      PYTHON_ARCH: "64"
+
 install:
   # Install miniconda Python
   - "powershell ./tools/install_python.ps1"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,10 @@ environment:
       PYTHON_VERSION: "3.5"
       PYTHON_ARCH: "64"
 
+    - PYTHON: "C:\\Python35-conda32"
+      PYTHON_VERSION: "3.5"
+      PYTHON_ARCH: "32"
+
 install:
   # Install miniconda Python
   - "powershell ./tools/install_python.ps1"


arokem:
OK - please ignore the license to ignore. I have a question: any particular reason the Appveyor thing is running under @Eric89GXL's Appveyor account (rather than an org account)?

Eric89GXL:
No real reason. Apparently [I added AppVeyor support] (https://github.com/nipy/nibabel/commit/03f21ad939f0ed3111a4534471068e9776779ccf) a while back. Let me know if I need to do anything to switch it over.

arokem:
Huh. It passed (after a few re-pushes). Feel free to merge this, if you think it's a good idea (Appveyor is admittedly a bit of a pain in the neck...)

lesteve:
Looking at https://ci.appveyor.com/project/Eric89GXL/nibabel/build/1.0.694/job/ntqqoxkcha3ywiji#L84, it looks like the python 3.5 AppVeyor build actually uses python 2.7.

effigies:
Did something change on the conda end between pushes? From what I've seen, passing is somewhat random, these days, and the errors are usually checksum failures. Maybe a package server that was/is out of sync?

arokem:
@lesteve : good catch! That's wild. Will investigate
@effigies : no change, as far as I can tell. The failures are probably due to random download failures on Appveyor (see discussion here: https://github.com/ContinuumIO/anaconda-issues/issues/71)

matthew-brett:
Hmm - that's weird - the buildbots have been testing Python 3.5 with Christoph Gohlke's wheels for a while without (these) problems : 

http://nipy.bic.berkeley.edu/builders/nibabel-bdist32-35

arokem:
Yes. It is weird. Any thoughts?

This might be obvious, but a lot of these look like overflow errors to me.
Something about the machine this is running on? Or Anaconda versions of
libraries being weird?

matthew-brett:
It's difficult to imagine the precision differences would be due to the machine.

Do you know - does numpy pass its own test suite on this setup?

arokem:
It seems that it does. For example: https://ci.appveyor.com/project/numpy/numpy/build/1.0.281/job/htein9lmbbgn116y

matthew-brett:
I've given you an account on the BIDS windows testing box, called "carly", I've sent details via email.

The way I use the machine is via remote desktop.  So I install the Microsoft Remote Desktop app on OSX, then forward the relevant port to my machine with:

ssh -L3389:localhost:3389 myuser@address-of-carly

and start remote desktop pointing with a session pointed to localhost:3389 . You'll need your password to log in though.

The least obtrusive way to get going is to work on a powershell command line inside a Windows docker container, like this:

```
docker run --rm -ti windowsservercore powershell
```

This gives you a fresh throwaway container to play in, for which you have admin permissions.   See: http://matthew-brett.github.io/pydagogue/windows_containers.html#example-of-installing-into-a-container

arokem:
OK - I've managed to install Anaconda in there (although not inside the docker container - instead I have installed it only for my user - I apologize if that's more intrusive than necessary), and I manage to replicate the same failures on Carly as the ones we see on Appveyor. It's quite hard for me to make any sense of them, though. The most I can make out is that a lot of them seem to have to do with nan's getting cast into the wrong values (for example, they're supposed to cast to 0, and end up in -128 instead, or they are supposed to cast to 0 and end up in some floating point value (~1.0039) instead. Especially on round-trips, things seem to go wrong (with overflow also happening - 0's going to `np.iinfo(np.int64).min` instead). Does this ring any bells?

matthew-brett:
Yes, sadly it does.   We had this kind of problem for an MSVC bug:

https://github.com/nipy/nibabel/pull/367

https://github.com/numpy/numpy/issues/6685

The tests are - I think - differences from the expected numbers when working with int max values and doing array scaling.

It was extremely hard to track the previous error down, partly because the tests are a little opaque.  You can see Ben C struggling manfully with the code in the first PR.

I ended up spending a lot of time taking the test input for working and not working versions (with and without MKL in that case) and comparing the inputs and outputs.   Ugh.

matthew-brett:
Ariel - any progress here?

arokem:
No progress. I haven't had time to look into this more.

lesteve:
I tracked down one of the failing test (test_nan2zero_scaling) to a numpy clip issue, which for some reason seems to be happening only on Windows with Python 3.5, see https://github.com/numpy/numpy/issues/7601.

lesteve:
Good news, with the numpy fix from the PR https://github.com/numpy/numpy/pull/7678, I can run the nibabel tests without any failurn my VM. So maybe there is no Python 3.5 specific problem within nibabel. If someone could double-check that would be great.

In the medium term, not sure whether there is an easy way to run the tests with AppVeyor using numpy from master. Alternatively you could wait for a numpy version with the fix to be relased before using Python 3.5 for continuous integration.

arokem:
I am +1 for waiting on a numpy release with the fix, before proceeding with this PR.

lesteve:
FYI the numpy fix is going to be in 1.11.1. The numpy release candidate has been announced a few hours ago so it may not be that long before a numpy release with the fix.

lesteve:
numpy 1.11.1 is available through conda now, so in principle it is just a matter of relaunching the CI tests to make sure the AppVeyor tests pass on Python 3.5 before this PR can be merged.

arokem:
Thanks for the alert. I just rebased this on top of current master and pushed, which should trigger the CI.

arokem:
Yep. Looks like everything is coming up green. 

I noticed that there is no Python 2/64 bot on Appveyor. Is there any reason for that, or should I add it?

Eric89GXL:
I think it was just for testing time, AppVeyor processes serially.

FYI a better solution is probably to use [ci-helpers] (https://github.com/astropy/ci-helpers/blob/master/appveyor.yml) to get Anaconda up on AppVeyor, but that could be another PR when someone gets motivated.

lesteve:
@matthew-brett I reckon this one should be merged.

matthew-brett:
Thanks for doing this - merging.

PR #447, opened 2016-04-22
BF: make MPL viewer tests require MPL 1.3

matthew-brett:
Viewer tests giving errors on very old MPL: see
http://nipy.bic.berkeley.edu/builders/nibabel-py2.x-sid-sparc/builds/129/steps/shell_5/logs/stdio

### DIFF
--- a/nibabel/optpkg.py
+++ b/nibabel/optpkg.py
@@ -1,4 +1,7 @@
 """ Routines to support optional packages """
+from distutils.version import LooseVersion
+
+from .externals.six import string_types, callable
 
 try:
     import nose
@@ -10,7 +13,17 @@ else:
 from .tripwire import TripWire
 
 
-def optional_package(name, trip_msg=None):
+def _check_pkg_version(pkg, min_version):
+    # Default version checking function
+    if isinstance(min_version, string_types):
+        min_version = LooseVersion(min_version)
+    try:
+        return min_version <= pkg.__version__
+    except AttributeError:
+        return False
+
+
+def optional_package(name, trip_msg=None, min_version=None):
     """ Return package-like thing and module setup for package `name`
 
     Parameters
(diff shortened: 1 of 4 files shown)


effigies:
If this is something that will be needed again, it might be good to add an optional `min_version` keyword in `optional_package`, that only returns `(pkg, True, lambda: None)` if `pkg.__version__ >= min_version`. But for a one-off requirement, this seems fine.

:+1: for merge.

matthew-brett:
Good point - I made optpkg take a `min_version` argument and used that instead.

nibabel/optpkg.py
```diff
@@ -66,6 +85,12 @@ def optional_package(name, trip_msg=None):
     >>> hasattr(subpkg, 'dirname')
     True
     """
+    if callable(min_version):
+        check_version = min_version
+    elif min_version is None:
+        check_version = lambda pkg : True
+    else:
+        check_version = lambda pkg : _check_pkg_version(pkg, min_version)
```
effigies: PEP8: No space before `:` in lines 91, 93.

nibabel/optpkg.py
```diff
@@ -75,7 +100,13 @@ def optional_package(name, trip_msg=None):
...
     else:  # import worked
         # top level module
-        return pkg, True, lambda: None
+        if check_version(pkg):
+            return pkg, True, lambda: None
+        if trip_msg is None:
+            trip_msg = ('We need package %s version >= %s for these '
+                        'functions, but package %s does not meet version '
+                        'requirements'
+                        % (name, name, name))
```
effigies: `(name, min_version, name)`?

nibabel/tests/test_optpkg.py
```diff
@@ -0,0 +1,63 @@
...
+    assert_false('fake_pkg' in sys.modules)
+    # Not inserted yet
+    assert_bad(fake_name)
+    try:
+        sys.modules[fake_name] = fake_pkg
+        # No __version__ yet
+        assert_good(fake_name)  # With no version check
+        assert_bad(fake_name, '1.0')
+        # We can make an arbitrary callable to check version
+        assert_good(fake_name, lambda pkg : True)
```
effigies: These aren't hitting the Travis tests, but are still PEP8 violations. Just out of curiosity, is this becoming a convention in other projects?

effigies:
Couple minor comments, but otherwise, LGTM.

matthew-brett:
Thanks for the review - I made some changes.

I wasn't sure what you were asking about when you asked whether this was standard in other projects ?

effigies:
Just was wondering about the `lambda arg : <body>` style with the space before the `:`, since you used it pretty consistently.

Changes look good. No further suggestions.

PR #455, opened 2016-06-02
Error for file not found

arokem:
Currently, when a file doesn't exist (for example, if you typo'd the file-name), you get a relatively cryptic error: "couldn't work out file-type for file foo.nii.gz". I propose that we check if the file exists first, and raise an informative error up front if it doesn't exist.

### DIFF
--- a/nibabel/loadsave.py
+++ b/nibabel/loadsave.py
@@ -20,6 +20,7 @@ from .imageclasses import all_image_classes
 from .arrayproxy import is_proxy
 from .py3k import FileNotFoundError
 
+
 def load(filename, **kwargs):
     ''' Load file given filename, guessing at file type
 
--- a/nibabel/loadsave.py
+++ b/nibabel/loadsave.py
@@ -9,6 +9,7 @@
 # module imports
 """ Utilities to load and save image objects """
 
+import os.path as op
 import numpy as np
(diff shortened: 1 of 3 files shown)


matthew-brett:
Seems reasonable.  Only possible problem would be if the extra check added significant time, for example on slower NFS systems.

matthew-brett:
`FileNotFoundError` seems to be Python 3 only.

arokem:
Bummer. I switched it to `IOError` instead (works on both 2.7 and 3.5 on my laptop). There are other hacks that can be done (along these lines: http://stackoverflow.com/a/21368622/3532933), but that seems like too much here.

matthew-brett:
Believe it or not - I hate to be annoying - but it would be nice to be able trap the file not found error specifically - maybe add it to `nibabel/py3k.py` as `class FileNotFoundError(IOError)` for Python 2.

arokem:
Not annoying at all. Seems like a good idea. Did I get it right?

matthew-brett:
Looks OK to me - just a PEP8 nitpick on travis - otherwise good to go.

arokem:
Cool. Added the necessary white-space.

matthew-brett:
Thanks for doing that - merging.

PR #457, opened 2016-06-05
DOC: Fieldtrip moved from Google Code to GitHub

effigies:
Update links for reference FreeSurfer matlab files.


matthew-brett:
Excellent - thanks for finding the exact lines, very helpful.

PR #458, opened 2016-06-07
Fs surf metadata

### LINKED ISSUE #456: BUG: write_geometry and read_geometry don't seem to be in the same coordinate system
As identified in https://github.com/mne-tools/mne-python/issues/3281, 

`nibabel.freesurfer.io.write_geometry` doesn't seem to write in the correct coordinate system. 

Weirdly, this is visible in freeviewer, but not in the python `viz` functions we use in [mne-python](https://github.com/mne-tools/mne-python).

To reproduce the error (you have to copy/paste `subjects/sample` into `subjects/sample_test` since it modifies the files):

```
import os.path as op
from nibabel.freesurfer.io import read_geometry, write_geometry

 # I'm not sure where nibabel example is stored, so I'm using MNE.
from mne.datasets import sample
data_path = sample.data_path()

# The paths to freesurfer reconstructions
subjects_dir = data_path + '/subjects'
subject = 'sample_test'

fname = op.join(subjects_dir, subject, 'bem', 'inner_skull.surf')
coord, faces = read_geometry(fname)
write_geometry(fname, coord, faces)
```

in `freeview`, we now see a misalignment:

bug_write_surface

Weirdly, it doesn't appear when plotting in in python:

`mne.viz.plot_bem` plots as expected:

write_surface_mne

agramfort:
cc @effigies @kingjr @jaeilepp

closes #456

### DIFF
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -4,9 +4,8 @@ import warnings
 import numpy as np
 import getpass
 import time
-from collections import OrderedDict
 
-
+from ..externals import OrderedDict
 from ..externals.six.moves import xrange
 from ..openers import Opener
 
@@ -176,7 +175,7 @@ def write_geometry(filepath, coords, faces, create_stamp=None,
         postlude = [b'\x00\x00\x00\x14']
         for key, val in volume_info.items():
             if key in ('voxelsize', 'xras', 'yras', 'zras', 'cras'):
-                val = '{:.3f} {:.3f} {:.3f}'.format(*val)
+                val = '{:.4f} {:.4f} {:.4f}'.format(*val)
             elif key == 'volume':
                 val = '{:d} {:d} {:d}'.format(*val)
             key = key.ljust(6)
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -3,6 +3,7 @@ from __future__ import division, print_function, absolute_import
 import numpy as np
 import getpass
 import time
+from collections import OrderedDict
 
 
 from ..externals.six.moves import xrange
@@ -44,13 +45,17 @@ def _fread3_many(fobj, n):
     return (b1 << 16) + (b2 << 8) + b3
 
 
-def read_geometry(filepath):
+def read_geometry(filepath, read_metadata=False, read_stamp=False):
     """Read a triangular format Freesurfer surface mesh.
 
     Parameters
     ----------
     filepath : str
         Path to surface file
+    read_metadata : bool
+        Read metadata as key-value pairs
+    read_stamp : bool
+        Return the comment from the file
 
     Returns
     -------
@@ -58,7 +63,14 @@ def read_geometry(filepath):
         nvtx x 3 array of vertex (x, y, z) coordinates
     faces : numpy array
         nfaces x 3 array of defining mesh triangles
+    volume_info : dict-like
+        If read_metadata is true, key-value pairs found in the geometry file
+    create_stamp : str
+        If read_stamp is true, the comment added by the program that saved
+        the file
     """
+    volume_info = None
+
     with open(filepath, "rb") as fobj:
         magic = _fread3(fobj)
         if magic == 16777215:  # Quad file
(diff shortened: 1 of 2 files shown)


Eric89GXL:
I see some debugging things in there, I assume you want someone to take over?

effigies:
Ugh. Running up against Python 2.6's lack of OrderedDict. I guess we could just do a list of 2-tuples? If someone wants to access by name on 2.7+, they can just do `OrderedDict(volume_info)` themselves.

jaeilepp:
``extra`` as well as the stuff at the end should be returned too, no? Otherwise you cannot recreate everything at ``write_geometry``.

nibabel/freesurfer/io.py
```diff
@@ -112,13 +162,27 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
...
     if create_stamp is None:
         create_stamp = "created by %s on %s" % (getpass.getuser(),
                                                 time.ctime())
 
+    postlude = b''
+    if volume_info is not None and len(volume_info) > 0:
+        postlude = [b'\x00\x00\x00\x14']
+        for key, val in volume_info.items():
+            if key in ('voxelsize', 'xras', 'yras', 'zras', 'cras'):
+                val = '{:.3f} {:.3f} {:.3f}'.format(*val)
```
effigies: Realized these should be `.4f`.

matthew-brett:
We have our own copy of ordereddict for Python 2.6 in externals.   But it's OK to drop Python 2.6 too.

nibabel/freesurfer/io.py
```diff
@@ -1,8 +1,10 @@
 from __future__ import division, print_function, absolute_import
 
+import warnings
 import numpy as np
 import getpass
 import time
+from collections import OrderedDict
```
matthew-brett: as in `from nibabel.externals import OrderedDict`

agramfort:
comments addressed. Let's see if travis is happy.

@Eric89GXL what debug statements do you see?

nibabel/freesurfer/io.py
```diff
@@ -86,20 +98,57 @@ def read_geometry(filepath):
...
+
+                if extra[:4] != b'\x00\x00\x00\x14':
+                    warnings.warn("Unknown extension code.")
+                else:
+                    try:
+                        for line in extra[4:].split(b'\n'):
+                            if len(line) == 0:
+                                continue
+                            key, val = map(bytes.strip, line.split(b'=', 1))
+                            print(key, val)
```
effigies: print statement

nibabel/freesurfer/io.py
```diff
@@ -112,13 +161,27 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
...
     if create_stamp is None:
         create_stamp = "created by %s on %s" % (getpass.getuser(),
                                                 time.ctime())
 
+    postlude = b''
+    if volume_info is not None and len(volume_info) > 0:
+        postlude = [b'\x00\x00\x00\x14']
+        for key, val in volume_info.items():
+            if key in ('voxelsize', 'xras', 'yras', 'zras', 'cras'):
+                val = '{:.4f} {:.4f} {:.4f}'.format(*val)
```
effigies: Python 2.6 fix: `val = '{0:.4f} {1:.4f} {2:.4f}'.format(*val)`

nibabel/freesurfer/io.py
```diff
@@ -112,13 +161,27 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
...
+            elif key == 'volume':
+                val = '{:d} {:d} {:d}'.format(*val)
```
effigies: Python 2.6 fix: `val = '{0:d} {1:d} {2:d}'.format(*val)`

nibabel/freesurfer/io.py
```diff
@@ -112,13 +161,27 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
...
+            key = key.ljust(6)
+            postlude.append('{} = {}'.format(key, val).encode('utf-8'))
```
effigies: Python 2.6 fix: `postlude.append('{0} = {1}'.format(key, val).encode('utf-8'))`

matthew-brett:
@agramfort - there are a few lines not covered by the tests here - see https://codecov.io/gh/nipy/nibabel/compare/791b10f7bc08c3f5ef5ae1e74d8b38e52158265e...63fd6c979cf27f51ad36a125973b60d8e40c7826

nibabel/freesurfer/io.py
```diff
@@ -44,21 +45,32 @@ def _fread3_many(fobj, n):
...
+    read_stamp : bool
+        Return the comment from the file
 
     Returns
     -------
     coords : numpy array
         nvtx x 3 array of vertex (x, y, z) coordinates
     faces : numpy array
         nfaces x 3 array of defining mesh triangles
+    volume_info : dict-like
```
matthew-brett: Say it's an ordered dict, given that we know that is the case?

nibabel/freesurfer/io.py
```diff
@@ -86,20 +98,57 @@ def read_geometry(filepath):
...
+            create_stamp = fobj.readline().rstrip(b'\n').decode('utf-8')
             fobj.readline()
             vnum = np.fromfile(fobj, ">i4", 1)[0]
             fnum = np.fromfile(fobj, ">i4", 1)[0]
             coords = np.fromfile(fobj, ">f4", vnum * 3).reshape(vnum, 3)
             faces = np.fromfile(fobj, ">i4", fnum * 3).reshape(fnum, 3)
+
+            extra = fobj.read() if read_metadata else b''
+            if extra:
+                volume_info = OrderedDict()
```
matthew-brett: Maybe neater to refactor into `_read_volume_info` method, as in

```
if extra:
    volume_info = self._read_volume_info(extra)
```

agramfort:
appveyor error seems unrelated

I have no idea why my last commit was necessary.

nibabel/freesurfer/io.py
```diff
@@ -44,21 +45,64 @@ def _fread3_many(fobj, n):
...
+                key = key.decode('utf-8')
+                if key in ('voxelsize', 'xras', 'yras', 'zras', 'cras'):
+                    val = np.fromstring(val, sep=' ')
+                    val = val.astype(np.float)
+                elif key == 'volume':
+                    val = np.fromstring(val, sep=' ', dtype=np.uint)
+                    val = val.astype(np.int)
+                volume_info[key] = val
+        except ValueError:
+            raise ValueError("Error parsing volume info")
```
Eric89GXL: This is a lot of lines to be put in `try`. Which line is expected to fail? If it's only one (hopefully) then it's better to only have that line in the `try` clause.

If you're trying to catch any possible error on the other hand, then `except Exception` is more appropriate, and it would be good to do `"Error parsing volume info: '%s'" % err`

nibabel/freesurfer/io.py
```diff
@@ -44,21 +45,64 @@ def _fread3_many(fobj, n):
...
+
+    if len(volume_info) == 0:
+        warnings.warn("Volume geometry info is either "
+                      "not contained or not valid.")
```
Eric89GXL: I don't think we need to warn if there is no volume info, that's clear from the dict being returned as empty right? Plus this ends up being redundant with the `warnings.warn()` call above, so at the very least this should go in the `else` clause above if you want to keep it (and the "unknown extension code" warning should probably have some of the same text, but come out in a single warning instead of 2 as it would be here).

Eric89GXL:
@agramfort while you're in there are you up for adding the `quad` stuff from MNE-Python?

agramfort:
take over if you have time. I have some more urgent matter to take care of
in the next couple of days :(

Eric89GXL:
Commit added to take care of my last comments (taking over for @agramfort). Ready for review/merge from my end.

Eric89GXL:
Whoops, didn't mean to make that comment. Just meant to close. Closing in favor of #460

PR #460, opened 2016-06-10
[MRG+2] Surf metadata

jaeilepp:
Takeover from https://github.com/nipy/nibabel/pull/458
cc: @agramfort @effigies 
closes https://github.com/nipy/nibabel/issues/456
I could not figure out the 'tail' of extra information, so I suggest we just dump it for the time being.

### DIFF
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -462,7 +462,7 @@ def _serialize_volume_info(volume_info):
             if not (np.array_equal(volume_info[key], [20]) or np.array_equal(
                     volume_info[key], [2, 0, 20])):
                 warnings.warn("Unknown extension code.")
-            strings.append(np.array(volume_info[key], dtype='>i4').tobytes())
+            strings.append(np.array(volume_info[key], dtype='>i4').tostring())
         elif key in ('valid', 'filename'):
             val = volume_info[key]
             strings.append('{0} = {1}\n'.format(key, val).encode('utf-8'))
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -462,7 +462,7 @@ def _serialize_volume_info(volume_info):
             if not (np.array_equal(volume_info[key], [20]) or np.array_equal(
                     volume_info[key], [2, 0, 20])):
                 warnings.warn("Unknown extension code.")
-            strings.append(np.array(volume_info[key], dtype='>i4').tostring())
+            strings.append(np.array(volume_info[key], dtype='>i4').tobytes())
         elif key in ('valid', 'filename'):
             val = volume_info[key]
             strings.append('{0} = {1}\n'.format(key, val).encode('utf-8'))
@@ -474,4 +474,4 @@ def _serialize_volume_info(volume_info):
             val = volume_info[key]
             strings.append('{0} = {1:f} {2:f} {3:f}\n'.format(
                 key.ljust(6), val[0], val[1], val[2]).encode('utf-8'))
-    return ''.join(strings)
+    return b''.join(strings)
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -78,17 +78,17 @@ def read_geometry(filepath, read_metadata=False, read_stamp=False):
     filepath : str
         Path to surface file
     read_metadata : bool
-        Read metadata as key-value pairs
+        Read metadata as key-value pairs.
         Valid keys:
-            'head' : array of int
-            'valid' : str
-            'filename' : str
-            'volume' : array of int, shape (3,)
-            'voxelsize' : array of float, shape (3,)
-            'xras' : array of float, shape (3,)
-            'yras' : array of float, shape (3,)
-            'zras' : array of float, shape (3,)
-            'cras' : array of float, shape (3,)
+            * 'head' : array of int
+            * 'valid' : str
+            * 'filename' : str
+            * 'volume' : array of int, shape (3,)
+            * 'voxelsize' : array of float, shape (3,)
+            * 'xras' : array of float, shape (3,)
+            * 'yras' : array of float, shape (3,)
+            * 'zras' : array of float, shape (3,)
+            * 'cras' : array of float, shape (3,)
     read_stamp : bool
         Return the comment from the file
 
@@ -173,17 +173,17 @@ def write_geometry(filepath, coords, faces, create_stamp=None,
     create_stamp : str
         User/time stamp (default: "created by <user> on <ctime>")
     volume_info : dict-like or None
-        Key-value pairs to encode at the end of the file
+        Key-value pairs to encode at the end of the file.
         Valid keys:
-            'head' : array of int
-            'valid' : str
-            'filename' : str
-            'volume' : array of int, shape (3,)
-            'voxelsize' : array of float, shape (3,)
-            'xras' : array of float, shape (3,)
-            'yras' : array of float, shape (3,)
-            'zras' : array of float, shape (3,)
-            'cras' : array of float, shape (3,)
+            * 'head' : array of int
+            * 'valid' : str
+            * 'filename' : str
+            * 'volume' : array of int, shape (3,)
+            * 'voxelsize' : array of float, shape (3,)
+            * 'xras' : array of float, shape (3,)
+            * 'yras' : array of float, shape (3,)
+            * 'zras' : array of float, shape (3,)
+            * 'cras' : array of float, shape (3,)
 
     """
     magic_bytes = np.array([255, 255, 254], dtype=np.uint8)
@@ -203,23 +203,8 @@ def write_geometry(filepath, coords, faces, create_stamp=None,
         faces.astype('>i4').reshape(-1).tofile(fobj)
 
         # Add volume info, if given
-        if volume_info is None or len(volume_info) == 0:
-            return
-
-        for key, val in volume_info.items():
-            if key == 'head':
-                if not (np.array_equal(val, [20]) or np.array_equal(
-                        val, [2, 0, 20])):
-                    warnings.warn("Unknown extension code.")
-                np.array(val, dtype='>i4').tofile(fobj)
-            elif key in ('valid', 'filename'):
-                fobj.write('{0} = {1}\n'.format(key, val).encode('utf-8'))
-            elif key == 'volume':
-                fobj.write('{0} = {1} {2} {3}\n'.format(
-                    key, val[0], val[1], val[2]).encode('utf-8'))
-            else:
-                fobj.write('{0} = {1:.4f} {2:.4f} {3:.4f}\n'.format(
-                    key.ljust(6), val[0], val[1], val[2]).encode('utf-8'))
+        if volume_info is not None and len(volume_info) > 0:
+            fobj.write(_serialize_volume_info(volume_info))
 
 
 def read_morph_data(filepath):
@@ -461,3 +446,32 @@ def read_label(filepath, read_scalars=False):
         scalar_array = np.loadtxt(filepath, skiprows=2, usecols=[-1])
         return label_array, scalar_array
     return label_array
+
+
+def _serialize_volume_info(volume_info):
+    """Helper for serializing the volume info."""
+    keys = ['head', 'valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras',
+            'zras', 'cras']
+    diff = set(volume_info.keys()).difference(keys)
+    if len(diff) > 0:
+        raise ValueError('Invalid volume info: %s.' % diff.pop())
+
+    strings = list()
+    for key in keys:
+        if key == 'head':
+            if not (np.array_equal(volume_info[key], [20]) or np.array_equal(
+                    volume_info[key], [2, 0, 20])):
+                warnings.warn("Unknown extension code.")
+            strings.append(np.array(volume_info[key], dtype='>i4').tostring())
+        elif key in ('valid', 'filename'):
+            val = volume_info[key]
+            strings.append('{0} = {1}\n'.format(key, val).encode('utf-8'))
+        elif key == 'volume':
+            val = volume_info[key]
+            strings.append('{0} = {1} {2} {3}\n'.format(
+                key, val[0], val[1], val[2]).encode('utf-8'))
+        else:
+            val = volume_info[key]
+            strings.append('{0} = {1:f} {2:f} {3:f}\n'.format(
+                key.ljust(6), val[0], val[1], val[2]).encode('utf-8'))
+    return ''.join(strings)
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -182,18 +182,20 @@ def write_geometry(filepath, coords, faces, create_stamp=None,
         faces.astype('>i4').reshape(-1).tofile(fobj)
 
         # Add volume info, if given
-        if volume_info is not None and len(volume_info) > 0:
-            for key, val in volume_info.items():
-                if key == 'head':
-                    np.array(val, dtype='>i4').tofile(fobj)
-                elif key in ('valid', 'filename'):
-                    fobj.write('{0} = {1}\n'.format(key, val).encode('utf-8'))
-                elif key == 'volume':
-                    fobj.write('{0} = {1} {2} {3}\n'.format(
-                        key, val[0], val[1], val[2]).encode('utf-8'))
-                else:
-                    fobj.write('{0} = {1:.4f} {2:.4f} {3:.4f}\n'.format(
-                        key.ljust(6), val[0], val[1], val[2]).encode('utf-8'))
+        if volume_info is None or len(volume_info) == 0:
+            return
+
+        for key, val in volume_info.items():
+            if key == 'head':
+                np.array(val, dtype='>i4').tofile(fobj)
+            elif key in ('valid', 'filename'):
+                fobj.write('{0} = {1}\n'.format(key, val).encode('utf-8'))
+            elif key == 'volume':
+                fobj.write('{0} = {1} {2} {3}\n'.format(
+                    key, val[0], val[1], val[2]).encode('utf-8'))
+            else:
+                fobj.write('{0} = {1:.4f} {2:.4f} {3:.4f}\n'.format(
+                    key.ljust(6), val[0], val[1], val[2]).encode('utf-8'))
 
 
 def read_morph_data(filepath):
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -4,9 +4,8 @@ import warnings
 import numpy as np
 import getpass
 import time
-from collections import OrderedDict
 
-
+from ..externals import OrderedDict
 from ..externals.six.moves import xrange
 from ..openers import Opener
 
@@ -176,7 +175,7 @@ def write_geometry(filepath, coords, faces, create_stamp=None,
         postlude = [b'\x00\x00\x00\x14']
         for key, val in volume_info.items():
             if key in ('voxelsize', 'xras', 'yras', 'zras', 'cras'):
-                val = '{:.3f} {:.3f} {:.3f}'.format(*val)
+                val = '{:.4f} {:.4f} {:.4f}'.format(*val)
             elif key == 'volume':
                 val = '{:d} {:d} {:d}'.format(*val)
             key = key.ljust(6)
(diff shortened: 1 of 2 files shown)


Eric89GXL:
@jaeilepp I didn't know you took this over. @agramfort suggested I do it, so I made a commit to #458. It looks like your code is a bit cleaner, but maybe you could make use of the test modifications I made.

nibabel/freesurfer/io.py
```diff
@@ -45,35 +45,20 @@ def _fread3_many(fobj, n):
...
-def _read_volume_info(extra):
+def _read_volume_info(fobj):
     volume_info = OrderedDict()
-
-    if extra is None:
-        return volume_info
-
-    if extra[:4] != b'\x00\x00\x00\x14':
+    head = np.fromfile(fobj, '>i4', 3)
+    if any(head != [2, 0, 20]):
```
effigies: Might do `if head not in (0, 2, 20):`

nibabel/freesurfer/io.py
```diff
@@ -45,35 +45,20 @@ def _fread3_many(fobj, n):
...
-        warnings.warn("Volume geometry info is either "
-                      "not contained or not valid.")
-
+        volume_info['head'] = head
+        for key in ['valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras',
+                    'zras', 'cras']:
+            pair = fobj.readline().split('=')
+            if pair[0].strip() != key or len(pair) != 2:
+                raise IOError('Error parsing volume info.')
+            volume_info[pair[0]] = pair[1]
```
effigies: Personal preference: I'd rather `volume_info` use keys without spaces, and then use `.ljust(6)` to get the old spacing back on write. If I wanted to pull out xras, I'd find it more intuitive to do `volume_info['xras']`, not `volume_info['xras  ']`.

agramfort:
travis is not happy

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,52 @@ def _fread3_many(fobj, n):
     return (b1 << 16) + (b2 << 8) + b3
 
 
-def read_geometry(filepath):
+def _read_volume_info(fobj):
+    volume_info = OrderedDict()
+    head = np.fromfile(fobj, '>i4', 3)
+    if any(head != [2, 0, 20]):
```
effigies: I think this will always fail, as the RHS is a list, not a numpy array. Might do `if head not in (0, 2, 20):`

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,52 @@ def _fread3_many(fobj, n):
...
+        warnings.warn("Unknown extension code.")
+    else:
+        volume_info['head'] = head
+        for key in ['valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras',
+                    'zras', 'cras']:
+            pair = fobj.readline().decode('utf-8').split('=')
+            if pair[0].strip() != key or len(pair) != 2:
+                raise IOError('Error parsing volume info.')
+            volume_info[pair[0]] = pair[1]
```
effigies: Personal preference: I'd rather `volume_info` use keys without spaces, and then use `.ljust(6)` to get the old spacing back on write.

If I wanted to pull out xras, I'd find it more intuitive to do `volume_info['xras']`, not `volume_info['xras  ']`.

(Looks like I commented on the wrong diff earlier. Sorry.)

nibabel/freesurfer/io.py
```diff
@@ -86,20 +115,30 @@ def read_geometry(filepath):
...
         elif magic == 16777214:  # Triangle file
-            fobj.readline()  # create_stamp
+            create_stamp = fobj.readline().rstrip(b'\n').decode('utf-8')
             fobj.readline()
             vnum = np.fromfile(fobj, ">i4", 1)[0]
             fnum = np.fromfile(fobj, ">i4", 1)[0]
             coords = np.fromfile(fobj, ">f4", vnum * 3).reshape(vnum, 3)
             faces = np.fromfile(fobj, ">i4", fnum * 3).reshape(fnum, 3)
+
+            volume_info = _read_volume_info(fobj)
```
effigies: I'd make this conditional on `read_metadata`, otherwise we potentially pop up irrelevant warnings.

nibabel/freesurfer/io.py
```diff
@@ -129,6 +170,14 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
         coords.astype('>f4').reshape(-1).tofile(fobj)
         faces.astype('>i4').reshape(-1).tofile(fobj)
 
+        # Add volume info, if given
+        if volume_info is not None and len(volume_info) > 0:
+            for key, val in volume_info.items():
+                if key == 'head':
+                    val.tofile(fobj)
+                    continue
+                fobj.write('{0}={1}'.format(key, val).encode('utf-8'))
```
effigies: Input has spaces around `=`, so to replicate the original file as closely as possible, I'd use `'{0} = {1}'`.

nibabel/freesurfer/io.py
effigies: Oh, I see, this is trying to check list equality. In that case, you'd want `if not np.array_equal(head, [2, 0, 20])`, but consider also [these files](https://github.com/nipy/nibabel/issues/456#issuecomment-223835825), which are just 20.

So maybe:

    if head == 20 or np.array_equal(head, [2, 0, 20]):

nibabel/freesurfer/io.py
effigies: And is the consensus now to simply use strings? At least [at one point](https://github.com/nipy/nibabel/issues/456#issuecomment-223836410) the idea was that `volume_info['*ras']` would be numpy arrays.

nibabel/freesurfer/tests/test_io.py
```diff
@@ -55,21 +56,31 @@ def test_geometry():
     assert_equal(coords.shape[0], faces.max() + 1)
 
     # Test quad with sphere
```
effigies: Also, a little out of topic, but I haven't found `lh.sphere` files in the quad format. I don't think we're actually testing quad files at all.

Eric89GXL: I hit that problem a year or two ago and managed to find one actually, feel free to use it:

https://github.com/mne-tools/mne-testing-data/tree/master/subjects/bert/surf

nibabel/freesurfer/io.py
Eric89GXL: @jaeilepp see my commit to @agramfort's PR, I fixed this issue there

nibabel/freesurfer/tests/test_io.py
effigies: Had a look at that file. It does have metadata at the end, but there is a big chunk (885354 bytes) before it. I'd say it's worth testing the quad file parsing, but not worth grabbing metadata from, since you had to go back to FSv3 to find it. So that can be handled in a different PR.

I think here I'd say just remove the misleading comment.

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,52 @@ def _fread3_many(fobj, n):
     return (b1 << 16) + (b2 << 8) + b3
 
 
-def read_geometry(filepath):
+def _read_volume_info(fobj):
+    volume_info = OrderedDict()
+    head = np.fromfile(fobj, '>i4', 3)
+    if any(head != [2, 0, 20]):
+        warnings.warn("Unknown extension code.")
```
matthew-brett: How about `return volume_info` here so we can get rid of the indentation below?  I think that makes it easier to read, because you don't have to wonder if the code is going to do something else for this case.

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,52 @@ def _fread3_many(fobj, n):
...
+    else:
+        volume_info['head'] = head
+        for key in ['valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras',
+                    'zras', 'cras']:
+            pair = fobj.readline().decode('utf-8').split('=')
```
matthew-brett: It's really utf-8 encoding?  Not say `latin1`?

jaeilepp: No idea. This works for all the data I have.

nibabel/freesurfer/io.py
```diff
@@ -129,6 +181,20 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
...
+                if key == 'head':
+                    np.array(val, dtype='>i4').tofile(fobj)
+                elif key in ('valid', 'filename'):
+                    fobj.write('{0} = {1}\n'.format(key, val).encode('utf-8'))
+                elif key == 'volume':
+                    fobj.write('{0} = {1} {2} {3}\n'.format(
+                        key, val[0], val[1], val[2]).encode('utf-8'))
+                else:
+                    fobj.write('{0} = {1:.4f} {2:.4f} {3:.4f}\n'.format(
+                        key.ljust(6), val[0], val[1], val[2]).encode('utf-8'))
```
jaeilepp: Note that currently this expects the data be in the same format that it was read. So no arbitrary extra data is allowed.

jaeilepp:
Comments addressed.

agramfort:
LGTM

jaeilepp:
I suppose Appveyor failures are unrelated?

agramfort:
yes it seems a conda pb

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,60 @@ def _fread3_many(fobj, n):
...
+            return volume_info
+
+    volume_info['head'] = head
+    for key in ['valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras',
+                'zras', 'cras']:
+        pair = fobj.readline().decode('utf-8').split('=')
+        if pair[0].strip() != key or len(pair) != 2:
+            raise IOError('Error parsing volume info.')
+        if key in ('valid', 'filename'):
+            volume_info[key] = pair[1].strip()
```
matthew-brett: Is there no way that the filename could have trailing whitespace, such as carriage returns?  Worth stripping the `pair` values before starting?

nibabel/freesurfer/io.py
```diff
@@ -129,6 +181,20 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
         coords.astype('>f4').reshape(-1).tofile(fobj)
         faces.astype('>i4').reshape(-1).tofile(fobj)
 
+        # Add volume info, if given
+        if volume_info is not None and len(volume_info) > 0:
```
matthew-brett: Avoid indentation, maybe clearer with:

```
if volume_info is None or len(volume_info) == 0:
    return
```

nibabel/freesurfer/tests/test_io.py
```diff
@@ -54,26 +55,43 @@ def test_geometry():
     assert_equal(0, faces.min())
     assert_equal(coords.shape[0], faces.max() + 1)
 
-    # Test quad with sphere
     surf_path = pjoin(data_path, "surf", "%s.%s" % ("lh", "sphere"))
-    coords, faces = read_geometry(surf_path)
+    warnings.filterwarnings('always', category=DeprecationWarning)
```
matthew-brett: Doesn't this need a `with` block to prevent the warning being ignored outside these tests?

nibabel/freesurfer/io.py
jaeilepp: The values are in effect stripped. See the conditional above.

nibabel/freesurfer/tests/test_io.py
jaeilepp: Oops. This isn't supposed to be here at all.

jaeilepp:
Comments addressed.

nibabel/freesurfer/io.py
```diff
@@ -129,6 +181,22 @@ def write_geometry(filepath, coords, faces, create_stamp=None):
         coords.astype('>f4').reshape(-1).tofile(fobj)
         faces.astype('>i4').reshape(-1).tofile(fobj)
 
+        # Add volume info, if given
+        if volume_info is None or len(volume_info) == 0:
+            return
+
+        for key, val in volume_info.items():
+            if key == 'head':
```
Eric89GXL: should we make sure that the first key is `head` and that it's the value that we expect?

nibabel/freesurfer/io.py
```diff
@@ -86,20 +123,33 @@ def read_geometry(filepath):
...
+                volume_info = _read_volume_info(fobj)
         else:
             raise ValueError("File does not appear to be a Freesurfer surface")
 
     coords = coords.astype(np.float)  # XXX: due to mayavi bug on mac 32bits
-    return coords, faces
 
+    ret = (coords, faces)
+    if read_metadata:
+        if len(volume_info) == 0:
```
Eric89GXL: It's possible to get to this line and have `volume_info = None`, i.e. if `magic != 16777214` and `read_metadata=True` in the call

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,60 @@ def _fread3_many(fobj, n):
...
         nvtx x 3 array of vertex (x, y, z) coordinates
     faces : numpy array
         nfaces x 3 array of defining mesh triangles
+    volume_info : OrderedDict
+        If read_metadata is true, key-value pairs found in the geometry file
+    create_stamp : str
+        If read_stamp is true, the comment added by the program that saved
+        the file
     """
+    volume_info = None
```
Eric89GXL: ...so I think you can just make this `volume_info = OrderedDict()` instead and it will take care of the `None` issue

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,60 @@ def _fread3_many(fobj, n):
...
+
+def read_geometry(filepath, read_metadata=False, read_stamp=False):
     """Read a triangular format Freesurfer surface mesh.
 
     Parameters
     ----------
     filepath : str
         Path to surface file
+    read_metadata : bool
+        Read metadata as key-value pairs
```
Eric89GXL: You should list the valid keys that we support reading, since it's not any arbitrary set.

Eric89GXL:
Otherwise LGTM

nibabel/freesurfer/io.py
```diff
@@ -79,6 +79,16 @@ def read_geometry(filepath, read_metadata=False, read_stamp=False):
...
+        Valid keys:
+            'head' : array of int
+            'valid' : str
+            'filename' : str
+            'volume' : array of int, shape (3,)
+            'voxelsize' : array of float, shape (3,)
+            'xras' : array of float, shape (3,)
+            'yras' : array of float, shape (3,)
+            'zras' : array of float, shape (3,)
+            'cras' : array of float, shape (3,)
```
Eric89GXL: I suspect this will look bad when rendered. Have you checked? This is more likely to work:
```
Valid keys:

    * ...
    * ...

```

nibabel/freesurfer/io.py
effigies: Or, perhaps better, iterate through the original list (`for key in ['valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras', 'zras', 'cras']`). This way we don't depend on the ordering of keys, if for some reason somebody wants to create their own `volume_info` dict.

I also think a strategy for missing values is a good idea. I'd be inclined to refuse to write an invalid chunk (since we'll fail to read it), so I would build the string and only write it on success. Perhaps this chunk could be re-worked into a sub-function like `_serialize_volume_info` that returns a string to write (or `b''` on failure, empty `volume_info`).

jaeilepp:
I just realized that there are some numerical errors because we read the values to arrays, but I suppose it's not a problem. For instance reading:
```
xras=[ -1.00000000e+00   8.56816911e-08   1.49011647e-08]
yras=[  1.15484021e-07   1.57160827e-08  -1.00000000e+00]
zras=[ -1.91852465e-07   1.00000000e+00   6.40284092e-09]
cras=[ -5.27361298   9.03908539 -27.28796387]
```
becomes:
```
xras=[-1.  0.  0.]
yras=[ 0.  0. -1.]
zras=[-0.  1.  0.]
cras=[ -5.273613   9.039085 -27.287964]
```
when written back to a file.

Eric89GXL:
> I just realized that there are some numerical errors because we read the values to arrays

Those values can be stored using double floating point just fine. Maybe you're writing them out in a bad format?

nibabel/freesurfer/io.py
```diff
@@ -369,3 +446,32 @@ def read_label(filepath, read_scalars=False):
...
+        elif key in ('valid', 'filename'):
+            val = volume_info[key]
+            strings.append('{0} = {1}\n'.format(key, val).encode('utf-8'))
+        elif key == 'volume':
+            val = volume_info[key]
+            strings.append('{0} = {1} {2} {3}\n'.format(
+                key, val[0], val[1], val[2]).encode('utf-8'))
+        else:
+            val = volume_info[key]
+            strings.append('{0} = {1:f} {2:f} {3:f}\n'.format(
```
Eric89GXL: This is the problem, try something like `{1:0.9g} ...` and you'll see it's written with higher precision.

jaeilepp:
If I change to general format I get ``cras=[ -5.27361298   9.03908539 -27.28796387]`` changed to ``cras=[ -5.27361   9.03909 -27.288  ]``. The others translate fine. I'm not sure if there's a format that works well for all values.

nibabel/freesurfer/io.py
jaeilepp: Okay, I missed this comment. ``0.10g`` seems to work.

jaeilepp:
Ready for review/merge.

nibabel/freesurfer/tests/test_io.py
```diff
@@ -54,26 +55,50 @@ def test_geometry():
...
                                                 time.ctime())
-        write_geometry(surf_path, coords, faces, create_stamp)
+        volume_info['cras'] = [1., 2., 3.]
+        write_geometry(surf_path, coords, faces, create_stamp, volume_info)
 
-        coords2, faces2 = read_geometry(surf_path)
+        coords2, faces2, volume_info2 = \
+            read_geometry(surf_path, read_metadata=True)
 
+        assert_equal(volume_info2['cras'], volume_info['cras'])
```
Eric89GXL: Now that you have improved the precision in writing, you can use something more complete like:
```
for key in ('xras', 'yras', 'zras', 'cras'):
    assert_allclose(volume_info2[key], volume_info[key], rtol=1e-7, atol=1e-30)
```
This would have failed under the `%f` writing but passes now with the `%0.10g` writing, right?

nibabel/freesurfer/io.py
```diff
@@ -44,24 +45,70 @@ def _fread3_many(fobj, n):
...
+    volume_info['head'] = head
+    for key in ['valid', 'filename', 'volume', 'voxelsize', 'xras', 'yras',
+                'zras', 'cras']:
+        pair = fobj.readline().decode('utf-8').split('=')
+        if pair[0].strip() != key or len(pair) != 2:
+            raise IOError('Error parsing volume info.')
+        if key in ('valid', 'filename'):
+            volume_info[key] = pair[1].strip()
+        elif key == 'volume':
+            volume_info[key] = np.array(pair[1].split()).astype(int)
```
Eric89GXL: But you could still get to this point with `pair[1]` having trailing whitespace, right? In fact it seems like it will have trailing whitespace based on the flow.

Eric89GXL: ...but I guess `.astype(int)` must take care of getting rid of that whitespace for you.

Eric89GXL:
Other than my last (valid) comment, LGTM

nibabel/freesurfer/tests/test_io.py
jaeilepp: In theory yes, but I couldn't find a test file that would have this kind of precision in the end.

nibabel/freesurfer/io.py
jaeilepp: This is now sub-functioned.

nibabel/freesurfer/io.py
jaeilepp: Fix on line 107.

nibabel/freesurfer/io.py
jaeilepp: Yeah, there should be no whitespaces anymore.

jaeilepp:
This is ready for review. Appveyor error seems unrelated.

agramfort:
+1 for merge on my side

Eric89GXL:
+1 for merge from me too

matthew-brett:
Thanks @jaeilepp for your patient work on this, and thanks too to y'all for the careful reviews.

PR #462, opened 2016-06-23
FIX: Copy affine_to_rasmm when slicing a Tractogram object

MarcCote:
This PR simply makes sure to set the `affine_to_rasmm` when creating a new `Tractogram` object that is the result of advanced indexing or slicing. It also adds a test to check it is working properly.

### DIFF
--- a/nibabel/streamlines/tests/test_tractogram.py
+++ b/nibabel/streamlines/tests/test_tractogram.py
@@ -374,7 +374,7 @@ class TestTractogram(unittest.TestCase):
                          DATA['tractogram'].data_per_streamline[::-1],
                          DATA['tractogram'].data_per_point[::-1])
 
-        # Make sure slicing conserve the affine_to_rasmm property.
+        # Make sure slicing conserves the affine_to_rasmm property.
         tractogram = DATA['tractogram'].copy()
         tractogram.affine_to_rasmm = DATA['rng'].rand(4, 4)
         tractogram_view = tractogram[::2]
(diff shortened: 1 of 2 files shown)


nibabel/streamlines/tests/test_tractogram.py
```diff
@@ -373,6 +374,13 @@ def test_tractogram_getitem(self):
                          DATA['tractogram'].data_per_streamline[::-1],
                          DATA['tractogram'].data_per_point[::-1])
 
+        # Make sure slicing conserve the affine_to_rasmm property.
```
matthew-brett: Typo `conserves`

matthew-brett:
Other than tiny typo, looks good, thanks.

nibabel/streamlines/tests/test_tractogram.py
MarcCote: Done

nibabel/streamlines/tests/test_tractogram.py
```diff
@@ -18,6 +18,7 @@
 
 def setup():
     global DATA
+    DATA['rng'] = np.random.RandomState(1234)
```
MarcCote: Unless nosetest sets the seed of np.random I prefer having reproducible tests. In practice, it should work with any seed, so there is no reason really. We also use a seed in `test_array_sequence.py`.

MarcCote:
Anything else @effigies ?

effigies:
Oh, yeah. Sorry. LGTM.

matthew-brett:
Thanks everyone for the input - merging.

PR #464, opened 2016-07-29
MRG: check for MINC using processing routines

matthew-brett:
The processing routines assume that image have spatial axes before time etc
axes, but this may well not be true for MINC images.

Barf on MINC images > 3D for now.  Later maybe we can investigate the image
axis names to work out which axes are spatial.

### DIFF
--- a/nibabel/imageclasses.py
+++ b/nibabel/imageclasses.py
@@ -105,6 +105,13 @@ ext_map = ExtMapRecoder((
     ('par', '.par'),
 ))
 
+# Image classes known to require spatial axes to be first in index ordering.
+# When adding an image class, consider whether the new class should be listed
+# here.
+KNOWN_SPATIAL_FIRST = (Nifti1Pair, Nifti1Image, Nifti2Pair, Nifti2Image,
+                       Spm2AnalyzeImage, Spm99AnalyzeImage, AnalyzeImage,
+                       MGHImage, PARRECImage)
+
 
 def spatial_axes_first(img):
     """ True if spatial image axes for `img` always preceed other axes
@@ -122,4 +129,4 @@ def spatial_axes_first(img):
     """
     if len(img.shape) < 4:
         return True
-    return not isinstance(img, Minc1Image)
+    return type(img) in KNOWN_SPATIAL_FIRST
--- a/nibabel/imageclasses.py
+++ b/nibabel/imageclasses.py
@@ -104,3 +104,22 @@ ext_map = ExtMapRecoder((
     ('mgz', '.mgz'),
     ('par', '.par'),
 ))
+
+
+def spatial_axes_first(img):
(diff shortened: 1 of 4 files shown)


effigies:
So I take it you don't want a `spatial_axes_first` attribute/property in `SpatialImage` classes, since this seems to be a temporary solution? It's a little bit of a footgun, if a new format comes along that should fail this test, given that you need to remember to add a special case to the function. But if it's a short term thing, no big deal.

Might update the doc on `resample_from_to` to require `from_img` to have `shape` attribute, and same for `smooth_img` and `img`.

matthew-brett:
Yes, was planning for this to be a short term thing.  I take your point about the footguniness - but then I guess the choices are either to add the attribute to all images - I guess with False as the default, to prevent false positives, or list all known image types where spatial axes are first, hoping that people with new image types will know to update that list.  What do you think?

matthew-brett:
I added the documentation on required `shape` attribute, thanks for pointing that out.

effigies:
Well, if you're okay with adding to the image class, this would be my first thought:

```Python
class SpatialImage:
    ...
    @property
    def spatial_axes_first(self):
        return len(self.shape) < 4
```

And then let known good formats say:

```Python
class Nifti1Image:
    spatial_axes_first = True
    ...
```

Worth considering whether this should go in the image or the header, but that's the general idea.

BUT, if you're thinking of a more long-term axis-labeling, you might not want to have to keep this API around. In which case the existing solution seems fine.

matthew-brett:
Yes, right - I was hoping to throw this away fairly soon - and avoid adding to the image API.  But of course the definitive fix might not come soon.

matthew-brett:
I made an opt-in list of image classes known to be OK.  It's a bit ugly but I guess it's a little safer.

nibabel/imageclasses.py
```diff
@@ -104,3 +104,29 @@ def __getitem__(self, *args, **kwargs):
...
+
+    Returns
+    -------
+    spatial_axes_first : bool
+        True if image only has spatial axes (number of axes < 4) or image type
+        known to have spatial axes preceeding other axes.
+    """
+    if len(img.shape) < 4:
+        return True
+    return type(img) in KNOWN_SPATIAL_FIRST
```
effigies: Just to be sure, this is meant to require exact classes and exclude subclasses?

matthew-brett: That was what I intended, on the basis that it is possible to imagine a sub-class that doesn't have the same behavior, but maybe that is a bit obscure.

effigies: Nah, that's fine. Just wanted to be sure that was the intended behavior. It makes sense to me to insist on explicit inclusion in the list.

effigies:
LGTM.

matthew-brett:
Thanks for the review.

PR #466, opened 2016-07-29
RF/ENH: remove file prior writing in to_filename

### LINKED ISSUE #465: saving image to a "new" file, which is a symlink... what should happen?
ATM

1. even if intend to write to a "new" file, if file exists and a symlink, it would write into a file pointed by the symlink... (well -- to say the truth -- this behavior is consistent with e.g. how `>` in shell works):

```shell
In [1]: import nibabel as nib

In [2]: !ln -s nonexistent out.nii.gz

In [3]: !ls -l nonexistent out.nii.gz
ls: cannot access 'nonexistent': No such file or directory
lrwxrwxrwx 1 yoh yoh 11 Jul 28 23:29 out.nii.gz -> nonexistent

*In [7]: nib.Nifti1Image(np.arange(12).reshape((2,2,3)), None, None).to_filename('out.nii.gz')

In [8]: !ls -l nonexistent out.nii.gz
-rw------- 1 yoh yoh 96 Jul 28 23:30 nonexistent
lrwxrwxrwx 1 yoh yoh 11 Jul 28 23:29 out.nii.gz -> nonexistent
```

2. if file pointed by a symlink exists but with permissions removed for writing (belongs to other user, or explicitly write bit removed, e.g. under annex):

```ipython

In [9]: !chmod a-w nonexistent

In [10]: nib.Nifti1Image(np.arange(12).reshape((2,2,3)), None, None).to_filename('out.nii.gz')
...
/home/yoh/proj/nipy/nipy-suite/nibabel/nibabel/openers.py in _gzip_open(fileish, *args, **kwargs)
     62 
     63 def _gzip_open(fileish, *args, **kwargs):
---> 64     gzip_file = BufferedGzipFile(fileish, *args, **kwargs)
     65 
     66     # Speedup for #209; attribute not present in in Python 3.5

/usr/lib/python2.7/gzip.pyc in __init__(self, filename, mode, compresslevel, fileobj, mtime)
     92             mode += 'b'
     93         if fileobj is None:
---> 94             fileobj = self.myfileobj = __builtin__.open(filename, mode or 'rb')
     95         if filename is None:
     96             # Issue #13781: os.fdopen() creates a fileobj with a bogus name

IOError: [Errno 13] Permission denied: 'out.nii.gz'
```

I wondered... would it be sensible to suggest to first remove the file intended to be written to (at least in to_filename) before writing?  that would allow to interact with git-annex'ed datasets more easily,

yarikoptic:
Closes #465 

Disclaimer: I am still not sure if that is the most kosher behavior. PR is primarily to test the idea, e.g. if it breaks any tests.

May be such behaviour could be controlled/enabled by some configuration/environment setting (e.g. `NIBABEL_REMOVE_BEFORE_WRITE=1`) so in those cases when it is clearly necessary we could at least easily enable it.

### DIFF
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -8,6 +8,7 @@
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 ''' Common interface for any image format--volume or surface, binary or xml.'''
 
+import os
 import warnings
 
 from .externals.six import string_types
@@ -347,7 +348,15 @@ class FileBasedImage(object):
         -------
         None
         '''
-        self.file_map = self.filespec_to_file_map(filename)
+        self.file_map = file_map = self.filespec_to_file_map(filename)
+        for _, fh in file_map.items():
+            if not isinstance(fh, FileHolder):
+                continue
+            if os.path.exists(fh.filename):
+                # Remove previous file where new file would be saved
+                # Necessary e.g. for cases where file is a symlink pointing
+                # to some non-writable file (e.g. under git annex control)
+                os.unlink(fh.filename)
         self.to_file_map()
 
     def to_filespec(self, filename):


nibabel/filebasedimages.py
```diff
@@ -347,7 +356,16 @@ def to_filename(self, filename):
...
+        remove_env = os.environ.get('NIBABEL_REMOVE_BEFORE_TO_FILENAME', None)
+        for _, fh in file_map.items():
+            if not isinstance(fh, FileHolder):
+                continue
+            if os.path.lexists(fh.filename) and (remove_env or _is_annex_symlink(fh.filename)):
+                # Remove previous file where new file would be saved
+                # Necessary e.g. for cases where file is a symlink pointing
+                # to some non-writable file (e.g. under git annex control)
+                os.unlink(fh.filename)
         self.to_file_map()
```
yarikoptic: or @matthew-brett do you think it would remain kosher to shift this functionality even deeper into to_file_map? ;)

matthew-brett:
This change makes me very nervous.

For example:

```
ln -s link_target.nii link.nii
```

Now consider:

```
python
>>> import numpy as np
>>> import nibabel as nib
>>> nib.save(nib.Nifti1Image(np.ones((2, 3, 4)), None), 'link.nii')
>>> data = img.load('link_target.nii').get_data()
```

This will give a different result after `export NIBABEL_REMOVE_BEFORE_TO_FILENAME=1`, or in a git-annex path.   I'm worried that this introduces significant global state making it less predictable what any given code fragment is going to do.

yarikoptic:
> This will give a different result after export NIBABEL_REMOVE_BEFORE_TO_FILENAME=1, or in a git-annex path.

if by 'different result' you mean that file no longer would be a symlink, that would be the feature not a bug if explicitly requested (env var) or it was a git annex path...  the only possible usecase I see which this would violate in case of git-annex is that may someone would want to rely on the fact that files are under annex and locked (i.e. can't be modified), and does expect script to fail in such a case.  So may be it shouldn't have annex specific handling after all, but imho env variable (or if there was a centralized configuration setting), should be ok, since asks for such behaviour explicitly

matthew-brett:
No, I meant that - when the environment variable is not set, the code above will give you the original contents of `link_target.nii`, but when the env var is set, it gives you the new contents of `link.nii`.  Or is my logic off?

yarikoptic:
ah, that way. yeah, sure... somewhat by design ;)

matthew-brett:
Right - I realize it is by design, but consider this code, where ``bar.nii`` is some image other than ``np.ones((2, 3, 4))``:

```
>>> import numpy as np
>>> import nibabel as nib
>>> nib.save(nib.Nifti1Image(np.ones((2, 3, 4)), None), 'foo.nii')
>>> data = img.load('bar.nii').get_data()
```

Now I have to work out if ``data`` is ``np.ones((2, 3, 4))`` or not.  

Before your change, the answer depends only on whether ``foo.nii`` is a symlink.

Now the answer depends on whether `foo.nii` is a symlink AND whether `NIBABEL_REMOVE_BEFORE_TO_FILENAME` is set.

yarikoptic:
You are writing/reading two different files. Result depends on either first I've is a symlink or not anyways. I could argue that with proposed modification logic would be more straightforward - writing to one Durant change the other (by design)

nibabel/filebasedimages.py
matthew-brett: Yes, to do this properly would mean going down into `to_file_map`, because it's possible to save images with filenames that way, but then, at the moment, you'd have to go into the implementation for each file type (`to_file_map` not implemented by default).

matthew-brett:
If I understand right, git-annex is the biggest (only?) motivation for this change?

Does git-annex symlink to files in the repository?  That seems like it would cause chaos for many operations.  As you said already, this:

```
$ echo "foo" > link_target
$ ln -s link_target link
$ echo "bar" > link
$ cat link_target
bar
```

will replace the contents of `link_target`, as for the current behavior of Python and nibabel.   Does it really make sense to modify nibabel to do something different from the shell?

nibotmi:
:umbrella: The latest upstream changes (presumably #479) made this pull request unmergeable. Please resolve the merge conflicts.

yarikoptic:
Omg, nibotmi buildbot talks to us now? How cute! Or it was a human?

matthew-brett:
A bot - but maybe a bot that just passed the Turing test ...

PR #468, opened 2016-07-30
RF+TST: restore GiftiDataArray num_dim setter

matthew-brett:
We removed the `GiftiDataArray.num_dim` attribute in favor of a
property, but pymvpa was expecting to be able to set the value.  Add a
setter that raises a ValueError if the value is incorrect, and gives a
deprecation warning regardless.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -369,6 +369,17 @@ class GiftiDataArray(xml.XmlSerializable):
     def num_dim(self):
         return len(self.dims)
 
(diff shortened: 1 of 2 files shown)


effigies:
Looks reasonable. If it fixes the PyMVPA tests, I'd say +1 for merge.

yarikoptic:
Thank you!  Meanwhile I will make it compatible with upcoming release.  Will it be 2.0.3 or 2.1 or 3? ;)

matthew-brett:
Next release will be 2.1

Issue #469, opened 2016-08-01
gifti to_xml  TypeError: cannot serialize 0 (type int)


yarikoptic:
happens with 2.0.1-634-g82a261c
After the fixup #468 ran into another failure which I am not sure at which stage should be addressed, so decided just to report with details:

```
======================================================================
ERROR: test_surf_gifti (mvpa2.tests.test_surfing.SurfTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/testing/tools.py", line 278, in newfunc
    func(*(arg + (filename,)), **kw)
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/tests/test_surfing.py", line 1104, in test_surf_gifti
    surf.write(fn, s)
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/support/nibabel/surf.py", line 2229, in write
    surf_gifti.write(fn, s, overwrite=overwrite)
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/support/nibabel/surf_gifti.py", line 256, in write
    xml = to_xml(s, fn)
  File "/home/yoh/proj/pymvpa/pymvpa/mvpa2/support/nibabel/surf_gifti.py", line 219, in to_xml
    xml = img.to_xml().encode('utf-8')
  File "/home/yoh/proj/nipy/nipy-suite/nibabel/nibabel/gifti/gifti.py", line 680, in to_xml
    """ + xml.XmlSerializable.to_xml(self, enc)
  File "/home/yoh/proj/nipy/nipy-suite/nibabel/nibabel/xmlutils.py", line 30, in to_xml
    return tostring(self._to_xml_element(), enc)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1126, in tostring
    ElementTree(element).write(file, encoding, method=method)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 820, in write
    serialize(write, self._root, encoding, qnames, namespaces)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 939, in _serialize_xml
    _serialize_xml(write, e, encoding, qnames, None)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 932, in _serialize_xml
    v = _escape_attrib(v, encoding)
[...]

PR #470, opened 2016-08-03
BF+TST: fix XML output from in-memory Gifti array

matthew-brett:
GiftiDataArrays created in memory (rather than loaded from disk) had
ExternalFileOffset attribute value as an integer, but for valid XML, the
values have to be strings.

Closes gh-469.

### DIFF
--- a/nibabel/gifti/tests/test_parse_gifti_fast.py
+++ b/nibabel/gifti/tests/test_parse_gifti_fast.py
@@ -99,6 +99,44 @@ DATA_FILE5_darr2 = np.array([[6402, 17923, 25602],
 DATA_FILE6_darr1 = np.array([9182740, 9182740, 9182740], dtype=np.float32)
 
 
+def assert_default_types(loaded):
+    default = loaded.__class__()
+    for attr in dir(default):
+        defaulttype = type(getattr(default, attr))
+        # Optional elements may have default of None
+        if defaulttype is type(None):
+            continue
+        loadedtype = type(getattr(loaded, attr))
+        assert loadedtype == defaulttype, \
+            "Type mismatch for attribute: {} ({!s} != {!s})".format(
+                attr, loadedtype, defaulttype)
+
+
+def test_default_types():
+    for fname in datafiles:
+        img = load(fname)
+        # GiftiImage
+        assert_default_types(img)
+        # GiftiMetaData
+        assert_default_types(img.meta)
+        # GiftiNVPairs
+        for nvpair in img.meta.data:
+            assert_default_types(nvpair)
+        # GiftiLabelTable
+        assert_default_types(img.labeltable)
+        # GiftiLabel elements can be None or float; skip
+        # GiftiDataArray
+        for darray in img.darrays:
+            assert_default_types(darray)
+            # GiftiCoordSystem
+            assert_default_types(darray.coordsys)
+            # GiftiMetaData
+            assert_default_types(darray.meta)
+            # GiftiNVPairs
+            for nvpair in darray.meta.data:
+                assert_default_types(nvpair)
+
+
 def test_read_ordering():
     # DATA_FILE1 has an expected darray[0].data shape of (3,3).  However if we
     # read another image first (DATA_FILE2) then the shape is wrong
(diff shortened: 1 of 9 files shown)


yarikoptic:
FWIW confirming that it does resolve #469 for PyMVPA.  Thanks!

matthew-brett:
Any comments here?

effigies:
Definitely fixes the bug, but given that `GiftiDataArray.ext_offset` is defined in the docstring to be an `int`, I'd think [this line](https://github.com/nipy/nibabel/blob/master/nibabel/gifti/parse_gifti_fast.py#L190) should be coercing to an `int`.

Maybe we should have a test to check types on parsed files?

matthew-brett:
Chris - yes, I think you're right.   Any ideas for a good general (or not general) way to check loaded attribute types?

effigies:
Simplest thing I can think of (in about 30 seconds) that isn't too class-specific is something on the order of:

```Python
default = klass()
loaded = klass.read_from_file(fname)
for attr in dir(default):
    assert isinstance(getattr(loaded, attr), type(getattr(default, attr)))
```

Can probably imagine some ways for this to fail, but the other options I can think of would be a lot of work.

yarikoptic:
>  default = klass()
>  loaded = klass.read_from_file(fname)
>  for attr in dir(default):
>      assert isinstance(getattr(loaded, attr), type(getattr(default, attr)))
>    Can probably imagine some ways for this to fail, but the other options I
>    can think of would be a lot of work.

why it shouldn't match type exactly but allow for subclassing?

-- 
Yaroslav O. Halchenko
Center for Open Neuroscience     http://centerforopenneuroscience.org
Dartmouth College, 419 Moore Hall, Hinman Box 6207, Hanover, NH 03755
Phone: +1 (603) 646-9834                       Fax: +1 (603) 646-1419
WWW:   http://www.linkedin.com/in/yarik

matthew-brett:
Chris - looks like a reasonable check to me - any interest in a quick PR?

effigies:
PR against your branch, Matthew. It'll fail the Python 2 checks as noted over there.

nibabel/gifti/tests/test_parse_gifti_fast.py
```diff
@@ -99,6 +99,45 @@
...
+def assert_default_types(loaded):
+    default = loaded.__class__()
+    for attr in dir(default):
+        defaulttype = type(getattr(default, attr))
+        # Optional elements may have default of None
+        if defaulttype is type(None):
+            continue
+        loadedtype = type(getattr(loaded, attr))
+        assert_equal(loadedtype, defaulttype,
+                     "Type mismatch for attribute: {} ({!s} != {!s})".format(
```
effigies: Sorry, forgot about 2.6:

`"Type mismatch for attribute: {0} ({1!s} != {2!s})".format(`

matthew-brett: Actually, now I'm trying to debug the errors on Python 2.6, I think it's time for 2.6 to go.  It's getting really hard to set up testing on 2.6.   OK with you?

effigies: Totally fine with that.

matthew-brett:
Ok - thanks for the feedback - it's certainly a better PR now.  Any more comments?

effigies:
LGTM.

requirements.txt
```diff
@@ -1 +1,8 @@
+# Minumum requirements
+#
+# Check these against
+#   nibabel/info.py
+#   .travis.yml
+#   doc/source/installation.rst
```
yarikoptic: Btw, what do you think about specifying minimal versioned requirements in setup.py, and then allow for structured optional dependencies (see eg how we have it in datalad)? See eg https://caremad.io/2013/07/setup-vs-requirement/ for the point

yarikoptic: Not for this PR but just saw these notes inspired by duplication, so decided to ask ;-)

matthew-brett:
OK to merge this one?

effigies:
:+1:

PR #472, opened 2016-08-06
BF: fix recursion error on Python 3.3

matthew-brett:
Python 3.3 seems to get upset with filling format string values from
byte strings.  See:
http://nipy.bic.berkeley.edu/builders/nibabel-py3.3/builds/240/steps/shell_5/logs/stdio

Fixed by this change - see ``try_branch.py`` run at:

http://nipy.bic.berkeley.edu/builders/nibabel-py3.3/builds/248/steps/shell_5/logs/stdio

### DIFF
--- a/nibabel/streamlines/trk.py
+++ b/nibabel/streamlines/trk.py
@@ -728,6 +728,10 @@ class TrkFile(TractogramFile):
         vars['property_names'] = "\n  ".join([asstr(s)
                                               for s in vars['property_name']
                                               if len(s) > 0])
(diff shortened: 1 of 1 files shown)


MarcCote:
LGTM (I'm curious what is the recursion limit for Python on that machine?)

matthew-brett:
Not sure on the recursion limit, but I think this will leads to infinite recursion, because the format string is failing and then it tries to make an error trace that calls the `str` method again.

MarcCote:
Gotcha.

PR #474, opened 2016-08-06
DOC: better installation instructions

matthew-brett:
Don't advise sudo, it's a terrible idea.

Make clear the difference between a "source" install and a "development"
install.

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -48,11 +48,16 @@ the source from `NiBabel pypi`_ .  Go to the pypi page and select the source
 distribution you want.  Download the distribution, unpack it, and then, from
 the unpacked directory, run::
 
-    python setup.py install
+    pip install .
 
-or (if you need root permission to install on a unix system)::
+If you get permission errors, this may be because ``pip`` is trying to install
+to the system directories.  You can solve this error by using ``sudo``, but we
+strongly suggest you either do an install into your "user" directories, like
+this::
 
-    sudo python setup.py install
+    pip install --user .
+
+or you work inside a virtualenv_.
 
 .. _install_debian:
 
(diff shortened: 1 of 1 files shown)


doc/source/installation.rst
```diff
@@ -48,11 +48,16 @@ the source from `NiBabel pypi`_ .  Go to the pypi page and select the source
...
-or (if you need root permission to install on a unix system)::
+If you get permission errors, this may be because ``pip`` is trying to install
+to the system directories.  You can solve this error by using ``sudo``, but we
+strongly suggest you either do an install into your "user" directories, like
+this::
 
-    sudo python setup.py install
+    pip install --user .
+
+or you work inside a virtualenv_.
```
yarikoptic: +1 fwiw ;-)

doc/source/installation.rst
```diff
@@ -128,8 +124,10 @@ It should look something like this::
     >>>
 
 
-To run the nibabel test suite, from the terminal run ``nosetests nibabel`` or ``python -c "import nibabel; nibabel.test()``.
+To run the nibabel test suite, from the terminal run ``nosetests nibabel`` or
+``python -c "import nibabel; nibabel.test()``.
 
-To run an extended test suite that validates ``nibabel`` for long-running and resource-intensive cases, please see :ref:`advanced_testing`.
+To run an extended test suite that validates ``nibabel`` for long-running and
+resource-intensive cases, please see :ref:`advanced_testing`.
```
yarikoptic: I didn't know... Need to read on ;-)

doc/source/installation.rst
```diff
@@ -66,60 +71,51 @@ their repositories. Once this is done, installing NiBabel is::
...
-*  `Python Imaging Library`_ (for PNG conversion in DICOMFS)
-*  nose_ 0.11 or greater (to run the tests)
-*  sphinx_ (to build the documentation)
-
-Get the sources
----------------
+*  SciPy_ (optional, for full SPM-ANALYZE support)
+*  PyDICOM_ 0.9.7 or greater (optional, for DICOM support)
+*  `Python Imaging Library`_ (optional, for PNG conversion in DICOMFS)
+*  nose_ 0.11 or greater (optionaal, to run the tests)
```
yarikoptic: Typo, optional

matthew-brett:
I'm not sure what to make of the Donald Stufft article - I guess that we should expect `requirements.txt` to be used by developers, mainly.

The datalad idea as I understand it is to add `full` installation group, and then use the local development idea of the `requirements.txt` file to trigger that.

So, yes, maybe.  I wish I had a clearer feeling for how `requirements.txt` is used in practice.  But maybe Donald Stufft is right, that it is really only useful for people writing web apps with very strict requirements for deployment.

yarikoptic:
On Sat, 06 Aug 2016, Matthew Brett wrote:

>    I'm not sure what to make of the Donald Stufft article - I guess that we
>    should expect requirements.txt to be used by developers, mainly.

or more "to pin point specific versions in particular deployments"

>    The datalad idea as I understand it is to add full installation group, and
>    then use the local development idea of the requirements.txt file to
>    trigger that.

yeap.  This way, for instance, pip install nibabel could install the
core of nibabel, and then additional  pieces could be pulled in   pip
install nibabel[dicom], or a full installation with everything
imaginable supported by nibabel  via 'nibabel[full]'

matthew-brett:
OK to leave better requirements handling for a later PR?

yarikoptic:
if you mean "let's merge this one now and improve later with another PR", then +1 on my side ;)

PR #479, opened 2016-08-10
MRG: add routine to deprecate with from/to versions

matthew-brett:
Add deprecation routine that allows you to specify at which version
deprecation began, and at which version deprecation will raise an error.

### DIFF
--- a/nibabel/deprecated.py
+++ b/nibabel/deprecated.py
@@ -1,7 +1,10 @@
-""" Module to help with deprecating classes and modules
+""" Module to help with deprecating objects and classes
 """
 
 import warnings
+
+from .deprecator import Deprecator
+from .info import cmp_pkg_version
 
 
 class ModuleProxy(object):
@@ -63,3 +66,18 @@ class FutureWarningMixin(object):
                       FutureWarning,
                       stacklevel=2)
         super(FutureWarningMixin, self).__init__(*args, **kwargs)
+
+
+class VisibleDeprecationWarning(UserWarning):
+    """ Deprecation warning that will be shown by default
+
+    Python >= 2.7 does not show standard DeprecationWarnings by default:
+
+    http://docs.python.org/dev/whatsnew/2.7.html#the-future-for-python-2-x
+
+    Use this class for cases where we do want to show deprecations by default.
+    """
+    pass
+
+
+deprecate_with_version = Deprecator(cmp_pkg_version)
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -269,14 +269,6 @@ class FileBasedImage(object):
         return klass.from_file_map(file_map)
 
     @classmethod
-    def from_filespec(klass, filespec):
-        warnings.warn('``from_filespec`` class method is deprecated\n'
-                      'Please use the ``from_filename`` class method '
-                      'instead',
-                      DeprecationWarning, stacklevel=2)
-        klass.from_filename(filespec)
-
-    @classmethod
     def from_file_map(klass, file_map):
         raise NotImplementedError
 
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -8,13 +8,12 @@
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 ''' Common interface for any image format--volume or surface, binary or xml.'''
 
-import warnings
-
 from .externals.six import string_types
 from .fileholders import FileHolder
 from .filename_parser import (types_filenames, TypesFilenamesError,
                               splitext_addext)
 from .openers import ImageOpener
+from .deprecated import deprecate_with_version
 
 
 class ImageFileError(Exception):
@@ -212,16 +211,13 @@ class FileBasedImage(object):
         '''
         raise TypeError("Cannot slice image objects.")
 
+    @deprecate_with_version('get_header method is deprecated.\n'
+                            'Please use the ``img.header`` property '
+                            'instead.',
+                            '2.1', '4.0')
     def get_header(self):
         """ Get header from image
-
-        Please use the `header` property instead of `get_header`; we will
-        deprecate this method in future versions of nibabel.
         """
-        warnings.warn('``get_header`` is deprecated.\n'
-                      'Please use the ``img.header`` property '
-                      'instead',
-                      DeprecationWarning, stacklevel=2)
         return self.header
 
     def get_filename(self):
@@ -273,11 +269,11 @@ class FileBasedImage(object):
         raise NotImplementedError
 
     @classmethod
+    @deprecate_with_version('from_files class method is deprecated.\n'
+                            'Please use the ``from_file_map`` class method '
+                            'instead.',
+                            '1.0', '3.0')
     def from_files(klass, file_map):
-        warnings.warn('``from_files`` class method is deprecated\n'
-                      'Please use the ``from_file_map`` class method '
-                      'instead',
-                      DeprecationWarning, stacklevel=2)
         return klass.from_file_map(file_map)
 
     @classmethod
@@ -318,11 +314,11 @@ class FileBasedImage(object):
         return file_map
 
     @classmethod
+    @deprecate_with_version('filespec_to_files class method is deprecated.\n'
+                            'Please use the "filespec_to_file_map" class '
+                            'method instead.',
+                            '1.0', '3.0')
     def filespec_to_files(klass, filespec):
-        warnings.warn('``filespec_to_files`` class method is deprecated\n'
-                      'Please use the ``filespec_to_file_map`` class method '
-                      'instead',
-                      DeprecationWarning, stacklevel=2)
         return klass.filespec_to_file_map(filespec)
 
     def to_filename(self, filename):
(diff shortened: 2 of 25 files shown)


nibabel/deprecated.py
```diff
@@ -63,3 +77,121 @@ def __init__(self, *args, **kwargs):
...
+            expired_class,
+            until))
+    message = '\n'.join(messages)
+
+    def deprecator(func):
+
+        @functools.wraps(func)
+        def deprecated_func(*args, **kwargs):
+            if expired:
+                raise expired_class(message)
```
yarikoptic: Isn't better place for this check within outer decorator itself? So you don't wait for it to be used but indicate to developers/yourself that it is time to really remove it completely?

matthew-brett: My worry is that someone (presumably me) will forget to remove them in tim, and some code will get out into the wild in that state.   In that case, then the error will only occur if they try and use the functions that they shouldn't be using, instead of completely breaking the module import.

I think all these functions have their deprecations checked, so they should generate a test error when they expire.

yarikoptic: That was kinda the point - now codebase has methods which should have been gone awhile back. Since bookings is using them, nothing is removed. As long as you run tests on 'to be released' version (easy to do by just pushing a PR with to be released changes), you will know that those functions must be removed. I guess I will also use it eg in duecredit where releases are done from successfully built tagged pushes to github. If build/release would fail - I would know that I had to remove some

yarikoptic: But overall out is up to you of cause! Just kinda defeats the original cause/purpose of my whining ;-)

matthew-brett: Right - but don't forget - tests will fail as soon as the version goes over the version where deprecation ends. So, even if I release a version where I forgot to remove the function, as soon as I start the next development version, on the commit after the release, the tests will fail.  That way I get to see that I forgot, without giving the user code which won't import.

yarikoptic: Ok. That is if you do have a test for every deprecated function. I didn't look at coverage of those functions, but overall coverage is not 100% (no pun intended, coverage is exemplary high) so some might be sneaking in.
But that is again is all fine with me ;-)

yarikoptic:
I guess note it is time also to annotate those which issue deprecation warning already. Paired within this PR would provide usecase testing ;-)

matthew-brett:
Just doing that now - should be done in an hour or so.

matthew-brett:
OK - I believe all deprecation warnings are covered by the tests, and are versioned.

nibabel/filebasedimages.py
```diff
@@ -212,16 +211,16 @@ def __getitem__(self):
         '''
         raise TypeError("Cannot slice image objects.")
 
+    @deprecate_with_version('get_header method is deprecated.\n'
+                            'Please use the ``img.header`` property '
+                            'instead.',
+                            '2.1', '4.0')
```
effigies: This was at least deprecated in the docs at [2.0.0](https://github.com/nipy/nibabel/blob/2.0.0/nibabel/spatialimages.py#L25). Is this meant to indicate the first release a warning was raised?

effigies:
Should be deprecating [SpatialImage.get_affine](https://github.com/matthew-brett/nibabel/blob/26e1ab4b7172865867ae13a4e0222317fb9145ba/nibabel/spatialimages.py#L632).

effigies:
Weird: I'm seeint `EcatImage` [overrode](https://github.com/matthew-brett/nibabel/blob/26e1ab4b7172865867ae13a4e0222317fb9145ba/nibabel/ecat.py#L848-L850) `SpatialImage.from_filespec`, so it was not raising deprecation warnings. Probably should be now?

Maybe it should be removed on a shorter time-scale than 4.0, though, since it's pretty unlikely people would be using that at this point.

nibabel/spatialimages.py
```diff
@@ -612,14 +612,15 @@ def uncache(self):
...
         return self._dataobj.shape
 
+    @deprecate_with_version('get_shape method is deprecated.\n'
+                            'Please use the ``img.shape`` property '
+                            'instead.',
+                            '1.2', '3.0')
     def get_shape(self):
         """ Return shape for image
 
         This function deprecated; please use the ``shape`` property instead
```
effigies: I think `deprecate_with_version` adds to the docstring?

nibabel/filebasedimages.py
matthew-brett: Yes, I was assuming that no-one was reading the docs, and so dated the deprecation from the first warning.

nibabel/spatialimages.py
matthew-brett: Good point.

matthew-brett:
Good catch with `from_filespec` in ECAT.   I am sure few people are using it, but I'd prefer to be gentle in removing, there isn't much cost to keeping it another version.

nibabel/filebasedimages.py
```diff
@@ -212,16 +211,16 @@ def __getitem__(self):
...
     def get_header(self):
         """ Get header from image
 
         Please use the `header` property instead of `get_header`; we will
         deprecate this method in future versions of nibabel.
```
effigies: Another docstring warning.

nibabel/loadsave.py
```diff
@@ -124,8 +124,10 @@ def save(img, filename):
     converted.to_filename(filename)
 
 
-@np.deprecate_with_doc('Please use ``img.dataobj.get_unscaled()`` '
-                       'instead')
+@deprecate_with_version('read_img_data deprecated.'
+                        'Please use ``img.dataobj.get_unscaled()`` instead.'
+                        '2.0.1',
+                        '4.0')
 def read_img_data(img, prefer='scaled'):
```
effigies: This function also has a docstring deprecation notice.

nibabel/parrec.py
```diff
@@ -822,6 +823,9 @@ def _get_unique_image_prop(self, name):
                               'not suppported.'.format(name, props))
         return props[0]
 
+    @deprecate_with_version('get_voxel_size deprecated. '
+                            'Please use "get_zooms" instead.',
+                            '2.0', '4.0')
     def get_voxel_size(self):
```
effigies: Docstring deprecation notice.

matthew-brett:
Good spots - keep 'em coming.

effigies:
LGTM.

matthew-brett:
I refactored the deprecation warning machinery - now in `nibabel/deprecator.py` - any chance of another quick review from the assembly?

nibabel/deprecated.py
```diff
@@ -63,3 +66,18 @@ def __init__(self, *args, **kwargs):
...
+    Python >= 2.7 does not show standard DeprecationWarnings by default:
+
+    http://docs.python.org/dev/whatsnew/2.7.html#the-future-for-python-2-x
+
+    Use this class for cases where we do want to show deprecations by default.
+    """
+    pass
+
+
+deprecate_with_version = Deprecator(cmp_pkg_version)
```
effigies: Did you want the warning class to be `VisibleDeprecationWarning`?

matthew-brett: I wasn't planning on it.  Numpy uses these for deprecations where the user is probably making a coding error.  We could promote all our deprecations, but that might be noisy.

effigies: Gotcha. Just didn't see where it was being used. I'll have a more thorough review of the whole thing, now I'm awake again.

nibabel/info.py
```diff
@@ -19,6 +23,51 @@
                               _version_micro,
                               _version_extra)
 
+
+_VERSION_RE = re.compile(r'([0-9.]*\d+)(.*)')
```
effigies: This doesn't seem to be used anywhere? (It's not the same regex as on L33...)

nibabel/info.py
```diff
@@ -19,6 +23,51 @@
...
+        lower.
+    """
+    version, extra = _parse_version(version_str)
+    pkg_version, pkg_extra = _parse_version(pkg_version_str)
+    if version != pkg_version:
+        return _cmp(StrictVersion(version), StrictVersion(pkg_version))
+    return (0 if extra == pkg_extra
+            else 1 if extra == ''
+            else -1 if pkg_extra == ''
+            else _cmp(extra, pkg_extra))
```
effigies: Looks like you're planning for extras like `'a'` and `'b'`? As long as you don't do `'-rc1'` or anything (where `'3.0-rc1'` should be <  `'3.0'`) this seems fine.

effigies: Wait, just saw the tests. Ignore that.

nibabel/tests/test_info.py
```diff
@@ -0,0 +1,47 @@
...
+                                       ('1.1', '1.1dev', 1),
+                                       ('1.1dev', '1.1', -1),
+                                       ('1.2.1', '1.2.1rc1', 1),
+                                       ('1.2.1rc1', '1.2.1', -1),
+                                       ('1.2.1rc1', '1.2.1rc', 1),
+                                       ('1.2.1rc', '1.2.1rc1', -1),
+                                       ('1.2.1rc1', '1.2.1rc', 1),
+                                       ('1.2.1rc', '1.2.1rc1', -1),
+                                       ('1.2.1b', '1.2.1a', 1),
+                                       ('1.2.1b', '1.2.1a', 1),
```
effigies: Repeat line. Probably meant to swap?

Considerations:

```
('1.2.1a', '1.2.1rc1')
('1.2.1a', '1.2.1')
```

My assumption would be `1.2.1rc1 < 1.2.1 < 1.2.1a`, which is definitely not true with this comparator.

nibabel/tests/test_deprecated.py
```diff
@@ -3,10 +3,26 @@
...
+from nibabel.deprecated import (ModuleProxy, FutureWarningMixin,
+                                deprecate_with_version)
 
-from ..deprecated import ModuleProxy, FutureWarningMixin
+from nose.tools import (assert_true, assert_equal)
+
+from nibabel.tests.test_deprecator import TestDeprecatorFunc as _TestDF
+
+
+_PKG_VERSION = info.__version__
```
effigies: Unused.

effigies:
Review complete.

nibabel/tests/test_info.py
matthew-brett: With this comparator, I'm assuming that any letter suffix to a version, makes the version earlier than the same version without the letter suffix.  So '-rc1' 'a' 'b' 'dev' all make the version earlier.  In particular, I'd take 'a' to mean 'alpha' and 'b' to mean 'beta'.  Of course someone could try and use 'post' as a suffix, and expect that to be later, but we haven't done that yet, and so I was hoping to ignore that.

nibabel/info.py
matthew-brett: Thanks, removed.

matthew-brett:
Thanks for the review - I've extended the docstrings / comments a bit to make it clearer what I was trying to do.

effigies:
Got it. LGTM.

matthew-brett:
Thanks a lot for the review.

PR #485, opened 2016-08-31
RF+TST: allow dtype specifiers as fileslice input

matthew-brett:
Fileslice function previously insisted on np.dtype object as dtype
specifier.  Allow any dtype specifier.


effigies:
LGTM.

matthew-brett:
Thanks for the review.

PR #391, opened 2015-12-03
Streamlines API

MarcCote:
This is take two at providing a general API to manage different streamlines file format (the code can all be found in `nibabel.streamlines.`). This is related to this somewhat outdated [BIAP5](https://github.com/nipy/nibabel/wiki/BIAP5). I initially wanted to make two PRs but as @Garyfallidis [pointed out](https://github.com/nipy/nibabel/pull/243#issuecomment-160810062), without the entire PR "it will make it difficult to try this with real data".

This new streamline API is divided in two main parts:

1. Representation of streamlines in memory (`Tractogram` and `ArraySequence` classes)
2. Loading and saving streamlines from/to different file format (`TractogramFile` class)

*We use the term 'Tractogram' to refer to a set of streamlines, usually the output of a tractography algorithm.*

streamlines_api

#### ArraySequence
The `ArraySequence` class needs to be explained first as it is at the heart of the streamlines representation. I guess it could be seen as a sort of jagged/ragged array to store a list of 2D arrays that have different length but a same number of dimensions (e.g. a list of list of points). For efficiency, we store a list of streamlines contiguously in memory using what we call a `ArraySequence`. This was needed as @Garyfallidis and I realized that a list of numpy arrays has too much overhead in term of memory when handling >1M streamlines (loading a tractogram of 2.5Gb could take up to 5Gb in RAM!).

A `ArraySequence` behaves exactly like a `list` of `2D ndarray`. That is, it supports appending, extending, indexing (supports slicing too), iterating, etc. as a `list` object. For example, indexing a `ArraySequence` object will return a 2D `ndarray`, as one would expect, representing the 3D points of the selected streamline. Note that, while `ArraySequence` does not support deletion (i.e. no `del` nor `remove`), one can efficiently use slicing to achieve the same goal.

A `ArraySequence` object can be instantiated using any iterable yielding `2D array`:
```
import numpy as np
import nibabel as nib
data = [np.random.rand(10, 3), np.random.rand(3, 3)]
seq = nib.streamlines.ArraySequence(data)
```

Slicing a `ArraySequence` object returns a new `ArraySequence` object acting as a view (*no memory duplication*). Of course, one can use the `copy` method to "force" the memory duplication.

#### Tractogram
This part handles streamlines in memory using the Python class `Tractogram`. This class provides a mechanism to keep data associated to points of every streamline (`data_per_point` property). It can also keep data associated to each streamline (`data_per_streamline` property). 

[...]

PR #486, opened 2016-09-13
ENH: Support TCK streamlines file format 

MarcCote:
This PR adds support for MRtrix streamlines file to the Streamlines API (#391).

Eventually, in a future PR, I will add some functions that allows easy conversion between all different formats. For now, going from TRK to TCK is easy. See
https://gist.github.com/MarcCote/ea6842cc4c3950f7596fc3c8a0be0154

Converting from TCK to TRK requires additional information from a reference anatomy. See
https://gist.github.com/MarcCote/ebf0ae83a9202eb96d1dfbb949cd98af

EDIT: Added links to some Gist code for streamlines conversion.


MarcCote:
@jchoude @arnaudbore @matthew-brett @FrancoisRheaultUS @samuelstjean 
This is not WIP anymore, is anyone up for a review? Thanks.

samuelstjean:
Can have a quick look after ismrm deadline.

arnaudbore:
So far:
- [x] Load
- [x] Save
- [x] Conversion (tck->trk and trk->tck) 

**Be really carefull with the software that you use to visualise your data**

Everything works perfectly (Great job @MarcCote)
I think it would be great to have an input from @jdtournier to check if we are missing something. Can you check as well the [the world space](https://github.com/nipy/nibabel/pull/486/files#diff-14a106e7412b371d3c333051564ccd3eR39) and the associated reference ?

Again, great job, very helpful !

jdtournier:
Just had a very brief look into this (why did you have to pick the week before ISMRM deadline?!?). I can't verify whether your handling of the coordinate system matches the expectations of the format, but yes, the positions are expected to be given in world/scanner/real coordinates in mm units, assuming the standard NIfTI coordinate system. Assuming that's what you mean by the 'world space' (`RAS+`?), then that should be fine. 

Only one minor thing to add: we tend to set the offset to the data such that it guarantees memory alignment for 32-bit floats when accessed via memory-mapping, just in case that affects performance (a lot of CPU optimisations rely on memory alignment). That means it's set to the lowest multiple of 4 bytes that exceeds your computed offset. It doesn't look like that's what you're doing? 

All good otherwise, but like I said, I've only skimmed through the code...

MarcCote:
@arnaudbore @jdtournier: Thank you both for your time (especially with ISMRM deadline closing in!).

@jdtournier: You are correct there is no memory alignment right now but I'll fix that and ping you once it is done (after ISMRM deadline).

@jdtournier: Is there any convention in MRtrix to save information alongside streamlines and streamlines' coordinates? For instance, some information we might want to keep are: the mean FA (or a bunch of metrics) for each streamline, some ID/name of the bundle a streamline belongs to, or simply a different RGB color for each point of the streamlines? In Trackvis/.trk those are known as `scalars` et `properties` but in this API we used respectively the terms `data_per_point` and `data_per_streamline`. 

@jdtournier: Regarding the 'world space `RAS+`, I'm using the following naming convention http://nipy.org/nibabel/coordinate_systems.html#naming-reference-spaces. So, yes the coordinates are in mm units and follow the standard NIfTI coordinate system. Also, it is worth mentioning that the streamline coordinate (0,0,0) refers to the center of the voxel whereas in Trackvis/.trk it refers to the corner of the voxel. Is it how streamlines are displayed in MRview?

When saving a TCK file, I make sure the streamlines are in ras+mm (world/scanner/real coordinates). See https://github.com/nipy/nibabel/pull/486/files#diff-14a106e7412b371d3c333051564ccd3eR205

jdtournier:
OK, good to hear wrt the alignment issue. It's not essential though, it's just that it _might_ make a difference it terms of performance, but I have to admit that so far, I don't think I've noticed any particular issues on that front. It's probably more relevant for images.

As to saving per vertex/streamline information, this is something that we discussed extensively a while back. We opted for storing this information separately in associated files, either _track scalar files_ (for the per-vertex case, e.g. as might be produced by [tcksample](http://mrtrix.readthedocs.io/en/latest/reference/commands/tcksample.html)) or straight ASCII files (for the per-streamline case, e.g. as used for the [SIFT2 weights](http://mrtrix.readthedocs.io/en/latest/reference/commands/tcksift2.html)). We figured that maintaining compatibility with the simple existing format, and the flexibility of separating the two types of information to be sufficient reasons _not_ to complicate the format (for example, we might want to have multiple per-vertex scalar information for the same streamlines, in which case storing everything in a single file means duplicating most of the contents of the file needlessly, etc). So if you want to add support for importing per vertex/streamline scalars, that's how we handle it in MRtrix3...

Just checked your definitions, and that matches my expectations. However, whether the origin of the coordinate system corresponds to the centre or corner of the voxel is unrelated to the streamlines storage - that really is about the _image_ storage convention. On that front, we also assume that the voxel position at (0,0,0) corresponds to the centre of the voxel, that's the standard adopted in both DICOM and NIfTI, and that's how images are displayed in MRView. But by design, in MRtrix, the streamlines are stored in scanner coordinates to make them _independent of whatever image was used to generate them_. So I'm slightly puzzled by your question. In MRtrix, the _world_ coordinate (0,0,0) is the origin of the world/scanner coordinate system, _irrespective_ of whether you're taking about streamlines or images, and whatever image you might be talking about. A given image's transformation matrix is what tells you how to figure out the position of its voxel centres in the world/scanner coordinate system. But it is absolutely _not_ the case that the "streamline coordinate (0,0,0) refers to the center of the voxel" - it refers to the isocentre of the scanner, which will rarely be aligned with any particular voxel. Just making sure we're on the same page...?

matthew-brett:
Where are we with this one?

Needs a rebase and PEP8 fixes at some point.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+                   "count", "datatype", "file"]  # Fields being replaced.
+
+        lines = []
+        lines.append(header[Field.MAGIC_NUMBER])
+        lines.append("count: {0:010}".format(header[Field.NB_STREAMLINES]))
+        lines.append("datatype: Float32LE")  # Always Float32LE.
+        lines.extend(["{0}: {1}".format(k, v)
+                      for k, v in header.items() if k not in exclude and not k.startswith("_")])
+        lines.append("file: . ")  # Manually add this last field.
+        out = "\n".join((asstr(line).replace('\n', '\t') for line in lines))
```
matthew-brett: Won't all the lines be strings already?

Why not `"\t".join(lines)` - are you expecting other `\n` characters that need replacing?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+        fileobj.write(asbytes(out))
+
+        # Compute offset to the beginning of the binary data.
+        # We add 5 for "\nEND\n" that will be added just before the data.
+        tentative_offset = len(out) + 5
```
matthew-brett: Logic a bit tough to follow, and it's a bit hard to think about the assumptions that went into the `+1` case below.  How about:

```
hdr_len_no_offset = len(out) + 5
# Need to add number of bytes to store offset as decimal string. We start with estimate without string,
# then update if the offset-as-decimal-string got longer after adding length of the offset string. 
new_offset = -1
old_offset = hdr_len_no_offset
while new_offset != old_offset:
    old_offset = new_offset
    new_offset = hdr_len_no_offset + len(str(old_offset))
```

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+        offset = tentative_offset + len(str(tentative_offset))
+
+        # The new offset might need one more character to write it in ASCII.
+        # e.g. offset = 98 (i.e. 2 char.), so offset += 2 = 100 (i.e. 3 char.)
+        # thus the final offset = 101.
+        if len(str(tentative_offset)) != len(str(offset)):
+            offset += 1  # +1, we need one more character for that new digit.
+
+        fileobj.write(asbytes(str(offset) + "\n"))
+        fileobj.write(asbytes(b"END\n"))
```
matthew-brett: Already bytes.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+            raise HeaderError("Missing 'file' attribute in TCK header.")
+
+        if hdr['file'].split()[0] != '.':
+            msg = ("TCK only supports single-file - in other words the"
+                   " filename part must be specified as '.' but '{}' was"
+                   " specified.").format(hdr['file'].split()[0])
+            raise HeaderError("Missing 'file' attribute in TCK header.")
+
+        # Set endianness and _dtype attributes in the header.
+        hdr[Field.ENDIANNESS] = '<'
```
matthew-brett: How about:

```
hdr[Field.ENDIANNESS] = '>' if hdr['datatype'].endswith('BE') else '<'
```

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+            # Set the file position at the beginning of the data.
+            f.seek(header["_offset_data"], os.SEEK_SET)
+
+            eof = False
+            buff = b""
+            pts = []
+
+            i = 0
+
+            while not eof or not np.all(np.isinf(pts)):
```
matthew-brett: I don't understand the `not np.all(np.isinf(pts))` part of this - can you explain?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+
+                # Make sure we've read enough to find a streamline delimiter.
+                if len(idx_nan) == 0:
+                    # If we've read the whole file, then fail.
+                    if eof and not np.all(np.isinf(pts)):
+                        msg = ("Cannot find a streamline delimiter. This file"
+                               " might be corrupted.")
+                        raise DataError(msg)
+
+                    # Otherwise read a bit more.
```
matthew-brett: What will happen if `eof == True` and `np.all(np.isinf(pts)) == True`?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+
+            while not eof or not np.all(np.isinf(pts)):
+
+                if not eof:
+                    bytes_read = f.read(buffer_size)
+                    buff += bytes_read
+                    eof = len(bytes_read) == 0
+
+                # Read floats.
+                pts = np.frombuffer(buff, dtype=dtype)
```
matthew-brett: Sorry for my confusion, but won't this code re-cast the first block of data over and over, for each iteration of the while loop?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,440 @@
...
+            f.seek(beginning, os.SEEK_SET)
+            self._write_header(f, header)
+
+    @staticmethod
+    def _write_header(fileobj, header):
+        """ Write TCK header to file-like object.
+
+        Parameters
+        ----------
+        fileobj : file-like object
```
MarcCote: Good point.

nibabel/streamlines/tck.py
matthew-brett: Bear in mind that generally nobody reads warnings.  I guess, raise an error if you can't work out what the user meant to do, or if they may be using the API incorrectly, but just warn if there are minor defects in the input.

matthew-brett:
@MarcCote - where are we with this one?

MarcCote:
@matthew-brett sorry for the delay. I'll have more time in the following weeks :).

nibabel/streamlines/tck.py
MarcCote: In the TCK format, the end of file is defined as `inf inf inf`. In my code, the variable `eof` indicates if there are data left to read (i.e. put in the buffer) in the file.

nibabel/streamlines/tck.py
MarcCote: Only a small portion will get re-casted. It would correspond to the last sequence of points that is partially contained in `buff`.
In short, the code is doing the following.
1. Read a chunk of data from the file (a buffer).
2. Determine how many points we've read (i.e. nb. of triples [NaN, NaN, NaN])
3.Yield points, one streamline at the time.
4. Flush the beginning of the buff (that is up to the last NaN triple we've detected).
5. Goto step 1 (making sure we concatenate the new buffer to what is remaining of the current buffer) until we detect a Inf triple (or there is nothing to read).

nibabel/streamlines/tck.py
MarcCote: Then the while loop should terminate. It means the only thing that is left in the buffer is the Inf triple which indicates the end of the TCK file. Did I miss something?

MarcCote:
@matthew-brett I forgot to ping you. I rebased and addressed your comments. It is ready for another round of review when you have time.

nibabel/streamlines/tck.py
matthew-brett: Add a comment here with that information (about inf, inf, inf)?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+        -------
+        header : dict
+            Metadata associated with this tractogram file.
+        """
+        # Record start position if this is a file-like object
+        start_position = fileobj.tell() if hasattr(fileobj, 'tell') else None
+
+        with Opener(fileobj) as f:
+            # Read magic number
+            magic_number = asstr(f.fobj.readline()).strip()
```
matthew-brett: Can't you leave this as bytes?   You still have `strip` and `endswith`.

nibabel/streamlines/tck.py
matthew-brett: Ah - I see - you are dropping the read part of buff at the end of the loop.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+
+                # Read floats.
+                pts = np.frombuffer(buff, dtype=dtype)
+
+                # Convert data to little-endian if needed.
+                if dtype != '<f4':
+                    pts = pts.astype('<f4')
+
+                pts = pts.reshape([-1, 3])
+                idx_nan = np.arange(len(pts))[np.isnan(pts[:, 0])]
```
matthew-brett: Or:

```
idx_nan, = np.nonzero(np.isnan(pts[:, 0]))
```

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+
+                # Make sure we've read enough to find a streamline delimiter.
+                if len(idx_nan) == 0:
+                    # If we've read the whole file, then fail.
+                    if eof and not np.all(np.isinf(pts)):
```
matthew-brett: Can the last streamline be delimited by the `inf, inf, inf` point?

nibabel/streamlines/tck.py
matthew-brett: No I don't think you missed anything - I see now that we'll go back to the while condition, then exit the loop.

matthew-brett:
See refactoring suggestion over at https://github.com/MarcCote/nibabel/pull/12

nibabel/streamlines/tests/test_tck.py
```diff
@@ -0,0 +1,159 @@
...
+
+    DATA['empty_tck_fname'] = pjoin(data_path, "empty.tck")
+    # simple.tck contains only streamlines
+    DATA['simple_tck_fname'] = pjoin(data_path, "simple.tck")
+    DATA['simple_tck_big_endian_fname'] = pjoin(data_path,
+                                                "simple_big_endian.tck")
+    # standard.tck contains only streamlines
+    DATA['standard_tck_fname'] = pjoin(data_path, "standard.tck")
+
+    DATA['streamlines'] = [np.arange(1*3, dtype="f4").reshape((1, 3)),
```
MarcCote: Right, after further reading (http://stackoverflow.com/questions/18171560/does-pep-require-whitespace-around-operators-in-function-arguments and https://www.python.org/dev/peps/pep-0008/) it seems it would better with spaces. I will modify it. Thanks.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+        ----------
+        tractogram : :class:`Tractogram` object
+            Tractogram that will be contained in this :class:`TckFile`.
+        header : dict (optional)
+            Metadata associated to this tractogram file.
+
+        Notes
+        -----
+        Streamlines of the tractogram are assumed to be in *RAS+*
+        and *mm* space where coordinate (0,0,0) refers to the center
```
MarcCote: Yes, but I don't know how to really phrase it. Any suggestion?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+                    yield TractogramItem(pts, {}, {})
+
+            tractogram = LazyTractogram.from_data_func(_read)
+
+        else:
+            tck_reader = cls._read(fileobj, hdr)
+            streamlines = ArraySequence(tck_reader)
+            tractogram = Tractogram(streamlines)
+
+        # By definition.
```
MarcCote: Correct.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+                       " alongside points. Dropping: {}".format(keys))
+                warnings.warn(msg, DataWarning)
+
+            # Make sure streamlines are in rasmm.
+            tractogram = self.tractogram.to_world(lazy=True)
+
+            for t in tractogram:
+                s = t.streamline.astype(dtype)
+                data = np.r_[s, self.FIBER_DELIMITER]
+                f.write(asbytes(data.tostring()))
```
MarcCote: You are right, I'll remove the `asbytes`.

nibabel/streamlines/tck.py
MarcCote: I tried leaving it as bytes, but it is failling at line 251 (`out = "\n".join(lines)`) in `_write_header`.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+    FIBER_DELIMITER = np.array([[np.nan, np.nan, np.nan]], '<f4')
+    EOF_DELIMITER = np.array([[np.inf, np.inf, np.inf]], '<f4')
+
+    def __init__(self, tractogram, header=None):
+        """
+        Parameters
+        ----------
+        tractogram : :class:`Tractogram` object
+            Tractogram that will be contained in this :class:`TckFile`.
+        header : dict (optional)
```
MarcCote: No, you are right, it doesn't have to. But maybe the header should be rethinked at some point.

nibabel/streamlines/tck.py
matthew-brett: Try `b"\n".join` ... ?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,444 @@
...
+        # Build header dictionary from the buffer.
+        hdr = dict(item.split(': ') for item in buf.rstrip().split('\n')[:-1])
+        hdr[Field.MAGIC_NUMBER] = magic_number
+
+        # Check integrity of TCK header.
+        if 'datatype' not in hdr:
+            raise HeaderError("Missing 'datatype' attribute in TCK header.")
+
+        if not hdr['datatype'].startswith('Float32'):
+            msg = ("TCK only supports float32 dtype but 'dataype: {}' was"
```
jchoude: dataype -> datatype

nibabel/streamlines/tests/test_streamlines.py
```diff
@@ -240,12 +244,17 @@ def test_save_complex_file(self):
...
 
                     # If streamlines format does not support saving data
-                    # per point or data per streamline, a warning message
+                    # per point or data per streamline, warning messages
                     # should be issued.
-                    if not (cls.SUPPORTS_DATA_PER_POINT and
-                            cls.SUPPORTS_DATA_PER_STREAMLINE):
-                        assert_equal(len(w), 1)
-                        assert_true(issubclass(w[0].category, Warning))
+                    nb_expected_warnings = 0
```
MarcCote: I'll go with the following as I found the "for loop" a bit overkill for two elements.
```
nb_expected_warnings = \
    ((not cls.SUPPORTS_DATA_PER_POINT) +
    (not cls.SUPPORTS_DATA_PER_STREAMLINE))
```

nibabel/streamlines/tck.py
MarcCote: I tried that, but doing so it is causing faillures elsewhere. :-/

matthew-brett: What failures?

MarcCote: I think it is because everything is in string, here:
```
        lines = []
        lines.append(header[Field.MAGIC_NUMBER])
        lines.append("count: {0:010}".format(header[Field.NB_STREAMLINES]))
        lines.append("datatype: Float32LE")  # Always Float32LE.
        lines.extend(["{0}: {1}".format(k, v)
                      for k, v in header.items()
                      if k not in exclude and not k.startswith("_")])
        lines.append("file: . ")  # Manually add this last field.
        out = b"\n".join(lines)
```

MarcCote:
```
======================================================================
ERROR: test_write_simple_file (nibabel.streamlines.tests.test_tck.TestTCK)
----------------------------------------------------------------------
  File "/data/research/src/nibabel/nibabel/streamlines/tests/test_tck.py", line 115, in test_write_simple_file
    tck.save(tck_file)
  File "/data/research/src/nibabel/nibabel/streamlines/tck.py", line 184, in save
    self._write_header(f, header)
  File "/data/research/src/nibabel/nibabel/streamlines/tck.py", line 251, in _write_header
    out = b"\n".join(lines)
TypeError: sequence item 0: expected a bytes-like object, str found
```

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,444 @@
...
+            buf = asstr(f.fobj.readline())
+            while not buf.rstrip().endswith("END"):
+                buf += asstr(f.fobj.readline())
+
+        # Build header dictionary from the buffer.
+        hdr = dict(item.split(': ') for item in buf.rstrip().split('\n')[:-1])
+        hdr[Field.MAGIC_NUMBER] = magic_number
+
+        # Check integrity of TCK header.
+        if 'datatype' not in hdr:
```
jchoude: Is this really mandatory? Might not be the best comparison, but for example your old TractConverter tool only wrote the ```count``` field. All other fields were discarded.

I think it is useful to have it, but should you really reject streamlines created in another tool that don't have it?

nibabel/streamlines/tck.py
matthew-brett: Any reason for the content of the lines to be strings?  Can they be bytes instead?

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,444 @@
...
+            raise HeaderError("Missing 'datatype' attribute in TCK header.")
+
+        if not hdr['datatype'].startswith('Float32'):
+            msg = ("TCK only supports float32 dtype but 'dataype: {}' was"
+                   " specified in the header.").format(hdr['datatype'])
+            raise HeaderError(msg)
+
+        if 'file' not in hdr:
```
jchoude: Same question here: is this really mandatory?

nibabel/streamlines/tck.py
MarcCote: Yes, I'm using `format` and `count` which are not `bytes` methods. I decided to just convert the magic number in string during the header creation. If you have a better, more uniform, way of doing all this str/bytes manipulation, I'm all ears.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+        # Fields to exclude
+        exclude = [Field.MAGIC_NUMBER,  # Handled separately.
+                   Field.NB_STREAMLINES,  # Handled separately.
+                   Field.ENDIANNESS,  # Handled separately.
+                   Field.VOXEL_TO_RASMM,  # Streamlines are always in RAS+ mm.
+                   "count", "datatype", "file"]  # Fields being replaced.
+
+        lines = []
+        lines.append(header[Field.MAGIC_NUMBER])
+        lines.append("count: {0:010}".format(header[Field.NB_STREAMLINES]))
```
MarcCote: Actually it would need more than 10 **Billions** streamlines. If that is an issuewe can simply bump this number up.

nibabel/streamlines/tck.py
MarcCote: Good point, I could raise a warning and try loading them using float32, i.e. the only dtype supported right now anyway.

nibabel/streamlines/tck.py
MarcCote: Good point. Now, raising a warning instead.

MarcCote: Thinking about this one, the `file` field indicates the offset where the data starts. In the case the fieldwas not provided, should I assume the data starts directly after END ?

nibabel/streamlines/tck.py
jchoude: Oups, my bad. I think we're safe for now :)

nibabel/streamlines/tck.py
jchoude: Yes, I think that would be better.

nibabel/streamlines/tck.py
jchoude: I would guess that's the best way to go, yes. Use it if there, else, use that heuristic.

MarcCote:
Thanks for the review @jchoude. @matthew-brett I've addressed you last comments and increase test coverage.

nibabel/streamlines/tck.py
matthew-brett: How about `where coordinate (i, j, k) refers to the center of the voxel`.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,454 @@
...
+from .tractogram_file import TractogramFile
+from .tractogram_file import HeaderWarning, DataWarning
+from .tractogram_file import HeaderError, DataError
+from .tractogram import TractogramItem, Tractogram, LazyTractogram
+from .header import Field
+
+MEGABYTE = 1024 * 1024
+
+
+def create_empty_header():
```
matthew-brett: What do you think about my suggestion to make this a method of the TckFile class?

See: https://github.com/nipy/nibabel/pull/486#discussion_r108033806

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,450 @@
...
+        fileobj.write(asbytes("END\n"))
+
+    @staticmethod
+    def _read_header(fileobj):
+        """ Reads a TCK header from a file.
+
+        Parameters
+        ----------
+        fileobj : string or file-like object
+            If string, a filename; otherwise an open file-like object
```
matthew-brett: Add something like `a file-like object opened for binary reads, pointing to` ...

MarcCote: I've already done this modification.

MarcCote:
What do you think of

         Streamlines of the tractogram are assumed to be in *RAS+* and *mm*
         space. It is also assumed that when streamlines are mapped back to
         voxel space, a streamline point located at an integer coordinate
         (i,j,k) is considered to be at the center of the corresponding 
voxel.
         This is in contrast with TRK's internal convention where it would
         have referred to a corner.


  2017-04-06 13:06 GMT-0400 Matthew Brett <notifications@github.com>

matthew-brett:
Sure, streamlines phrasing is fine.

MarcCote:
@matthew-brett I moved `create_empty_header` into `TractogramFile` and modified `TrkFile` accordingly. I modified the docstring of `Tractogram` and `LazyTractogram` to better explain the convention used for streamlines coordinates. Let me know if there is anything else.

nibabel/streamlines/tck.py
```diff
@@ -0,0 +1,464 @@
...
+        return header
+
+    @classmethod
+    def load(cls, fileobj, lazy_load=False):
+        """ Loads streamlines from a filename or file-like object.
+
+        Parameters
+        ----------
+        fileobj : string or file-like object
+            If string, a filename; otherwise an open file-like object
```
MarcCote: I don't know what happened, but you are right,t it was missing some docstrings that weren't specifying binary mode. Should be fixed by commit 0d0ece18b83b0e03079ca6bedac9bef6d5c3a3e8.

matthew-brett:
Thanks for your patience - in it goes.

PR #488, opened 2016-09-29
doc: Properly find the version

jehane:
The conf.py can't import nibabel if nibabel isn't installed, as is the
case when installing nibabel for the first time.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -22,8 +22,6 @@
 import sys
 import os
 
-import nibabel
-
 # Check for external Sphinx extensions we depend on
 try:
     import numpydoc
@@ -89,7 +87,7 @@ copyright = u'2006-2016, %(MAINTAINER)s <%(AUTHOR_EMAIL)s>' % rel
 # built documents.
 #
 # The short X.Y version.
-version = nibabel.__version__
+version = rel['VERSION']
 # The full version, including alpha/beta/rc tags.
 release = version
 


bochecha:
In case you wonder why @jehane sent a pull-request for my commit, that's because I wrote the patch to help them and then completely forgot to open this pull-request. :smile:

effigies:
Looks reasonable.

:+1: for merge.

matthew-brett:
The change is reasonable, but I believe the docs will not build correctly without an installation of nibabel.   What do you get if you try and build the docs when nibabel is not installed?

bochecha:
Seems to work just fine.

In a virtualenv, I did:

```
$ pip install -r doc-requirements.txt
...
$ make htmldoc
...
Build finished. The HTML pages are in ../build/html.
make[1]: Leaving directory '/home/mathieu/Workspace/nibabel/doc'
```

Then if I open `$(pwd)/build/html/index.html` in Firefox I get what looks like your documentation, with the user manual, API doc, etc.

matthew-brett:
I think in that case that `import nibabel` will also work because you are running from the nibabel root directory.   But, if I do:

```
cd doc
make html
```

I get lots of errors like this in the doc build

```
/Users/mb312/dev_trees/nibabel/doc/source/neuro_radio_conventions.rst:83: WARNING: Exception occurred in plotting neuro_radio_conventions-2
 from /Users/mb312/dev_trees/nibabel/doc/source/neuro_radio_conventions.rst:
  File "/Users/mb312/.virtualenvs/test/lib/python2.7/site-packages/matplotlib/sphinxext/plot_directive.py", line 517, in run_code
    six.exec_(code, ns)
  File "/Users/mb312/.virtualenvs/test/lib/python2.7/site-packages/matplotlib/externals/six.py", line 672, in exec_
    exec("""exec _code_ in _globs_, _locs_""")
ImportError: No module named nibabel
```

and the API part of the documentation lists the submodules, but the content is empty.

matthew-brett:
@bochecha - any comments here?   For example, I wonder whether we just need a better error message if nibabel is not installed.

bochecha:
> I think in that case that import nibabel will also work because you are running from the nibabel root directory. But, if I do:
> 
>     cd doc
>     make html

I think this is the root of the problem.

Indeed, `make htmldoc` works just fine without nibabel installed, and without this patch, since it sets `PYTHONPATH` correctly.

I think when we had the problem originally with @jehane was because we had `cd doc && make html`.

So maybe `make htmldoc` needs to be emphasized as the right way to build the docs.

> For example, I wonder whether we just need a better error message if nibabel is not installed.

... perhaps that error message could just tell people to run `make htmldoc` from the root dir?

matthew-brett:
Thanks for analyzing the problem a bit more.   How about https://github.com/nipy/nibabel/pull/503 ?

Issue #489, opened 2016-09-29
Time-slicing spatial images


effigies:
RE: #271 and #404

So, I recognize the issues involved with spatial slicing of `SpatialImages`, but a pretty handy thing to be able to do would be: `img[...,slicer].orthoview()`

Almost always I'm loading the file on the spot, so really it's: `nib.load(fname)[...,slicer].orthoview()`

Being a one-liner from filename to display is the real beauty of this, but the options are:

``` Python
x = nib.load(fname)
Nifti1Image(x.dataobj[...,slicer], x.affine, x.header).orthoview()
# or
nib.viewers.OrthoSlicer3D(x.dataobj[...,slicer], x.affine)
```

Would it be acceptable to modify `SpatialImage.__getitem__` to accept a time slice and return a `SpatialImage`, but reject spatial slices?

effigies:
I suppose an acceptable alternative that doesn't involve mucking with `__getitem__` would be:

```Python
nib.load(fname).orthoview(timeslice)
```

The only annoyance I see there is the need to manually define the `slice`, instead of getting the `start:stop:step` syntax.

PR #491, opened 2016-10-16
ENH: Add manual value limits to OrthoSlicer3D/SpatialImages.orthoview

effigies:
Setting constant value limits for time series and spatial displays can be useful for comparing two images side-by-side.

Edit: Slicing has been removed from this PR in favor of #550.

### DIFF
--- a/nibabel/spatialimages.py
+++ b/nibabel/spatialimages.py
@@ -588,8 +588,17 @@ class SpatialImage(DataobjImage):
             "slicing image array data with `img.dataobj[slice]` or "
             "`img.get_data()[slice]`")
 
-    def orthoview(self, **kwargs):
+    def orthoview(self, axes=None, vlim=None):
         """Plot the image using OrthoSlicer3D
+
+        Parameters
+        ------------------
+        axes : tuple of mpl.Axes or None, optional
+            3 or 4 axes instances for the 3 slices plus volumes,
+            or None (default).
+        vlim : array-like or None, optional
+            Value limits to display image and time series. Can be None
+            (default) to derive limits from data.
 
         Returns
         -------
@@ -602,8 +611,8 @@ class SpatialImage(DataobjImage):
         consider using viewer.show() (equivalently plt.show()) to show
         the figure.
         """
-        return OrthoSlicer3D(self.dataobj, self.affine,
-                             title=self.get_filename(), **kwargs)
+        return OrthoSlicer3D(self.dataobj, self.affine, axes=axes,
+                             title=self.get_filename(), vlim=vlim)
 
     def as_reoriented(self, ornt):
         """Apply an orientation change and return a new image
--- a/nibabel/spatialimages.py
+++ b/nibabel/spatialimages.py
@@ -588,7 +588,7 @@ class SpatialImage(DataobjImage):
             "slicing image array data with `img.dataobj[slice]` or "
             "`img.get_data()[slice]`")
 
-    def orthoview(self):
+    def orthoview(self, **kwargs):
         """Plot the image using OrthoSlicer3D
 
         Returns
@@ -603,7 +603,7 @@ class SpatialImage(DataobjImage):
         the figure.
         """
         return OrthoSlicer3D(self.dataobj, self.affine,
-                             title=self.get_filename())
+                             title=self.get_filename(), **kwargs)
 
     def as_reoriented(self, ornt):
         """Apply an orientation change and return a new image
--- a/nibabel/viewers.py
+++ b/nibabel/viewers.py
@@ -63,7 +63,8 @@ class OrthoSlicer3D(object):
             title.
         vlim : array-like or None, optional
             Value limits to display image and time series. Can be None
-            (default) to derive limits from data.
+            (default) to derive limits from data. Bounds can be of the
+            form ``'x%'`` to use the ``x`` percentile of the data.
         """
         # Use these late imports of matplotlib so that we have some hope that
         # the test functions are the first to set the matplotlib backend. The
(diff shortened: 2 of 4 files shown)


effigies:
If this seems reasonable, what about adding a `subslice` parameter to OrthoViewer3D that takes an object that can be passed to `ndarray.__getitem__`? That would take care of #489 without further disruption to `SpatialImage`.

matthew-brett:
Good idea.   Can you add some tests?

I know this adds some duplication, but it would be good to write out the arguments specifically, to avoid `**kwargs` - it makes the function signature less useful.

Maybe make these keyword only?

effigies:
Looks like keyword-only is Python 3+.

Exposed `axes` in `SpatialImage.orthoview` as well, though I can remove that if that's unwanted.

nibabel/viewers.py
```diff
@@ -61,6 +61,9 @@ def __init__(self, data, affine=None, axes=None, title=None):
         title : str or None, optional
             The title to display. Can be None (default) to display no
             title.
+        vlim : array-like or None, optional
```
matthew-brett: What do you think of also allowing strings like `10%` `90%` as input, meaning percentile?  Then the default could be `vlim=('1%', '99%')`.

effigies: That works for `clim`, but `ylim`'s default is a bit more complicated.

effigies:
Pushed a couple commits. Will add tests when I get another minute.

nibabel/spatialimages.py
```diff
@@ -663,9 +663,21 @@ def __getitem__(self):
...
-    def orthoview(self):
+    def orthoview(self, axes=None, vlim=None, slicer=None):
         """Plot the image using OrthoSlicer3D
 
+        Parameters
+        ----------
+        axes : tuple of mpl.Axes or None, optional
+            3 or 4 axes instances for the 3 slices plus volumes,
+            or None (default).
+        vlim : array-like or None, optional
```
matthew-brett: Copy `vlim` docstring entry from OrthoSlicer3D?

nibabel/viewers.py
matthew-brett: Aha.  I see there's a heuristic for the colorbar around line 205 - is that what you mean?

How about throwing that away and just using the data limits as set by `vlim`?  As in something like:

```
if vlim is None:
    vlim = ('1%', '99%')
```

etc.

matthew-brett:
PR to this branch at https://github.com/effigies/nibabel/pull/1

nibabel/viewers.py
effigies: I believe the purpose of that heuristic is so that the time-series is guaranteed to fit in the pane, with a slight buffer on top and bottom for visibility. If we set `ylim(1%, 99%)`, then some values won't fit in the pane, by default.

matthew-brett: Won't that be true for any not-default limits that we set?  I mean, don't we need to apply the heuristic for any limits?

effigies: I'm not sure I fully understand this comment.

The three use-cases I see, myself, are (1) exploration, (2) comparison, (3) figure-generation. I think the existing default behavior is good for (1). I think the proposed changes are good for (2)+(3), where consistency/predictability is more valuable than smart estimates.

The problem I see in your proposed change is that you're changing the default behavior to something a little less friendly by setting `ylim == clim` even in the default case.

Apologies if I'm not making much sense.

matthew-brett: Sure - I agree that the ``ylim == clim thing`` is bad in the default case, the question is what to do in the not-default case. 

For example - I guess that people will want to use `vlim` to get better contrast in the images, but they don't intend to change the time-series pane.   Do you think that's right?  In which case, should `ylim` be another input argument, that defaults to the original heuristic, regardless of `vlim`?

matthew-brett:
@effigies - any thoughts on my PR?

effigies:
I had some when I first saw it, but was in the midst of writing. Will have another look when I get a chance.

nibabel/viewers.py
matthew-brett: @effigies - any comments here?

matthew-brett:
@effigies - any chance of an update here?

nibabel/viewers.py
effigies: Hey @matthew-brett. Sorry, busy times.

To clarify my usage when I was working on this: I would have multiple files with different ranges. By setting `vlim`, I would be able to directly compare them, both the images and the time series. So for my case, I definitely think keeping them the same was preferable.

However, if what we want is a separate `ylim`, we can do that. In which case, should we change `vlim` to `clim` to drive home that one is for images, one is for time series? I chose `vlim` initially specifically to abstract from `y/c`.

What do you think?

effigies:
Oh, and one other point about `ylim` is that it specifically backs off a bit from the min/max in order to include the entire line on screen, a bit away from the border. effigies#1 eliminates this default behavior, which I think is bad.

matthew-brett:
Just to clarify, as I return to this:

* `clim` is a tuple setting the minimum and maximum of the whole 3D image for the purpose of the image display.   It's useful to set this when futzing with the image contrast;
* `ylim` is a name that could be used for the minimum and maximum values to display on the 4th time-series plane

At the moment, in this PR`vlim` changes `clim`, and `ylim`.  `clim` can be changed using the `clim` property, but not `ylim`.  The set of `ylim` is different in the case where `vlim` is default (None), because the default allows some margin at the top and bottom of the y axis, to make the line easier to see.  Is that about right?

effigies:
Hi Matthew. Sorry about the delay. Yes, these statements are all correct.

nibotmi:
:umbrella: The latest upstream changes (presumably #562) made this pull request unmergeable. Please resolve the merge conflicts.

nibotmi:

nibotmi:
:umbrella: The latest upstream changes (presumably #695) made this pull request unmergeable. Please resolve the merge conflicts.

PR #493, opened 2016-11-05
Miscellaneous MINC reader fixes

rdvincent:
This PR is a potpourri of small fixes for the MINC1 and MINC2 readers.

Mostly they deal with edge cases, although I managed to trigger each of these issues with actual files I have on my system. These fixes do not change the results of "make test" on my system, although I do see "FAILED (errors=1, failures=6)" when I run with either master or this branch.


nibabel/externals/netcdf.py
```diff
@@ -607,7 +607,10 @@ def _read_var_array(self):
                     pos = self.fp.tell()
                     self.fp.seek(begin_)
                     data = fromstring(self.fp.read(a_size), dtype=dtype_)
-                    data.shape = shape
+                    # Avoid exception for empty variables.
+                    if len(data.shape) != 1 or data.shape[0] != 0:
+                        data.shape = shape
```
effigies: This is equivalent to:

```Python
if data.shape != (0,):
    data.shape = shape
````

Are there other cases to check against?

Looking at the surrounding code, I *think* this is doing an awful lot to read zero bytes. Maybe it would make more sense to drop this change and set line 606 to:

```Python
elif asize > 0:
```

And add the following to the end of the block:

```Python
else:
    data = np.array([], dtype=dtype_)
```

nibabel/minc2.py
```diff
@@ -77,7 +77,7 @@ def _get_dimensions(self, var):
             dimorder = var.attrs['dimorder'].decode()
         except KeyError:  # No specified dimensions
             return []
-        return dimorder.split(',')
+        return dimorder.split(',')[:len(var.shape)]
```
effigies: Can you give an example of a failure here?

nibabel/minc1.py
```diff
@@ -111,8 +117,14 @@ def get_affine(self):
...
-            steps[i] = dim.step
-            starts[i] = dim.start
+            try:
+                steps[i] = dim.step
+            except AttributeError:
+                steps[i] = 1.0
+            try:
+                starts[i] = dim.start
+            except AttributeError:
+                starts[i] = 0.0
```
rdvincent: As above.

nibabel/externals/netcdf.py
rdvincent: The only case I've seen is the "empty" netcdf case with zero bytes allocated. I prefer the minimal change, happy to simplify the test.

nibabel/minc2.py
rdvincent: I'll check in a test case. Some MINC files contain stray "dimorder" attributes that are longer than the variable itself, e.g. a file may have a scalar image-min with a dimorder=zspace attribute.

matthew-brett: Test case would be good.  Could you add a comment explaining as well?   Otherwise it might be puzzling for someone coming fresh to the code.

nibabel/minc1.py
```diff
@@ -95,8 +95,7 @@ def get_data_shape(self):
     def get_zooms(self):
         """ Get real-world sizes of voxels """
         # zooms must be positive; but steps in MINC can be negative
-        return tuple(
-            [abs(float(dim.step)) for dim in self._dims])
+        return tuple([abs(float(dim.step)) if hasattr(dim, 'step') else 1.0 for dim in self._dims])
```
matthew-brett: I guess this line is too long for PEP8 (78 chars)?

matthew-brett:
On Sat, Nov 5, 2016 at 8:18 AM, Chris Markiewicz <notifications@github.com>
wrote:

> *@effigies* commented on this pull request.
> ------------------------------
>
> In nibabel/minc1.py <https://github.com/nipy/nibabel/pull/493>:
>
> > @@ -111,8 +117,14 @@ def get_affine(self):
>              except AttributeError:
>                  dir_cos = _default_dir_cos[name]
>              rot_mat[:, i] = dir_cos
> -            steps[i] = dim.step
> -            starts[i] = dim.start
> +            try:
> +                steps[i] = dim.step
> +            except AttributeError:
> +                steps[i] = 1.0
> +            try:
> +                starts[i] = dim.start
> +            except AttributeError:
> +                starts[i] = 0.0
>
> Ah, I see, I didn't look at the surrounding context. Still, I think this
> is easier to read, so hopefully @matthew-brett
> <https://github.com/matthew-brett> agrees with me. :-)
>

Yes - I agree :)

nibabel/minc2.py
rdvincent: The MINC 2 format file I added to the test cases will exercise this.

nibabel/minc1.py
rdvincent: Not seeing an error, but I'll modify the line to make it shorter.

effigies:
LGTM.

nibabel/externals/netcdf.py
```diff
@@ -607,7 +607,10 @@ def _read_var_array(self):
                     pos = self.fp.tell()
                     self.fp.seek(begin_)
                     data = fromstring(self.fp.read(a_size), dtype=dtype_)
-                    data.shape = shape
+                    # Avoid exception for empty variables.
```
matthew-brett: Should this fix go into the scipy version of the file, that we copied from?

Is there a test for this change?

rdvincent: This is a generic issue with this NetCDF library, in that a NetCDF variable can have a data offset pointing to the end of the file. I've created a test case, but the current code fails if mmap is enabled. I have a more elaborate fix that avoids these failures.

nibabel/externals/netcdf.py
```diff
@@ -210,6 +210,13 @@ def __init__(self, filename, mode='r', mmap=None, version=1):
             self.fp = open(self.filename, '%sb' % mode)
             if mmap is None:
                 mmap = True
+        try:
+            self.fp.seek(0, 2)
```
matthew-brett: I guess the ValueError comes here?  Best to isolate that line with `try`, as in:

```
try:
    self.fp.seek(0, 2)
except ValueError:
    self.file_bytes = -1  # Unknown file length (gzip).
else:
    self.file_bytes = self.fp.tell()
    self.fp.seek(0)
```

nibabel/externals/netcdf.py
```diff
@@ -599,18 +606,20 @@ def _read_var_array(self):
             else:  # not a record variable
                 # Calculate size to avoid problems with vsize (above)
                 a_size = reduce(mul, shape, 1) * size
-                if self.use_mmap:
+                if self.file_bytes >= 0 and begin_+a_size > self.file_bytes:
```
matthew-brett: PEP8 requires spaces around the `+`.

nibabel/externals/netcdf.py
```diff
@@ -599,18 +606,20 @@ def _read_var_array(self):
...
                             buffer=mm, offset=begin_, order=0)
                 else:
                     pos = self.fp.tell()
                     self.fp.seek(begin_)
-                    data = fromstring(self.fp.read(a_size), dtype=dtype_)
-                    # Avoid exception for empty variables.
-                    if data.shape != (0,):
-                        data.shape = shape
-
+                    buf = self.fp.read(a_size)
```
matthew-brett: Comment here?  I guess this is for a broken file where there's an expected variable truncated at the end?  Do you really want to allow that case without raising an error?

nibabel/externals/netcdf.py
```diff
@@ -599,14 +607,22 @@ def _read_var_array(self):
             else:  # not a record variable
                 # Calculate size to avoid problems with vsize (above)
                 a_size = reduce(mul, shape, 1) * size
-                if self.use_mmap:
+                if self.file_bytes >= 0 and begin_ + a_size > self.file_bytes:
+                    data = fromstring(b'\x00'*a_size, dtype=dtype_)
```
effigies: Spaces around the `*`.

nibabel/externals/netcdf.py
```diff
@@ -599,14 +607,22 @@ def _read_var_array(self):
...
                 else:
                     pos = self.fp.tell()
                     self.fp.seek(begin_)
-                    data = fromstring(self.fp.read(a_size), dtype=dtype_)
+                    # Try to read file, which may fail because the data is
+                    # at or past the end of file. In that case, we treat
+                    # this data as zeros.
+                    buf = self.fp.read(a_size)
+                    if len(buf) < a_size:
+                        buf = b'\x00'*a_size
```
effigies: Spaces around the `*`.

matthew-brett:
Great - thanks for the edits - merging.

Would you consider putting a PR to do the same changes for `scipy/io/netcdf.py`?

PR #496, opened 2016-11-10
Format string {} not allowed in Python 2.6

SyamGadde:
pypi indicates that the current version of nibabel works for any version of Python, but the use of a literal '{}' (i.e. without indices) in a format string is only supported started in Python 2.7.

### DIFF
--- a/nibabel/deprecator.py
+++ b/nibabel/deprecator.py
@@ -146,7 +146,7 @@ class Deprecator(object):
         if since:
             messages.append('* deprecated from version: ' + since)
         if until:
-            messages.append('* {} {} as of version: {}'.format(
+            messages.append('* {0} {1} as of version: {2}'.format(
                 "Raises" if self.is_bad_version(until) else "Will raise",
                 error_class,
                 until))


matthew-brett:
Sorry - that's probably a bug in our specifications to pypi - we dropped Python 2.6 support for this last release.

effigies:
This was the only way I could see in pypi to indicate supported versions: aa4a727

SyamGadde:
Irrespective of nibabel's stated requirements, this change to the format string allows people who are stuck with Python 2.6 (RHEL/Centos 6) to at least use the portions of the library that work with Python 2.6.  Nothing I've tried breaks quite yet.

matthew-brett:
Fair enough - in it goes.   In general though, the point of dropping Python 2.6 testing and compatibility is so we can write code with the modern idioms, so expect to have more stuff break soon, and we can't guarantee the results are correct, with Python 2.6.

SyamGadde:
Understood.  Thanks!

PR #500, opened 2016-12-22
BF: boolean indexing now requires matching dimension

### LINKED ISSUE #499: test_orientation is failing with new Numpy's pre-release
In the next Numpy's release, boolean index must match indexed array along specified dimension.
See https://travis-ci.org/nipy/nibabel/jobs/186167556#L1230

From https://github.com/nipy/nibabel/blob/master/nibabel/orientations.py#L71
```python
    P, S, Qs = npl.svd(RS)
    # Threshold the singular values to determine the rank.
    if tol is None:
        tol = S.max() * max(RS.shape) * np.finfo(S.dtype).eps
    keep = (S > tol)
    R = np.dot(P[:, keep], Qs[keep])
```
When `RS` is not square, `keep` will have a dimension less than either `P`'s columns or `Qs`' rows, therefore indexing using `keep` will raise an error. I haven't examine the code but maybe we could use `full_matrices=False` when calling `npl.svd()` (see https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.svd.html)

MarcCote:
Fixes #499

The tests are passing on my machine, so it doesn't seem to break anything (that is currently tested). Let's see what happen with the Numpy pre-release.

### DIFF
--- a/nibabel/benchmarks/bench_streamlines.py
+++ b/nibabel/benchmarks/bench_streamlines.py
@@ -91,5 +91,6 @@ def bench_load_trk():
         for s1, s2 in zip(scalars_new, scalars_old):
             assert_array_equal(s1, s2)
 
+
 if __name__ == '__main__':
     bench_load_trk()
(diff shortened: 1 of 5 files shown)


effigies:
Simple changes, I'm inclined to merge tomorrow, barring objections.

Any comments? @matthew-brett?

PR #501, opened 2017-01-03
Update get_csa_header to return none on CSAReadError

parneshraniga:
Some non-MRI dicom datasets have just strings in CSA Header tags e.g Siemens PET dicoms . This causes the read on line 68 to throw an error because the string is not a CSA header. Unfortunately this throws out other packages such as dcmstack which use nibabel. The proposed fix returns none if a read error is encountered when such strings are parsed. It is possible such strings may pass the initial check for 0 < n_tags <=128 which throws the CSAReadError but fail later in which case the except block can be be modified to handle those exceptions as well.

### DIFF
--- a/nibabel/nicom/csareader.py
+++ b/nibabel/nicom/csareader.py
@@ -65,10 +65,11 @@ def get_csa_header(dcm_data, csa_type='image'):
     element_no = section_start + element_offset
     try:
         tag = dcm_data[(0x29, element_no)]
-    except KeyError:
+        return read(tag.value)
+    except (KeyError,CSAReadError):
         # The element could be missing due to anonymization
         return None
-    return read(tag.value)
+    
 
 
 def read(csa_str):


matthew-brett:
Thanks for this - have you got an example DICOM you can share, that we can use to test?   Is there anything about the string that allows us to detect its presence in a more specific way than an error in the csa read routine?

parneshraniga:
Thus far it seems to fail for Siemens PET data. I will check to see if the data can be shared or is already available and let you know. On the two datasets that I checked, there wasn't anything in particular that stood out but I will go through the data I have to see what I can find.

matthew-brett:
You can delete the data in the DICOM file if you want, and replace it with zeros, so the file compresses well, and doesn't reveal private brain data.

parneshraniga:
Done.

effigies:
Great! Thanks. Any luck getting an anonymized DICOM that was failing before?

parneshraniga:
Yes I can grab a couple of files, anonymise them and clear the data out. Whats the best way to forward them to you?

effigies:
Any place we can grab them from. Google drive or dropbox are common. If they're small, they can go in [nibabel/tests/data](https://github.com/nipy/nibabel/tree/master/nibabel/tests/data) in a new commit; if they're big, see http://nipy.org/nibabel/devel/add_test_data.html.

matthew-brett:
Guys - now is an excellent time to get this one finished up - I have lots of free time at the moment to work on this.

matthew-brett:
@parneshraniga - would you mind sharing those files?  That would be very helpful.

matthew-brett:
@parneshraniga - anything we can do to help?  Can you give us access to the unmodified files?  We can anonymize them for you.

matthew-brett:
@parneshraniga - another reminder in hope that we can get this one in.

effigies:
@parneshraniga Do you have any time to help get this one finished?

effigies:
Hi @parneshraniga, is there any chance you still have access to files that can reproduce this issue?

PR #502, opened 2017-01-07
TST: add test for fix in PR 500

matthew-brett:
Orientations calculation did not take into account what would happen
with not-square matrices, 500 fixes this.  These are some tests for
fixed behavior.

### DIFF
--- a/nibabel/tests/test_orientations.py
+++ b/nibabel/tests/test_orientations.py
@@ -206,6 +206,7 @@ def test_io_orientation():
     # Test drop of rows, columns
     mat, vec = to_matvec(def_aff)
     aff_extra_col = np.zeros((4, 5))
+    aff_extra_col[-1, -1] = 1  # Not strictly necessary, but for completeness
     aff_extra_col[:3, :3] = mat
     aff_extra_col[:3, -1] = vec
     assert_array_equal(io_orientation(aff_extra_col, tol=1e-5),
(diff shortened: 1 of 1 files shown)


nibabel/tests/test_orientations.py
```diff
@@ -203,6 +203,23 @@ def test_io_orientation():
     assert_array_equal(io_orientation(def_aff, tol=0), pass_tol)
     def_aff[1, 1] = eps * 10
     assert_array_equal(io_orientation(def_aff, tol=1e-5), fail_tol)
+    # Test drop of rows, columns
+    mat, vec = to_matvec(def_aff)
+    aff_extra_col = np.zeros((4, 5))
+    aff_extra_col[:3, :3] = mat
+    aff_extra_col[:3, -1] = vec
```
effigies: Is that supposed to be -1 with 5 columns? Or `aff_extra_col[:3, 3]`?

effigies: (Assuming I understand what's going on...)

For clarity, maybe an easier way to do this would be to have:

```Python
base_affine = np.zeros((4,4))
base_affine [:3, :3] = mat
base_affine[:3, -1] = vec

aff_extra_col[:4, :4] = base_affine
aff_extra_row[:4, :4] = base_affine
```

matthew-brett: In both cases I was aiming for a 4th column / row with all zero, with the last column always being the translations, and non-zero.  Is that what you thought?

effigies: Ah, no, I misunderstood. Carry on, then.

matthew-brett:
Thanks for the review.

PR #503, opened 2017-01-12
DOC: better warning when nibabel not on path

matthew-brett:
Inspired by https://github.com/nipy/nibabel/pull/488 with thanks.


bochecha:
Thanks, this should help avoiding the problem which made us run into #488. :+1:

PR #504, opened 2017-01-19
Clarified error messages in LazyTractogram

MarcCote:
This PR clarifies what "generator function" means in relevant error messages. It also makes sure all values contained in a `LazyDict` object are callable.

### DIFF
--- a/nibabel/streamlines/tractogram.py
+++ b/nibabel/streamlines/tractogram.py
@@ -167,7 +167,7 @@ class LazyDict(collections.MutableMapping):
     def __setitem__(self, key, value):
         if not callable(value):
             msg = ("Values in a `LazyDict` must be generator functions."
-                   " Those are functions which whenever are called return an"
+                   " These are functions which, when called, return an"
                    " instantiated generator.")
             raise TypeError(msg)
         self.store[key] = value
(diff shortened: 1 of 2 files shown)


nibabel/streamlines/tractogram.py
```diff
@@ -165,8 +165,11 @@ def __getitem__(self, key):
         return self.store[key]()
 
     def __setitem__(self, key, value):
-        if value is not None and not callable(value):  # TODO: why None?
-            raise TypeError("`value` must be a generator function or None.")
+        if not callable(value):
+            msg = ("Values in a `LazyDict` must be generator functions."
+                   " Those are functions which whenever are called return an"
```
matthew-brett: Maybe "These are functions which, when called, return an instantiated generator."

nibabel/streamlines/tractogram.py
```diff
@@ -604,7 +607,10 @@ def _apply_affine():
 
     def _set_streamlines(self, value):
         if value is not None and not callable(value):
-            raise TypeError("`streamlines` must be a generator function.")
+            msg = ("`streamlines` must be a generator function. That is a"
```
matthew-brett: Ditto.

nibabel/streamlines/tractogram.py
MarcCote: Thank for the suggestion, it is indeed way better.

MarcCote:
Completely forgot about this PR :O.

MarcCote:
@matthew-brett This PR should be good to go. I've modified some outputs messages as you suggested.

matthew-brett:
Thanks - in it goes.

PR #506, opened 2017-01-24
MAINT: update numpy dependency; add Python 3.6

matthew-brett:
This will also trigger tests against the latest wheels.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,12 @@ env:
         - DEPENDS="numpy scipy matplotlib h5py pillow"
         - PYDICOM=1
         - INSTALL_TYPE="setup"
+        - EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
+        - EXTRA_PIP_FLAGS="--find-links=$EXTRA_WHEELS"
 python:
-    - 3.3
     - 3.4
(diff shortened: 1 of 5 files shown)


matthew-brett:
I think this one is a fairly straightforward win - any takers for a quick review?

.travis.yml
```diff
@@ -23,10 +23,12 @@ env:
         - DEPENDS="numpy scipy matplotlib h5py pillow"
         - PYDICOM=1
         - INSTALL_TYPE="setup"
+        - EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
+        - PIP_FLAGS="--find-links=$EXTRA_WHEELS"
```
grlee77: Is this PIP_FLAGS always included by pip in all cases then?  I was unfamiliar with that environment variable and don't see it mentioned in the manual?  I would have guessed it wanted PIP_FIND_LINKS.

grlee77:
LGTM.  the comment above is mainly a question for my own education

.travis.yml
matthew-brett: Oops - sorry - that should be EXTRA_PIP_FLAGS, used explicitly below.  I've changed it...

matthew-brett:
Thanks for the review - merging as I think this stuff is uncontroversial.

Issue #505, opened 2017-01-20
parrec load, different keywords


steso:
Hi
I try to read par/rec files, however some of the keywords in my par header differ. (e.g. Repetition time [msec] instead of [ms]). I tried to modify my parrec.py accordingly, however if I change the keywords in parrec.py _hdr_key_dict, nothing changes (still the old values instead of the corrected). Obviously I don't understand the code (yet), maybe somebody could point me into the right direction? And what woud be the best way account for these slight changes in the header?
Thanks in advance

matthew-brett:
Can you post at least an example .PAR file somewhere we can have a look?   You're right, it's not particularly easy to deal with small variations in the format, we'll have to put in some explicit regular expressions or similar.

steso:
Ok here is one of my par-files:
https://www.dropbox.com/s/q5ilcqclvw36b75/invivo.par?dl=0
The differences are in the following properties:

Series_data_type
Patient Position
Repetition time [msec]
Diffusion echo time [msec]

I think I didn't forget any. The error occurs in parrec.py in parse_PAR_header(fObj) in the following line:
# get props for this hdr field
props = _hdr_key_dict[key]
However, what is still doesn't understand, if I modify the _hdr_key_dict in that particular file (parrec.py) and re-run my code, the key values in _hdr_key_dict are still the old ones. Is the _hdr_key_dict stored somewhere else and loaded?
I just wanted to change the mentioned _hdr_key values as a quick fix instead of changing all of the par-files.

grlee77:
I just took a quick look at the .PAR file.  I see that it is still labelled as a V4.2 PAR which we nominally support.  Disappointing that apparently being V4.2 doesn't guarantee identical field names, etc.  Do you know which Philips software release the scanner that acquired this is running?  I haven't looked at .PAR files from the fairly recent V5.3 release, yet, but perhaps that is when the change occurred?

Fortunately there are no entirely new header fields or data columns, so I don't think it should be too hard to fix.  will take a quick look now.

matthew-brett:
Looking at our example files, most of them seem to be V4.2 - so this must be a variation in the V4.2 format - I agree, it would be good to understand where this variation comes from...

steso:
They should be from a release 3-ish (I guess 3.2.1) system. I can check and compare the 5.3 release par-files next week if it helps, still not sure why the simple renaiming worksround doesn't work...

matthew-brett:
[...]

PR #507, opened 2017-01-25
ENH: support alternative header field name variants in .PAR files

grlee77:
This is a workaround for the alternative header naming variants observed in #505.

I modified one of the existing .PAR files.  I just truncated it and changed 4 header lines to have the entries from the .PAR file in #505.  We can swap the file out for that other actual .par file prior to merging if the license terms permit.


nibabel/parrec.py
```diff
@@ -206,6 +206,14 @@
     'Max. number of gradient orients': ('max_gradient_orient', int),
     # Line below added for par / rec version > 4.1
     'Number of label types   <0=no ASL>': ('nr_label_types', int),
+
```
matthew-brett: Vague preferernce for removing blank line here, comments do enough of a job of spacing.

nibabel/parrec.py
```diff
@@ -206,6 +206,14 @@
...
+    # The following are duplicates of the above fields, but with slightly
+    # different abbreviation, spelling, or capatilization.  Both variants have
+    # been observed in the wild in V4.2 PAR files.
```
matthew-brett: Add link to issue?

nibabel/tests/data/variant_v4_2_header.par
```diff
@@ -0,0 +1,101 @@
+# === DATA DESCRIPTION FILE ======================================================
```
matthew-brett: Do we have permission to distribute this file?   Add to `COPYING` file if so.

grlee77:
made the suggested changes.  haven't determined why there was one test failure for each Appveyor run last time around.  There were several other warnings there as well.

grlee77:
The Windows test failure is real and is because I truncated the PAR file too much so that the data doesn't match the number of slices in the header.  It only passes on Linux/OSX because there apparently the *.PAR search is case sensitive and doesn't find the new ".par" file.

matthew-brett:
Sorry for the delay - looks good -merging.

PR #508, opened 2017-01-25
MRG: removing some vendored dependencies

matthew-brett:
We don't need OrderedDict copy now we don't support Python 2.6.

Depend on six package - it's very stable at this point, and pip works much
better than in the past.


matthew-brett:
This one now ready for review.

nibabel/externals/__init__.py
```diff
@@ -1,5 +1,5 @@
 # init for externals package
-try:
-    from collections import OrderedDict
-except ImportError:  # < Python 2.7
-    from .ordereddict import OrderedDict
+from collections import OrderedDict
+
+from ..deprecated import ModuleProxy as _ModuleProxy
+six = _ModuleProxy('nibabel.externals.six')
```
grlee77: As nibabel directly uses six everywhere now, this is just here to support 3rd party code that may still be using nibabels.external.six?

matthew-brett: @grlee77 - right - this clause is just in case someone is importing six from nibabel.externals .

matthew-brett:
Thanks for the review - made changes - OK to merge?

effigies:
Looks like callable edits didn't make it into your last commit?

matthew-brett:
Oops - thanks - added callable.

effigies:
Good to merge.

matthew-brett:
Thanks for the review.

PR #512, opened 2017-02-13
BF: Supports reading TRK (version 1)

MarcCote:
Right now, the streamlines API is not able to load TRK (version 1) since it misses the `vox_to_ras` field. This PR fixes that.

With this PR, when loading a TRK (version 1), the `vox_to_ras` field is set so it contains all zeros. The rest of the code can already handle this case since in TRK (version 2) whenever `vox_to_ras[3][3]` is 0, it means the matrix is not recorded. The way it is curently handled is we assume the affine transformation is the identity matrix and we warn the user.

In addition, this PR corrects some typos and changes a bit how we print the TRK header such that only the right amount of scalars and properties names are being displayed.

### DIFF
--- a/nibabel/streamlines/tests/test_trk.py
+++ b/nibabel/streamlines/tests/test_trk.py
@@ -156,6 +156,20 @@ class TestTRK(unittest.TestCase):
         new_trk_file = trk_file[:254] + noise + trk_file[254+4:]
         assert_raises(HeaderError, TrkFile.load, BytesIO(new_trk_file))
 
+    def test_load_trk_version_1(self):
+        trk_file = open(DATA['simple_trk_fname'], 'rb').read()
+
+        # Simulate a TRK (version 1).
+        version = np.array(1, dtype=np.int32).tostring()
+        new_trk_file = trk_file[:992] + version + trk_file[992+4:]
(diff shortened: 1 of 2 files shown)


matthew-brett:
Thanks - have a look at : https://github.com/MarcCote/nibabel/pull/11 - seem reasonable to you?

MarcCote:
I rebased, it should be all good.

nibabel/streamlines/tests/test_trk.py
```diff
@@ -1,10 +1,11 @@
 import os
+import sys
 import copy
 import unittest
 import numpy as np
 from os.path import join as pjoin
 
-from six import BytesIO
+from nibabel.externals.six import BytesIO
```
matthew-brett: Would you mind reverting this to `from six import BytesIO`?  We're trying to remove use of the `nibabel.externals.six` module.

matthew-brett:
OK - looks good apart from the `six` import - thanks.

nibabel/streamlines/tests/test_trk.py
MarcCote: Sure no problem.

matthew-brett:
Thanks, in it goes.

PR #514, opened 2017-03-06
DOC: Fix minor typo in the documentation

iglpdc:
When talking about the scaling of NIfTI images, the docs mention that ``scl_slope`` is added to the output image, but I think it should be ``scl_inter`` instead.


matthew-brett:
Thanks!

PR #515, opened 2017-03-07
DOC: Update authors, changelog

effigies:
Ran through the first few steps of http://nipy.org/nibabel/devel/make_release.html

Can be merged as-is or I can update as we get closer to release.


matthew-brett:
Thanks for doing that - let's merge and do new PRs as we need to.

PR #516, opened 2017-03-08
Bugfix to WrapStruct.get

pauldmccarthy:
A minor bug in `WrapStruct.get` - it fails for header fields with values that evaluates to `False` (e.g. `0`):

```
(fslpy) PMcMBPr15:nibabel paulmc$ fslhd MNI152_T1_2mm_nosqform.nii.gz | grep "_code"
slice_code     0
intent_code    0
qform_code     0
sform_code     0
file_code      1
(fslpy) PMcMBPr15:nibabel paulmc$ ipython
imPython 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 12:54:16)
Type "copyright", "credits" or "license" for more information.

IPython 5.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import nibabel as nib
nibabel/cifti2/parse_cifti2.py:24: FutureWarning: We no longer carry a copy of the 'six' package in nibabel; Please import the 'six' package directly

In [2]: img = nib.load('MNI152_T1_2mm_nosqform.nii.gz')

In [3]: print img.header['sform_code']
0

In [4]: print img.header.get('sform_code')
None

In [5]: print img.header.get('sform_code', -1)
-1

In [6]: print nib.__version__
2.2.0dev
```

### DIFF
--- a/nibabel/tests/test_wrapstruct.py
+++ b/nibabel/tests/test_wrapstruct.py
@@ -183,6 +183,22 @@ class _TestWrapStructBase(TestCase):
         assert_equal(hdr.get(keys[0]), vals[0])
         assert_equal(hdr.get(keys[0], 'default'), vals[0])
 
+        # make sure .get returns values 
+        # which evaluate to False. We
+        # have to use a different falsy
+        # value depending on the data
+        # type of the first header field.
+        if np.issubdtype(hdr_dt[0], np.number):
+            falsyval = 0
+        else:
+            falsyval = ''
+
+        hdr[keys[0]] = falsyval
+        assert_equal(hdr[keys[0]], falsyval)
+        assert_equal(hdr.get(keys[0]), falsyval)
+        assert_equal(hdr.get(keys[0], -1), falsyval)
+
+
     def test_endianness_ro(self):
         # endianness is a read only property
         ''' Its use in initialization tested in the init tests.
--- a/nibabel/tests/test_wrapstruct.py
+++ b/nibabel/tests/test_wrapstruct.py
@@ -191,7 +191,7 @@ class _TestWrapStructBase(TestCase):
         if np.issubdtype(hdr_dt[0], np.number):
             falsyval = 0
         else:
-            falsyval = ''
+            falsyval = b''
 
         hdr[keys[0]] = falsyval
         assert_equal(hdr[keys[0]], falsyval)
--- a/nibabel/tests/test_wrapstruct.py
+++ b/nibabel/tests/test_wrapstruct.py
@@ -183,15 +183,10 @@ class _TestWrapStructBase(TestCase):
         assert_equal(hdr.get(keys[0]), vals[0])
         assert_equal(hdr.get(keys[0], 'default'), vals[0])
 
-        # make sure .get returns values 
-        # which evaluate to False. We
-        # have to use a different falsy
-        # value depending on the data
-        # type of the first header field.
-        if np.issubdtype(hdr_dt[0], np.number):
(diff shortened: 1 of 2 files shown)


nibabel/wrapstruct.py
```diff
@@ -343,7 +343,8 @@ def items(self):
 
     def get(self, k, d=None):
         ''' Return value for the key k if present or d otherwise'''
-        return (k in self.keys()) and self._structarr[k] or d
+        if k in self.keys(): return self._structarr[k]
+        else:                return d
```
effigies:
```Python
return self._structarr[k] if k in self.keys() else d
```

effigies:
Do you think you could update the test with a value that evaluates `False`, and making sure it behaves as expected?

https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_wrapstruct.py#L166-L184

pauldmccarthy:
Hi Chris,

No problem, test updated ... Are there any data types other than numeric and string that need to be considered in the test case?

nibabel/tests/test_wrapstruct.py
```diff
@@ -183,6 +183,22 @@ def test_mappingness(self):
...
 
+        # make sure .get returns values 
+        # which evaluate to False. We
+        # have to use a different falsy
+        # value depending on the data
+        # type of the first header field.
+        if np.issubdtype(hdr_dt[0], np.number):
+            falsyval = 0
+        else:
+            falsyval = ''
```
effigies: Judging by Travis, I think this should fix the error: `falsyval = b''`

effigies:
I think numeric and bytes are all the relevant types, but @matthew-brett will know better.

pauldmccarthy:
Oops sorry - I still spend most of my time with  python 2 :)

nibabel/tests/test_wrapstruct.py
```diff
@@ -183,6 +183,22 @@ def test_mappingness(self):
         assert_equal(hdr.get(keys[0]), vals[0])
         assert_equal(hdr.get(keys[0], 'default'), vals[0])
 
+        # make sure .get returns values
```
matthew-brett: Can you wrap these to fill out to 78 characters?

nibabel/wrapstruct.py
```diff
@@ -343,7 +343,10 @@ def items(self):
 
     def get(self, k, d=None):
         ''' Return value for the key k if present or d otherwise'''
-        return (k in self.keys()) and self._structarr[k] or d
```
matthew-brett: How about just `return self._structarr[k] if k in self.keys()) else d`?

nibabel/tests/test_wrapstruct.py
```diff
@@ -183,6 +183,22 @@ def test_mappingness(self):
         assert_equal(hdr.get(keys[0]), vals[0])
         assert_equal(hdr.get(keys[0], 'default'), vals[0])
 
+        # make sure .get returns values 
+        # which evaluate to False. We
+        # have to use a different falsy
+        # value depending on the data
+        # type of the first header field.
+        if np.issubdtype(hdr_dt[0], np.number):
```
matthew-brett: Nicer as `falsyval = 0 if np.issubdtype(hdr_dt[0], np.number) else b''`

matthew-brett:
Great - thanks - let's see what breaks!

PR #517, opened 2017-03-20
Fix rotation matrix around the first array axis

fepegar:
The "rotation of γ radians around the first array axis" matrix looked like a 3 by 4 matrix. Removed the zeros so that it's 3 by 3.


matthew-brett:
Well spotted - thanks.

fepegar:
No problem!

When will this changes be visible on the web?

matthew-brett:
Oops - I thought I'd pushed it for some reason.    Will push up a version of your quaternion link fix and then push the docs.  Thanks for the reminder.

matthew-brett:
Should be OK now.

fepegar:
It is, thank you!

PR #519, opened 2017-03-22
Import BytesIO from io

effigies:
```Python
>>> import nibabel
[...]/nibabel/cifti2/parse_cifti2.py:24: FutureWarning: We no longer carry a copy of the 'six' package in nibabel; Please import the 'six' package directly
```

This PR also updates a bunch of `from six import BytesIO` imports I found while making sure that was the only thing still importing from `externals.six`.

### DIFF
--- a/nibabel/cifti2/parse_cifti2.py
+++ b/nibabel/cifti2/parse_cifti2.py
@@ -11,6 +11,7 @@ from __future__ import division, print_function, absolute_import
 from distutils.version import LooseVersion
 
 import numpy as np
+from io import BytesIO
 
 from .cifti2 import (Cifti2MetaData, Cifti2Header, Cifti2Label,
                      Cifti2LabelTable, Cifti2VertexIndices,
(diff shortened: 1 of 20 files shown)


matthew-brett:
Looks good - are you happy with just this change, or are there more to come?

Would you mind rebasing?

effigies:
Yeah, I'll rebase before this goes in. I'm a little curious why that first one failed, so I want to run these tests. I'll ping you once I've stopped tinkering.

matthew-brett:
I did try this refactoring a while back, but ran into the same trouble and backed off.

The differences come from the fact that`from six import BytesIO, StringIO` both result in `StringIO.StringIO` in Python 2.7.   `StringIO.StringIO` accepts and returns not-unicode strings; as expected by the tests - that causes some of the errors.  `io.BytesIO` can't be pickled, causing another error.  I can't remember the explanation for the other errors, but I concluded that it wasn't worth refactoring round them.  Maybe you'll come to a different conclusion though...

effigies:
Yeah, not planning on refactoring, but I do want to see which ones are safe to switch. Latest commit looks to be [building fine](https://travis-ci.org/nipy/nibabel/builds/213966410) on Travis. At least now we're down to:

```Shell
$ grep -rI IO nibabel | grep 'six import'
nibabel/externals/tests/test_six.py:        from six import BytesIO
nibabel/tests/test_analyze.py:from six import BytesIO, StringIO
nibabel/tests/test_batteryrunners.py:from six import StringIO
nibabel/tests/test_wrapstruct.py:from six import StringIO
```

If someone decides to tackle those last few imports, at least the scope will be clearer.

Squashing...

effigies:
Rebased. I'll merge when (assuming) the tests pass, unless you'd rather do the honors?

matthew-brett:
Go for it.

PR #521, opened 2017-03-25
ENH: ArrayProxy reshape

### LINKED ISSUE #520: Problem loading large CIFTI2 files in lastest updates
Hi!

it seems that one of the post-"Merge pull request #249" updates to the codebase is forcing the `load` function of cifti2 to bring into memory the whole loaded matrix. For large files (I'm working on  files of ~10Gb) this forces a "cannot allocate memory" exception. 

I don't know how is it possible to add such large files to the test-base such that we can have a control for this. In any case it will be great if the error could be fixed.

effigies:
Might as well start a PR. Are there any reasons this is a terrible idea?

Closes #520.

(Will close this if it doesn't actually resolve the issue.)

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -162,6 +162,17 @@ class ArrayProxy(object):
         # Upcast as necessary for big slopes, intercepts
         return apply_read_scaling(raw_data, self._slope, self._inter)
 
+    def reshape(self, shape):
+        size = np.prod(self._shape)
+        if np.prod(shape) != size:
+            raise ValueError("cannot reshape array of size {:d} into shape "
+                             "{!s}".format(size, shape))
+        new_ap = ArrayProxy(file_like=self.file_like,
+                            header=self._header,
+                            mmap=self._mmap)
+        new_ap.shape = shape
+        return new_ap
+
 
 def is_proxy(obj):
     """ Return True if `obj` is an array proxy
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -194,13 +194,23 @@ class ArrayProxy(object):
         ``np.reshape(array_proxy, shape)``
         '''
         size = np.prod(self._shape)
-        if np.prod(shape) != size:
+
+        # Calculate new shape if not fully specified
+        shape_arr = np.asarray(shape)
+        unknowns = shape_arr == -1
+        if len(unknowns) > 1:
+            raise ValueError("can only specify one unknown dimension")
+        elif len(unknowns) == 1:
+            uk_val = size // np.prod(shape_arr[~unknowns])
+            shape_arr[unknowns] = uk_val
+
+        if shape_arr.prod() != size:
             raise ValueError("cannot reshape array of size {:d} into shape "
                              "{!s}".format(size, shape))
-        return ArrayProxy(file_like=self.file_like,
-                          spec=(shape, self._dtype, self._offset,
-                                self._slope, self._inter),
-                          mmap=self._mmap)
+        return self.__class__(file_like=self.file_like,
+                              spec=(shape, self._dtype, self._offset,
+                                    self._slope, self._inter),
+                              mmap=self._mmap)
 
 
 def is_proxy(obj):
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -195,11 +195,10 @@ class ArrayProxy(object):
         if np.prod(shape) != size:
             raise ValueError("cannot reshape array of size {:d} into shape "
                              "{!s}".format(size, shape))
-        new_ap = ArrayProxy(file_like=self.file_like,
-                            header=self._header,
-                            mmap=self._mmap)
-        new_ap._shape = shape
-        return new_ap
+        return ArrayProxy(file_like=self.file_like,
+                          spec=(shape, self._dtype, self._offset,
+                                self._slope, self._inter),
+                          mmap=self._mmap)
 
 
 def is_proxy(obj):
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -188,23 +188,21 @@ class ArrayProxy(object):
         return apply_read_scaling(raw_data, self._slope, self._inter)
 
     def reshape(self, shape):
-        ''' Return an ArrayProxy with a new shape, without modifying data
-
-        ``array_proxy.reshape(shape)`` is equivalent to
-        ``np.reshape(array_proxy, shape)``
-        '''
+        ''' Return an ArrayProxy with a new shape, without modifying data '''
         size = np.prod(self._shape)
 
         # Calculate new shape if not fully specified
-        shape_arr = np.asarray(shape)
-        unknowns = shape_arr == -1
-        if len(unknowns) > 1:
+        from operator import mul
+        from functools import reduce
+        n_unknowns = len([e for e in shape if e == -1])
+        if n_unknowns > 1:
             raise ValueError("can only specify one unknown dimension")
-        elif len(unknowns) == 1:
-            uk_val = size // np.prod(shape_arr[~unknowns])
-            shape_arr[unknowns] = uk_val
+        elif n_unknowns == 1:
+            known_size = reduce(mul, shape, -1)
+            unknown_size = size // known_size
+            shape = tuple(unknown_size if e == -1 else e for e in shape)
 
-        if shape_arr.prod() != size:
+        if np.prod(shape) != size:
             raise ValueError("cannot reshape array of size {:d} into shape "
                              "{!s}".format(size, shape))
         return self.__class__(file_like=self.file_like,
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -78,7 +78,7 @@ class ArrayProxy(object):
             File-like object or filename. If file-like object, should implement
             at least ``read`` and ``seek``.
         spec : object or tuple
-            Tuple must have length 2-5, with the following fields:
+            Tuple must have length 2-5, with the following values.
                 - shape : tuple
                     tuple of ints describing shape of data
                 - storage_dtype : dtype specifier
@@ -90,7 +90,7 @@ class ArrayProxy(object):
                 - slope : float
                     Scaling factor for resulting data (default: 1.0)
                 - inter : float
-                    Intercept for rescaled dadta (default: 0.0)
+                    Intercept for rescaled data (default: 0.0)
             OR
             Header object implementing ``get_data_shape``, ``get_data_dtype``,
             ``get_data_offset``, ``get_slope_inter``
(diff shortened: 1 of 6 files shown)


demianw:
To close #520 
we need to also include this patch
```
--- a/nibabel/cifti2/parse_cifti2.py
+++ b/nibabel/cifti2/parse_cifti2.py
@@ -147,7 +147,6 @@ class _Cifti2AsNiftiImage(Nifti2Image):
         if self.cifti_img is None:
             raise ValueError('Nifti2 header does not contain a CIFTI2 '
                              'extension')
-        self.cifti_img.data = self.get_data()


 class Cifti2Parser(xml.XmlParser):
```

matthew-brett:
Thanks for taking the initiative on this.

I've been thinking about this for a bit.

Reshape is a good idea in general.

The immediate problem is that we can expect the headers to be pretty big, and the ArrayProxy constructor copies them.

More generally, I would really like to add an alternative constructor signature to ArrayProxy, as in

```
    # alternative signature 0 (current signature)
    def __init__(self, file_like, header, mmap=True):
```

```
    # alternative signature 1 (new signature)
    def __init__(self, file_like, shape, storage_dtype, offset=0, slope_inter=None, mmap=True)
```

Then we could use the second signature for the reshaped arrayproxy.  The `mmap` argument is already keyword only, so this won't mess up calls with the current signature.  

Then - it would be good to have more tests for new constructor and the reshape.

nibabel/cifti2/parse_cifti2.py
```diff
@@ -147,7 +147,6 @@ def __init__(self, dataobj, affine, header=None,
         if self.cifti_img is None:
             raise ValueError('Nifti2 header does not contain a CIFTI2 '
                              'extension')
-        self.cifti_img.data = self.get_data()
```
matthew-brett: Or could be `self.cifti_img.dataobj = self.dataobj`.  This doesn't load the data, but keeps it as an array proxy.

effigies: I'm not sure the `cifti_img` attribute is doing any work at all. Even the check on the header is replicated [here](https://github.com/effigies/nibabel/blob/c6192ad20b392c13e4f5f1ccdb2a8046f4a2ac0d/nibabel/cifti2/cifti2.py#L1334-L1341). I think it may be a refactoring hold-over that never got cleaned up. I'm pretty sure we could, without losing functionality, rewrite the entire class as:

```Python
class _Cifti2AsNiftiImage(Nifti2Image):
    header_class = _Cifti2AsNiftiHeader
    makeable = False
```

effigies:
@demianw Great! Until this PR is finished, you can use c6192ad.

@matthew-brett I'll see what I can do. I've got a lot on my plate at the moment, so this may take a little bit of time to iterate through.

effigies:
@matthew-brett It looks like the `FutureWarning` has been on `ArrayProxy.header` since the 2.0 release, so I think it's reasonable to fast-track deprecation to `3.0`.

How about this approach:

* Add (and document) signature 1
* Add `.from_header` `classmethod`
* Update signature 0 docs to point to `.from_header`, raise a `FutureWarning` if signature 0 is used
  * 3.0 (along with `.header` removal)
    * Move signature 0 to `DeprecationWarning`
    * Remove from docstring
  * 4.0
    * Remove signature 0 support

effigies:
No tests yet. Want to make sure I haven't broken old tests first.

matthew-brett:
I like the deprecation scheme...

nibabel/arrayproxy.py
```diff
@@ -45,15 +45,22 @@ class ArrayProxy(object):
...
+
+            * get_data_shape
+            * get_data_dtype
+            * get_data_offset
+            * get_slope_inter
+
+            The header should also have a 'copy' method.  This requirement will
+            go away when the deprecated 'header' property goes away.
+
+        (2) ArrayProxy(file_like, shape, storage_dtype,
```
matthew-brett: Now I think of it, maybe the second argument can be either:

* the header (as now)
* a tuple of length 2 - 5 containing (shape, storage_dtype, [offset [slope, [inter]]]).

Then the signature can stay the same length, at the expense of slightly greater complexity of explanation / signature.  What do you think?

effigies: I think that's a little less awkward. Do we still want to deprecate the `header` path, then?

matthew-brett: No, I think we can leave the header path.  I suppose we might rename the input argument, hoping that no-one out there is using `ArrayProxy(header=something)`...

effigies:
@matthew-brett Ready for review.

nibabel/arrayproxy.py
```diff
@@ -90,22 +107,30 @@ def __init__(self, file_like, header, mmap=True):
...
+            par = (spec.get_data_shape(),
+                   spec.get_data_dtype(),
+                   spec.get_data_offset(),
+                   1. if slope is None else slope,
+                   0. if inter is None else inter)
+            # Reference to original header; we will remove this soon
+            self._header = spec.copy()
+        elif 2 <= len(spec) <= 5:
+            optional = (0, 1., 0.)
+            par = spec + optional[len(spec) - 2:]
```
matthew-brett: Nice.

nibabel/arrayproxy.py
```diff
@@ -64,17 +66,32 @@ class ArrayProxy(object):
...
         """ Initialize array proxy instance
 
         Parameters
         ----------
         file_like : object
             File-like object or filename. If file-like object, should implement
             at least ``read`` and ``seek``.
-        header : object
+        spec : object or tuple
+            Tuple must have length 2-5, with the following fields:
```
matthew-brett: Suggest 'values' instead of 'fields' - 'fields' (to me) implies names, which might be confusing.

matthew-brett: Anal, I know, but I prefer full-stops at the end of sentences in the docstring.

nibabel/arrayproxy.py
```diff
@@ -90,22 +107,30 @@ def __init__(self, file_like, header, mmap=True):
...
-        self._inter = 0.0 if self._inter is None else self._inter
+        self._shape, self._dtype, self._offset, self._slope, self._inter = par
+        # Permit any specifier that can be interpreted as a numpy dtype
+        self._dtype = np.dtype(self._dtype)
         self._mmap = mmap
-        # Reference to original header; we will remove this soon
-        self._header = header.copy()
 
     @property
+    @deprecate_with_version('ArrayProxy.header deprecated', '2.2', '3.0')
```
matthew-brett: Nice.

nibabel/arrayproxy.py
```diff
@@ -162,6 +187,16 @@ def __getitem__(self, slicer):
         # Upcast as necessary for big slopes, intercepts
         return apply_read_scaling(raw_data, self._slope, self._inter)
 
+    def reshape(self, shape):
```
matthew-brett: One-liner docstring?

matthew-brett: Would be good to allow -1 as element in shape, for compatibility with numpy array reshape.

nibabel/arrayproxy.py
```diff
@@ -162,6 +187,16 @@ def __getitem__(self, slicer):
...
+        size = np.prod(self._shape)
+        if np.prod(shape) != size:
+            raise ValueError("cannot reshape array of size {:d} into shape "
+                             "{!s}".format(size, shape))
+        return ArrayProxy(file_like=self.file_like,
```
matthew-brett: Maybe `self.__class__(file_like=self.file_like ` ... to allow subclassing.

nibabel/cifti2/parse_cifti2.py
```diff
@@ -121,33 +121,9 @@ def _chk_pixdims(hdr, fix=False):
...
 
 class _Cifti2AsNiftiImage(Nifti2Image):
+    """ Load a NIfTI2 image with a Cifti2 header """
     header_class = _Cifti2AsNiftiHeader
-    files_types = (('image', '.nii'),)
-    valid_exts = ('.nii',)
     makeable = False
-    rw = True
-
-    def __init__(self, dataobj, affine, header=None,
```
matthew-brett: I guess this is because the `cifti_img` attribute is not being used elsewhere?

Do we want to raise an error if no cifti extension found?

What is this subclass now for?

effigies: The point of the subclass is to be a `Nifti2Image` with [this header](https://github.com/effigies/nibabel/blob/ad353403ef2747e9813c4ea4c8a666742bf0817e/nibabel/cifti2/parse_cifti2.py#L79-L120), which has different constraints from a standard `Nifti2Header`.

matthew-brett:
We might also want to set `self._header = None` for the tuple input argument case, and test that the `header` attribute correctly returns None.

effigies:
I do set the default `_header` attribute to `None`: https://github.com/nipy/nibabel/pull/521/files#diff-644c66bb38fc2e28c018a29d87e96d36R69

Added a test.

nibabel/arrayproxy.py
effigies: Done.

nibabel/arrayproxy.py
effigies: Used a period here. The rest seemed to be more fragments than sentences, but I can add periods there, if they're preferred, stylistically.

nibabel/arrayproxy.py
```diff
@@ -162,6 +187,31 @@ def __getitem__(self, slicer):
...
+
+        ``array_proxy.reshape(shape)`` is equivalent to
+        ``np.reshape(array_proxy, shape)``
+        '''
+        size = np.prod(self._shape)
+
+        # Calculate new shape if not fully specified
+        shape_arr = np.asarray(shape)
+        unknowns = shape_arr == -1
+        if len(unknowns) > 1:
```
matthew-brett: This will be boolean array of same length as `shape`.

matthew-brett: Maybe

```
from operator import mul
from functools import reduce
n_unknowns = len([e for e in shape if e == -1])
if n_unknowns > 1:
    raise ValueError("can only specify one unknown dimension")
elif n_unknowns == 1:
    known_size = reduce(mul, shape, -1)
    unknown_size = size // known_size
    shape = [unknown_size if e == -1 else e for e in shape]

File: nibabel/tests/test_arrayproxy.py [hunk 10]
```diff
@@ -185,6 +210,14 @@ def __array__(self):
     assert_equal(arr.shape, shape)
 
 
+def test_reshaped_is_proxy():
+    shape = (1, 2, 3, 4)
+    hdr = FunkyHeader(shape)
+    bio = BytesIO()
+    prox = ArrayProxy(bio, hdr)
+    assert_true(isinstance(prox.reshape((2, 3, 4)), ArrayProxy))
```
matthew-brett: Need tests for -1 element, errors for incorrect size, to many -1s.

effigies:
Thanks for the review. Kind of rushed that last push without testing first.

matthew-brett:
Excellent - thanks.   Good to go from my side.  Any other comments?

effigies:
I'm all set. Thanks for the review. I'll let you merge.

matthew-brett:
An extra thought - what should we do for the other ArrayProxy API objects such as MincImageArray, EcatImageArray, PARRECArrayProxy ?   Leave reshape to generate an AttributeError?  Raise a NotImplementedError?

effigies:
Given that the way to reshape has been with `reshape_dataobj`, I was intending to leave it be. If we do add `reshape` in whatever  form to the other proxies, I think we'd need to switch that to a `try`/`catch` block, rather than a `hasattr` check, though.

matthew-brett:
Thanks - let's put this in now and reflect at leisure about the other classes.

PR #525, opened 2017-03-31
Fix link to Wikipedia/Quaternion

fepegar:


doc/source/nifti_images.rst
```diff
@@ -263,7 +263,7 @@ The qform affine
 
 This affine can be calculated from a combination of the voxel sizes (entries 1
 through 4 of the ``pixdim`` field), a sign flip called ``qfac`` stored in
-entry 0 of ``pixdim``, and a `quaternion <wikipedia quaternion>`_ that can be
```
matthew-brett: Actually I think the correct fix is 

```
`quaternion <wikipedia quaternion_>`_
```

which picks up the pre-defined link in the `links_names.txt` file...

matthew-brett:
Will push up a fix for this - thanks for pointing it out.  Closing this one in favor of the fix above.

Issue #518, opened 2017-03-20
Discrepancy between nibabel's as_closest_canonical and FreeSurfer's mri_convert


oesteban:
Playing around orientations, I have an example image for which `mri_info` gives me this:
```
          type: nii
    dimensions: 256 x 256 x 186
   voxel sizes: 0.8984, 0.8984, 0.9000
          type: FLOAT (3)
           fov: 229.990
           dof: 0
        xstart: -115.0, xend: 115.0
        ystart: -115.0, yend: 115.0
        zstart: -83.7, zend: 83.7
            TR: 7.24 msec, TE: 0.00 msec, TI: 0.00 msec, flip angle: 0.00 degrees
       nframes: 1
       PhEncDir: UNKNOWN
ras xform present
    xform info: x_r =   0.0000, y_r =   0.0000, z_r =  -1.0000, c_r =    -1.8686
              : x_a =   0.0000, y_a =  -1.0000, z_a =   0.0000, c_a =    15.5558
              : x_s =  -1.0000, y_s =   0.0000, z_s =   0.0000, c_s =     9.8568
Orientation   : IPL
Primary Slice Direction: sagittal

voxel to ras transform:
                0.0000   0.0000  -0.9000    81.8314
                0.0000  -0.8984   0.0000   130.5510
               -0.8984   0.0000   0.0000   124.8520
                0.0000   0.0000   0.0000     1.0000

voxel-to-ras determinant 0.72641

ras to voxel transform:
               -0.0000  -0.0000  -1.1131   138.9715
                0.0000  -1.1131   0.0000   145.3150
               -1.1111   0.0000   0.0000    90.9238
                0.0000   0.0000   0.0000     1.0000
```
So, I start with an image in **IPL** orientation and aligned with the standard axes. Using nibabel to read these parameters, I get exactly the same orientation code and affine.

If I use `mri_convert --out_orientation RAS -i input.nii.gz -o output_RAS.nii.gz`, I get the following:
```
          type: nii
    dimensions: 256 x 256 x 186
   voxel sizes: 0.8984, 0.8984, 0.9000
          type: FLOAT (3)
           fov: 229.990
           dof: 0
        xstart: -115.0, xend: 115.0
        ystart: -115.0, yend: 115.0
        zstart: -83.7, zend: 83.7
            TR: 7.24 msec, TE: 0.00 msec, TI: 0.00 msec, flip angle: 0.00 degrees
       nframes: 1
       PhEncDir: UNKNOWN
[...]

PR #526, opened 2017-04-01
Improve as_closest_canonical

markhymers:
Hi,
At present, as_closest_canonical leaves dim_info as it was before any axis re-ordering.  This patch fixes that and adds tests.

Thanks,

Mark


effigies:
Possibly related to #518? cc @oesteban

nibabel/nifti1.py
```diff
@@ -1260,7 +1260,7 @@ def set_dim_info(self, freq=None, phase=None, slice=None):
         This is stored in one byte in the header
         '''
         for inp in (freq, phase, slice):
-            if inp not in (None, 0, 1, 2):
+            if inp is not None and inp not in (0, 1, 2):
```
matthew-brett: Why this change?

nibabel/funcs.py
```diff
@@ -224,7 +224,18 @@ def as_closest_canonical(img, enforce_diag=False):
     # we need to transform the data
     arr = img.get_data()
     t_arr = apply_orientation(arr, ornt)
-    return img.__class__(t_arr, out_aff, img.header)
+    # Also apply the transform to the dim_info fields
+    new_hdr = img.header.copy()
```
matthew-brett: This is a bit tricky, because the output image does not have to be Nifti type, and therefore the header need not have 'get_dim_info'.

I wonder whether we need a new image method, `transpose`, that deals with this case.  The default implementation could go in `SpatialImage` and be overridden by Nifti.

nibabel/funcs.py
```diff
@@ -224,7 +224,18 @@ def as_closest_canonical(img, enforce_diag=False):
...
+    new_dim = list(new_hdr.get_dim_info())
+    for idx, value in enumerate(new_dim):
```
matthew-brett: Sorry if I'm missing something obvious, by why would we not also permute the `None`s?

Maybe something like this:

```
    indices = ornt[:, 0].astype(np.intp)
    new_dim = np.array(new_hdr.get_dim_info(), dtype=object)[indices]
```

nibabel/tests/test_funcs.py
```diff
@@ -150,3 +157,22 @@ def test_closest_canonical():
...
+    xyz_img = as_closest_canonical(img)
+    assert_false(img is xyz_img)
+    # Check both the original and new objects
+    assert_true(img.header.get_dim_info() == (0, 1, 2))
+    assert_true(xyz_img.header.get_dim_info() == (0, 2, 1))
+    out_arr = xyz_img.get_data()
+    assert_array_equal(out_arr, np.transpose(arr, (0, 2, 1, 3)))
+
+    # same axis swap but with None dim info (except for slice dim)
+    img.header.set_dim_info(None, None, 2)
```
matthew-brett: This probably the same question as before, but why isn't the correct answer `(None, 2, None)` after this permutation?

nibabel/funcs.py
markhymers: Hi Matthew,

We don't permute the None's because this array is (freq, phase, slice), *not* the order of the actual data axes.

As an example, if we have a "normal" MRI image in which
freq = 0, phase = 1, slice = 2
and we permute dimensions 1 and 2, we obviously end up with
freq = 0, phase = 2, slice = 1

If we have an image in which one or more of freq, phase and slice are None (e.g. spiral images):
freq = None, phase = None, slice = 1
if we then permute axes 1 and 2, we move the slice dimension to dimension 2, but we still don't have any mapping for the freq and phase attributes:
freq = None, phase = None, slice = 2

I don't know if I'm doing a good job of explaining this...

nibabel/tests/test_funcs.py
markhymers: See my comment above.  We still don't have information about frequency and phase in this case, but the slice dimension has now been switched from dimension 2 to dimension 1.

nibabel/nifti1.py
markhymers: As the commit message says, it came up whilst I was testing the code below.

At least with python 3.4, every time this function gets used, we get python FutureWarnings about comparing None with == (which is what in does).  The only way I could see to work around it was to split None out of the in - better suggestions welcome.

nibabel/funcs.py
markhymers: Once I get agreement that the rest of the comments make sense and you're happy with the general idea, I can try and implement this.

nibabel/nifti1.py
matthew-brett: I don't get that warning myself, but that's fine - maybe add a comment above to explain.

nibabel/funcs.py
matthew-brett: Ah - yes - I remember.   Values are (which axis is frequency, which axis is phase, which axis is slice) where None has the obvious meaning.

nibabel/funcs.py
markhymers: Ok, new version which factors out the logic into a transpose function pushed.

markhymers:
Hi @matthew-brett ,

Just a quick check to see if there's anything else you would like to do with this branch.

Thanks,

Mark

matthew-brett:
Nice job.  Now I see the code, I wonder about two things:

* can we get rid of the `new_aff` input argument?  I think we can just calculate the affine using the algorithm in the `as_closest_canonical` function;
* or - would this method actually be better placed in the header, as an `as_oriented` method or similar.  In this case, the `as_closest_canonical` function would look for an `as_oriented` method in the header, and apply it if present.  In that case, the `as_oriented` code could just go in the Nifti1 header object.

What do you think?

markhymers:
We could get rid of the new_aff argument, but I'm not sure it's worthwhile - this way, the function is general; it can use whichever orientation / affine the user required.

As for the second point, I'm not really sure what it wins us  Unless you're entirely tied to doing it that way, I'd rather leave it as it-is for now.

matthew-brett:
I still think the `new_aff` argument is redundant.   When we are `transpose`ing the image, we will nearly always want to use the corresponding affine, and when we don't, we can just set the affine in a second step, as in:

```
transposed = img.transpose(ornt)
reoriented = img.__class__(transposed.dataobj, new_affine, transposed.header)
```

Let's keep the transposing in the image then, but I prefer the name `as_oriented` for the method, to distinguish it from the numpy `transpose` method, which takes axis indices rather than the `ornt` argument.

effigies:
Just a bump on this. How are you each feeling about the state of this and any more work before merging?

matthew-brett:
Sorry - updating on this - I am happy to make the changes I have suggested if @markhymers prefers.

effigies:
I'm having a look at this now.

I feel that expecting users to supply the updated affine, rather than updating the affine to match the new orientation, is asking for trouble. I'm okay with @matthew-brett's solution, though I recognize that making users update the affine in a second step is a little clunky. Another option would be to set `new_aff=None` as the default, and derive the updated affine if one isn't supplied.

Possibly I'm missing something; this comment is based off a pretty brief study of the code.

matthew-brett:
My feeling is that changing to an arbitrary affine is a different job from reorienting the image, so changing to an arbitrary affine shouldn't be part of the reorientation.  But I could accept an off-by-default affine argument, if y'all feel very strongly about it.

effigies:
I was reading this as though swapping out the affine would for some reason be a common pattern. But I agree that it's really a separate process. For the same reason, I'd prefer dropping `new_aff`, but I'm also okay with the off-by-default argument.

PR #528, opened 2017-04-10
BF: Fix Nifti1Header.get/set_intent for unknown intent codes

pauldmccarthy:
The `nibabel.nifti1.Nifti1Header.get_intent` and `set_intent` methods fail for intent codes which are not listed in `nibabel.nifti1.intent_codes`. For example:

```
In [1]: import nibabel as nib
   ...: import numpy   as np
   ...:
   ...: data = np.random.random((10, 10, 10))
   ...: xform = np.eye(4)
   ...:
   ...: img = nib.nifti1.Nifti1Image(data, xform)
   ...: hdr = img.header
   ...:
   ...: # Getting/setting the intent code directly works fine
   ...: hdr['intent_code'] = 2006
   ...: print(hdr['intent_code'])
   ...:
2006

In [2]: print(hdr.get_intent())
KeyError                                  Traceback (most recent call last)
<ipython-input-2-b920d8c34544> in <module>()
----> 1 print(hdr.get_intent())

/Users/paulmc/Projects/nibabel/nibabel/nifti1.pyc in get_intent(self, code_repr)
   1305             label = code
   1306         elif code_repr == 'label':
-> 1307             label = recoder.label[code]
   1308         else:
   1309             raise TypeError('repr can be "label" or "code"')

KeyError: 2006

In [3]: hdr.set_intent(2006)
KeyError                                  Traceback (most recent call last)
<ipython-input-3-c617589adad8> in <module>()
----> 1 hdr.set_intent(2006)

/Users/paulmc/Projects/nibabel/nibabel/nifti1.pyc in set_intent(self, code, params, name)
   1357         '''
   1358         hdr = self._structarr
-> 1359         icode = intent_codes.code[code]
   1360         p_descr = intent_codes.parameters[code]
   1361         if len(params) and len(params) != len(p_descr):

KeyError: 2006
```

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -1347,7 +1347,7 @@ class Nifti1Header(SpmAnalyzeHeader):
             defaults to ().  Unspecified parameters are set to 0.0
         name : string
             intent name (description). Defaults to ''
-        allow_unknown : bool
+        allow_unknown : {False, True}, optional
             Allow unknown integer intent codes. If False (the default),
             a KeyError is raised on attempts to set the intent
             to an unknown code.
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -1375,6 +1375,8 @@ class Nifti1Header(SpmAnalyzeHeader):
         >>> hdr.get_intent()
         ('f test', (0.0, 0.0), '')
         >>> hdr.set_intent(9999, allow_unknown=True) # unknown code
+        >>> hdr.get_intent()
+        ('unknown code 9999', (), '')
         '''
         hdr = self._structarr
         known_intent = code in intent_codes
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -1301,13 +1301,20 @@ class Nifti1Header(SpmAnalyzeHeader):
         hdr = self._structarr
         recoder = self._field_recoders['intent_code']
         code = int(hdr['intent_code'])
+        known_intent = code in recoder
         if code_repr == 'code':
             label = code
         elif code_repr == 'label':
-            label = recoder.label[code]
+            if known_intent:
+                label = recoder.label[code]
+            else:
+                label = ''
         else:
             raise TypeError('repr can be "label" or "code"')
-        n_params = len(recoder.parameters[code])
+        if known_intent:
+            n_params = len(recoder.parameters[code])
+        else:
+            n_params = 0
         params = (float(hdr['intent_p%d' % (i + 1)]) for i in range(n_params))
         name = asstr(np.asscalar(hdr['intent_name']))
         return label, tuple(params), name
@@ -1356,8 +1363,13 @@ class Nifti1Header(SpmAnalyzeHeader):
         ('f test', (0.0, 0.0), '')
         '''
         hdr = self._structarr
-        icode = intent_codes.code[code]
-        p_descr = intent_codes.parameters[code]
+        known_intent = code in intent_codes
+        if known_intent:
+            icode = intent_codes.code[code]
+            p_descr = intent_codes.parameters[code]
+        else:
+            icode = code
+            p_descr = 3
         if len(params) and len(params) != len(p_descr):
             raise HeaderDataError('Need params of form %s, or empty'
                                   % (p_descr,))
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -13,6 +13,7 @@ NIfTI1 format defined at http://nifti.nimh.nih.gov/nifti-1/
 from __future__ import division, print_function
 import warnings
 from io import BytesIO
+from six import string_types
 
 import numpy as np
 import numpy.linalg as npl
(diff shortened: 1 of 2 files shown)


nibabel/nifti1.py
```diff
@@ -1301,13 +1301,20 @@ def get_intent(self, code_repr='label'):
         hdr = self._structarr
         recoder = self._field_recoders['intent_code']
         code = int(hdr['intent_code'])
+        known_intent = code in recoder
         if code_repr == 'code':
             label = code
         elif code_repr == 'label':
-            label = recoder.label[code]
+            if known_intent:
```
matthew-brett: Maybe:

```
label = recoder.label[code] if known_intent else ''
```

nibabel/nifti1.py
```diff
@@ -1301,13 +1301,20 @@ def get_intent(self, code_repr='label'):
...
+                label = recoder.label[code]
+            else:
+                label = ''
```
matthew-brett: Maybe label something more descriptive like `'unknown code ' + str(code)`?

nibabel/nifti1.py
```diff
@@ -1301,13 +1301,20 @@ def get_intent(self, code_repr='label'):
...
         else:
             raise TypeError('repr can be "label" or "code"')
-        n_params = len(recoder.parameters[code])
+        if known_intent:
```
matthew-brett: Maybe

```
n_params = len(recoder.parameters[code]) if known_intent else 0
```

pauldmccarthy:
Hi Matthew,

My immediate use case is FSL's FNIRT, which uses some custom intent codes for storing displacement fields, spline coefficients.

But more generally, as far as I can tell, the NIFTI spec (starting at `nifti1.h:567`), doesn't appear to disallow the use of new, or custom, intent codes. So, going by the spec, codes which are not in the header should not be considered "invalid".

This PR is admittedly a bit moot, as it is possible to just access the fields directly through e.g. `hdr['intent_code']`. So I don't really mind if the changes do not get made. But I thought it was worth raising the issue.

I have nothing against any of your suggestions - keeping the error type the same, and adding an `allow_invalid` (or perhaps we could call it `allow_unknown` :) ) flag. I'm happy to make the changes if you think it's worth including.

matthew-brett:
Ah - I didn't know FSL was in the habit of using custom codes.   In that case, I think it's OK to go ahead as you have here, and allow unknown codes.

I would prefer a key error for invalid labels though, to be compatible with previous code.

effigies:
@pauldmccarthy Does FSL have names for these intent codes, e.g. `FSL_INTENT_*`? For documentation purposes, it'd be nice to have explicit support for known non-standard codes.

pauldmccarthy:
Hi Chris,

Yes - here are all of the ones that I am aware of:
```python
# FSL-specific intent codes

# FNIRT (warpfns/fnirt_file_reader.h:104)
FSL_FNIRT_DISPLACEMENT_FIELD            = 2006
FSL_CUBIC_SPLINE_COEFFICIENTS           = 2007
FSL_DCT_COEFFICIENTS                    = 2008
FSL_QUADRATIC_SPLINE_COEFFICIENTS       = 2009

# TOPUP (topup/topup_file_io.h:104)
FSL_TOPUP_CUBIC_SPLINE_COEFFICIENTS     = 2016
FSL_TOPUP_QUADRATIC_SPLINE_COEFFICIENTS = 2017
FSL_TOPUP_FIELD                         = 2018
```

effigies:
Oh great! Would you mind adding those to [`nifti1.intent_codes`](https://github.com/nipy/nibabel/blob/master/nibabel/nifti1.py#L235)?

@matthew-brett Any objections here?

pauldmccarthy:
Sure, I will do this over the weekend :)

pauldmccarthy:
I've added an `allow_unknown` flag to `set_intent`; one issue with this is that `set_intent(..., allow_unknown=True)` will work when an integer code is specified, but will *not* work when the code is specified as a string. 

The only way this could be made to work would be to use a default integer code (e.g. `0`) for unknown string intents. I thought it better to just disallow setting the intent to an unknown string, even if `allow_unknown=True`.

nibabel/nifti1.py
```diff
@@ -1354,19 +1374,30 @@ def set_intent(self, code, params=(), name=''):
         >>> hdr.set_intent('f test')
         >>> hdr.get_intent()
         ('f test', (0.0, 0.0), '')
+        >>> hdr.set_intent(9999, allow_unknown=True) # unknown code
```
matthew-brett: Add `get_intent` line to show output.

effigies:
This LGTM. Comments addressed, it looks like. Will merge tomorrow AM if you don't have any objections, @matthew-brett.

nibabel/nifti1.py
```diff
@@ -1331,6 +1347,10 @@ def set_intent(self, code, params=(), name=''):
             defaults to ().  Unspecified parameters are set to 0.0
         name : string
             intent name (description). Defaults to ''
+        allow_unknown : bool
```
matthew-brett: Should be `allow_unknown : {False, True}, optional` to signal default (first value in curlies) and the fact that it's optional.

matthew-brett:
One last comment, but yes, otherwise good to go.

pauldmccarthy:
Sorry for the delay ... docstring updated.

nibabel/nifti1.py
```diff
@@ -232,6 +233,17 @@
     (2003, 'rgb vector', (), "NIFTI_INTENT_RGB_VECTOR"),
     (2004, 'rgba vector', (), "NIFTI_INTENT_RGBA_VECTOR"),
     (2005, 'shape', (), "NIFTI_INTENT_SHAPE"),
+    # FSL-specific intent codes - FNIRT
```
matthew-brett: Sorry to be a pain - but do you have a reference for where these codes can be found?

pauldmccarthy: The FSL 5.0.9 source code ... there is a good chance that, in the near future, it will be available/linkable online via a git repository. But for the time being my reference is  http://fsl.fmrib.ox.ac.uk/fsldownloads :)

matthew-brett: Ouch!   Maybe a pointer to the file / line number where you found it, so we can put in the link, when there's something to link to?

pauldmccarthy: Oh, sure - the FNIRT codes can be found in `$FSLDIR/src/warpfns/fnirt_file_reader.h:104`, and the TOPUP codes in `$FSLDIR/src/topup/topup_file_io.h:104`. Note that this is for 5.0.9 - I haven't downloaded 5.0.10 yet (released a few days ago), so they may have changed (although probably haven't). I will check tomorrow, and update here if the line numbers have changed in 5.0.10.

matthew-brett: Thanks - or put the 5.09 data in there - approximate should be good enough to find it again.

matthew-brett:
Excellent - many thanks for your persistence.

PR #529, opened 2017-04-14
MRG: move test_utils to test_volumeutils

matthew-brett:
Module tested is volumeutils, so test file badly named.


matthew-brett:
Thanks @effigies

PR #533, opened 2017-05-15
Improve error handling for `img.__getitem__`

arokem:
Usually, when using `__getitem__`, one passes an index/slice to the function. In the current implementation, this would raise the relatively cryptic: 

    TypeError: __getitem__() takes 1 positional argument but 2 were given

And the much more explanatory error message implemented in `__getitem__` would remain unraised. 

My proposal: implement an `idx` kwarg here, so that indexing gets through to the error message. I also implemented a test (and learned a few things about `assert_raises` along the way!).

### DIFF
--- a/nibabel/spatialimages.py
+++ b/nibabel/spatialimages.py
@@ -656,7 +656,7 @@ class SpatialImage(FileBasedImage):
                      klass.header_class.from_header(img.header),
                      extra=img.extra.copy())
 
-    def __getitem__(self):
+    def __getitem__(self, idx=None):
         ''' No slicing or dictionary interface for images
         '''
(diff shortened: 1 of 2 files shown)


nibabel/spatialimages.py
```diff
@@ -656,7 +656,7 @@ def from_image(klass, img):
                      klass.header_class.from_header(img.header),
                      extra=img.extra.copy())
 
-    def __getitem__(self):
+    def __getitem__(self, idx=None):
```
matthew-brett: Oops - well spotted.  But - is `__getitem__` ever called without arguments?  I mean, why not just:

```
def __getitem__(self, idx):
```

?

matthew-brett:
Thanks Ariel.

PR #536, opened 2017-05-21
fix: inferior-posterior to inferior-superior

R3DDY97:
There will be inferior-superior axis 
not  inferior-posterior in human brain anatomy


matthew-brett:
Well spotted - and thanks for taking the time to do a PR.

PR #539, opened 2017-06-07
DOC: fix dicom link

matthew-brett:
Thanks to Joan Massich for pointing out the broken link and Chris M for
finding the corresponding link.


matthew-brett:
Thanks - merging.

PR #544, opened 2017-06-22
RF: Move reorientation into image classes

effigies:
#526 seems to have stalled. I'd like to incorporate these changes, as well as those discussed by @matthew-brett and myself.

On renaming `transpose` to `as_oriented`, while I agree that distinguishing from `np.transpose` may be desirable, I disagree that this would be a more appropriate name. That is, the `ornt` matrix doesn't describe an orientation, but a transformation. I would expect `as_oriented` to be idempotent for a given orientation, which it is not. With `ornt = [[0, -1], [2, 1], [1, 1]]`, I could flip back and forth between `RAS` and `LSA` all day long.

I also moved the identity transform check into `transpose`, and let `Nifti1Pair.transpose` call `SpatialImage.transpose`, which I think helps clarify which parts of the operation are specific to NIfTIs.

Closes #526.


effigies:
Can't request a review, but @markhymers, if you have any feedback, please jump in.

nibabel/nifti1.py
```diff
@@ -1933,6 +1934,40 @@ def set_sform(self, affine, code=None, **kwargs):
             else:
                 self._affine[:] = self._header.get_best_affine()
 
+    def transpose(self, ornt):
+        """Apply an orientation change and return a new image
+
+        If image already has orientation, return the original image, unchanged
```
matthew-brett: I guess you mean, if `ornt` is the identity?

nibabel/spatialimages.py
```diff
@@ -483,3 +484,34 @@ def orthoview(self):
         """
         return OrthoSlicer3D(self.dataobj, self.affine,
                              title=self.get_filename())
+
+
+    def transpose(self, ornt):
+        """Apply an orientation change and return a new image
+
+        If image already has orientation, return the original image, unchanged
```
matthew-brett: See comment above.

nibabel/nifti1.py
effigies: You're right. I wrote that bit while I thought `ornt` was the target orientation, not the transformation.

effigies:
I agree transpose isn't great. What about `reorient`?

matthew-brett:
How do you feel about `as_reoriented`?

effigies:
Fine by me.

matthew-brett:
Then fine by me too.  Merge when tests pass?

effigies:
Sounds good. I'll add a changelog entry, as well.

PR #550, opened 2017-08-18
ENH: Add image slicing

### LINKED ISSUE #489: Time-slicing spatial images

effigies:
See https://mail.python.org/pipermail/neuroimaging/2017-August/001501.html

Closes #489.


nibabel/funcs.py
```diff
@@ -219,3 +219,39 @@ def _aff_is_diag(aff):
...
+    if mask is None:
+        mask = img
+    elif not np.allclose(img.affine == mask.affine):
+        raise ValueError('Affine for image does not match affine for mask')
+
+    bounds = np.sort(np.vstack(np.nonzero(mask.get_data())))[:, [0, -1]]
+    x, y, z = bounds
+    new_origin = np.vstack((bounds[:, [0]], [1]))
+
+    new_data = img.get_data()[x[0]:x[1] + 1, y[0]:y[1] + 1, z[0]:z[1] + 1]
```
effigies: @mwaskom Note the slicing... I'm pretty sure the lines I put in the email were dropping one too many voxels in each dimension.

mwaskom:
This is very nice. It might make sense to expose a slightly more general API that accepts the bounds. In my case I wanted to crop to a volume mask, but others might want to crop based on bound values that they obtain in some other fashion. Does that seem reasonable?

mwaskom:
Also it might be nice to have a `margins` (or similar parameter) that will add the number of voxels to pad on each side. In my case I have added that so that I don't lose data if there is a slight registration error between the cropped image and the image I am going to register to it.

effigies:
@matthew-brett Want to have a quick look at this? I haven't incorporated @mwaskom's comments, yet, but if this seems reasonable, I can polish it up before the release.

effigies:
@mwaskom I made the changes you suggested, if you'd like to have a look. Does it make sense to also include a `crop_image_to_mask`, which finds the bounds for you and calls `crop_image`? Or just leave that functionality to users?

matthew-brett:
Guys - what do you think about pulling over the Nipy image slicing.  It would look something like:

```
cropped = img.slicer[1:-1, 2:-2, ...]
```

to drop the left and rightmost pixels and 2 posterior and anterior pixels?  See: https://github.com/practical-neuroimaging/analysis-clinic/blob/master/nipy_image_slicing.ipynb

satra:
@matthew-brett - i think being able to slice would be nice. my worry is that this affords "easy" but potentially incorrect downsampling, and that users really should know how the image was acquired in the scanner (line, epi, 3d, etc.,.) before applying the operation.

see: https://mail.python.org/pipermail/neuroimaging/2017-September/001528.html

matthew-brett:
@satra - do you mean, you think we shouldn't implement this in case people mis-use it?  I guess we can make them do it by hand?  Or would you be OK with a big warning in the docstring?

satra:
@matthew-brett - it's more a worry than me saying we shouldn't implement this. 

slicing is so easy in numpy that as soon as you have the data in a numeric array, people will end up doing it. and in many cases they may have done the right thing (smoothing/filtering) prior to the slicing step. so we shouldn't take away easy slicing for that reason. the key that this functionality adds is the adjustment of headers and other things to reflect the outcome of that operation.

putting a note in the docstring is a good starting point.

mwaskom:
If people are going to have to call a filtering function anyway, it seems to erase the convenience of having the slice notation. At that point, why not just have one function that filters and downsamples?

effigies:
Using slices for cropping would be fine with me, but I think I'm less comfortable with adding downsampling. That makes more sense to me as a function where options like filtering can be specified.

I like the nipy image approach, but I also like that nibabel added enough friction to make me think about whether I wanted to switch to nipy to use it, or use some other tool to filter.

effigies:
Before I put work into this, my approach would be to only permit slices and ellipses, and constrain slices not to specify step (except 1 or `None`). I'd allow cropping in the temporal dimension as well, since that's trivial.

This would be in `SpatialImage`, but I think the affine update step would need to be overridden in images whose first three dimensions aren't spatial.

Seem reasonable?

satra:
@effigies - sounds good to me.

if there is a function that's being created to support a functional form one could potentially allow something like this:

`img.slice(..., pre_filtered=True)`

under that setting you can allow steps.

effigies:
@satra IMO, by talking actual filtering, we're getting into nipy/nilearn territory (possibly nistats? Not sure, haven't looked closely enough yet). Obviously there's a blurry line, but my idea of nibabel's scope is just wrangling data formats, while leaving stats and signal processing to other tools.

I don't think it would be good to get into a position where users have to make a choice between downsampling in the high-level or low-level package, and consider the cascading effects of that decision. If it's needed here, then I think we'll want broad input on how to implement it so that higher level libraries use our algorithms, rather than offer alternatives.

@matthew-brett I'm going to declare down-sampling out-of-scope for this PR, but I'm okay with adding bounding slices. And I'm okay with allowing `img[[0], ...]` but not `img[0, ...]`, since the former keeps the dimensions intact while being consistent with numpy indexing, but the latter has to sacrifice numpy consistency for coherency as a `SpatialImage`.

@mwaskom This PR was kicked off by your idea. Would `SpatialImage.__getitem__` be a sufficient interface, or should `funcs.crop_image` stay around? The bounding behavior of `__getitem__` would need to be different from what I currently have (using inclusive bounds), and would require the user to calculate margins.

satra:
@effigies - how about the following keyword instead `allow_downsampling=True` with `False` being the default.  the intent here is to allow regular options for slicing including steps, but it serves as a point to indicate to the potential user that this means something specific.

matthew-brett:
Hum  .  So this is something that I often do for various reasons so I would
miss it if I could not do it.

I think we need to think of use cases.   I often do this when I want a
quick and dirty smaller version of a high resolution file.   Our I'm going
to smooth it later down the line.   Of course this can also be used to
subsample in time and other dimensions.

Would you propose trying to work out the optimal smoothing by default?
Also in time etc?

satra:
@matthew-brett - that's why i propose having the keyword (`allow_downsampling`) - precisely for those situations where you know what you are doing, whether for quick and dirty, or my data are band-limited.

there are use cases where downsampling is never an issue and smoothing/filtering does not need to be used. for example, with slice removal (if the same slices are removed for all volumes)

satra:
@matthew-brett - regarding the comment of trying to figure out optimal filters, i would stay away from that in nibabel, unless nibabel wanted to have utilities to show power spectrum (space/time), etc.,.

effigies:
I guess I'm not sure what API you're proposing.

```Python
class SpatialImage(...):
    def slice(*args, allow_downsampling=False):
        ...
```

What are `*args`, and what's the effect of `allow_downsampling`? Don't raise an exception when someone tries to slice with a non-1 step? My biggest issue here is just that the syntactic sugar for slicing is only available to `__getitem__`, which can't take these sorts of arguments.

You could do something like:

```Python
class UnsafeSlicer:
    def __init__(self, img):
        self.img = img

    def __getitem__(self, idx):
        dataobj = self.img.dataobj[idx]
        affine = self.img.slice_affine(idx)
        return self.img.__class__(dataobj, affine, self.img.header)
```

Which could then be called:

```Python
new_img = UnsafeSlicer(img)[::2, -20:-2:3, ..., [0]]
```

The main thing that would need to be added is a function that modifies the affine with bounds updates and subsampling.

satra:
@effigies - perhaps something like this.

```
class SpatialImage(...):
    def slice(*args, allow_downsampling=False):
        ...
    def __getitem__(self, idx):
        if requires_downsampling(idx):
            raise ValueError('slicing requires downsampling, please use img.slice(..., allow_downsampling=True) instead')
        dataobj = self.img.dataobj[idx]
        affine = self.img.slice_affine(idx)
        return self.img.__class__(dataobj, affine, self.img.header)
```
and `*args` would be exactly the same as `idx` in `__getitem__`

effigies:
@satra Okay, that seems manageable.

nibabel/spatialimages.py
```diff
@@ -461,12 +461,118 @@ def from_image(klass, img):
...
+        scales = np.eye(3, dtype=int)
+
+        for i, slicer in enumerate(idx[:3]):
+            new_origin[i] = (slicer.start or 0 if isinstance(slicer, slice)
+                             else np.asanyarray(slicer)[0])
+            if slicer.step is not None:
+                scales[i] = slicer.step
+
+        affine = self.affine.copy()
+        affine[:3, :3] = scales.dot(self.affine[:3, :3])
```
effigies: I need to check that this isn't backwards.

matthew-brett:
I think this is a question of philosophy.   One the one end of the spectrum we go out of our way to stop users doing things that might be incorrect; at the other we try and get out of the user's way.   I think nibabel is / should be at the second end of the spectrum; it's a low-level, if-you-say-so-boss type of library. 

For example, we let the user  ``img.dataobj``, which is fairly low-level object, and we let them slice into this object, without getting in their way.

So, I don't think we should recommend downsampling this way, or even mention it in the docs, but I also don't think we should go out of our way not to do the obvious thing, just in case.   So my vote would be be for allowing `img.slicer[:::2, ...]` and so on.

About implementing ``__getitem__`` - hmm - I don't know.  I have a vague instinct that the slicing shouldn't be a normal operation, and that attaching it to `img.slicer`` will give the correct amount of warning that this is a slightly unusual thing to do, about which you should take some care.

mwaskom:
FWIW in my original mailing list message a while back that started this, I did want to downsample the image, but only to get a volume in anatomical space with the same resolution as my functionals so that I could have a target geometry for transformation. So that's a case where I would be annoyed if nibabel didn't let me downsample without filtering, because I didn't actually care about any artifacts that might have shown up in the low-res anatomy image.

With that said, I don't personally find the slice notation that much more clear or convenient than a function/method.

effigies:
@matthew-brett So that would basically be my `UnsafeSlicer` from https://github.com/nipy/nibabel/pull/550#issuecomment-332556463, with:

```Python
class SpatialImage(...):
    ...

    @property
    def slicer(self):
        return UnsafeSlicer(self)
```

Sound about right? (We can use another besides `UnsafeSlicer`.)

matthew-brett:
@effigies - right.

satra:
@matthew-brett - i'm personally completely fine with nibabel being an "as-you-say-so-boss" library :)  and would be happy to see the `UnsafeSlicer` route.

i just thought that given what was done with `get_fdata`, we went a little out of our way to save people from shooting themselves in the foot.

matthew-brett:
@satra - yes - fair comment - but still I think the `get_data` / `get_fdata` case is a bit different, in that `get_data` was the top-level standard interface, revealing accidental complexity to the user, to do with the scaling factors in the header.   Here I think we have a less visible interface, where the user will have to work out for themselves that they can subsample, and so we can reasonable expect them to have some idea what they are doing.   We should certainly put something about subsampling in the docs though.  I don't suppose you have a few minutes to throw something together do you?   Any old `.rst` file would be fine, I can plumb it into the docs properly later.

satra:
@matthew-brett - will try to put something together. perhaps i can create a notebook that can be reused for this.

matthew-brett:
Notebook is fine - thanks.

nibabel/fileslice.py
```diff
@@ -113,9 +117,16 @@ def canonical_slicers(sliceobj, shape, check_inds=True):
...
-        # int / slice indexing cases
+        # int / [int] / slice indexing cases
         dim_len = shape[n_real]
         n_real += 1
+
+        # Treat [x] as "x:x+1"
+        ary = np.asanyarray(slicer)
+        if ary.shape == (1,):
+            can_slicers.append(slice(int(ary[0]), int(ary[0]) + 1))
+            continue
```
effigies: @matthew-brett I made this change to `canonical_slicers` (and a concomitant one to `is_fancy`) because it's a simple enough transform for a common idiom. It allowed me to reuse a lot of the machinery in these to make slicing and affine calculation straightforward.

Hope that's okay.

nibabel/spatialimages.py
```diff
@@ -325,6 +326,23 @@ class SpatialImage(DataobjImage):
...
+            idx = self._validate(idx)
+            dataobj = self.img.dataobj[idx]
+            affine = self.img.slice_affine(idx)
+            return self.img.__class__(dataobj, affine, self.img.header)
+
+        def _validate(self, idx):
+            idx = canonical_slicers(idx, self.img.shape)
+            for slicer in idx:
+                if isinstance(slicer, int):
+                    raise IndexError("Cannot slice image with integer")
```
effigies: Just pointing this out: I don't want to slice an image like `img.slicer[0, :, :]`, since a numpy user would expect to lose a dimension in that case, and we don't want to drop dimensions. So I'm pushing the `[0]` syntax.

Let me know if y'all disagree.

nibabel/spatialimages.py
```diff
@@ -461,8 +479,66 @@ def from_image(klass, img):
...
+        transform = np.eye(4, dtype=int)
+
+        for i, slicer in enumerate(idx):
+            if isinstance(slicer, slice):
+                if slicer.step == 0:
+                    raise ValueError("slice step cannot be 0")
+                transform[i, i] = slicer.step if slicer.step is not None else 1
+                transform[i, 3] = slicer.start or 0
+            elif isinstance(slicer, int):
+                transform[i, 3] = slicer
```
effigies: Here I'm allowing the possibility that someone might use `slice_affine` with integer indices, even though I don't use it here.

nibabel/spatialimages.py
```diff
@@ -325,6 +326,23 @@ class SpatialImage(DataobjImage):
...
 
+    class Slicer(object):
+        def __init__(self, img):
+            self.img = img
+
+        def __getitem__(self, idx):
+            idx = self._validate(idx)
+            dataobj = self.img.dataobj[idx]
+            affine = self.img.slice_affine(idx)
+            return self.img.__class__(dataobj, affine, self.img.header)
```
effigies: At present, this *always* returns a new image. That's consistent with the `array[:]` idiom, but in other cases we save memory by returning the same image if no change is made. So, if we want to check indices for the case where we're retrieving the whole dataobj, we can do that.

nibabel/spatialimages.py
```diff
@@ -461,8 +479,66 @@ def from_image(klass, img):
...
+            Affine with updated scale and intercept
+        """
+        idx = canonical_slicers(idx, self.shape, check_inds=False)[:3]
+
+        # Transform:
+        # sx  0  0 tx
+        #  0 sy  0 ty
+        #  0  0 sz tz
+        #  0  0  0  1
+        transform = np.eye(4, dtype=int)
```
effigies: By my algebra, to rescale, we do:

```Python
new_affine[:3, :3] = affine[:3, :3].dot([[sx, 0, 0], [0, sy, 0], [0, 0, sz]])
new_affine[:, [3]] = affine.dot([[tx], [ty], [tz], [1]])
```

Where `s*` are the steps, and `t*` are the starts.

Which I'm pretty sure is just equivalent to building this one transform matrix and doing `affine.dot(transform)`. If someone wants to check my math.

nibabel/fileslice.py
```diff
@@ -54,6 +54,10 @@ def is_fancy(sliceobj):
         # slice or Ellipsis or None OK for  basic
         if isinstance(slicer, slice) or slicer in (None, Ellipsis):
             continue
+        # Allow single items in a dimension to be sliced
```
matthew-brett: So here you're allowing a single element array to be considered as an integer - could you say a bit more about why you need this?

I think fancy indexing always does a copy - I don't know whether the user will be surprised if this particular fancy indexing does not:

```
In [14]: a = np.arange(3)

In [15]: b = a[[0]]

In [16]: b
Out[16]: array([0])

In [17]: b[0] = 99

In [18]: a
Out[18]: array([0, 1, 2])

In [20]: c = a[0:1]

In [21]: c
Out[21]: array([0])

In [22]: c[0] = 99

In [23]: a
Out[23]: array([99,  1,  2])
```

nibabel/fileslice.py
matthew-brett: See above - is this just to allow the user to slice the images with single element lists?

nibabel/spatialimages.py
matthew-brett: I think returning a new image is fine.

nibabel/spatialimages.py
matthew-brett: We don't want to drop dimensions because Nifti expects at least three dimensions?  I could well imagine wanting to drop the fourth dimension with `img[..., 1]` for example.

nibabel/funcs.py
```diff
@@ -219,3 +219,69 @@ def _aff_is_diag(aff):
     ''' Utility function returning True if affine is nearly diagonal '''
     rzs_aff = aff[:3, :3]
     return np.allclose(rzs_aff, np.diag(np.diag(rzs_aff)))
+
+
+def crop_image(img, bounds, margin=0):
```
matthew-brett: It looks like the bounds only apply to the first three dimensions.   This are nearly always the spatial dimensions, but not so for Minc.

I'm not sure how margin interacts with bounds .   What do you get for bounds of [[2, -2], [2, -2], [-2, 2]] and margins of 2?

Do we need this function as well as the slicing functionality?

nibabel/spatialimages.py
```diff
@@ -461,8 +479,66 @@ def from_image(klass, img):
                      klass.header_class.from_header(img.header),
                      extra=img.extra.copy())
 
+    def slice_affine(self, idx):
```
matthew-brett: Should this be public?

nibabel/spatialimages.py
```diff
@@ -461,8 +479,66 @@ def from_image(klass, img):
...
+        """ Retrieve affine for current image, if sliced by a given index
+
+        Applies scaling if down-sampling is applied, and adjusts the intercept
+        to account for any cropping.
+
+        Parameters
+        ----------
+        idx : numpy-compatible slice index
```
matthew-brett: `slicer` maybe?  Can be anything that slices an array / image?

nibabel/spatialimages.py
matthew-brett: Right - the transform you're compiling is the transform from voxels in the cropped / subsampled array to voxels in the original array.   To get the final affine you apply the original affine to this transformation.

nibabel/tests/test_funcs.py
```diff
@@ -193,3 +193,22 @@ def test_closest_canonical():
     img.header.set_dim_info(None, None, 2)
     xyz_img = as_closest_canonical(img)
     assert_true(xyz_img.header.get_dim_info() == (None, None, 1))
+
+
+def test_crop_image():
```
matthew-brett: It would be good to test a series of crops / subsamples here.

Am I right that you're not testing the `img.slicer` functionality here?

nibabel/fileslice.py
effigies: I'm allowing a single element list/vector to be considered as a slice, not as an integer. I'm just testing whether the thing in it can be treated as an integer. The reason is that if I want to take a single slice of an image while preserving the dimensions, the expected syntax would be for `img.slicer[[i]]` to be equivalent to `img.slicer[i:i+1]`.

Again, I have only been considering the expectations for result shape. That's a valid point about copying. I'm not sure about the best way to resolve this.

nibabel/fileslice.py
effigies: Yes.

nibabel/funcs.py
effigies: I'm currently leaving this untouched as I work on the slicing. If it turns out this is no longer useful,I'll drop it.

But for your example, that would end up giving you the image back, untouched.

And yes, it doesn't work for Minc, at the moment.

nibabel/spatialimages.py
effigies: That's fair. I think we could restrict integer indices to non-spatial.

In Minc, does a single volume collapse the first dimension, or is would the shape be `(1, x, y, z)`?

nibabel/spatialimages.py
effigies: It seemed potentially useful, but I can make it private.

nibabel/tests/test_funcs.py
effigies: Not yet, no. Wanted to get the API before working on the tests.

effigies:
Thanks for the feedback.A bit I can work on when I find time, but let me know what you think about how we want to handle slicing.

nibabel/spatialimages.py
matthew-brett: Could we work out the indexing, and then allow image formats to say whether that kind of indexing is allowed or not, maybe with a call to something like `img._check_slicing()`.  Nifti and Analyze etc could then barf with the integer indices on the spatial axes.

nibabel/fileslice.py
matthew-brett: Do we need to allow slicing the images with single element lists?  There is some benefit I suppose to sticking to the fancy vs not-fancy distinction - at least it's also what numpy does.

satra:
@matthew-brett - here is a notebook: https://nbviewer.jupyter.org/urls/dl.dropbox.com/s/s0nw827nc4kcnaa/Aliasing.ipynb?flush_cache=true

effigies:
Okay, thinking more about slicing, I won't modify `is_fancy` or `canonical_slicers`, but `Slicer.__getitem__` will convert `[x]` to `slice(x, x+1, None)` and **always** return a new image with a full copy of the sliced `dataobj`. This will shield the original image from any work done on the slice and be consistent with `numpy` fancy indexing.

As before, if users need un-copied slices of the dataobj, they can work with it directly, rather than through the `.slicer` interface.

`Slicer.__getitem__` will also leave slicer validation to the underlying image, which will handle the Minc issue.

nibabel/freesurfer/mghformat.py
```diff
@@ -330,13 +330,10 @@ def set_data_shape(self, shape):
...
-        # needs 4 dimensions.
-        if len(shape) == 3:
-            shape = list(shape)
-            shape.append(1)
-            shape = tuple(shape)
-        dims[:] = shape
+        if len(shape) > 4:
+            raise ValueError("Cannot set data shape with more than 4 "
+                             "dimensions")
+        dims[:] = tuple(shape) + (1,) * (4 - len(shape))
```
effigies: There was previously not a test that tried to create an `MGHImage` with 5 dimensions. `dims[:] = shape` would throw an error in recent versions of numpy, but not the minimum 1.7.1. This change upgrades any <4 dimension objects to 4D, though if you'd prefer I could revert to 3D or 4D only.

nibabel/spatialimages.py
```diff
@@ -324,6 +325,43 @@ class ImageDataError(Exception):
...
+        def __getitem__(self, slicer):
+            klass = self.img.__class__
+            if not klass.makeable:
+                raise NotImplementedError(
+                    "Cannot slice un-makeable image types")
+
+            try:
+                slicer = self.img._check_slicing(self._arr_to_slice(slicer))
+            except ValueError as err:
+                raise IndexError(*err.args)
```
effigies: `canonical_slicers` raises `ValueError`s on things that would be `IndexError`s in numpy. Since they're being used in a `__getitem__` context, I make the conversion here so the expected exception type is raised without changing the API elsewhere.

nibotmi:
:umbrella: The latest upstream changes (presumably #562) made this pull request unmergeable. Please resolve the merge conflicts.

effigies:
Welp, turns out even the simplest fancy indexing can alter the shape in unintuitive ways. I've decided to follow `dataobj`'s lead and just disallow it.

```Python
In [22]: img.get_data()[[0], [0]].shape
Out[22]: (1, 6, 2)

In [23]: img.get_data().shape
Out[23]: (4, 5, 6, 2)

In [24]: img.get_data()[:1, :1].shape
Out[24]: (1, 1, 6, 2)
```

effigies:
@matthew-brett This is ready for review. Know you're busy, so no rush. Just a bump.

effigies:
@matthew-brett With #569 in, I think this is ready to go. Updated the changelog.

effigies:
Finally had a couple minutes to figure out why the tests broke. Turned out to be a tiny fix due to the #569 refactor.

@matthew-brett @satra I know it's been months, so sorry to ask you to re-familiarize yourselves with this PR, but I think this is ready to go in barring any last minute concerns (or surprise test failures... it passes locally).

satra:
@effigies - looks good to me. 

i think the docstring makes it clear that this doesn't do any filtering to avoid aliasing. however, i'm not sure how many neuroimagers really understand that and potential dangers. would it be useful to point to the wikipedia article on aliasing?

nibabel/spatialimages.py
```diff
@@ -324,6 +325,24 @@ class ImageDataError(Exception):
 class SpatialImage(DataobjImage):
     ''' Template class for volumetric (3D/4D) images '''
     header_class = SpatialHeader
+    # Make a strong assumption that the first three dimensions are spatial
```
matthew-brett: Should this go in the parent?  Or in the children that do require this?

nibabel/minc1.py
```diff
@@ -319,6 +319,49 @@ def from_file_map(klass, file_map):
...
+        -------
+        slicer : object
+            Validated slicer object that will slice image's `dataobj`
+            without collapsing spatial dimensions
+        '''
+        slicer = canonical_slicers(slicer, self.shape)
+        try:
+            all_slices = super(Minc1Image, self)._check_slicing(slicer, False)
+            sp_dims = self._spatial_dims
+        except IndexError:
```
matthew-brett: This is where all my pain is - the problem of which axis is which.   I think we need a better solution, like axis names : https://github.com/nipy/nibabel/wiki/BIAP6 ...

effigies: Okay. So I've seen BIAP6 come up in one form or another a number of times over the last 5-6 years. What would it take to make a final decision on how to handle axes (mostly: who do we need input from?) and do you have an estimate of the effort involved to make a change here? I'm happy to contribute some effort in hacking and review, if there's a reasonably defined scope.

In the short term, does it make sense to simply raise `NotImplementedError` in MINC, and punt on this issue entirely, as far as this PR is concerned? Or do you think working out the spatial dimensions (as suggested in https://github.com/nipy/nibabel/pull/550/files#r169142529) is small enough to make handling this in MINC feasible?

I recognize that it's dissatisfying to keep adding features that aren't available in MINC, but even if it's only available in NIFTI, this feature seems valuable (and having it in NIFTI might encourage users more familiar with MINC to help extend the functionality).

matthew-brett: Yes, fair point (about the 5-6 years).   How about a developer hangout soonish to see if we can scope this out?   If we find a proper fix is not practical, we can drop MINC, that's what I did in practice in the [processing module](https://github.com/nipy/nibabel/blob/master/nibabel/processing.py#L158), using the [spatial_axes_first function](https://github.com/nipy/nibabel/blob/master/nibabel/imageclasses.py#L120).

effigies: That sounds reasonable to me. I'm pretty available between 9am-4pm EST for the next three weeks (also other times, but harder to describe briefly, and less predictable). I'm not sure who else should be involved, but assume you have a better sense of the field of relevant/interested parties? I have an admittedly basic grasp on MINC (I found the standard unsatisfyingly vague, though that may be because it depends on familiarity with HDF5), so that might be worth bearing in mind in terms of the expertise you want involved in the hangout.

matthew-brett: For now - do you think you can re-use the `spatial_axes_first` function etc from `processing`?  And therefore, disallow MINC?

effigies: Sure thing.

effigies:
At long last, @matthew-brett, I think this one is ready for review.

nibabel/spatialimages.py
```diff
@@ -325,6 +326,21 @@ class SpatialImage(DataobjImage):
     ''' Template class for volumetric (3D/4D) images '''
     header_class = SpatialHeader
 
+    class Slicer(object):
```
matthew-brett: Remind me why the Slicer class is internal to the SpatialImage class?  I think that will mean that you can no longer pickle images.

nibabel/spatialimages.py
```diff
@@ -325,6 +326,21 @@ class SpatialImage(DataobjImage):
...
+        def __init__(self, img):
+            self.img = img
+
+        def __getitem__(self, slicer):
+            try:
+                slicer = self.img._check_slicing(slicer)
+            except ValueError as err:
+                raise IndexError(*err.args)
+            dataobj = self.img.dataobj[slicer]
+            affine = self.img._slice_affine(slicer)
```
matthew-brett: It's a bit scary using this apparently private method.  Did we discuss this before?  I forget?

nibabel/spatialimages.py
effigies: The reason is that we wanted something that could be used like:

```Python
img.slicer[sliceobj]
```

Some other way of organizing the code would be fine with me.

matthew-brett: I think it would make pickle happy if the class definition was at the top level, like:

```
class ImageSlicer(object):
```

Does that work?

nibabel/spatialimages.py
effigies: https://github.com/nipy/nibabel/pull/550#discussion_r141847919

matthew-brett:
We're not currently checking for empty slices.  Do you think we should?

```
In [22]: anat = nib.load(pjoin(data_dir, 'anatomical.nii'))

In [23]: anat.affine
Out[23]: 
array([[ -2.,   0.,   0.,  32.],
       [  0.,   2.,   0., -40.],
       [  0.,   0.,   2., -16.],
       [  0.,   0.,   0.,   1.]])

In [24]: anat2 = anat.slicer[:0, :, :]

In [25]: anat2.shape
Out[25]: (0, 41, 25)
```

nibabel/spatialimages.py
effigies: Sure. In which case we definitely shouldn't use a private method like `img._slice_affine`. Does it make more sense to place it here:

```Python
class ImageSlicer(object):
    ...

    @classmethod
    def slice_affine(affine, slicer):
        ...
```

I'm mostly unsure how much sense `img.slice_affine` as part of the image API.

matthew-brett:
Ah yes - I remember now about `spatial_axes_first` - thanks.

So, why not put `Slicer` outside the `SpatialImage` class, as `SpatialFirstSlicer` along with the slice checking and the spatial axes first check, and have something like:

```
class SpatialImage(...)

    ImageSlicer = SpatialFirstSlicer

    @property
    def slicer(self):
        return self.ImageSlicer(self)
```

?

effigies:
Sure. I'll see if I can get this done quickly, but probably not tonight.

Thanks for your time, Matthew.

matthew-brett:
No problem - sorry to be so slow.

effigies:
@matthew-brett Changes made, if you've got time for another review.

nibabel/spatialimages.py
```diff
@@ -321,9 +322,99 @@ class ImageDataError(Exception):
     pass
 
 
+class SpatialFirstSlicer(object):
+    ''' Slicing interface that returns a new image with an updated affine
+
+    Checks that an image's first three axes are spatial
+    '''
+    def __init__(self, img):
+        from .imageclasses import spatial_axes_first
```
matthew-brett: Can this go at the top of the file now?  Or would that make a circular import?  If so, maybe worth a comment.

nibabel/spatialimages.py
```diff
@@ -321,9 +322,99 @@ class ImageDataError(Exception):
...
+            return only slices along spatial dimensions (x, y, z)
+
+        Returns
+        -------
+        slicer : object
+            Validated slicer object that will slice image's `dataobj`
+            without collapsing spatial dimensions
+        '''
+        slicer = canonical_slicers(slicer, self.img.shape)
+        spatial_slices = slicer[:3]
```
matthew-brett: Maybe worth a comment here about this line being specific to spatial-first images. Or generalize somehow to allow overriding and working out spatial slices from the image.

matthew-brett:
Thanks for your patience, in it goes ...

PR #551, opened 2017-08-24
MRG: implementing / testing get_fdata

matthew-brett:
get_fdata the new proposed default method for getting the data from the image.
It always returns floating point.

### DIFF
--- a/nibabel/dataobj_images.py
+++ b/nibabel/dataobj_images.py
@@ -340,9 +340,10 @@ class DataobjImage(FileBasedImage):
         dtype = np.dtype(dtype)
         if not issubclass(dtype.type, np.inexact):
             raise ValueError('{} should be floating point type'.format(dtype))
-        if (self._fdata_cache is not None and
-            self._fdata_cache.dtype.type == dtype.type):
-            return self._fdata_cache
+        # Return cache if cache present and of correct dtype.
+        if self._fdata_cache is not None:
+            if self._fdata_cache.dtype.type == dtype.type:
+                return self._fdata_cache
         data = np.asanyarray(self._dataobj).astype(dtype)
         if caching == 'fill':
             self._fdata_cache = data
--- a/nibabel/dataobj_images.py
+++ b/nibabel/dataobj_images.py
@@ -138,7 +138,7 @@ class DataobjImage(FileBasedImage):
         (no reference to the array).  If the cache is full, "unchanged" leaves
         the cache full and returns the cached array reference.
 
-        The cache can effect the behavior of the image, because if the cache is
+        The cache can affect the behavior of the image, because if the cache is
         full, or you have an array image, then modifying the returned array
         will modify the result of future calls to ``get_data()``.  For example
         you might do this:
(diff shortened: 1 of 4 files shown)


nibabel/dataobj_images.py
```diff
@@ -191,11 +204,155 @@ def get_data(self, caching='fill'):
...
+        >>> img.in_memory
+        False
+        >>> data[0, 0, 0] = 99
+        >>> data_again = img.get_fdata(caching='unchanged')
+        >>> data_again is data
+        False
+        >>> data_again[0, 0, 0, 0]
+        0.0
+        """
+        if caching not in ('fill', 'unchanged'):
```
effigies: Want to test `caching='unknown'`?

nibabel/dataobj_images.py
```diff
@@ -191,11 +204,155 @@ def get_data(self, caching='fill'):
             self._data_cache = data
         return data
 
+    def get_fdata(self, caching='fill', dtype=np.float64):
```
satra: any reason why float64 over float32 is the default?

satra:
@matthew-brett - other than my question, which should prevent this from being merged, and on top of @effigies review, the code and documentation looks good.

effigies:
Shouldn't, maybe?

satra:
thanks @effigies - that was definitely a lapso

nibabel/dataobj_images.py
matthew-brett: @satra - the default up till now has been to return float64 for images with scaling.  Also, some images do have float64 and it seems a shame to downgrade them with this call.  So float32 seems to me like an option rather than a default.

matthew-brett:
Guys - thanks for the reviews.  I've updated to respond to your comments, and I've added some more tests of the caching, along with a fix to the code.  To help reviewing, can I suggest that y'all look at commit c729e06 separately (it has the caching changes, and tests)?

matthew-brett:
Oops sorry - I missed a changed I needed to make in the tests, to reflect the change in caching behavior.  I rebased.  New commit with all the relevant changes is c800493.

satra:
+1 from me as well.

matthew-brett:
@effigies - any interests in being release manager again?  It's ready as far as I am concerned - but I've got 15 hours of completely new teaching to prepare in the next couple of weeks, and I'm really slammed for time.

effigies:
Sure. I'll aim to release by the end of the week. I'll let you know if I have any issues.

mwaskom:
Was this discussed more somewhere else? The release notes link here for information about this change, but there's no information in the PR (aside from the code itself!) that motivates the new function. (I ask as it's my vague sense that `get_fdata` is intended to replace `get_data`).

effigies:
@mwaskom This thread is probably the better link... https://mail.python.org/pipermail/neuroimaging/2017-April/001383.html

mwaskom:
Thanks @effigies. Perhaps the PR intro could be edited to include this link.

effigies:
Done.

PR #552, opened 2017-09-05
Conditionally use indexed_gzip

pauldmccarthy:
Howdy all,

I've been [sitting on this one](https://mail.python.org/pipermail/neuroimaging/2016-March/000822.html) for a while. This PR proposes to conditionally use [`indexed_gzip`](https://github.com/pauldmccarthy/indexed_gzip) instead of `gzip` to allow for faster random access to `.gz` files. The default behaviour of `nibabel` is unchanged by this PR - to take advantage of `indexed_gzip`, one needs to load an image with a new flag, like so:

```python
img = nib.load('large_gzipped_image.nii.gz', keep_file_open=True)
```

This PR also proposes changes to `ArrayProxy` and `fileslice.read_segments` which makes paired calls to `seek` and `read` thread-safe - see [here](https://mail.python.org/pipermail/neuroimaging/2017-February/001321.html) for a discussion surrounding this change.

**Changes**

1. The `ArrayProxy` has been modified so that it will optionally create and use a single `ImageOpener` for its lifetime, rather than creating/destroying an `ImageOpener` every time the image data is accessed. This is controlled via the `keep_file_open` flag.

2. The `fileslice.read_segments` function optionally accepts a `threading.Lock` so that paired calls to the `seek`/`read` functions can be protected against changes to the running thread.

3. The `openers._gzip_open` function has been changed to conditionally use `indexed_gzip` when available.

4. (Unrelated) There was a bug in `_gzip_open` which was causing the `max_read_chunk` fix to have no effect. This has been fixed.

**Outstanding issues**

 - `indexed_gzip` is not available for windows platforms. I don't have a windows machine to test on, so would need somebody to help out in making it work on windows. See pauldmccarthy/indexed_gzip#1. Note that this PR in its initial state does not break `nibabel` on windows, as the use of `indexed_gzip` is conditional upon its presence.

 - `indexed_gzip` does not support writing. If a file is opened in write mode, the built-in `GzipFile` class is used. I am not against adding write-support into `indexed_gzip`, but to date have not considered it a necessity (and don't know how much time it would take!).

 - The `MGHFormat` class uses the `ArrayProxy`, but I have not updated it to use the `keep_file_open` flag, so it does not benefit from `indexed_gzip`. Adding this would be trivial, but I thought I'd just start by proposing changes to NIFTI/ANALYZE access.

 - The `PARRECImage` class implements its own `ArrayProxy`, so does not benefit from `indexed_gzip`. The `PARRECArrayProxy` class would need to be updated similarly, or some refactorings made so that it uses the standard `ArrayProxy`. I've not looked too closely at the `PARRECArrayProxy`, but on the surface it seems to be quite similar to the `ArrayProxy` class, so this should not be too difficult.

 - `indexed_gzip` is currently available as a binary wheel for macos, but only a source package for other platforms. Hence cython and a C compiler must be available for it to be installed. Is this a concern? I can look into generating linux builds if you think it is needed.

**Other notes**

 - I have not updated `fileslice._simple_fileslice` to accept a `Lock`, because it is not used by anything (apart from unit tests)

 - I decided against using `pread` for the thread-safe fix (as discussed in the mailing list thread) because the built-in `GzipFile` class doesn't have a `pread` method.

 - I have run the `nibabel` unit tests locally, both with and without `indexed_gzip`. It may be desirable to build both of these scenarios into the standard `nibabel` CI test suite.

### DIFF
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -959,8 +959,8 @@ class AnalyzeImage(SpatialImage):
             ``'auto'`` (the default), and the optional ``indexed_gzip``
             dependency is present, a single file handle is created and
             persisted. If ``indexed_gzip`` is not available, behaviour is the
-            same as if ``keep_file_open is False``. If ``file_like`` is an
-            open file handle, this setting has no effect.
+            same as if ``keep_file_open is False``. If ``file_map`` refers
+            to an open file handle, this setting has no effect.
 
         Returns
         -------
@@ -1011,8 +1011,7 @@ class AnalyzeImage(SpatialImage):
             ``'auto'`` (the default), and the optional ``indexed_gzip``
             dependency is present, a single file handle is created and
             persisted. If ``indexed_gzip`` is not available, behaviour is the
-            same as if ``keep_file_open is False``. If ``file_like`` is an
-            open file handle, this setting has no effect.
+            same as if ``keep_file_open is False``.
 
         Returns
         -------
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -934,7 +934,7 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_file_map(klass, file_map, mmap=True, keep_file_open=False):
+    def from_file_map(klass, file_map, mmap=True, keep_file_open='indexed'):
         ''' class method to create image from mapping in `file_map ``
 
         Parameters
@@ -950,11 +950,17 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
-        keep_file_open: If ``file_like`` is a file name, the default behaviour
-            is to open a new file handle every time the data is accessed. If
-            this flag is set to `True``, the file handle will be opened on the
-            first access, and kept open until this ``ArrayProxy`` is garbage-
-            collected.
+        keep_file_open : { 'indexed', True, False }, optional, keyword only
+            `keep_file_open` controls whether a new file handle is created
+            every time the image is accessed, or a single file handle is
+            created and used for the lifetime of this ``ArrayProxy``. If
+            ``True``, a single file handle is created and used. If ``False``,
+            a new file handle is created every time the image is accessed. If
+            ``'indexed'`` (the default), and the optional ``indexed_gzip``
+            dependency is present, a single file handle is created and
+            persisted. If ``indexed_gzip`` is not available, behaviour is the
+            same as if ``keep_file_open is False``. If ``file_like`` is an
+            open file handle, this setting has no effect.
 
         Returns
         -------
@@ -982,7 +988,7 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_filename(klass, filename, mmap=True, keep_file_open=False):
+    def from_filename(klass, filename, mmap=True, keep_file_open='indexed'):
         ''' class method to create image from filename `filename`
 
         Parameters
@@ -996,11 +1002,17 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
-        keep_file_open: If ``file_like`` is a file name, the default behaviour
-            is to open a new file handle every time the data is accessed. If
-            this flag is set to `True``, the file handle will be opened on the
-            first access, and kept open until this ``ArrayProxy`` is garbage-
-            collected.
+        keep_file_open : { 'indexed', True, False }, optional, keyword only
+            `keep_file_open` controls whether a new file handle is created
+            every time the image is accessed, or a single file handle is
+            created and used for the lifetime of this ``ArrayProxy``. If
+            ``True``, a single file handle is created and used. If ``False``,
+            a new file handle is created every time the image is accessed. If
+            ``'indexed'`` (the default), and the optional ``indexed_gzip``
+            dependency is present, a single file handle is created and
+            persisted. If ``indexed_gzip`` is not available, behaviour is the
+            same as if ``keep_file_open is False``. If ``file_like`` is an
+            open file handle, this setting has no effect.
 
         Returns
         -------
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -950,7 +950,7 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
-        keep_file_open : { 'auto', True, False }, optional, keyword only
+        keep_file_open : { None, 'auto', True, False }, optional, keyword only
             `keep_file_open` controls whether a new file handle is created
             every time the image is accessed, or a single file handle is
             created and used for the lifetime of this ``ArrayProxy``. If
@@ -960,8 +960,9 @@ class AnalyzeImage(SpatialImage):
             present, a single file handle is created and persisted. If
             ``indexed_gzip`` is not available, behaviour is the same as if
             ``keep_file_open is False``. If ``file_map`` refers to an open
-            file handle, this setting has no effect. The default value is
-            set to the value of ``nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT``.
+            file handle, this setting has no effect. The default value
+            (``None``) will result in the value of
+            ``nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT`` being used.
 
         Returns
         -------
@@ -1003,8 +1004,7 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
-
-        keep_file_open : { 'auto', True, False }, optional, keyword only
+        keep_file_open : { None, 'auto', True, False }, optional, keyword only
             `keep_file_open` controls whether a new file handle is created
             every time the image is accessed, or a single file handle is
             created and used for the lifetime of this ``ArrayProxy``. If
@@ -1013,8 +1013,9 @@ class AnalyzeImage(SpatialImage):
             ``'auto'``, and the optional ``indexed_gzip`` dependency is
             present, a single file handle is created and persisted. If
             ``indexed_gzip`` is not available, behaviour is the same as if
-            ``keep_file_open is False``. The default value is set to the value
-            of ``nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT``.
+            ``keep_file_open is False``. The default value (``None``) will
+            result in the value of
+            ``nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT`` being used.
 
         Returns
         -------
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -934,8 +934,8 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_file_map(klass, file_map, mmap=True, keep_file_open='auto'):
-        ''' class method to create image from mapping in `file_map ``
+    def from_file_map(klass, file_map, mmap=True, keep_file_open=None):
+        '''class method to create image from mapping in `file_map ``
 
         Parameters
         ----------
@@ -956,11 +956,12 @@ class AnalyzeImage(SpatialImage):
             created and used for the lifetime of this ``ArrayProxy``. If
             ``True``, a single file handle is created and used. If ``False``,
             a new file handle is created every time the image is accessed. If
-            ``'auto'`` (the default), and the optional ``indexed_gzip``
-            dependency is present, a single file handle is created and
-            persisted. If ``indexed_gzip`` is not available, behaviour is the
-            same as if ``keep_file_open is False``. If ``file_map`` refers
-            to an open file handle, this setting has no effect.
+            ``'auto'``, and the optional ``indexed_gzip`` dependency is
+            present, a single file handle is created and persisted. If
+            ``indexed_gzip`` is not available, behaviour is the same as if
+            ``keep_file_open is False``. If ``file_map`` refers to an open
+            file handle, this setting has no effect. The default value is
+            set to the value of ``nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT``.
 
         Returns
         -------
@@ -988,8 +989,8 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_filename(klass, filename, mmap=True, keep_file_open='auto'):
-        ''' class method to create image from filename `filename`
+    def from_filename(klass, filename, mmap=True, keep_file_open=None):
+        '''class method to create image from filename `filename`
 
         Parameters
         ----------
@@ -1002,16 +1003,18 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
+
         keep_file_open : { 'auto', True, False }, optional, keyword only
             `keep_file_open` controls whether a new file handle is created
             every time the image is accessed, or a single file handle is
             created and used for the lifetime of this ``ArrayProxy``. If
             ``True``, a single file handle is created and used. If ``False``,
             a new file handle is created every time the image is accessed. If
-            ``'auto'`` (the default), and the optional ``indexed_gzip``
-            dependency is present, a single file handle is created and
-            persisted. If ``indexed_gzip`` is not available, behaviour is the
-            same as if ``keep_file_open is False``.
+            ``'auto'``, and the optional ``indexed_gzip`` dependency is
+            present, a single file handle is created and persisted. If
+            ``indexed_gzip`` is not available, behaviour is the same as if
+            ``keep_file_open is False``. The default value is set to the value
+            of ``nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT``.
 
         Returns
         -------
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -934,7 +934,7 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_file_map(klass, file_map, mmap=True, keep_file_open='indexed'):
+    def from_file_map(klass, file_map, mmap=True, keep_file_open='auto'):
         ''' class method to create image from mapping in `file_map ``
 
         Parameters
@@ -950,13 +950,13 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
-        keep_file_open : { 'indexed', True, False }, optional, keyword only
+        keep_file_open : { 'auto', True, False }, optional, keyword only
             `keep_file_open` controls whether a new file handle is created
             every time the image is accessed, or a single file handle is
             created and used for the lifetime of this ``ArrayProxy``. If
             ``True``, a single file handle is created and used. If ``False``,
             a new file handle is created every time the image is accessed. If
-            ``'indexed'`` (the default), and the optional ``indexed_gzip``
+            ``'auto'`` (the default), and the optional ``indexed_gzip``
             dependency is present, a single file handle is created and
             persisted. If ``indexed_gzip`` is not available, behaviour is the
             same as if ``keep_file_open is False``. If ``file_like`` is an
@@ -988,7 +988,7 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_filename(klass, filename, mmap=True, keep_file_open='indexed'):
+    def from_filename(klass, filename, mmap=True, keep_file_open='auto'):
         ''' class method to create image from filename `filename`
 
         Parameters
@@ -1002,13 +1002,13 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
-        keep_file_open : { 'indexed', True, False }, optional, keyword only
+        keep_file_open : { 'auto', True, False }, optional, keyword only
             `keep_file_open` controls whether a new file handle is created
             every time the image is accessed, or a single file handle is
             created and used for the lifetime of this ``ArrayProxy``. If
             ``True``, a single file handle is created and used. If ``False``,
             a new file handle is created every time the image is accessed. If
-            ``'indexed'`` (the default), and the optional ``indexed_gzip``
+            ``'auto'`` (the default), and the optional ``indexed_gzip``
             dependency is present, a single file handle is created and
             persisted. If ``indexed_gzip`` is not available, behaviour is the
             same as if ``keep_file_open is False``. If ``file_like`` is an
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -934,7 +934,7 @@ class AnalyzeImage(SpatialImage):
 
     @classmethod
     @kw_only_meth(1)
-    def from_file_map(klass, file_map, mmap=True):
+    def from_file_map(klass, file_map, mmap=True, keep_file_open=False):
         ''' class method to create image from mapping in `file_map ``
 
         Parameters
@@ -950,6 +950,11 @@ class AnalyzeImage(SpatialImage):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
+        keep_file_open: If ``file_like`` is a file name, the default behaviour
+            is to open a new file handle every time the data is accessed. If
+            this flag is set to `True``, the file handle will be opened on the
+            first access, and kept open until this ``ArrayProxy`` is garbage-
+            collected.
 
         Returns
         -------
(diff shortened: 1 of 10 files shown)


nibabel/arrayproxy.py
```diff
@@ -125,6 +131,27 @@ def __init__(self, file_like, spec, mmap=True):
...
         self._dtype = np.dtype(self._dtype)
         self._mmap = mmap
+        self._keep_file_open = keep_file_open
+        self._lock = Lock()
+
+    def __del__(self):
+        '''If this ``ArrayProxy`` was created with ``keep_file_open=True``,
+        the open file object is closed if necessary.
+        '''
+        if hasattr(self, '_opener') and not self._opener.closed:
```
matthew-brett: What happens if the user passes an open file to the original image creation routine?  Does that get closed too?

nibabel/arrayproxy.py
```diff
@@ -155,12 +182,26 @@ def inter(self):
     def is_proxy(self):
         return True
 
+    @contextmanager
+    def _get_fileobj(self):
+        '''Create and return a new ``ImageOpener``, or return an existing one.
```
matthew-brett: Clean up docstring for numpy docstring standard?

nibabel/fileslice.py
```diff
@@ -634,29 +634,49 @@ def read_segments(fileobj, segments, n_bytes):
...
     Returns
     -------
     buffer : buffer object
         object implementing buffer protocol, such as byte string or ndarray or
         mmap or ctypes ``c_char_array``
     """
+    # Make a dummy lock-like thing to make the code below a bit nicer
+    if lock is None:
+
+        class DummyLock(object):
```
matthew-brett: Push this out of the method to the module top-level, maybe as `class _NullLock` ?

pauldmccarthy:
Hopefully all of the code comments have been addressed with the recent push - well spotted on the handling of open file objects - the `ArrayProxy` was actually closing them if `keep_file_open=True` :) I've added another test case to check for this.

I will set up `indexed_gzip`linux wheel builds later today. But Windows is a problem - I think that [some work](https://github.com/pauldmccarthy/indexed_gzip/issues/1) is needed for `indexed_gzip` to compile on windows, but I don't have a test machine, nor do I know anything about windows development. I could do with some help on this one!

pauldmccarthy:
Hey @effigies, I guess my line of thinking was that the `keep_files_open` flag was conceptually independent of the use of `indexed_gzip`, and would be explicitly selected by the user. 

But to be honest I can't think of a specific use case where one would want or need `keep_file_open=True`, except when using something like `indexed_gzip`. So in principle I am ok with your proposal.

Except for one thing - couldn't there be a situation where `have_indexed_gzip is True`, and the user explicitly wants`keep_file_open=False`?

effigies:
Yes, that's definitely a possibility.

Perhaps what makes sense is a different default to `keep_file_open`, such as `keep_file_open='indexed_gzip'`, which can then be forced one way or the other with `True` or `False`. I don't really like the idea of having the default value change from one install to another... that is `pydoc` showing `True` or `False`, depending on whether another library is installed.

In this case, if we have `keep_file_open='indexed_gzip'` and we discover that we don't have a `.gz` file OR we don't have `indexed_gzip`, then we can set it to `False` and get standard behavior. If the user specifies `False, we also get standard behavior. In the very rare case that the user specifies `True`, they'll get a bunch of filehandles, but at least they'll be expecting them.

pauldmccarthy:
Right - I had overlooked the fact that you were only talking about changing the default behaviour, rather than hiding the `keep_file_open` option from the user.

I think your new proposal sounds good - sensible/common usage default, but still allowing the user full control. I'll push some new commits either tonight or tomorrow accordingly.

nibabel/fileslice.py
```diff
@@ -17,6 +17,21 @@
 SKIP_THRESH = 2 ** 8
 
 
+
```
effigies: One too many blank lines, here, causing the style checks to fail.

nibabel/tests/test_fileslice.py
```diff
@@ -689,6 +691,60 @@ def test_read_segments():
...
+
+    def runtest():
+        seg, nbytes = random_segments(1)
+        expected = get_expected(seg)
+        _check_bytes(read_segments(fobj, seg, nbytes, lock), expected)
+
+        seg, nbytes = random_segments(10)
+        expected = get_expected(seg)
+        _check_bytes(read_segments(fobj, seg, nbytes, lock), expected)
+        numpassed[0] += 1
```
effigies: This line should be locked, too, as read/update/store on a list element is not thread-safe.

nibabel/openers.py
```diff
@@ -60,13 +60,38 @@ def readinto(self, buf):
...
-    # See https://github.com/nipy/nibabel/pull/210 for discussion
-    if hasattr(gzip_file, 'max_chunk_read'):
+def _gzip_open(fileish, mode='rb', compresslevel=9):
+
+    # is indexed_gzip present?
+    try:
+        from indexed_gzip import SafeIndexedGzipFile
+        have_indexed_gzip = True
+    except:
+        have_indexed_gzip = False
```
effigies: Let's move this block to the top of `openers.py`.

nibabel/arrayproxy.py
```diff
@@ -125,6 +137,71 @@ def __init__(self, file_like, spec, mmap=True):
...
+        # if keep_file_open is True/False, we do what the user wants us to do
+        if keep_file_open != 'auto':
+            return bool(keep_file_open)
+        # Otherwise, if file_like is gzipped, and we have_indexed_gzip, we set
+        # keep_file_open to True, else we set it to False
+        try:
+            import indexed_gzip
+            have_indexed_gzip = True
+        except ImportError:
+            have_indexed_gzip = False
```
effigies: Remove this block here, and import `have_indexed_gzip` from openers at the top of the file.

nibabel/analyze.py
```diff
@@ -990,6 +1002,17 @@ def from_filename(klass, filename, mmap=True):
...
+            `keep_file_open` controls whether a new file handle is created
+            every time the image is accessed, or a single file handle is
+            created and used for the lifetime of this ``ArrayProxy``. If
+            ``True``, a single file handle is created and used. If ``False``,
+            a new file handle is created every time the image is accessed. If
+            ``'auto'`` (the default), and the optional ``indexed_gzip``
+            dependency is present, a single file handle is created and
+            persisted. If ``indexed_gzip`` is not available, behaviour is the
+            same as if ``keep_file_open is False``. If ``file_like`` is an
+            open file handle, this setting has no effect.
```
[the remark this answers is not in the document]
pauldmccarthy: Aah, I was just copying/pasting that comment block. 

I was wondering about this sort of thing. In my own projects, I tend to define and document arguments on the function/methods/classes that they are used in, and then pass them through with `**kwargs` for functions/methods/classes which call them. 

But for this PR, I was following what seems to be the convention in `nibabel`, where arguments are explicitly defined at each level. What is your take on this?

nibabel/arrayproxy.py
```diff
@@ -155,12 +232,34 @@ def inter(self):
...
     def get_unscaled(self):
-        ''' Read of data from file
+        """ Read of data from file
 
         This is an optional part of the proxy API
-        '''
-        with ImageOpener(self.file_like) as fileobj:
+        """
+        with self._get_fileobj() as fileobj:
             raw_data = array_from_file(self._shape,
```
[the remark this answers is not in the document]
pauldmccarthy: Yep, you're right, well spotted.

nibabel/tests/test_fileslice.py
pauldmccarthy: If only it were easier to share variables with nested functions :)

nibabel/openers.py
pauldmccarthy: Sorry for the continually occurring style issues - I just discovered that my emacs was not liking the file-local variables in e.g. `arrayproxy.py`, so linting was not running.

nibabel/analyze.py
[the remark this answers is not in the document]
effigies: Not sure quite what you mean by duplication. (Maybe a concrete example of the kind of change you're proposing would be useful?) But on the whole, I think this is looking good as is. I'll have another read through now.

nibabel/arrayproxy.py
```diff
@@ -125,6 +137,65 @@ def __init__(self, file_like, spec, mmap=True):
...
+        The value of ``keep_file_open`` that will be used by this
+        ``ArrayProxy``.
+        """
+
+        # file_like is a handle - keep_file_open is irrelevant
+        if hasattr(file_like, 'read') and hasattr(file_like, 'seek'):
+            return False
+        # if keep_file_open is True/False, we do what the user wants us to do
+        if keep_file_open != 'auto':
+            return bool(keep_file_open)
```
effigies: We should validate that `keep_file_open` is one of `(True, False, 'auto')`. Maybe make this:

```Python
if isinstance(keep_file_open, bool):
    return keep_file_open

if keep_file_open != 'auto':
    raise ValueError("Unknown value for `keep_file_open`: {!r}".format(keep_file_open))
```

And there should be a test to ensure that an invalid value raises an exception.

nibabel/arrayproxy.py
```diff
@@ -155,12 +226,34 @@ def inter(self):
...
+
     def get_unscaled(self):
-        ''' Read of data from file
+        """ Read of data from file
 
         This is an optional part of the proxy API
-        '''
-        with ImageOpener(self.file_like) as fileobj:
+        """
+        with self._get_fileobj() as fileobj, self._lock:
```
effigies: I was thinking `array_from_file` would take a lock and block its critical section. Do you have a sense over what kind of difference in slowdown could result from blocking the whole call instead? If it's trivial, then this is fine.

nibabel/analyze.py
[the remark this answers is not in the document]
pauldmccarthy: By duplication, I mean that the `mmap=True, keep_file_open='auto'` arguments, and associated docstrings, are defined multiple times, but only really need to be defined in the `ArrayProxy`, and in `analyze.py`, `spm99analyze.py` (and future classes) could be passed through with `**kwargs` (and have docstrings saying "See ArrayProxy for details on the other keyword arguments). 

I just like to remove as much duplication as possible. But in this case it's a minor point.

pauldmccarthy:
Great - I should be able to find some over the weekend to finish this off - building the linux wheels has taken much longer than I would have liked!

nibabel/analyze.py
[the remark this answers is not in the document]
effigies: I see. I think the idea is to keep docstrings as informative as possible. It's brittle, but I think that's good because API changes shouldn't be easy to do accidentally. And it's harder for your docs to get out of sync with your behavior.

effigies:
Sounds good. I'm going to be on vacation next week, so my reviews may be a bit sparse.

nibabel/arrayproxy.py
pauldmccarthy: I think, given that the `array_from_file` function reads an entire image, so where the lock occurs is not going to make much of a difference to performance. In fact, I am wondering whether this locking should occur in the `ArrayProxy` at all, or if it should be left to the responsibility of calling code.

This situation is a bit different to the locking needed in the `read_segments` function, which can't easily be managed by calling code  ... unless we delegate responsibilty entirely to the calling code, and allow a `Lock` to be passed into `ArrayProxy.__init__`?

effigies:
@pauldmccarthy Hi, sorry. Back from vacation and getting back into things. I'll have a look at this tomorrow and see where we stand.

@matthew-brett Do we want to block release on this? It seems pretty useful to me, and given how infrequent nibabel releases are, I'm inclined to try to get this in.

nibabel/openers.py
```diff
@@ -14,6 +14,13 @@
 import sys
 from os.path import splitext
 
+# is indexed_gzip present?
+try:
+    from indexed_gzip import SafeIndexedGzipFile
+    HAVE_INDEXED_GZIP = True
+except:
```
effigies: `except ImportError`

pauldmccarthy: Yikes - I got bitten by this the other week - i learnt the hard way that `except:` will also catch `SystemExit`!

effigies:
Thanks. @matthew-brett any more comments?

matthew-brett:
Hi guy - sorry to be quiet - I've been teaching this week, crash course, finishes tomorrow.

I hope I'll be in a fit state to have a look on Friday.

effigies:
Sounds good. Good luck with the end of the course.

matthew-brett:
:) - I will need it!

matthew-brett:
Hi guys.  Sorry, now I have grant deadline - for Monday afternoon UK time.

Can I check my understanding here?

Before this PR, if you open an image with a filename, there's a new file object created and deleted for each read you do.

After this PR, you'll get the same behavior only if `indexed_gzip` is not installed?

So, with this PR, if could happen that your code is working, opening lots of files without running out of file handles, until you install `indexed_gzip` at which point you do start running out of file handles?

If that's right - can you think of a way round it?   What about having module-level constants to control the behavior?

pauldmccarthy:
Hey @matthew-brett,

This is why I originally had the `keep_file_open` flag default to `False`, which would preserve the original behaviour of opening a new file handle for each access.  

But I think we already have a solution - in the PR's current state, the use of `indexed_gzip`, and whether file handles are kept open or not, is dictated by the module-level `nibabel.openers.HAVE_INDEXED_GZIP` flag. So this flag could be overridden by a user who has `indexed_gzip` installed, but wants the original behaviour of using a new file handle on each access.

What do you think?

matthew-brett:
I'm worried that the user might hit a 'too many files open' error without knowing why.  So I think the default behavior should stay the same, but it can be overridden by the module level default.

Can we just have `keep_file_open` as the thing that can be changed by default, and we always use `indexed_gzip` if available, and this flag is set to True, either by the module default, or by the user with the explicit call?  Is there a reason not to use `indexed_gzip` if it is present, and the open command is compatible?

pauldmccarthy:
If I understand you correctly, these changes should do the trick:

 -  Add a module-level `KEEP_FILE_OPEN` flag to `nibabel.arrayproxy`, which takes one of `True, False, 'auto'`, and is initally set to `False`
 - Change `ArrayProxy.__init__` so the default value for `keep_file_open` is set to the current value of `nibabel.arrayproxy.KEEP_FILE_OPEN`

It may be better to default `keep_file_open` arguments to `None`, and then have `ArrayProxy.__init__` clobber it with the current value of the module-level `KEEP_FILE_OPEN`. This would avoid problems with other modules (e.g. `analyze`, `mghformat`) importing a stale version of `KEEP_FILE_OPEN` on startup.

How does that sound?

matthew-brett:
I'm not sure what you mean by the "stale version", but yes, that sounds like a reasonable way to go.

pauldmccarthy:
Great, I'll make the changes now :)

By 'stale', I just mean avoiding having multiple copies of this module level `KEEP_FILE_OPEN` flag floating about. So rather than `analyze.py` doing:
```python
from .arrayproxy import ArrayProxy, KEEP_FILE_OPEN
...
    def from_file_map(..., keep_file_open=KEEP_FILE_OPEN):
```

it would be better to do this:
```python
    def from_file_map(..., keep_file_open=None):
```

and then have `arrayproxy.py` do this:
```python
KEEP_FILE_OPEN_DEFAULT = False
class ArrayProxy(object):
...
    def __init__(..., keep_file_open=None):
        if keep_file_open is None:
            keep_file_open = KEEP_FILE_OPEN_DEFAULT
```

Because in the former there would be multiple versions of `KEEP_FILE_OPEN` floating around.

matthew-brett:
I see - thanks - yes, that sounds right.

nibabel/arrayproxy.py
```diff
@@ -37,6 +37,24 @@
 from .openers import ImageOpener, HAVE_INDEXED_GZIP
 
 
+KEEP_FILE_OPEN_DEFAULT = False
+"""This flag controls whether a new file handle is created every time an image
```
matthew-brett: I'd prefer the comment above, though I know it's a horrible nit-pick.

nibabel/tests/test_arrayproxy.py
```diff
@@ -423,18 +424,75 @@ def test_keep_file_open_default():
...
         assert proxy._keep_file_open
     # If no have_indexed_gzip, then keep_file_open should be False
     with mock.patch('nibabel.openers.HAVE_INDEXED_GZIP', False), \
          mock.patch('nibabel.arrayproxy.HAVE_INDEXED_GZIP', False), \
          mock.patch('nibabel.openers.SafeIndexedGzipFile', None, create=True):
+        proxy = ArrayProxy(fname, ((10, 10, 10), dtype), keep_file_open='auto')
+        assert not proxy._keep_file_open
+
+
+def test_keep_file_open_default():
```
matthew-brett: Nice.

nibabel/analyze.py
```diff
@@ -950,6 +950,18 @@ def from_file_map(klass, file_map, mmap=True):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
+        keep_file_open : { 'auto', True, False }, optional, keyword only
```
matthew-brett: I think you now need `{None, 'auto', True, False}`, and to say that None is the default, and gives the KEEP_FILE_OPEN_DEFAULT value.

matthew-brett: Likewise for the other docstrings.

nibabel/arrayproxy.py
```diff
@@ -125,6 +156,70 @@ def __init__(self, file_like, spec, mmap=True):
...
+        ``ArrayProxy``.
+        """
+        if keep_file_open is None:
+            keep_file_open = KEEP_FILE_OPEN_DEFAULT
+        # if keep_file_open is True/False, we do what the user wants us to do
+        if isinstance(keep_file_open, bool):
+            return keep_file_open
+        if keep_file_open != 'auto':
+            raise ValueError(
+                'keep_file_open should be one of {\'auto\', True, False }')
```
matthew-brett: Add None here too I think.

nibabel/arrayproxy.py
```diff
@@ -155,12 +250,34 @@ def inter(self):
     def is_proxy(self):
         return True
 
+    @contextmanager
+    def _get_fileobj(self):
+        """Create and return a new ``ImageOpener``, or return an existing one.
+
+        The specific behaviour depends on the value of the ``keep_file_open``
+        flag that was passed to ``__init__``.
+
```
matthew-brett: Extra blank line.

nibabel/fileslice.py
```diff
@@ -17,6 +17,20 @@
 SKIP_THRESH = 2 ** 8
 
 
+class _NullLock(object):
+    """The ``_NullLock`` is an object which can be used in place of a
```
matthew-brett: Single standalone line as first line in docstring.  Maybe:

```
Can be used as no-function dummy object in place of ``threading.lock``
```

nibabel/fileslice.py
```diff
@@ -634,29 +648,41 @@ def read_segments(fileobj, segments, n_bytes):
         absolute file offset in bytes and number of bytes to read
     n_bytes : int
         total number of bytes that will be read
+    lock : threading.Lock
```
matthew-brett: Can actually be any context manager - or at least the NullLock, no?

Maybe:

```
lock : {None, threading.Lock, lock-a-like}, optional

... None (the default) gives a lock-a-like object that does no locking.   
```

nibabel/fileslice.py
```diff
@@ -737,6 +763,9 @@ def fileslice(fileobj, sliceobj, shape, dtype, offset=0, order='C',
         returning one of 'full', 'contiguous', None.  See
         :func:`optimize_slicer` and see :func:`threshold_heuristic` for an
         example.
+    lock: threading.Lock, optional
```
matthew-brett: As for `read_segments` docstring.

nibabel/fileslice.py
```diff
@@ -750,7 +779,7 @@ def fileslice(fileobj, sliceobj, shape, dtype, offset=0, order='C',
     segments, sliced_shape, post_slicers = calc_slicedefs(
         sliceobj, shape, itemsize, offset, order)
     n_bytes = reduce(operator.mul, sliced_shape, 1) * itemsize
-    bytes = read_segments(fileobj, segments, n_bytes)
+    bytes = read_segments(fileobj, segments, n_bytes, lock)
```
matthew-brett: Maybe ``arr_data`` to avoid clash with `bytes` class?

nibabel/freesurfer/mghformat.py
```diff
@@ -491,6 +491,18 @@ def from_file_map(klass, file_map, mmap=True):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
+        keep_file_open : { 'auto', True, False }, optional, keyword only
```
matthew-brett: Add None?  See comments above.

nibabel/freesurfer/mghformat.py
```diff
@@ -523,6 +536,17 @@ def from_filename(klass, filename, mmap=True):
...
```
matthew-brett: None, from comments above.

nibabel/openers.py
```diff
@@ -60,13 +67,31 @@ def readinto(self, buf):
...
 
 
-def _gzip_open(fileish, *args, **kwargs):
-    gzip_file = BufferedGzipFile(fileish, *args, **kwargs)
+def _gzip_open(fileish, mode='rb', compresslevel=9):
+
+    # is this a file? if not we assume it is a string
+    is_file = hasattr(fileish, 'read') and hasattr(fileish, 'write')
+
+    # If we've been given a file object, we can't change its mode.
```
matthew-brett: Does this mean we silently ignore the passed mode in this case?  Raise error instead maybe?  Or error if `mode is not None`?

nibabel/spm99analyze.py
```diff
@@ -261,13 +261,26 @@ def from_file_map(klass, file_map, mmap=True):
             `mmap` value of True gives the same behavior as ``mmap='c'``.  If
             image data file cannot be memory-mapped, ignore `mmap` value and
             read array from file.
+        keep_file_open : { 'auto', True, False }, optional, keyword only
```
matthew-brett: None as option, from comments above.

nibabel/tests/test_arrayproxy.py
```diff
@@ -11,19 +11,24 @@
...
+from ..arrayproxy import (ArrayProxy, KEEP_FILE_OPEN_DEFAULT, is_proxy,
+                          reshape_dataobj)
+from ..openers import ImageOpener
 from ..nifti1 import Nifti1Header
 
 from numpy.testing import assert_array_equal, assert_array_almost_equal
 from nose.tools import (assert_true, assert_false, assert_equal,
                         assert_not_equal, assert_raises)
 from nibabel.testing import VIRAL_MEMMAP
+import mock
```
matthew-brett: I think this should go above the other less-standard test-related modules like numpy.testing.

nibabel/tests/test_arrayproxy.py
```diff
@@ -327,3 +332,188 @@ def check_mmap(hdr, offset, proxy_class,
             # Check invalid values raise error
             assert_raises(ValueError, proxy_class, fname, hdr, mmap='rw')
             assert_raises(ValueError, proxy_class, fname, hdr, mmap='r+')
+
+
+# An image opener class which counts how many instances of itself have been
+# created
+class CountingImageOpener(ImageOpener):
+
+    numOpeners = 0
```
matthew-brett: I believe this should be `num_openers` (mixedCase not allowed unless it's already the default naming scheme : https://www.python.org/dev/peps/pep-0008/

nibabel/tests/test_arrayproxy.py
```diff
@@ -327,3 +332,188 @@ def check_mmap(hdr, offset, proxy_class,
...
+                assert CountingImageOpener.numOpeners == i + 1
+            CountingImageOpener.numOpeners = 0
+            proxy_kfp = ArrayProxy(fname, ((10, 10, 10), dtype),
+                                   keep_file_open=True)
+            assert proxy_kfp._keep_file_open
+            for i in range(voxels.shape[0]):
+                x , y, z = [int(c) for c in voxels[i, :]]
+                assert proxy_kfp[x, y, z] == x * 100 + y * 10 + z
+                assert CountingImageOpener.numOpeners == 1
+        # Test that the keep_file_open flag has no effect if an open file
```
matthew-brett: Nice comments explaining the tests, thanks.

nibabel/tests/test_openers.py
```diff
@@ -20,6 +20,7 @@
 from nose.tools import (assert_true, assert_false, assert_equal,
                         assert_not_equal, assert_raises)
 from ..testing import error_warnings
+import mock
```
matthew-brett: mock before other testing imports.

nibabel/tests/test_arrayproxy.py
```diff
@@ -327,3 +332,188 @@ def check_mmap(hdr, offset, proxy_class,
...
+            assert not proxy._keep_file_open
+        with mock.patch('nibabel.openers.HAVE_INDEXED_GZIP', True), \
+             mock.patch('nibabel.arrayproxy.HAVE_INDEXED_GZIP', True), \
+             mock.patch('nibabel.openers.SafeIndexedGzipFile', gzip.GzipFile,
+                        create=True):
+            proxy = ArrayProxy(fname, ((10, 10, 10), dtype))
+            assert not proxy._keep_file_open
+        # KEEP_FILE_OPEN_DEFAULT=True should cause keep_file_open to be True,
+        # regardless of whether or not indexed_gzip is present
+        with mock.patch('nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT', True), \
```
matthew-brett: Any way of making this decorator so you can use it without having to retype it?

nibabel/tests/test_arrayproxy.py
pauldmccarthy: A throw-back from many years as a java programmer :)

nibabel/openers.py
pauldmccarthy: Yes - currently if `fileish` is a file-like object, the mode (if passed in) is ignored. 

We can't raise an error `if mode is not None` because its default value is `'rb'` - the `Opener.__init__` will set the `mode` argument if it is not provided by the caller.

How about I change the above behaviour, so that `Opener` passes `mode` through without changing/setting it, and then in `_gzip_open` do something like this:

```python
def _gzip_open(..., mode=None, ...):
    if hasattr(fileish, 'mode') and mode is None:
        mode = fileish.mode
    if hasattr(fileish, 'mode') and fileish.mode != mode:
        raise ValueError('Open file mode does not match requested mode')
```

pauldmccarthy: Actually, I think this is a problem which doesn't exist ... `_gzip_open` only ever gets called for file names - if `Opener.__init__` is passed an open file handle, it just stores it as-is, and returns.

So we don't need to worry about this at all :)

pauldmccarthy:
It's my turn to apologise - I'm on holiday from tomorrow for a week, without laptop. So there'll be a delay on my part.

effigies:
Nice! Thanks for suffering so many rounds of review, @pauldmccarthy.

PR #556, opened 2017-09-27
DOCTEST: Use legacy float array printing for now

effigies:
In numpy's pre-release, float arrays now have different printing strategies. Adding `#doctest: +NORMALIZE_WHITESPACE` doesn't resolve the issue, so for the moment we need to use their option to revert printing behavior to keep doctests from breaking.

Using `.. testsetup::` should avoid visual changes to the docs.

Introduced in https://github.com/numpy/numpy/commit/710e0327687b9f7653e5ac02d222ba62c657a718

### DIFF
--- a/nibabel/nicom/dwiparams.py
+++ b/nibabel/nicom/dwiparams.py
@@ -18,6 +18,11 @@ The B matrix ``B`` is a symmetric positive semi-definite matrix.  If
 
    B ~ (q_est . q_est.T) / norm(q_est)
 
+.. testsetup::
+
+>>> from distutils.version import LooseVersion
+>>> if LooseVersion(np.__version__) > LooseVersion('1.13.1'):
+...     np.set_printoptions(sign='legacy')
 '''
 import numpy as np
 import numpy.linalg as npl
--- a/nibabel/nicom/dwiparams.py
+++ b/nibabel/nicom/dwiparams.py
@@ -20,9 +20,9 @@ The B matrix ``B`` is a symmetric positive semi-definite matrix.  If
 
 .. testsetup::
 
->>> from distutils.version import LooseVersion
->>> if LooseVersion(np.__version__) > LooseVersion('1.13.1'):
-...     np.set_printoptions(sign='legacy')
+    from distutils.version import LooseVersion
+    if LooseVersion(np.__version__) > LooseVersion('1.13.1'):
+        np.set_printoptions(sign='legacy')
 '''
 import numpy as np
 import numpy.linalg as npl
--- a/nibabel/nicom/dwiparams.py
+++ b/nibabel/nicom/dwiparams.py
@@ -18,14 +18,10 @@ The B matrix ``B`` is a symmetric positive semi-definite matrix.  If
 
    B ~ (q_est . q_est.T) / norm(q_est)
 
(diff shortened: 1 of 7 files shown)


effigies:
@matthew-brett I think you'll know more about how to deal with [this error](https://travis-ci.org/nipy/nibabel/jobs/280429752) than I do. Is this an issue on matplotlib's end, or some buildbot that needs fixing?

```
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    matplotlib from https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com/matplotlib-2.1.0rc1+104.g4792425-cp27-cp27mu-manylinux1_x86_64.whl#sha256=0b50fa1925772cf0b6d7a386774c9f7671c758bdc3c2472e06f0e319da841848:
        Expected sha256 0b50fa1925772cf0b6d7a386774c9f7671c758bdc3c2472e06f0e319da841848
             Got        b49660821b0a93fd9c1dd8c0aafd5489b86524edd907e032f9347eebb84f47f1
```

nibabel/nifti1.py
```diff
@@ -9,6 +9,12 @@
 ''' Read / write access to NIfTI1 image format
 
 NIfTI1 format defined at http://nifti.nimh.nih.gov/nifti-1/
+
```
matthew-brett: Can we make this two separate strings, so the user doesn't get this testsetup in the docstring help?

nibabel/nicom/dwiparams.py
```diff
@@ -18,6 +18,11 @@
 
    B ~ (q_est . q_est.T) / norm(q_est)
 
+.. testsetup::
```
matthew-brett: Can we put this in a separate docstring?  See comment below.

nibabel/quaternions.py
```diff
@@ -23,6 +23,12 @@
 >>> M = quat2mat(q) # from this module
 >>> vec = np.array([1, 2, 3]).reshape((3,1)) # column vector
 >>> tvec = np.dot(M, vec)
+
```
matthew-brett: Ditto.

effigies:
Hopefully this works. I don't know all the rules to doctests.

effigies:
Is there some way to do a package-wide pre-doctest setup so we can have this in one place? Hard to google.

matthew-brett:
I don't know how to set up project wide doctest initialization - other than writing a specific nibabel plugin for the doctests / nose.

effigies:
I think I've resolved this. If there's a `setup_test()` function, that will get run before doctests (see [doctest fixtures](https://nose.readthedocs.io/en/latest/doc_tests/test_doctest_fixtures/doctest_fixtures.html)). At least locally, I was able to write this function in one place `nibabel.testing` and import it in others. We'll see if it does what we want with the pre-release installed.

effigies:
@matthew-brett Does this look reasonable?

matthew-brett:
Thanks - looks good - in it goes.

Issue #310, opened 2015-04-29
Provide at least reading capability for AFNI's (BRIK/HEAD) format


yarikoptic:
AFNI is quite widely used, and ATM I know no library to load those volumetric files in Python.  It would be really great if at least reading, even if limited (e.g. for a specific subbrick) capabilities were introduced.

matthew-brett:
Yo - do you have some files in this format you can share?

eamoncaddigan:
Hi! I'm also interested in this feature and would be happy to (privately) share data files.

eamoncaddigan:
AFNI is open source, but I don't know what license it has; it is likely less permissive than NiBabel's, so just translating its C code into Python is shady. It would *not* be a violation of AFNI's copyright if one programmer were to read their source and document the file format in order for a second programmer to write a python implementation. 

I would happy to be either of these two programmers!

yarikoptic:
On Tue, 22 Sep 2015, Eamon Caddigan wrote:

> AFNI is open source, but I don't know what license it has; it is likely less

FWIW AFNI core developments since 2000 is a product of NIH, ie.
USA goverment, so can't be copyrighted, thus public domain see
http://neuro.debian.net/debian/extracts/afni/copyright so depending on
which piece you are talking about it might be even more permissive than
nibabel ;)

-- 
Yaroslav O. Halchenko
Center for Open Neuroscience     http://centerforopenneuroscience.org
Dartmouth College, 419 Moore Hall, Hinman Box 6207, Hanover, NH 03755
Phone: +1 (603) 646-9834                       Fax: +1 (603) 646-1419
WWW:   http://www.linkedin.com/in/yarik

effigies:
Might also be worth noting that there are publicly available .BRIK/.HEAD files in http://afni.nimh.nih.gov/pub/dist/src/

bcipolli:
Just received some `BRIK`/`HEAD` files to work with. Has anybody looked at code snippets to make this happen, or should I find them myself?

bcipolli:
I was able to load my file using a MATLAB implementation here:
http://afni.nimh.nih.gov/pub/dist/src/matlab/BrikLoad.m

Perhaps we can port this to Python? Does anybody know reliable tools to convert code to get a head start?

yarikoptic:
On Sun, 20 Dec 2015, Ben Cipollini wrote:

> I was able to load my file using a MATLAB implementation here:
[...]

PR #561, opened 2017-10-02
NF: Functionality for reading AFNI BRIK/HEAD

rmarkello:
In partial response to #310, this is a work-in-progress for providing functionality to read AFNI BRIK/HEAD files. 

For example, the following should work with these changes:

```
>>> import nibabel as nib
>>> brik = nib.load('example_4d+orig.BRIK')
>>> brik.shape
(33, 41, 25, 3)
```

I tried to model the framework off other interfaces (e.g., copying functions from PARREC and MINC&mdash;hopefully that's okay!), only modifying what was necessary to suit the specifics of BRIK/HEAD format. As of now this is pretty bare bones, but it does seem to work on a variety of manual tests and a few automated ones adopted from `nibabel/tests/test_minc1.py`.

I'd appreciate any feedback or insight, including on what functionality/tests should be added.


effigies:
Hi, thanks for this. We've fixed some tests in master, if you want to rebase or merge.

matthew-brett:
Sorry to get to this so slowly - I've just started a new job ...

I noticed the reference to https://github.com/florisvanvugt/afnipy/blob/master/afni.py - but that project is GPL licensed - at a superficial first glance, you haven't used that code - is that right?  If you have, we have to get permission to use it without the GPL.

Just at a superficial glance, it looks like a very good job of a first pass, and pretty close to ready.  For my taste (and my interpretation of PEP8) you have too many blank lines, and I'd like to see some doctests, and eventually, we could negotiate some Image API tests for this image type (`tests/test_image_api.py`), but  generally - excellent - thanks.

rmarkello:
No need to apologize -- I've been meaning to add some more tests to get it out of its current failing state. I assumed that was the reason behind any delays!

As for the reference to [`afni.py`](https://github.com/florisvanvugt/afnipy/blob/master/afni.py): I didn't use any of that code, but rather used it for some insight into how AFNI data was stored in the BRIK file. I left the reference there to indicate the source of that insight, but it can certainly be removed since the official AFNI documentation is just as valid.

I'll work on paring down some of the blank lines and adding some doctests, but do you have a format you'd like followed for the latter? I (sort of) based this on an interpretation of [`parrec.py`](https://github.com/nipy/nibabel/blob/master/nibabel/parrec.py) and [`minc1.py`](https://github.com/nipy/nibabel/blob/master/nibabel/minc1.py), neither of which have doctests. I see some minimal ones in [`nifti1.py`](https://github.com/nipy/nibabel/blob/master/nibabel/nifti1.py) that I can follow, but even there most of the functions aren't tested. Thanks for the guidance!

Shotgunosine:
I emailed Floris and got him to update the license to MIT. I'm glad you had already found his code.

Shotgunosine:
Hey, we'd love to see this PR added. What needs to be done? Anything we can do to help?

effigies:
Doctests that generally mirror `nifti1.py` (when appropriate) seem like a good start. And from a quick glance through, I wonder if we can subclass `AFNIArrayProxy` from `ArrayProxy`, to avoid rewriting more than absolutely necessary.

Nothing else really stuck out to me from a cursory read through. Merging master into this branch should make it easier for the coverage checks to identify which lines aren't being tested, and we can see about improving coverage to make sure everything's behaving as it should.

Examples of invalid BRIK/HEAD images would also be useful for checking corner cases.

rmarkello:
Apologies for my neglect on this!

I was able to successfully subclass `AFNIArrayProxy` from `ArrayProxy`, getting rid of a few unnecessary functions in the process, and am currently working on some doc-tests. I'll take a look around for some invalid images, but might ask for some help from @Shotgunosine on that front.

Regarding coverage: would you prefer a merge or a rebase from master? I only ask given the explicit guidelines in the [documentation](http://nipy.org/nibabel/gitwash/development_workflow.html).

effigies:
Merges and rebases are equally acceptable. If you're comfortable with rebases, they're often cleaner to review on a commit-by-commit basis.

nibabel/brikhead.py
```diff
@@ -0,0 +1,499 @@
...
+    >>> brik = load('example4d+orig.BRIK')
+    >>> brik.shape
+    (33, 41, 25, 3)
+    >>> brik.affine
+    array([[ -3.    ,  -0.    ,  -0.    ,  49.5   ],
+           [ -0.    ,  -3.    ,  -0.    ,  82.312 ],
+           [  0.    ,   0.    ,   3.    , -52.3511],
+           [  0.    ,   0.    ,   0.    ,   1.    ]])
+    >>> head = load('example4d+orig.HEAD')
+    >>> (head.get_data() == brik.get_data()).all()
```
effigies: Breaking on memmap for old numpy. The following should work, though you may need to add an `import numpy as np` at the top of the doctest.

```Python
>>> np.array_equal(head.get_data(), brik.get_data())
True
```

rmarkello:
Pretty sure this is broken due to the new testing file I added. Will reduce its size, hopefully seeing the Appveyor fixed, later today!

effigies:
Seems to be specific to Python 3.5+, so I suspect a pretty small change will fix it.

effigies:
Oh, I see. You meant the test BRIK file, not the Python test files. Carry on. :-)

Shotgunosine:
I've asked around here and it doesn't look like anyone's got any bad brik-head files handy that we could test on.

matthew-brett:
Could you make some bad ones by setting bad values into the header?  Maybe you could save space by building and testing the bad BRIK files during the tests?

rmarkello:
I've added a few broken HEAD files where I manually changed the attributes, as suggested. Thankfully, since the HEAD file gets read in first, there's no need for an accompanying BRIK file since it will error out before getting there.

However, taking a closer look at the AppVeyor error, I'm a bit confused. From some digging, I think it's specific to the `example4d+orig.BRIK.gz` file (the doc-test that uses that file is failing, too). The only change made to that file prior to the AppVeyor failing was that I gzipped it (87873eef1203c3407b0f6f21636dbd68b442bc95) in order to test capabilities for detecting compressed files. I'm unclear why that would have screwed things up (and only on Python 3.5+, as @effigies pointed out). I'll keep digging, but if you have any ideas they'd be appreciated!

effigies:
From a [quick search](https://duckduckgo.com/?q=appveyor+%22Not+enough+storage+is+available+to+process+this+command%22+python&atb=v90-7_j&ia=web), it seems others have run up against this on AppVeyor when using `mmap`ed files. It's resolved by explicitly closing the file handles.

This seems like probably any easy fix to attempt before we dig deeper.

nibabel/brikhead.py
```diff
@@ -0,0 +1,487 @@
...
+            `mmap` value of True gives the same behavior as ``mmap='c'``.  If
+            image data file cannot be memory-mapped, ignore `mmap` value and
+            read array from file.
+        """
+        with file_map['header'].get_prepare_fileobj('rt') as hdr_fobj:
+            hdr = klass.header_class.from_fileobj(
+                hdr_fobj)
+        brik_fobj = file_map['image'].get_prepare_fileobj()
+        data = klass.ImageArrayProxy(brik_fobj, hdr,
+                                     mmap=mmap)
```
effigies: I think the issue might be here. I don't think you need to `get_prepare_fileobj` in this method.

```Python
data = klass.ImageArrayProxy(file_map['image'], hdr, mmap=mmap)
```

I think this will help release the filehandles between reads. Please see analyze.py for probably the most exercised (as the Nifti1Image superclass) variant of `from_file_map`.

nibabel/brikhead.py
```diff
@@ -0,0 +1,487 @@
...
+            `mmap` controls the use of numpy memory mapping for reading image
+            array data.  If False, do not try numpy ``memmap`` for data array.
+            If one of {'c', 'r'}, try numpy memmap with ``mode=mmap``.  A
+            `mmap` value of True gives the same behavior as ``mmap='c'``.  If
+            image data file cannot be memory-mapped, ignore `mmap` value and
+            read array from file.
+        """
+        file_map = klass.filespec_to_file_map(filename)
+        # only BRIK can be compressed, but `filespec_to_file_map` doesn't
+        # handle that case; remove potential compression suffixes from HEAD
```
effigies: As a point of personal preference, I would rather improve `filespec_to_filemap` to handle the case where header may not be compressed. (But if that makes more sense as a separate PR, just to get this in, that's fine by me.)

rmarkello: I think it'd be good to see this PR merged, first, if that's all right! Once this is in I'll be happy to raise an issue to discuss this change, since, as far as I can tell `filespec_to_file_map` is used by most image classes in nibabel.

effigies:
How's everybody feeling about this PR?

@rmarkello @Shotgunosine are you using this regularly and it's operating as you'd hope? Any further additions you're thinking of making?

@matthew-brett Have you had a chance to look over this? A lot of this plumbing is stuff you're much more familiar with than I am.

rmarkello:
I haven't run into any issues with my own use of this, but, admittedly, I've been utilizing it in some pretty pedestrian workflows (though I suppose that will be the normal use case!). 

In terms of further additions: I think, as mentioned above, modifying the `filespec_to_file_map` will be the next step.

nibabel/brikhead.py
matthew-brett: How about a little wrapper around filespec_to_filemap in this PR, to point the way, and keep that part of the logic separate?

nibabel/brikhead.py
```diff
@@ -0,0 +1,488 @@
...
+        head_fname = file_map['header'].filename
+        if not os.path.exists(head_fname):
+            for ext in klass._compressed_suffixes:
+                head_fname = re.sub(ext, '', head_fname)
```
matthew-brett: I think this is going to remove `gz` or whatever inside filenames - for example - consider `head_fname = 'foogz_bar.nii.gz'`. 

I think you want something like:

```
for ext in klass._compressed_suffixes:
     head_fname = head_fname[:-len(ext)] if head_fname.endswith(ext) else head_fname
```

nibabel/brikhead.py
```diff
@@ -0,0 +1,488 @@
...
+    >>> var = "type = string-attribute\\nname = TEMPLATE_SPACE\\ncount = 5\\n'ORIG~"
+    >>> name, attr = _unpack_var(var)
+    >>> print(name, attr)
+    TEMPLATE_SPACE ORIG
+    """
+    err_msg = ('.HEAD file appears to contain a misformed attribute. Please '
+               'check .HEAD file and try again.')
+    # get data type and key; if error, bad attribute/HEAD file
+    try:
+        atype = TYPE_RE.findall(var)[0]
```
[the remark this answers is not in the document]
rmarkello: There should never be more than one TYPE associated with an attribute in a HEAD file. If there is, the HEAD file is corrupted and an error should be raised. I'll add that check and make the error messages more explicit.

nibabel/brikhead.py
```diff
@@ -0,0 +1,488 @@
...
+        floatfacs = self.info.get('BRICK_FLOAT_FACS', None)
+        if floatfacs is None or not np.any(floatfacs):
+            return None
+        scale = np.ones(self.info['DATASET_RANK'][1])
+        floatfacs = np.atleast_1d(floatfacs)
+        scale[floatfacs.nonzero()] = floatfacs[floatfacs.nonzero()]
+
+        return scale
+
+    def get_slope_inter(self):
```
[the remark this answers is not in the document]
rmarkello: Because `AFNIArrayProxy` inherits from `ArrayProxy`, this needs to be included. `ArrayProxy` demands that the provided "`spec`" (i.e., `AFNIHeader` in this case) implement this (see: arrayproxy.py:140). I added the inheritance in response to @effigies suggestion, above.

nibabel/brikhead.py
```diff
@@ -0,0 +1,488 @@
...
+
+        Returns
+        -------
+        (x, y, z, t) : tuple of int
+        """
+        dset_rank = self.info['DATASET_RANK']
+        shape = tuple(self.info['DATASET_DIMENSIONS'][:dset_rank[0]])
+        n_vols = dset_rank[1]
+
+        return shape + (n_vols,) if n_vols > 1 else shape
```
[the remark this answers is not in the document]
rmarkello: Well, no. By default, AFNI datasets are actually always ndim=4 of shape (x, y, z, >=1), because of how the dimensions are stored in the HEAD file (see response below). You'll never get an (x, y, z) dataset because that would preclude you from subsetting along the "time" axis, and AFNI programs feature heavily in permitting you to do this. I'll make sure to remedy this in my next update.

nibabel/brikhead.py
```diff
@@ -0,0 +1,488 @@
...
+        Examples
+        --------
+        >>> header = AFNIHeader(parse_AFNI_header('scaled+tlrc.HEAD'))
+        >>> header.get_data_scaling()
+        array([  3.88336300e-08])
+        """
+        floatfacs = self.info.get('BRICK_FLOAT_FACS', None)
+        if floatfacs is None or not np.any(floatfacs):
+            return None
+        scale = np.ones(self.info['DATASET_RANK'][1])
```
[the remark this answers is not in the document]
rmarkello: Correct. DATASET_RANK[0] must be set to 3, denoting a 3D image, and DATASET_RANK[1] will always determine how many "sub-bricks" (in AFNI parlance) / volumes there are along the fourth (traditionally, but not exclusively) time axis. As in my above response, DATASET_RANK[1] will (at least as far as I'm aware) always be >=1.

rmarkello:
To do:
- [ ] Add relevant API tests to `test_image_api.py`
- [ ] Wrap `filename_parser.types_filenames` to account for BRIK/HEAD idiosyncracies currently being handled by `AFNIImage.filespec_to_file_map`

nibabel/brikhead.py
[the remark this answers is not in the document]
matthew-brett: Thanks - would you mind putting this in comments somewhere?  Maybe in an extra extended docstring at the top?  As in:

```
""" The current module docstring

Some comments about this module for public consumption
"""

""" More notes on the AFNI format

DATASET_RANK[0] must be set to 3, denoting a 3D image, and DATASET_RANK[1] will always determine how many "sub-bricks" (in AFNI parlance) / volumes there are along the fourth (traditionally, but not exclusively) time axis. As in my above response, DATASET_RANK[1] will (at least as far as I (RM) am aware) always be >=1.
"""
```

nibabel/brikhead.py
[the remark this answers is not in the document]
matthew-brett: Maybe add comment to this effect?

nibabel/brikhead.py
```diff
@@ -0,0 +1,558 @@
...
+            dict with keys ``image`` and ``header`` where values are fileholder
+            objects for the respective BRIK and HEAD files
+
+        Raises
+        ------
+        ImageFileError
+            If `filespec` is not recognizable as being a filename for this
+            image type.
+        """
+        # copied from filebasedimages.py
```
matthew-brett: Can you post-process instead?  As in, inherit from `FileBasedImage` (sorry - why aren't you doing that? - I'm sure there's a good reason), and post-process the result?

```
file_map = super(AFNIImage, klass).filespec_to_file_map(filespec)
```

and then use `file_map['image'].filename` and `file_map['header'].filename`?

Do you agree that it seems cleaner to raise an error for `my_image.HEAD.gz`, even if `my_image.BRIK.gz` exists?

nibabel/brikhead.py
```diff
@@ -0,0 +1,558 @@
...
+        """
+        file_map = klass.filespec_to_file_map(filename)
+        return klass.from_file_map(file_map, mmap=mmap)
+
+    @classmethod
+    def filespec_to_file_map(klass, filespec):
+        """
+        Make `file_map` from filename `filespec`
+
+        Deals with idiosyncracies of AFNI BRIK / HEAD formats
```
matthew-brett: Maybe add something like:

AFNI BRIK files can be compressed, but HEAD files cannot - see https://afni.nimh.nih.gov/pub/dist/doc/program_help/README.compression.html.  Let's say you have an AFNI file you want to load, with files `my_image.HEAD` and `my_image.BRIK.gz`. To identify the AFNI BRIK / HEAD pair, you can specify:

* The HEAD filename - e.g. ``my_image.HEAD``
* The BRIK filename without compressed extension, e.g. ``my_image.BRIK``
* The full BRIK filename - e.g. ``my_image.BRIK.gz``.

As a side note, I guess you're not proposing to support `.Z` compressed files.  I bet they are rare.

rmarkello:
Responded to most of @matthew-brett's comments, so this might be ready for another look! I was a bit skeptical of the coverage reports so I rebased onto master, too, just to ensure that wasn't impacting things.

rmarkello:
Hi all, just wanted to check in on this PR. I know thing are pretty busy in the repo but would love any feedback on this; it should be ready for review!

effigies:
Hi Ross. For my part, I was content with this PR when I last commented, so I haven't really been following the latest changes. I can make a point to review it in the next couple days, and let you know if I see anything, but I'll still leave it to Matthew to make the final call.

The only comment I can make right now is that you can check the [Codecov report](https://codecov.io/gh/nipy/nibabel/pull/561/diff) to see if there are any coverage failures that you think are worth covering.

nibabel/brikhead.py
```diff
@@ -0,0 +1,559 @@
...
+    def __array__(self):
+        raw_data = self.get_unscaled()
+        # datatype may change if applying self._scaling
+        return raw_data if self.scaling is None else raw_data * self.scaling
+
+    def __getitem__(self, slicer):
+        raw_data = super(AFNIArrayProxy, self).__getitem__(slicer)
+        # apply volume specific scaling (may change datatype!)
+        if self._scaling is not None:
+            scaling = self._scaling.copy()
```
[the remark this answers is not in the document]
rmarkello: I don't remember anymore! It doesn't seem necessary so I'll fix it asap.

rmarkello:
This should be ready to go, @matthew-brett ! Thanks so much for the helpful feedback.

effigies:
@matthew-brett Sorry to keep bothering you, but is this ready to go? I see you made some contributions... not sure if that was final touch-ups or not.

matthew-brett:
Chris - if you're happy for this to go in the release, please merge away.

effigies:
So. Happy.

Issue #558, opened 2017-09-28
Very slow slicing with indexed_gzip


effigies:
To reproduce, you can test on an OpenFMRI dataset:

```Shell
datalad install -r ///openfmri/ds000114
cd ds000114
git annex get sub-07/ses-test/anat/sub-07_ses-test_T1w.nii.gz
```

Install `indexed_gzip`:

```Python
>>> import nibabel as nib
>>> img = load('ds000114/sub-07/ses-test/anat/sub-07_ses-test_T1w.nii.gz')
>>> img.shape
(256, 156, 256)
>>> img.dataobj[0]
```

I'm not sure how long it takes. Slicing along the last dimension is fast, so it's not reading the file that's the issue.

I think the fix will be to disable `indexed_gzip` if `nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT is False`.

matthew-brett:
How long does it take if `indexed_gzip` is not installed?  I'd expect that to be very slow on a gzipped file in general, because you're doing lots of little reads far apart.

effigies:
```
In [23]: %time img.dataobj[0, :, :]
CPU times: user 3min 21s, sys: 3.79 s, total: 3min 25s
Wall time: 3min 25s
Out[23]: 
array([[ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       ..., 
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.]], dtype=float32)
```

```
In [26]: nib.openers.HAVE_INDEXED_GZIP=False

In [27]: x = nib.load('/data/bids/ds000114/sub-07/ses-test/anat/sub-07_ses-test_T1w.nii.gz')

In [28]: %time x.dataobj[0, :, :]
CPU times: user 376 ms, sys: 4 ms, total: 380 ms
Wall time: 380 ms
Out[28]: 
array([[ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       [ 0.,  0.,  0., ...,  0.,  0.,  0.],
       ..., 
[...]

Issue #557, opened 2017-09-28
Gzip file load error on latest master


solitaryangler:
Hi,
I needed the TCK streamline format so I installed the latest master branch with:
```pip install -e git+https://github.com/nipy/nibabel.git#egg=nibabel```
I can now load TCK streamlines:
```
nib.streamlines.FORMATS
Out[2]: 
{'.tck': nibabel.streamlines.tck.TckFile,
 '.trk': nibabel.streamlines.trk.TrkFile}
```
But I cannot load gzipped nifti files any more. If I gunzip the nifti, then it works fine. 
The error I get:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-4-6e9570ba8f5d> in <module>()
----> 1 nib.load('Diffusion_brain_mask.nii.gz')

.../Softwares/nibabel/src/nibabel2/nibabel/loadsave.py in load(filename, **kwargs)
     41     sniff = None
     42     for image_klass in all_image_classes:
---> 43         is_valid, sniff = image_klass.path_maybe_image(filename, sniff)
     44         if is_valid:
     45             img = image_klass.from_filename(filename, **kwargs)

.../Softwares/nibabel/src/nibabel2/nibabel/filebasedimages.py in path_maybe_image(klass, filename, sniff, sniff_max)
    508         sniff = klass._sniff_meta_for(filename,
    509                                       max(klass._meta_sniff_len, sniff_max),
--> 510                                       sniff)
    511         if sniff is None or len(sniff[0]) < klass._meta_sniff_len:
    512             return False, sniff

.../Softwares/nibabel/src/nibabel2/nibabel/filebasedimages.py in _sniff_meta_for(klass, filename, sniff_nbytes, sniff)
    455         # Attempt to sniff from metadata location
    456         try:
--> 457             with ImageOpener(meta_fname, 'rb') as fobj:
    458                 binaryblock = fobj.read(sniff_nbytes)
    459         except IOError:

.../Softwares/nibabel/src/nibabel2/nibabel/openers.py in __init__(self, fileish, *args, **kwargs)
    135         if 'compresslevel' in arg_names and 'compresslevel' not in kwargs:
    136             kwargs['compresslevel'] = self.default_compresslevel
--> 137         self.fobj = opener(fileish, *args, **kwargs)
    138         self._name = fileish
    139         self.me_opened = True
.../Softwares/nibabel/src/nibabel2/nibabel/openers.py in _gzip_open(filename, mode, compresslevel)
[...]

PR #562, opened 2017-10-02
Only use `indexed_gzip` when explicitly requested

pauldmccarthy:
This PR aims to address issue #558 - it should have been a part of PR #552 

For small files, the current version of `indexed_gzip` is much slower to use than the built-in `GzipFile` class. Therefore, `indexed_gzip` should not be used unless it is requested (via the `keep_file_open` flag, which is interpreted by the `nibabel.arrayproxy.ArrayProxy` class). In all other circumstances, the built-in `gzip.GzipFile` class should be used instead.

I have made this change by having the `ArrayProxy` pass a "hint" to the `Opener` class, indicating whether the file handle is going to be kept open for multiple accesses, or whether it is just for a one-time access. The `Opener` class (actually, the `nibabel.openers._gzip_open` function) then decides whether or not to use `indexed_gzip`.

### DIFF
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -113,8 +113,7 @@ class Opener(object):
         passed to opening method when `fileish` is str.  ``mode``, if not
         specified, is `rb`.  ``compresslevel``, if relevant, and not specified,
         is set from class variable ``default_compresslevel``. ``keep_open``, if
-        relevant, and not specified, is set from class variable
-        ``default_keep_open``.
+        relevant, and not specified, is ``False``.
     \*\*kwargs : keyword arguments
         passed to opening method when `fileish` is str.  Change of defaults as
         for \*args
@@ -130,9 +129,6 @@ class Opener(object):
     default_compresslevel = 1
     #: whether to ignore case looking for compression extensions
     compress_ext_icase = True
-    #: hint which tells us whether the file handle will be kept open for
-    #  multiple reads/writes, or just for one-time access.
-    default_keep_open = False
 
     def __init__(self, fileish, *args, **kwargs):
         if self._is_fileobj(fileish):
@@ -153,8 +149,7 @@ class Opener(object):
             kwargs['compresslevel'] = self.default_compresslevel
         # Default keep_open hint
         if 'keep_open' in arg_names:
-            if 'keep_open' not in kwargs:
-                kwargs['keep_open'] = self.default_keep_open
+            kwargs.setdefault('keep_open', False)
         # Clear keep_open hint if it is not relevant for the file type
         else:
             kwargs.pop('keep_open', None)
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -13,11 +13,19 @@ import bz2
 import gzip
 import sys
 from os.path import splitext
+from distutils.version import StrictVersion
 
-# is indexed_gzip present?
+# is indexed_gzip present and modern?
 try:
-    from indexed_gzip import SafeIndexedGzipFile
+    from indexed_gzip import SafeIndexedGzipFile, __version__ as version
+
+    if StrictVersion(version) < StrictVersion("0.6.0"):
+        raise ImportError('indexed_gzip is present, but too old '
+                          '(>= 0.6.0 required): {})'.format(version))
+
+    del version
     HAVE_INDEXED_GZIP = True
+
 except ImportError:
     HAVE_INDEXED_GZIP = False
 
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -12,6 +12,7 @@
 import bz2
 import gzip
 import sys
+import warnings
 from os.path import splitext
 from distutils.version import StrictVersion
 
@@ -19,12 +20,14 @@ from distutils.version import StrictVersion
 try:
     from indexed_gzip import SafeIndexedGzipFile, __version__ as version
 
-    if StrictVersion(version) < StrictVersion("0.6.0"):
-        raise ImportError('indexed_gzip is present, but too old '
-                          '(>= 0.6.0 required): {})'.format(version))
+    HAVE_INDEXED_GZIP = True
+
+    if StrictVersion(version) < StrictVersion('0.6.0'):
+        warnings.warn('indexed_gzip is present, but too old '
+                      '(>= 0.6.0 required): {})'.format(version))
+        HAVE_INDEXED_GZIP = False
 
     del version
-    HAVE_INDEXED_GZIP = True
 
 except ImportError:
     HAVE_INDEXED_GZIP = False
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -67,10 +67,10 @@ class BufferedGzipFile(gzip.GzipFile):
             return n_read
 
 
-def _gzip_open(filename, mode='rb', compresslevel=9):
+def _gzip_open(filename, mode='rb', compresslevel=9, keep_open=False):
 
     # use indexed_gzip if possible for faster read access
-    if mode == 'rb' and HAVE_INDEXED_GZIP:
+    if keep_open and mode == 'rb' and HAVE_INDEXED_GZIP:
         gzip_file = SafeIndexedGzipFile(filename)
 
     # Fall-back to built-in GzipFile (wrapped with the BufferedGzipFile class
@@ -101,12 +101,14 @@ class Opener(object):
     \*args : positional arguments
         passed to opening method when `fileish` is str.  ``mode``, if not
         specified, is `rb`.  ``compresslevel``, if relevant, and not specified,
-        is set from class variable ``default_compresslevel``
+        is set from class variable ``default_compresslevel``. ``keep_open``, if
+        relevant, and not specified, is set from class variable
+        ``default_keep_open``.
     \*\*kwargs : keyword arguments
         passed to opening method when `fileish` is str.  Change of defaults as
         for \*args
     """
-    gz_def = (_gzip_open, ('mode', 'compresslevel'))
+    gz_def = (_gzip_open, ('mode', 'compresslevel', 'keep_open'))
     bz2_def = (bz2.BZ2File, ('mode', 'buffering', 'compresslevel'))
     compress_ext_map = {
         '.gz': gz_def,
@@ -117,6 +119,9 @@ class Opener(object):
     default_compresslevel = 1
     #: whether to ignore case looking for compression extensions
     compress_ext_icase = True
+    #: hint which tells us whether the file handle will be kept open for
+    #  multiple reads/writes, or just for one-time access.
+    default_keep_open = False
 
     def __init__(self, fileish, *args, **kwargs):
         if self._is_fileobj(fileish):
@@ -132,8 +137,16 @@ class Opener(object):
         # Set default mode
         if 'mode' not in full_kwargs:
             kwargs['mode'] = 'rb'
+        # Default compression level
         if 'compresslevel' in arg_names and 'compresslevel' not in kwargs:
             kwargs['compresslevel'] = self.default_compresslevel
+        # Default keep_open hint
+        if 'keep_open' in arg_names:
+            if 'keep_open' not in kwargs:
+                kwargs['keep_open'] = self.default_keep_open
+        # Clear keep_open hint if it is not relevant for the file type
+        else:
+            kwargs.pop('keep_open', None)
         self.fobj = opener(fileish, *args, **kwargs)
         self._name = fileish
         self.me_opened = True
--- a/nibabel/tests/test_openers.py
+++ b/nibabel/tests/test_openers.py
@@ -260,7 +260,7 @@ def test_compressed_ext_case():
                 except ImportError:
                     SafeIndexedGzipFile = GzipFile
                 assert_true(isinstance(fobj.fobj, (GzipFile,
-                                                   SafeIndexedGzipFile))
+                                                   SafeIndexedGzipFile)))
             else:
                 assert_true(isinstance(fobj.fobj, BZ2File))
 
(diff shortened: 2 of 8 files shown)


effigies:
Thanks for this. It looks good to me on a first pass, but I don't think I'm thinking optimally this morning, so I'll have a more detailed look later, unless Matthew beats me to it.

effigies:
The failing Travis tests are related to #556. Don't worry about those.

pauldmccarthy:
In light of a discussion over at #557, I've also added a check to the present version of `indexed_gzip` - if it is older than 0.6.0, it is not used. 0.6.0 is the current latest version, and is also the first version which supports Windows.

nibabel/openers.py
```diff
@@ -117,6 +130,9 @@ class Opener(object):
     default_compresslevel = 1
     #: whether to ignore case looking for compression extensions
     compress_ext_icase = True
+    #: hint which tells us whether the file handle will be kept open for
+    #  multiple reads/writes, or just for one-time access.
+    default_keep_open = False
```
effigies: Is this class variable intended to be changed by a power user? Under what circumstances might I want to change that?

nibabel/openers.py
```diff
@@ -132,8 +148,16 @@ def __init__(self, fileish, *args, **kwargs):
         # Set default mode
         if 'mode' not in full_kwargs:
             kwargs['mode'] = 'rb'
+        # Default compression level
         if 'compresslevel' in arg_names and 'compresslevel' not in kwargs:
             kwargs['compresslevel'] = self.default_compresslevel
+        # Default keep_open hint
+        if 'keep_open' in arg_names:
+            if 'keep_open' not in kwargs:
+                kwargs['keep_open'] = self.default_keep_open
```
effigies: Can replace the internal `if` block with:

```Python
kwargs.setdefault('keep_open', self.default_keep_open)
```

nibabel/tests/test_openers.py
```diff
@@ -96,38 +97,55 @@ def test_BinOpener():
...
+        tests = [
+            (False, {'mode' : 'rb', 'keep_open' : True},  GzipFile),
+            (False, {'mode' : 'rb', 'keep_open' : False}, GzipFile),
+            (False, {'mode' : 'wb', 'keep_open' : True},  GzipFile),
+            (False, {'mode' : 'wb', 'keep_open' : False}, GzipFile),
+            (True,  {'mode' : 'rb', 'keep_open' : True},  MockIndexedGzipFile),
+            (True,  {'mode' : 'rb', 'keep_open' : False}, GzipFile),
+            (True,  {'mode' : 'wb', 'keep_open' : True},  GzipFile),
+            (True,  {'mode' : 'wb', 'keep_open' : False}, GzipFile),
+        ]
```
effigies: Nice.

nibabel/openers.py
pauldmccarthy: I'm not actually sure - I was just going with the convention set from the other kwargs. I can't really think of a use-case for changing it, so will change it default it to `False` in `__init__`.

effigies:
Since 0.6.1 was released, now getting issues with this line:

https://github.com/nipy/nibabel/blob/dbe74e1d86255a1ff93e1483e1cbc740dd091075/nibabel/tests/test_openers.py#L241

Should probably make it `SafeIndexedGzipFile`.

pauldmccarthy:
Aah true - I changed the inheritance hierarchy so that `SafeIndexedGzipFile` is no longer a sub-class of `IndexedGzipFile` (but rather, a subclass of `io.BufferedReader`).

You were testing against the master branch, right? I didn't pick this up on the PR branch, because `indexed_gzip` will no longer get used unless requested.

I've been quite busy this week, so still have a bit of work to do on the benchmarking code (and have had to make a few changes to other benchmark modules to get them to run). But I should be able to finalise the outstanding issues before the end of the week.

pauldmccarthy:
Ok, sorry for the delay ... I've knocked up a benchmarking script which compares `ArrayProxy`-based slicing on:
 - `GzipFile`/`keep_file_open=False`
 - `GzipFile`/`keep_file_open=True`
 - `SafeIndexedGzipFile`/`keep_file_open=True`

Each of these are compared against the time taken to perform an equivalent `ArrayProxy`-based slice on an uncompressed/mem-mapped image.

In its current form the script takes about 30 minutes to run. Some representative results are as follows:

```
+-------------------------------------------------------+------------------+------------------+------------------+------------------+
|                                                       |       Time       |  Baseline time   |    Time ratio    | Memory deviation |
+=======================================================+==================+==================+==================+==================+
| Type gzip, keep_open False, slice [?, :, :, :]        |  6.48            |  2.28            |  2.84            |  4.89            |
| Type gzip, keep_open False, slice [:, :, :, ?]        |  0.51            |  0.00            | 683.49           |  0.00            |
| Type gzip, keep_open False, slice [?, ?, ?, :]        |  1.01            |  0.00            | 1588.11          |  0.00            |
| Type gzip, keep_open True, slice [?, :, :, :]         |  6.53            |  2.27            |  2.88            |  0.06            |
| Type gzip, keep_open True, slice [:, :, :, ?]         |  0.36            |  0.00            | 476.13           |  0.00            |
| Type gzip, keep_open True, slice [?, ?, ?, :]         |  1.00            |  0.00            | 1639.86          |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, :, :, :] |  4.26            |  2.25            |  1.89            | 16.61            |
| Type indexed_gzip, keep_open True, slice [:, :, :, ?] |  0.02            |  0.00            | 18.37            |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, ?, ?, :] |  0.62            |  0.00            | 975.92           |  0.00            |
+-------------------------------------------------------+------------------+------------------+------------------+------------------+
```

I found it interesting that slices of the form `[?, ?, ?, :]` (e.g. time series for one voxel) are faster, but not that much faster, with `indexed_gzip`. I might be able to get better results by tuning some parameters (e.g. seek point spacing, readbuffer sizes).

The last commit in this push contains some changes that I had to make to get all of the existing `nibabel/benchmarks` scripts running on my system. I'm not too sure about these changes, as I don't know the typical environment in which the benchmarks are executed. If these are not necessary, let me know and I'll delete this commit.

effigies:
Yeah, I would see if upping your buffer size can improve performance a bit. On my system, the default is 8KiB, while we set (when possible) the GzipFile max read chunk at [100MiB](https://github.com/pauldmccarthy/nibabel/blob/8e1723fc2a15ebd064b3137d7ae5acb133bac66b/nibabel/openers.py#L37).

pauldmccarthy:
Although that fix is only applied to python < 3.4 - in 3.5 and above, the `GzipFile` uses `io.BufferedReader` with a default buffer size (also 8KiB on my system). In `indexed_gzip` 0.6.1 I set this buffer size to 1MB (this can be customised via `SafeIndexedGzipFile.__init__`).

Here are results after changing `_gzip_open` to do this (confusing parameter names to `__init__`, I know):
```python
        gzip_file = SafeIndexedGzipFile(filename,
                                        readbuf_size=GZIP_MAX_READ_CHUNK,
                                        buffer_size=GZIP_MAX_READ_CHUNK)
```

```
+-------------------------------------------------------+------------------+------------------+------------------+------------------+
|                                                       |       Time       |  Baseline time   |    Time ratio    | Memory deviation |
+=======================================================+==================+==================+==================+==================+
| Type gzip, keep_open False, slice [?, :, :, :]        |  6.42            |  2.27            |  2.83            |  5.57            |
| Type gzip, keep_open False, slice [:, :, :, ?]        |  0.51            |  0.00            | 648.70           |  0.00            |
| Type gzip, keep_open False, slice [?, ?, ?, :]        |  1.01            |  0.00            | 1469.82          |  0.00            |
| Type gzip, keep_open True, slice [?, :, :, :]         |  6.46            |  2.31            |  2.80            | -3.82            |
| Type gzip, keep_open True, slice [:, :, :, ?]         |  0.34            |  0.00            | 402.97           |  0.00            |
| Type gzip, keep_open True, slice [?, ?, ?, :]         |  1.03            |  0.00            | 1426.10          |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, :, :, :] |  2.89            |  2.25            |  1.28            | 115.96           |
| Type indexed_gzip, keep_open True, slice [:, :, :, ?] |  0.13            |  0.00            | 179.51           |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, ?, ?, :] |  0.71            |  0.00            | 1041.26          |  0.00            |
+-------------------------------------------------------+------------------+------------------+------------------+------------------+
```

It's made volume access slower! I guess because way more data is being read than necessary.

effigies:
Hmm. Okay. Still, failing back to standard GzipFile speeds isn't bad. As long as we're not significantly under-performing vanilla gzip for common operations, I'm okay with "This only improves some operations."

pauldmccarthy:
Well, it does improve performance on all of the tested slices, just not as much as I would like :). 

It's not shown directly in the benchmark output, but the most interesting comparison is between `gzip`/`keep_file_open=True`, and `indexed_gzip`/`keep_file_open=True`. Here are the speed-ups when using `indexed_gzip`:

```
slice [?, :, :, :] indexed_gzip / gzip: 6.53 / 4.26 = 1.53
slice [:, :, :, ?] indexed_gzip / gzip: 0.36 / 0.02 = 18.0
slice [?, ?, ?, :] indexed_gzip / gzip: 1.00 / 0.62 = 1.61
```

The biggest gain is for volume access (x18 speed-up, what I originally wrote `indexed_gzip` for). We get improvements in the other slice types, but they're not quite as impressive.

I'll try and put aside a few days over Christmas to do some more digging, and see if I can get some bigger speed-ups.

pauldmccarthy:
Well I am going to have to eat my words there - no change in results with more iterations - these results were with 200 iterations, on a shape of `(100, 100, 100, 100)`:

```
+-------------------------------------------------------+------------------+------------------+------------------+------------------+
|                                                       |       Time       |  Baseline time   |    Time ratio    | Memory deviation |
+=======================================================+==================+==================+==================+==================+
| Type gzip, keep_open False, slice [?, :, :, :]        |  6.48            |  2.26            |  2.86            |  7.24            |
| Type gzip, keep_open False, slice [:, :, :, ?]        |  0.50            |  0.00            | 622.59           |  0.00            |
| Type gzip, keep_open False, slice [?, ?, ?, :]        |  1.00            |  0.00            | 1407.09          |  0.00            |
| Type gzip, keep_open True, slice [?, :, :, :]         |  6.54            |  2.28            |  2.87            |  0.09            |
| Type gzip, keep_open True, slice [:, :, :, ?]         |  0.34            |  0.00            | 467.24           |  0.00            |
| Type gzip, keep_open True, slice [?, ?, ?, :]         |  1.01            |  0.00            | 1523.59          |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, :, :, :] |  4.22            |  2.22            |  1.90            | 16.72            |
| Type indexed_gzip, keep_open True, slice [:, :, :, ?] |  0.02            |  0.00            | 16.78            |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, ?, ?, :] |  0.64            |  0.00            | 1037.29          |  0.00            |
+-------------------------------------------------------+------------------+------------------+------------------+------------------+


slice [?, :, :, :] gzip / indexed_gzip: 6.54 / 4.22 =  1.55
slice [:, :, :, ?] gzip / indexed_gzip: 0.34 / 0.02 = 17.00
slice [?, ?, ?, :] gzip / indexed_gzip: 1.01 / 0.64 =  1.58
```

**But** if I run the benchmark on a bigger image (`(200, 200, 200, 100) float32`, ~ 3GiB uncompressed) things look a bit more promising:

```
+-------------------------------------------------------+------------------+------------------+------------------+------------------+
|                                                       |       Time       |  Baseline time   |    Time ratio    | Memory deviation |
+=======================================================+==================+==================+==================+==================+
| Type gzip, keep_open False, slice [?, :, :, :]        | 31.33            | 10.00            |  3.13            | 124.61           |
| Type gzip, keep_open False, slice [:, :, :, ?]        |  4.46            |  0.01            | 466.79           |  0.25            |
| Type gzip, keep_open False, slice [?, ?, ?, :]        |  7.45            |  0.00            | 11706.92         |  0.00            |
| Type gzip, keep_open True, slice [?, :, :, :]         | 30.54            | 10.55            |  2.89            | 15.26            |
| Type gzip, keep_open True, slice [:, :, :, ?]         |  3.17            |  0.01            | 328.21           |  0.25            |
| Type gzip, keep_open True, slice [?, ?, ?, :]         |  8.42            |  0.00            | 13253.83         |  0.00            |
| Type indexed_gzip, keep_open True, slice [?, :, :, :] | 27.22            | 10.82            |  2.52            |  1.68            |
| Type indexed_gzip, keep_open True, slice [:, :, :, ?] |  0.07            |  0.01            |  6.82            |  0.21            |
| Type indexed_gzip, keep_open True, slice [?, ?, ?, :] |  0.69            |  0.00            | 920.69           |  0.00            |
+-------------------------------------------------------+------------------+------------------+------------------+------------------+

slice [?, :, :, :] gzip / indexed_gzip: 30.54 / 27.22 =  1.12
slice [:, :, :, ?] gzip / indexed_gzip:  3.17 /  0.07 = 45.29
slice [?, ?, ?, :] gzip / indexed_gzip:  8.42 /  0.69 = 12.20
```

Performance for the first slice type `[?, :, :, :]` has gone down - it's still better than gzip, but not by much. But for the slice types that matter (`[:, :, :, ?]` == volume, and `[?, ?, ?, :]` == time course), we get big improvements in performance.

effigies:
I'm good with this if you want to merge or rebase to master to fix the tests.

effigies:
Thanks for the quick work, @pauldmccarthy.

PR #564, opened 2017-10-14
FIX: Set L/R labels in orthoview correctly

### LINKED ISSUE #559: flip dimensions in OrthoSlicer3D
Hey,
I'm loading sample subject (from freesurfer) T1.mgz file, both using OrthoSlicer3D and freeview (same slice: 133, 84, 122). It looks like in the coronal and axial views, the left and right are flipped:

OrthoSlicer3D:

orthoslicer3d

freeview:

freeview

Any idea why?
Thanks!

effigies:
Orthoview displays in neurological convention (left on the left, etc.) but places labels as if the display is radiological. This fix simply changes the labels.

If radiological convention is desired, we may want to consider adding an option, but that seems like it would make sense with a more significant refactor to simplify (or at least clarify) the visualization code.

Closes #559.


effigies:
Think this one should be pretty uncontroversial. Will merge tomorrow AM (EDT) unless there's an objection.

PR #565, opened 2017-10-18
[WIP] ENH: Load/save linear transform array (LTA) files

effigies:
This is a first pass at supporting loading and saving of LTA files. This is a pretty flexible structure that encodes the geometry (equivalent to the VOX-RAS affine) of the source and target spaces, and an affine transform between those spaces.

I'm using `LabeledWrapStruct` essentially as a way to enforce numpy typing on a dict. The `VolumeGeometry` type seems like it may recur a bit in FreeSurfer, as it [shows up in](https://github.com/nipy/nibabel/blob/b7eee37000de4c612624fdf9a61843cc20994fcd/nibabel/freesurfer/io.py#L50-L73) geometry files, as well.

The end goal is to support conversion between different affine formats. Because I'm working with the three formats regularly, I'll aim for LTA (.lta), FSL (.mat) and ITK (.txt) formats first.

A not-incidental secondary goal will be to document these different formats, as most of their existing documentation is the parsing/emitting code in the various packages that support these formats.

TODO:

* [ ] Support conversion between LTA types (`VOX_TO_VOX` and `RAS_TO_RAS`, at minimum)
* [ ] Support concatenation of LTA transforms

### DIFF
--- a/nibabel/freesurfer/lta.py
+++ b/nibabel/freesurfer/lta.py
@@ -1,0 +1,233 @@
+import numpy as np
+from ..wrapstruct import LabeledWrapStruct
+from ..volumeutils import Recoder
+
+transform_codes = Recoder((
+    (0, 'LINEAR_VOX_TO_VOX'),
+    (1, 'LINEAR_RAS_TO_RAS'),
+    (2, 'LINEAR_PHYSVOX_TO_PHYSVOX'),
+    (14, 'REGISTER_DAT'),
+    (21, 'LINEAR_COR_TO_COR')),
+    fields=('code', 'label'))
+
+
+class StringBasedStruct(LabeledWrapStruct):
+    def __init__(self,
+                 binaryblock=None,
+                 endianness=None,
+                 check=True):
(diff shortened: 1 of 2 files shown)


matthew-brett:
Chris - this may not be relevant - but there's some stuff on the FSL mat format here : https://github.com/nipy/dipy/blob/master/dipy/external/fsl.py#L69

matthew-brett:
Chris - where are we with this one?

effigies:
The status here is that it's stalled. Now I have a better understanding of the different affine conventions (i.e. FSL using a voxel-based origin, ITK ignoring sforms) the urgency of working through all of this has fallen away a little bit.

I'd be up for trying to make a roadmap for handling affine transforms, but I think that's going to involve settling on a common internal representation. I actually like the LTA approach, which includes both the source and target spaces, which means that this information is sufficient to produce an FSL or ITK matrix.

But #567 is my highest priority when I have time, followed by #550.

oesteban:
@effigies do you want to move this one to poldracklab/nitransforms until that is finished and we pull everything together in nibabel's proper?

PR #567, opened 2017-10-24
ENH: Add get_norm_zooms for zooms in mm/s units

effigies:
Because most NIfTI files are in mm+s units, it's pretty easy to write code that breaks on nonstandard units. This PR adds a `get_norm_zooms` method to spatial image headers that always returns zooms scaled to mm+s, based on metadata when available.

The basic API is:

```Python
img.header.get_norm_zooms()  # Zooms in mm+s, assuming already in mm+s if missing
img.header.get_norm_zooms(raise_unknown=True)  # Raise error if units not explicitly encoded
```

Before going further, just want to check if this seems in scope, here. If so, I'll add tests and check around the other image formats to make sure that, where they can encode units, we use it.

One consideration: Would it be reasonable to permit `img.header.get_norm_zooms(units=('meter', 'usec'))`? While this seems like a handy thing to let nibabel calculate, behavior on unknown units (e.g. when not set, or in formats that don't include unit data) seems problematic.


effigies:
This is ready for review. All behavior tested.

effigies:
Just a bump on this, now the holidays are over.

matthew-brett:
I hate to say it - but would you mind putting a test into `test_image_api.py`?

I'm suffering slightly about this one, because we otherwhere make our outputs canonical - for example, we make our affines point to RAS+ space, for DICOM.   So it seems kindve odd to let the formats do whatever with the zooms, and fix it with this function.

I guess it's too late to switch nifti / freesurfer to (mm, mm, mm, seconds) ?

How about `get_canonical_zooms`?

matthew-brett:
Or a flag to `get_zooms` ?

effigies:
Not too late for MGHHeader, since we haven't released with TR as zooms[3], but yeah, NIFTI uses raw zooms without correcting for units. Unless we want to treat that as a bug, but I think people will be depending on it.

We could deprecate get_zooms, or switch to a keyword argument to specify raw or canonical.

Finally, I'm okay with get_canonical_zooms, but it is getting long. Does it make sense to move to a canonical zooms property? Or even header.zooms that behaves canonically, combined with a get/set_zooms deprecation that warns of new semantics.

And sure, I'll add tests.

matthew-brett:
I believe I avoided using a property on the basis that the setter would also modify the header (at least, it does now).   So, it would break the property manifesto : https://github.com/nipy/nibabel/wiki/property_manifesto

I could imagine another model, where we left the header unmodified from its load state, and stored things like zooms etc inside the image, but that would be an API break.

Could you say more about what people are depending on?  Do you think that some people currently have meters set in their header, but are using the units as mm (for example)?

effigies:
Okay. So no property. I'd be okay with doing the option to `get_zooms`, if that seems cleaner. We'd presumably want a deprecation schedule to move from a default of raw to default of canonical.

And an example of depending on current behavior:

```Python
zooms = np.array(reoriented.header.get_zooms()[:3])
xyz_unit = img.header.get_xyzt_units()[0]
atol = {'meter': 1e-5, 'mm': 0.01, 'micron': 10}[xyz_unit]
rescale = not np.allclose(zooms, target_zooms, atol=atol)
```

(Adapted from https://github.com/poldracklab/fmriprep/blob/682b507c141254a51316b2367f9834708eb4409e/fmriprep/interfaces/images.py#L236-L256)

Here we took into account the fact that the zooms may not be in mm. If the zooms start being mm but `get_xyzt_units` still says they're in meters, then we'll be doing the wrong thing.

matthew-brett:
Right - but I guess the snippet leaves open whether anyone actually has micron etc nifti images ...

A flag seems like the right way to go.   I think we should also default Freesurfer to using seconds - what do you think?

effigies:
Yes, that's true. I've seen someone say they have ms TR images (they ran into issues with an FSL tool that assumed sec), but I don't know if I've ever heard of a meter image in the wild.

So how about a `units` parameter to `get_zooms`. I'd be inclined to make the options `'raw'` and `'canonical'`, given this discussion, but open to alternatives. We can default to `canonical` on any image type known to only allow `mm/s`, and FreeSurfer, which we'll coerce.

For nifti and any others where we are not guaranteed that `raw == canonical`, we should default to `None`, which will provide raw results but warn that the default will switch to canonical (version 4)?

Do you know MINC units off the top of your head? I think I had a look but couldn't find them.

effigies:
Does bring up a question of `set_zooms`. What's the appropriate analog of the `units` flag there?

effigies:
A thing I just thought about: If the zooms are in microns, then the affine is presumably from VOX -> RAS (um), right? Do we need to account for that at all?

effigies:
@matthew-brett I have some unpushed changes in the pipeline*, but if you have a minute to give your thoughts on the API issues I brought up in the last three comments, I'd appreciate it.

`*` Moving to `get_zooms(units='raw'|'canonical')`; holdup is in trying to prevent a huge number of `FutureWarning`s littering the test outputs, and even more so building up a will to dig into `warnings` context managers.

nibabel/tests/test_nifti1.py
```diff
@@ -1049,6 +1074,123 @@ def test_write_scaling(self):
...
+        img_klass = self.image_class
+        arr = np.arange(120, dtype=np.int16).reshape((2, 3, 4, 5))
+        aff = np.eye(4)
+        img = img_klass(arr, aff)
+
+        # Unknown units = 2 warnings
+        with clear_and_catch_warnings() as warns:
+            warnings.simplefilter('always')
+            assert_array_almost_equal(img.header.get_zooms(units='norm'),
+                                      (1, 1, 1, 1))
```
effigies: It's unclear to me why (in Python 2 only) this is failing to raise warnings.

nibabel/analyze.py
```diff
@@ -402,7 +402,7 @@ def from_header(klass, header=None, check=True):
                                      klass))
         obj.set_data_dtype(header.get_data_dtype())
         obj.set_data_shape(header.get_data_shape())
-        obj.set_zooms(header.get_zooms())
+        obj.set_zooms(header.get_zooms(units='raw'), units='raw')
```
effigies: Perhaps using `units='norm'` in both cases would be better, as it would also entail setting sensible units for NIfTI. But that may be making a bad assumption if you're in a non-temporal 4th dimension case (ppm, rads, hz).

nibabel/ecat.py
```diff
@@ -581,8 +581,10 @@ def get_frame_affine(self, frame=0):
                                                               z_off])
         return aff
 
-    def get_zooms(self, frame=0):
+    def get_zooms(self, frame=0, units='norm', raise_unknown=False):
         """returns zooms  ...pixdims"""
+        if units not in ('norm', 'raw'):
+            raise ValueError("`units` parameter must be 'norm' or 'raw'")
         subhdr = self.subheaders[frame]
         x_zoom = subhdr['x_pixel_size'] * 10
```
effigies: I wasn't sure whether to make `'raw'` change this factor to 1.

effigies:
This is ready for review, though there's one major thing to resolve with the tests (see comment).

I've moved to `units='norm'` instead of `units='canonical'` for brevity, and added it to `set_zooms`.

effigies:
Tests passing, and full coverage for new code.

effigies:
Just a bump on this one, which is ready for review.

nibabel/freesurfer/mghformat.py
```diff
@@ -237,40 +237,75 @@ def _ndims(self):
...
             zooms
+        units : {'norm', 'raw'}, optional
+            Zooms are specified in normalized units of mm/sec for
+            spatial/temporal dimensions or as raw values to be stored in
+            header.
+
+        .. _mghformat: https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat#line-82
         '''
+        if units not in ('norm', 'raw'):
+            raise ValueError("`units` parameter must be 'norm' or 'raw'")
```
yarikoptic: too frequent of a code pattern.  In current design I guess the only way to mitigate is to come up with `self._check_units(units)` to centralize it

nibabel/nifti1.py
```diff
@@ -1621,12 +1621,159 @@ def set_xyzt_units(self, xyz=None, t=None):
...
+        Returns
+        -------
+        zooms : tuple
+            tuple of header zoom values
+
+        '''
+        if units is None:
+            units = 'raw'
+            warnings.warn('Units not specified in `{}.get_zooms`. Returning '
+                          'raw zooms, but default will change to normalized.\n'
```
yarikoptic: might be worth announcing in what version this change is to happen (4.0)?

Also I am not sure if it would be best to keep default of get_zooms here (`raw`) be different from all other ones (`norm`)

effigies: I'm okay with following whatever schedule is considered best. If I understand the overall philosophy, it's okay to break the API at 3.0; it's just that if we were planning to remove something entirely we'd need to schedule final removal for 4.0. I could be wrong though.

Also, the reason to keep this as `raw` for NIFTI is that anybody who has been correctly handling non-mm/s units will find they're suddenly double-correcting if we start correcting for them. I do agree that diverging defaults is dissatisfying, but it seemed least bad at the time. Open to other approaches.

nibabel/freesurfer/mghformat.py
effigies: We could do something like that.

effigies:
@yarikoptic @matthew-brett Can we go ahead and argue this out a little bit?

When adding an argument for `.get_zooms()`, the default behavior needs to be the existing behavior, to avoid breaking things that were correctly written. For NIFTI, that means raw zooms. However, we want the default to become normalized (mm/s) in future versions (3.0?), which means we need a period of bugging users/downstream devs so that they don't get stung by the change. But because users cannot change to setting `units='norm'` or similar until this change goes in, immediately warning will make it very difficult for downstream libraries to support multiple versions of nibabel without annoying users.

In addition, the recently augmented `MGHImage` has TRs in ms units, which will be surprising to pretty much anybody who expects zooms to be consistent across images.

So the NIFTI/MGH images seem to demand opposite defaults: raw/normalized, respectively. Now, given that we'll be annoying users, is it sensible to go ahead and set the default to `units='norm'` for all images, and just put up a `UserWarning` (or something that will be filtered to display only once) noting that the behavior has changed?

I'm not coming to a very satisfying conclusion in any direction, and starting to consider whether we might want a new normalized-zoom getter method after all, but before I convince myself to try something like that, I want your input. What's the least bad way to go here?

nibotmi:
:umbrella: The latest upstream changes (presumably #630) made this pull request unmergeable. Please resolve the merge conflicts.

nibotmi:

effigies:
Just a bump on this. @yarikoptic @matthew-brett I'm not sure how best to resolve the issues brought up in https://github.com/nipy/nibabel/pull/567#issuecomment-386336861.

effigies:
Rethinking this API with an eye to consistency, I have an alternative approach:

```Python
class SpatialHeader:
    def _get_raw_zooms(self):
        # old get_zooms goes here

    def _get_zoom_units(self):
        # Return units in the same order as zooms, one per zoom
        # Simple default:
        return tuple('unknown' for _ in self._get_raw_zooms())
        # Hard-coded formats can return a static tuple

    def get_zooms(self, *, units=None):
        raw_zooms = self._get_raw_zooms()
        if units is None:
            return raw_zooms
        # units is tuple of (spatial, temporal, other)
        zoom_units = self._get_zoom_units()
        return _rescale(raw_zooms, zoom_units, units)
```

For the `units` parameter, I think we should make the inputs simple for the very common cases where we have 3D images or 4D time series. For instance, I think both of the following should work intuitively:

```Python
img.get_zooms(units='mm')           # Rescale spatial zooms only
img.get_zooms(units=('mm', 'sec'))  # Rescale spatial and temporal zooms
```

For less common cases, more complex specifications seem acceptable. One option to make sure that all possibilities can be specified would be to accept a `units` tuple with a unit per dimension.

PR #569, opened 2017-10-25
RF: Simplify MGHImage and add footer fields

effigies:
This PR includes a number of simplifications to the `MGHHeader` and `MGHImage` data structures.

The most significant is subclassing `MGHHeader` from `LabeledWrapStruct`. It appeared to have been a reimplementation of `LabeledWrapStruct` in many places without the `endianness` option. I have simply raised an exception on attempts to instantiate as little-endian.

The next most significant change is moving from `Mdc` (matrix of direction cosines) and `Pxyz_c` (center point in world coordinates) to the more explicit `x/y/z/c_ras` column vectors. These are consistent with most FreeSurfer literature and program outputs, and correctly encode the shape of the direction cosines as column-major rather than row-major (needing to be transposed when reading/writing).

Somewhat confusingly, there are two ways of referring to the voxel and world coordinates, and `XYZ` means world in one, and voxel in the other, and in both cases the alternative includes "R" and "S".

| Voxel | World |
|-------|-------|
| **C**olumn **R**ow **S**lice | XYZ |
| XYZ | **R**ight **A**nterior **S**uperior |

As CRS/XYZ appears to be FreeSurfer-unique, I'd prefer to settle on XYZ/RAS, and perhaps consider non-XYZ variable names when they can reduce confusion.

TODO:

* [x] Update tests with new header fields
* [x] Annotate header dtype with descriptions of fields (possibly further updating field names, shapes)
* [x] Improve footer handling (not targeting full features, but access to TR would be good)
* [x] Fix `get_data_shape` handling (see #550)


effigies:
This is ready for review, if anybody's got a few minutes.

nibabel/freesurfer/mghformat.py
```diff
@@ -188,58 +157,21 @@ def from_fileobj(klass, fileobj, check=True):
...
-        pxyz_0 = hdr['Pxyz_c'] - np.dot(Mdc, np.dot(d, pcrs_c))
-        M = np.eye(4, 4)
-        M[0:3, 0:3] = np.dot(Mdc, d)
-        M[0:3, 3] = pxyz_0.T
-        return M
+        affine = np.eye(4)
+        hdr = self._structarr
+        MdcD = np.hstack((hdr['x_ras'], hdr['y_ras'], hdr['z_ras'])) * hdr['voxelsize']
+        vol_center = MdcD.dot(hdr['dims'][:3].reshape(-1, 1)) / 2
+        affine[:3, :3] = MdcD
```
matthew-brett:
```
return nib.affines.from_matvec(MdcD, hdr['c_ras'] - vol_center)
```

?

nibabel/freesurfer/mghformat.py
```diff
@@ -384,32 +311,37 @@ def get_slope_inter(self):
...
-        hdr['delta'][:] = np.array([1, 1, 1])
-        hdr['Mdc'][0][:] = np.array([-1, 0, 0])  # x_ras
-        hdr['Mdc'][1][:] = np.array([0, 0, -1])  # y_ras
-        hdr['Mdc'][2][:] = np.array([0, 1, 0])   # z_ras
-        hdr['Pxyz_c'][:] = np.array([0, 0, 0])   # c_ras
-        return hdr
+
+        Ignores byte order; always big endian
+        '''
+        structarr = super(MGHHeader, klass).default_structarr()
```
matthew-brett: Here might be worth checking and error if endian not big.

nibabel/freesurfer/mghformat.py
```diff
@@ -604,33 +541,22 @@ def _write_data(self, mghfile, data, header):
...
-        hdr['delta'][:] = delta
-        hdr['Mdc'][:, :] = Mdc.T
-        hdr['Pxyz_c'][:] = Pxyz_c[:3]
+        voxelsize = voxel_sizes(self._affine)
+        Mdc = self._affine[:3, :3] / voxelsize
+        c_ras = self._affine.dot(np.vstack((shape / 2.0, [1])))[:3]
+
+        # Assign after we've had a chance to raise exceptions
+        hdr['voxelsize'] = voxelsize
+        hdr['x_ras'] = Mdc[:, [0]]
```
matthew-brett:
```
hdr['x_ras'], hdr['y_ras'], hdr['z_ras'] = Mdc.T
```

?

nibabel/freesurfer/mghformat.py
```diff
@@ -13,30 +13,38 @@
...
-    ('delta', '>f4', (3,)),
-    ('Mdc', '>f4', (3, 3)),
-    ('Pxyz_c', '>f4', (3,))
+    ('version', '>i4'),             # 0; must be 1
+    ('dims', '>i4', (4,)),          # 4; width, height, depth, nframes
+    ('type', '>i4'),                # 20; data type
+    ('dof', '>i4'),                 # 24; degrees of freedom
+    ('ras_good', '>i2'),            # 28; *_ras fields valid
+    ('voxelsize', '>f4', (3,)),     # 30; zooms (X, Y, Z)
+    ('x_ras', '>f4', (3, 1)),       # 42; X direction cosine column
```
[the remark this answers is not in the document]
effigies: By including all three columns as a single, row-ordered `Mdc` matrix, we're exposing a transposed matrix. It was being transposed correctly in and out, but if someone does decide to use it directly, they need to know this. I prefer explicit to silently misleading...

I could use `(3,)` if that's important. But it will mean users *knowing* that they're columns and not row vectors (can add to the docs).

For context, this is the comparable portion of the C data structure, which is just 16 individual `float` fields: https://github.com/freesurfer/freesurfer/blob/5367eec84621b331271e97f4e1f1502c3f6d0d87/include/mri.h#L184-L191

nibabel/freesurfer/mghformat.py
```diff
@@ -13,30 +13,38 @@
...
-    ('dof', '>i4'),
-    ('goodRASFlag', '>i2'),
-    ('delta', '>f4', (3,)),
-    ('Mdc', '>f4', (3, 3)),
-    ('Pxyz_c', '>f4', (3,))
+    ('version', '>i4'),             # 0; must be 1
+    ('dims', '>i4', (4,)),          # 4; width, height, depth, nframes
+    ('type', '>i4'),                # 20; data type
+    ('dof', '>i4'),                 # 24; degrees of freedom
+    ('ras_good', '>i2'),            # 28; *_ras fields valid
```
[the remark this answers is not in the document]
effigies: Yeah, I can re-add a `_header_data` structure (which I've replaced with `LabeledWrapStruct._structarr`) that will expose the old fields and raise a `DeprecationWarning`.

nibabel/freesurfer/mghformat.py
```diff
@@ -84,6 +92,7 @@ class MGHHeader(object):
 
     def __init__(self,
                  binaryblock=None,
+                 endianness='>',
```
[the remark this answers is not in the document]
effigies: I did this because of places where `LabeledWrapStruct` calls `__init__` assuming the type signature of the base class. I can instead make sure to override any methods that make such calls, just figured this would be a little shorter.

nibabel/freesurfer/mghformat.py
```diff
@@ -188,58 +157,21 @@ def from_fileobj(klass, fileobj, check=True):
...
-        M = np.eye(4, 4)
-        M[0:3, 0:3] = np.dot(Mdc, d)
-        M[0:3, 3] = pxyz_0.T
-        return M
+        affine = np.eye(4)
+        hdr = self._structarr
+        MdcD = np.hstack((hdr['x_ras'], hdr['y_ras'], hdr['z_ras'])) * hdr['voxelsize']
+        vol_center = MdcD.dot(hdr['dims'][:3].reshape(-1, 1)) / 2
+        affine[:3, :3] = MdcD
+        affine[:3, [3]] = hdr['c_ras'] - vol_center
```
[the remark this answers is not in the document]
effigies: The RHS is a column, so the `[3]` keeps the LHS expecting a column. Is there a trick I don't know here?

effigies:
Re: Renaming, the biggest motivation is that a lot of the field names seem only to align with one slideshow that documents the spaces/transforms, and not at all with either the FreeSurfer source code or the outputs of `mri_info` or other FreeSurfer tools, let alone how anybody outside FreeSurfer refers to these spaces.

So my strong preference is to move to a more consistent naming scheme (I'm open to alternative suggestions), but I'm also perfectly happy to provide a backwards-compatible interface to access the data using the old field names to get back the old shapes.

nibabel/freesurfer/mghformat.py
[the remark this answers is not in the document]
matthew-brett: OK - fine to leave as is (3,1) etc.

nibabel/freesurfer/mghformat.py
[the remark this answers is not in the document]
matthew-brett: I played with the idea of making WrapStruct have a parent or similar that had forced endianness, but I seem to remember it got a bit messy.

effigies:
Okay. I've pinged the Python neuroimaging list and the FreeSurfer support list. Hopefully we can find somebody willing to share their opinions on the Internet. :-)

mwaskom:
I don't have a strong opinion here, but I do think it's more likely that a user has encountered the fscoordinates slide deck than read through the freesurfer source code.

satra:
@effigies - perhaps chatting with doug would be a good idea as well. also folks at corticometrics may have some suggestions as well.

however, i do not have a strong opinion either. since i rarely do much with those files than reading into an array, doing some processing, and writing. or just visualization.

effigies:
@satra Can I expect them to pay respond to post on the FreeSurfer list, or is there a preferred way of contacting them more directly?

satra:
@effigies - i hadn't seen that the mail was on the freesurfer list. hopefully they respond - otherwise it's your call :)

effigies:
Okay, it sounds like there's a slight preference from @mwaskom to follow the FS documentary slideshow and a preference (of unknown strength) from @matthew-brett to avoid changing names without a Very Good Reason. I'll defer to this for now and revert `Mdc` and `Pxyz_c`, because I don't think it's critical to my more pressing goals here of getting zooms for 4D images to include TR (to match the behavior of Analyze-like images) and enforcing the 3/4D constraint.

However, I do think I would like to move to more "standard" FreeSurfer naming, because this construct ("volume geometry", or dimensions + x/y/z/c_ras) shows up in several places, including `m3z` warp files, LTA source and destination geometries (see #565), and in a text appendage to surface files:

https://github.com/nipy/nibabel/blob/2139ce0d24e65a83295bb6b3eaaf005eaeaebb5f/nibabel/freesurfer/io.py#L50-L73

https://github.com/nipy/nibabel/blob/2139ce0d24e65a83295bb6b3eaaf005eaeaebb5f/nibabel/freesurfer/io.py#L462-L488

But I think it does make sense to postpone the discussion of field naming to an actual attempt to provide a common interface to volume geometries (which is actually re-implemented in several FreeSurfer data structures).

matthew-brett:
Chris - sorry - I should have said - I don't care too much about the name changes, especially if there's a deprecation route.  I should have said before, but if you think these names are more canonical, that's OK - we can always explain in the docstring and comments what the relationship is to the slide deck.

effigies:
Okay. I think I'll still push it off for now. If/when we add support for LTA/M3Z etc, it will be time enough to consider a consistent naming scheme, rather than deprecating the old one now and finding later that we would have preferred something slightly different.

I will add docs though noting that `Mdc` is a transposed matrix. (It's a little annoying that I can't specify row/column order in the dtype.)

effigies:
@matthew-brett Got back to this. I think I've addressed all of your comments, and I've fleshed out some more tests. Ready for another review.

effigies:
Hi @matthew-brett, just bugging you about this. No worries if you're busy.

nibabel/freesurfer/mghformat.py
```diff
@@ -282,44 +223,71 @@ def set_data_dtype(self, datatype):
...
-        return tuple(zooms[:])
+        # Do not return time zoom (TR) if 3D image
+        tzoom = (self['tr'],)[:self._ndims() > 3]
+        return tuple(self._structarr['delta']) + tzoom
 
     def set_zooms(self, zooms):
         ''' Set zooms into header fields
 
-        See docstring for ``get_zooms`` for examples
+        Sets the spaing of voxels in the x, y, and z dimensions.
```
matthew-brett: Typo 'spacing'

nibabel/freesurfer/mghformat.py
```diff
@@ -282,44 +223,71 @@ def set_data_dtype(self, datatype):
...
+        For four-dimensional files, a temporal zoom (repetition time, or TR, in
```
matthew-brett: Ouch - it's really in milliseconds?  That's unfortunate (compared to nifti).

effigies: Annoyingly. Fortunately #567 proposes a way to get normalized zooms. (It was working on that that led to this PR in the first place.)

effigies: Reference: https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat#line-82

matthew-brett: Add reference to docstring?

matthew-brett:
Thanks for your patience, in it goes.

matthew-brett:
Oops - a couple of PEP8 failures : https://travis-ci.org/nipy/nibabel/builds/318192252

effigies:
Fixed in 7afc6d2680c2a15c4ebcbe99a88878e7361b8075. Feel free to cherry-pick into master if #550 isn't ready to merge.

PR #580, opened 2017-11-12
set sform with set_sform not get_sform

justbennet:
Small typo correction in the Working with NIfTI images doc.


pauldmccarthy:
There's another typo there - two 'with's :)

justbennet:
I added a  second commit to remove the second with.  I hope that was the right procedure.

effigies:
Perfect.

PR #582, opened 2017-11-18
BF: fix doctest failure with pre-release numpy

matthew-brett:
Numpy output printing has changed in upcoming 1.14 - trigger legacy
printing for doctests.

Chris - I had to do `from nibabel.testing import setup_test as setup_module` to
make this work. Is this also true for the other imports of this function?


effigies:
That wasn't the case before, but things may have changed, for all I know. Testing this locally.

effigies:
The issues in Travis seem to have been pip-related, which have resolved themselves. This works on my end, too. The AppVeyor issue seems entirely unrelated. Merging.

PR #584, opened 2017-11-22
MAINT: Remove conda from Appveyor script

matthew-brett:
We can get away without conda, now wheels are available.

Simplify file otherwise, by removing unnecessary quotes and backslashes.


effigies:
LGTM, if it passes.

PR #587, opened 2017-12-20
MAINT: Use SSH address to use key-based auth

effigies:
I've had issues pushing updated docs to the HTTPS URL. Anybody that has HTTPS write access should have SSH, so this seems safe.


matthew-brett:
Thanks ...

Issue #586, opened 2017-12-13
First point of `LazyTractogram` is not saved


nilgoyette:
This test should pass

    def test_save_from_generator(self):
        tractogram = Tractogram(DATA['streamlines'],
                                affine_to_rasmm=np.eye(4))

        # Just to create a generator
        filtered = (s for s in tractogram.streamlines if True)
        lazy_tractogram = LazyTractogram(lambda: filtered,
                                         affine_to_rasmm=np.eye(4))

        for ext, _ in FORMATS.items():
            with InTemporaryDirectory():
                filename = 'streamlines' + ext
                nib.streamlines.save(lazy_tractogram, filename)
                tfile = nib.streamlines.load(filename, lazy_load=False)
                assert_tractogram_equal(tfile.tractogram, tractogram)

The problem comes from

    first_item = next(iter(self.tractogram))
    ...
    for t in tractogram:

When `self.tractogram` is a generator, `first_item` is not in the generator anymore.

I took 10 minutes to fix it but I couldn't find a clean way to do it. A simple `itertools.chain([first_item], tractogram)` is not enough :)

effigies:
@MarcCote Do you have time to look into this?

nilgoyette:
I fixed it, but I really don't like what I did.

    reinsert_first_point = isinstance(self.tractogram, LazyTractogram)
    ...
    try:
        first_item = next(iter(self.tractogram))
    ...
    if reinsert_first_point:
        backup = self.tractogram._streamlines

        def chain_first_and_all_other():
            yield from itertools.chain([first_item.streamline], backup())

        self.tractogram._streamlines = chain_first_and_all_other
    ...
    tractogram = self.tractogram.to_world(lazy=True)

If @MarcCote or anyone have a better solution, go for it!

MarcCote:
@nilgoyette the streamlines you provided to LazyTractogram is not a generator function, i.e. calling it should provide new generators every time. For instance,
```python
In [3]: gen = (s for s in range(5))

In [4]: lmb = lambda: gen

In [5]: print(list(lmb()))
[0, 1, 2, 3, 4]

In [6]: print(list(lmb()))
[]
```

MarcCote:
As @nilgoyette told me offline, this particular scenario happens when you do tracking and you want to dump streamlines into a file as you generate them. This is obviously a valid reason. We definitively want to support **writing** in "lazy" mode.

[...]

PR #588, opened 2017-12-21
BF: #596 First point of `LazyTractogram` is not saved

MarcCote:
With this PR, we now assume tractogram objects can be iterated over only once at saving time.

This PR fixes issue #586.
Thanks to @nilgoyette for providing the missing unit test and part of the solution :+1:.

### DIFF
--- a/nibabel/streamlines/trk.py
+++ b/nibabel/streamlines/trk.py
@@ -7,7 +7,6 @@ import os
 import struct
 import string
 import warnings
-import itertools
 
 import numpy as np
 import nibabel as nib
--- a/nibabel/streamlines/trk.py
+++ b/nibabel/streamlines/trk.py
@@ -424,11 +424,16 @@ class TrkFile(TractogramFile):
             i4_dtype = np.dtype("<i4")  # Always save in little-endian.
             f4_dtype = np.dtype("<f4")  # Always save in little-endian.
 
-            # Make sure streamlines are in rasmm then send them to voxmm.
+            # Since the TRK format requires the streamlines to be saved in
+            # voxmm, we first transform them accordingly. The transformation
+            # is performed lazily since `self.tractogram` might be a
+            # LazyTractogram object, which means we might be able to loop
+            # over the streamlines only once.
             tractogram = self.tractogram.to_world(lazy=True)
             affine_to_trackvis = get_affine_rasmm_to_trackvis(header)
             tractogram = tractogram.apply_affine(affine_to_trackvis, lazy=True)
-            # Assume looping over the streamlines can be done only once.
+
+            # Create the iterator we'll be using for the rest of the funciton.
             tractogram = iter(tractogram)
 
             try:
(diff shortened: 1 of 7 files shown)


nilgoyette:
Much cleaner than my version. Thank you.

MarcCote:
This PR is ready for reviews. :-)
@nilgoyette @effigies @matthew-brett

nilgoyette:
I don't have any concern with this PR; it fixes the problem and it's clean. +1 to merge.

nibabel/streamlines/trk.py
```diff
@@ -423,8 +424,20 @@ def save(self, fileobj):
...
+            # Make sure streamlines are in rasmm then send them to voxmm.
+            tractogram = self.tractogram.to_world(lazy=True)
+            affine_to_trackvis = get_affine_rasmm_to_trackvis(header)
+            tractogram = tractogram.apply_affine(affine_to_trackvis, lazy=True)
+            # Assume looping over the streamlines can be done only once.
+            tractogram = iter(tractogram)
+
             try:
-                first_item = next(iter(self.tractogram))
+                # Use the first element to check
```
matthew-brett: I guess you could refactor this out into:

```
def peek_next(iterable):
    next_item = next(iterable)
    return next_item, itertools.chain([next_item], iterable)
```

nibabel/streamlines/trk.py
```diff
@@ -423,8 +424,20 @@ def save(self, fileobj):
             i4_dtype = np.dtype("<i4")  # Always save in little-endian.
             f4_dtype = np.dtype("<f4")  # Always save in little-endian.
 
+            # Make sure streamlines are in rasmm then send them to voxmm.
```
matthew-brett: Could you say a bit more about this change?

nibabel/streamlines/trk.py
MarcCote: Good idea. Will do.

nibabel/streamlines/trk.py
MarcCote: Because `self.tractogram` could be a LazyTractogram, peeking the first element requires to call `iter()` on it (i.e. start looping through the TractogramItems), but later I have to send the streamlines to voxmm, that is calling `apply_affine` on `self.tractogram` which would have been advanced by one element (because of peek_next).    

The alternative would have been to create a new LazyTractogram from the iterable return by peek_next, then call the apply_affine function. I just thought it would be simpler to do the transformation first. It shouldn't change the overall logic of the method: peek element to get some data type information, apply affine and save to disk.

nibabel/streamlines/trk.py
MarcCote: Done

nibabel/streamlines/trk.py
matthew-brett: Could you add something like this as an explanation in the code?

MarcCote: Better?

matthew-brett:
Yes, looks good - in it goes.

PR #590, opened 2018-01-10
BF: if voxel order in header is lowercase, .trk can't load

kesshijordan:
If the voxel_order field of the header is lowercase in a track file (this is the case when I re-save streamlines using trackvis on my mac: Trackvis Version 0.6.1 (Build 2017.01.17) Public), it will not load using the new nib.streamlines.load because axcodes2ornt returns nan's when given a lower case voxel order.

### DIFF
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -350,9 +350,8 @@ def axcodes2ornt(axcodes, labels=None):
            [ 2.,  1.]])
     """
 
-    axcodes = tuple(c.upper() for c in axcodes)
-
     if labels is None:
+        axcodes = tuple(c.upper() for c in axcodes)
         labels = list(zip('LPI', 'RAS'))
     n_axes = len(axcodes)
     ornt = np.ones((n_axes, 2), dtype=np.int8) * np.nan
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -349,9 +349,8 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
-
     if labels is None:
-        axcodes = tuple(c.upper() for c in axcodes)
+        axcodes = tuple(c.upper() if c is not None else None for c in axcodes)
         labels = list(zip('LPI', 'RAS'))
     n_axes = len(axcodes)
     ornt = np.ones((n_axes, 2), dtype=np.int8) * np.nan
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -349,7 +349,12 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
-    axcodes = axcodes.upper()
+
+    upper_str = []
+    for c in axcodes:
+        upper_str.append(c.upper())
+    axcodes = tuple(upper_str)
+
     if labels is None:
         labels = list(zip('LPI', 'RAS'))
     n_axes = len(axcodes)
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -349,6 +349,7 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
+    axcodes = axcodes.upper()
     if labels is None:
         labels = list(zip('LPI', 'RAS'))
     n_axes = len(axcodes)
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -350,10 +350,7 @@ def axcodes2ornt(axcodes, labels=None):
            [ 2.,  1.]])
     """
 
-    upper_str = []
-    for c in axcodes:
-        upper_str.append(c.upper())
-    axcodes = tuple(upper_str)
+    axcodes = tuple(c.upper() for c in axcodes)
 
     if labels is None:
         labels = list(zip('LPI', 'RAS'))
--- a/nibabel/streamlines/tests/test_trk.py
+++ b/nibabel/streamlines/tests/test_trk.py
@@ -17,7 +17,7 @@ from ..tractogram import Tractogram
 from ..tractogram_file import HeaderError, HeaderWarning
 
 from .. import trk as trk_module
-from ..trk import TrkFile, encode_value_in_name, decode_value_from_name
+from ..trk import TrkFile, encode_value_in_name, decode_value_from_name, get_affine_trackvis_to_rasmm
 from ..header import Field
 
 DATA = {}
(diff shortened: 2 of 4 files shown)


matthew-brett:
Thanks a lot for this - @MarcCote - any reason we should insist on lower-case voxel orientations?

@kesshijordan - any chance of adding a test?

kesshijordan:
Thanks, @matthew-brett. Sure, though do you know how to fix the travis failures, above? The spacing on the arrays disagrees on expected vs got (though the numbers match) in the DOC_DOC_TEST=1 builds.

effigies:
Looks like that's an effect of numpy releasing 1.14 (related: #556, #582), and us not running `DOC_DOC_TEST` on prereleases. I wouldn't worry about it in this PR.

kesshijordan:
Thanks, @effigies

@matthew-brett I added a test comparing the trackvis to rasmm affines between two simulated trk files (voxel order las and LAS). I put it in the test_load_file_with_wrong_information() set. Is that fine?

MarcCote:
@matthew-brett no, there is no reason why we couldn't support both. That said, shouldn't we fix that in `nibabel.orientations.axcodes2ornt`? Is there any reason why we couldn't support lower case letter for the orientation?

matthew-brett:
Good point - lower case should be OK for the axis codes in orientations.py ...

kesshijordan:
@matthew-brett and @MarcCote : I moved the change to `nibabel.orientations.axcodes2ornt` by inserting `axcodes = axcodes.upper()` as the first line. I think the test I added before should still work, since it would break if 'las' and 'LAS' were treated differently. Is this okay?

nibabel/orientations.py
```diff
@@ -349,6 +349,7 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
+    axcodes = axcodes.upper()
```
MarcCote: According to the docsring, `axcodes` is a tuple not a str. So, `.upper()` does not exist.

kesshijordan: Thanks; I hadn't noticed that : ) It fixed my loading issue because the .trk files I'm trying to load have orientation stored as a string, but that does break on tuples. I exchanged that line for a loop that generates an upper-case tuple from either a string or a tuple (see below). How's this?

```
upper_str = []
    for c in axcodes:
        upper_str.append(c.upper())
    axcodes = tuple(upper_str)
```

effigies: How about the following?

```Python
axcodes = tuple(c.upper() for c in axcodes)
```

kesshijordan: Thanks, @effigies

effigies:
Sorry to push more work on you, but do you think you could quickly add some test cases to `axcodes2ornt`?

https://github.com/nipy/nibabel/blob/ee4ce2a646cd49ea382d24f73e58b429d4795a49/nibabel/tests/test_orientations.py#L278-L314

Something like:

```Python
lia_ornt = [[0, -1], [2, -1], [1, 1]]
for lia_axcodes in ('LIA', 'lia', ('L', 'I', 'A'), ('l', 'i', 'a')):
    assert_array_equal(axcodes2ornt(lia_axcodes), lia_ornt)
```

nibabel/orientations.py
```diff
@@ -349,6 +349,9 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
+
+    axcodes = tuple(c.upper() for c in axcodes)
```
effigies: This is breaking on `labels is not None`. Probably the least disruptive approach is to move this into the `if labels is None` branch immediately below.

My reasoning is that, if somebody's passing labels, they'll also be passing axis codes with labels selected from that set. (The alternative would be to also apply `.upper()` to the labels, but I don't like the smell of that.)

effigies:
Hi @kesshijordan, just checking in. I think this is very nearly ready, if you have a few minutes. If you're swamped, I can submit a PR to your branch with my suggested changes.

kesshijordan:
Hi, @effigies. Thanks for checking in; I've been swamped but I have a bit of time now. I'll submit changes soon.

nibabel/orientations.py
```diff
@@ -349,7 +349,9 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
+
```
effigies: Just for a clean diff, could you remove this empty line?

nibabel/orientations.py
```diff
@@ -349,7 +349,9 @@ def axcodes2ornt(axcodes, labels=None):
...
     if labels is None:
+        axcodes = tuple(c.upper() for c in axcodes)
```
effigies: Looks like we did not consider `c is None`. Maybe update this to:

```Python
axcodes = tuple(c.upper() if c is not None else None for c in axcodes)
```

kesshijordan:
Absolutely. Thanks for all the pointers, @effigies. I have made those corrections and resubmitted.

effigies:
Looks ready to me. @MarcCote @matthew-brett any further comments?

nibabel/orientations.py
```diff
@@ -350,6 +350,7 @@ def axcodes2ornt(axcodes, labels=None):
            [ 2.,  1.]])
     """
     if labels is None:
+        axcodes = tuple(c.upper() if c is not None else None for c in axcodes)
```
matthew-brett: Should this be run only when `labels is None`?

effigies: Yes. Unless the labels themselves are also to be filtered. See https://github.com/nipy/nibabel/pull/590#discussion_r163340494.

matthew-brett: Doesn't that smell off though - to be case insensitive only when the labels are not specified?

I guess the ideal would be to be case insensitive only for codes labels in `LRAPIS`.

effigies: I see, so the issue is if someone passes an alternative default ordering of the normal codes, but not custom labels, we want the overall behavior to be unchanged. I was under the impression that the only time you wanted to specify labels would be to use some custom labeling.

Then yeah, that might make sense. What if it becomes:

```Python
if labels is None:
    labels = list(zip('LPI', 'RAS'))

# Enable case-insensitive axis codes for standard direction labels
if set(elem for pair in labels for elem in pair) == set('LRAPIS'):
    axcodes = tuple(...)
```

effigies: @matthew-brett Can you please confirm whether I've summarized your concern and proposed a reasonable solution? If so, I think we can get this fix in pretty soon.

matthew-brett: Sorry to be a bit imprecise.

Now I think about this more, it still seems ugly to me to special case for characters `LRAPIS`.  Why these and not other codes?   For example, what if someone uses `F` and `B` for front and back?  Wouldn't they expect the same behavior?

I've very sorry to keep changing, but can't we go back to expecting upper case for the codes from the caller, and, in particular, fixing this in the trackvis caller rather than here?

kesshijordan: @matthew-brett would it make sense to put in a "header format check" that returns a more informative error? That way the user can see why the load is failing and modify the header themself?

matthew-brett:
@kesshijordan - I've made an alternative PR, as a suggestion.  It's got your initial changes, plus (I think) some code to give a more informative error, as you suggested above.  Is that OK with you?

matthew-brett:
Sorry - the PR is linked below, but for completeness - #600

kesshijordan:
That sounds good, @matthew-brett. Thanks!

nibotmi:
:umbrella: The latest upstream changes (presumably #600) made this pull request unmergeable. Please resolve the merge conflicts.

matthew-brett:
@kesshijordan - OK to close this one?

MarcCote:
Yes, I'm good with the changes of #600.

kesshijordan:
Yes, @matthew-brett. Okay to close this one.

PR #591, opened 2018-01-12
DOC: fix doctests for numpy 1.14 array printing.

matthew-brett:
Numpy 1.14 has a different output format for the array.  There's a bug
in `numpy.set_printoptions(legacy='1.13')` so we can't use that to set
the print format back to its previous output:

https://github.com/numpy/numpy/issues/10383

Depend on numpy 1.14 for the doctests.

### DIFF
--- a/doc/source/coordinate_systems.rst
+++ b/doc/source/coordinate_systems.rst
@@ -104,7 +104,7 @@ middle voxel in the EPI data array like this:
     (26, 30, 16)
     >>> center_vox_value = epi_img_data[center_i, center_j, center_k]
     >>> center_vox_value
-    81.549287796020508
+    81.5492877960205...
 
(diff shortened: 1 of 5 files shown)


effigies:
Does it make sense to go ahead and update all the docstring tests, and get rid of the setup test function?

Actually, it's not clear why those doctests aren't breaking, if legacy printing is buggy.

matthew-brett:
I guess we didn't hit the particular case that exposes the bug.

I'm easy either way for converting the doctests.  Not converting means we can run the doctests on various versions, which does keep our test coverage up for versions of numpy < 1.14.  Maybe we can convert if we hit this bug in the doctests at some point.

effigies:
Sounds good. +1 for merge.

matthew-brett:
Thanks - I'll go ahead, so we've got all green on master again.

PR #592, opened 2018-01-24
Fix freesurfer `read_annot`, and enhance `write_annot`

pauldmccarthy:
This PR comprises two sections, related to the `nibabel.freesurfer.io.read_annot` and `write_annot` functions. The first involves changes to `read_annot`, to fix some bugs, and the second involves enhancements to `write_annot`, to make saving `.annot` files a little easier.

Bug fixes to `read_annot`
-------------------------

After a few hours of having my mind bent by the [craziness of the Freesurfer `.annot` file format](https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation), I discovered a few bugs in the `nibabel.freesurfer.io.read_annot` function. The following simple test case will trigger the buggy behaviour:

```python
import numpy as np
from nibabel.freesurfer import read_annot, write_annot
from nibabel.tmpdirs import InTemporaryDirectory

def test_annot_readback():

    # This annot file will store a LUT for a mesh made of 10 vertices, with
    # 3 colours in the LUT.
    nvertices = 10
    nlabels = 3

    names = ['label {}'.format(l) for l in range(1, nlabels + 1)]

    # randomly generate a label for each vertex, making sure
    # that at least one of each label value is present. Label
    # values are in the range (0, nlabels-1) - they are used
    # as indices into the lookup table (generated below).
    labels = list(range(nlabels)) + \
             list(np.random.randint(0, nlabels, nvertices - nlabels))
    labels = np.array(labels, dtype=np.int32)
    np.random.shuffle(labels)

    # Generate some random colours for the LUT
    rgbal = np.zeros((nlabels, 5), dtype=np.int32)
    rgbal[:, :4] = np.random.randint(0, 255, (nlabels, 4))

    # But make sure we have at least one large alpha, to make sure that when
    # it is packed into a signed 32 bit int, it results in a negative value
    # for the annotation value.
    rgbal[0, 3] = 255

    # Generate the annotation values for each LUT entry
    rgbal[:, 4] = (rgbal[:, 0] +
                   rgbal[:, 1] * (2 ** 8) +
                   rgbal[:, 2] * (2 ** 16) +
                   rgbal[:, 3] * (2 ** 24))

    annot_path = 'c.annot'

    with InTemporaryDirectory():
        write_annot(annot_path, labels, rgbal, names)
        labels2, rgbal2, names2 = read_annot(annot_path)
        assert np.all(np.isclose(rgbal2, rgbal))
```

This fails against the current implementation of `read_annot` because:

 - The annotation value for new-format `.annot` files is being incorrectly generated - the alpha value is not being included.
 - All alpha values are being overwritten with 255 for some unknown reason
 - `np.int` is used rather than `np.int32`, when generating the annotation values. When the top bit is 1, an `int32` will roll around to a negative value (as it should in this instanace), but the use of `np.int` causes such a value to be interpreted as a large positive integer.

I'm guessing that these errors are a throwback from the development period for this code.

Enhancements to `write_annot`
-----------------------------

The Freesurfer `.annot` file format is mildly insane and awkward to use. Alongside each vertex ID is stored the ABRG colour, packed into a signed 32 bit integer. So in order to identify the LUT entry that corresponds to a given vertex, you have to actually scan through the LUT until you find the matching colour<sup>*</sup>.

The `read_annot` function does this for us, so it's no big deal. But, if you want to generate an `.annot` file (as I did, just to write a couple of unit tests for my own code), you have to do the reverse, and generate the "annotation values" for every vertex. This procedure is shown in the above example.

So I'm proposing a small change to `write_annot`, to get it to do this for us, and that this is the new default behaviour - it accepts a new `fill_ctab` parameter, which defaults to `True`. So instead of having to do:

```python
rgbal = np.zeros((nlabels, 5), dtype=np.int32)
rgbal[:, :4] = np.random.randint(0, 255, (nlabels, 5))
rgbal[:,  4] = (rgbal[:, 0] +
                rgbal[:, 1] * (2 ** 8) +
                rgbal[:, 2] * (2 ** 16) +
                rgbal[:, 3] * (2 ** 24))

write_annot(annot_path, labels, rgbal, names)
```

With this PR, you can now do the following (and I am proposing that `fill_ctab` defaults to `True`):

```python
rgbal = np.random.randint(0, 255, (nlabels, 4), dtype=np.int32)
write_annot(annot_path, labels, rgbal, names, fill_ctab=True)
```

<sup>*</sup> Although this only allows you to get the corresponding structure name, so you don't need to bother if all you want is the colour.

### DIFF
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -518,7 +518,7 @@ def write_annot(filepath, labels, ctab, names, fill_ctab=True):
             np.array([num]).astype(dtype).tofile(fobj)
 
         def write_string(s):
-            s = s.encode() + b'\00'
+            s = (s if isinstance(s, bytes) else s.encode()) + b'\x00'
             write(len(s))
             write(s, dtype='|S%d' % len(s))
 
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -402,7 +402,7 @@ def read_annot(filepath, orig_ids=False):
     return labels, ctab, names
 
 
-def write_annot(filepath, labels, ctab, names):
+def write_annot(filepath, labels, ctab, names, fill_ctab=True):
     """Write out a Freesurfer annotation file.
 
     See:
@@ -418,6 +418,11 @@ def write_annot(filepath, labels, ctab, names):
         RGBA + label id colortable array.
     names : list of str
         The names of the labels. The length of the list is n_labels.
+    fill_ctab : bool
+        If True, the annotation values for each vertex  are automatically
+        generated. In this case, the provided `ctab` may have shape
+        (n_labels, 4) or (n_labels, 5) - if the latter, the final column is
+        ignored.
     """
     with open(filepath, "wb") as fobj:
         dt = ">i4"
@@ -429,6 +434,16 @@ def write_annot(filepath, labels, ctab, names):
         def write_string(s):
             write(len(s))
             write(s, dtype='|S%d' % len(s))
+
+        # Generate annotation values for each ctab entry
+        if fill_ctab:
+            new_ctab = np.zeros((ctab.shape[0], 5), dtype=np.int32)
+            new_ctab[:, :4] = ctab[:, :4]
+            ctab = new_ctab
+            ctab[:, 4] = (ctab[:, 0] +
+                          ctab[:, 1] * (2 ** 8) +
+                          ctab[:, 2] * (2 ** 16) +
+                          ctab[:, 3] * (2 ** 24))
 
         # vtxct
         write(vnum)
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -12,6 +12,14 @@ from six.moves import xrange
 from ..openers import Opener
 
 
+_ANNOT_DT = ">i4"
+"""Data type for Freesurfer `.annot` files.
+
+Used by :func:`read_annot` and :func:`write_annot`.  All data (apart from
+strings) in an `.annot` file is stored as big-endian int32.
+"""
+
+
 def _fread3(fobj):
     """Read a 3-byte int from an open binary file object
 
@@ -74,8 +82,10 @@ def _read_volume_info(fobj):
 
 
 def _pack_rgba(rgba):
-    """Used by :meth:`read_annot` and :meth:`write_annot` to pack an RGBA
-    sequence into a single integer.
+    """Pack an RGBA sequence into a single integer.
+
+    Used by :func:`read_annot` and :func:`write_annot` to generate
+    "annotation values" for a Freesuerfer `.annot` file.
 
     Parameters
     ----------
@@ -86,7 +96,7 @@ def _pack_rgba(rgba):
     Returns
     -------
 
-    out : ndarray, shape (n, )
+    out : ndarray, shape (n, 1)
         Annotation values for each colour.
     """
     bitshifts = 2 ** np.array([[0], [8], [16], [24]], dtype=rgba.dtype)
@@ -316,7 +326,20 @@ def write_morph_data(file_like, values, fnum=0):
 
 
 def read_annot(filepath, orig_ids=False):
-    """Read in a Freesurfer annotation from a .annot file.
+    """Read in a Freesurfer annotation from a `.annot` file.
+
+    An `.annot` file contains a sequence of vertices with a label (also known
+    as an "annotation value") associated with each vertex, and then a sequence
+    of colours corresponding to each label.
+
+    The colour table itself may be stored in either an "old-style" format, or
+    a "new-style" format - the :func:`_read_annot_ctab_old_format` and
+    :func:`_read_annot_ctab_new_format` functions are respectively used to
+    read in the colour table.
+
+    See:
+    https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation
+    https://github.com/freesurfer/freesurfer/blob/dev/matlab/read_annotation.m
 
     Parameters
     ----------
@@ -336,11 +359,10 @@ def read_annot(filepath, orig_ids=False):
         RGBA + label id colortable array.
     names : list of str
         The names of the labels. The length of the list is n_labels.
+
     """
     with open(filepath, "rb") as fobj:
-        # all data (apart from strings) in an .annot file is stored as
-        # big-endian int32
-        dt = ">i4"
+        dt = _ANNOT_DT
 
         # number of vertices
         vnum = np.fromfile(fobj, dt, 1)[0]
@@ -355,50 +377,16 @@ def read_annot(filepath, orig_ids=False):
             raise Exception('Color table not found in annotation file')
 
         # in old-format files, the next field will contain the number of
-        # entries in the colour table. In new-format files, this will be
+        # entries in the colour table. In new-format files, this must be
         # equal to -2
         n_entries = np.fromfile(fobj, dt, 1)[0]
 
         # We've got an old-format .annot file.
         if n_entries > 0:
-
-            # orig_tab string length + string
-            length = np.fromfile(fobj, dt, 1)[0]
-            orig_tab = np.fromfile(fobj, '>c', length)
-            orig_tab = orig_tab[:-1]
-            names = list()
-            ctab = np.zeros((n_entries, 5), dt)
-            for i in xrange(n_entries):
-                # structure name length + string
-                name_length = np.fromfile(fobj, dt, 1)[0]
-                name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
-                names.append(name)
-                # read RGBA for this entry
-                ctab[i, :4] = np.fromfile(fobj, dt, 4)
+            ctab, names = _read_annot_ctab_old_format(fobj, n_entries)
         # We've got a new-format .annot file
         else:
-            # file version number
-            ctab_version = -n_entries
-            if ctab_version != 2:
-                raise Exception('Color table version not supported')
-            # maximum LUT index present in the file
-            max_index = np.fromfile(fobj, dt, 1)[0]
-            ctab = np.zeros((max_index, 5), dt)
-            # orig_tab string length + string
-            length = np.fromfile(fobj, dt, 1)[0]
-            np.fromfile(fobj, "|S%d" % length, 1)[0]  # Orig table path
-            # number of LUT entries present in the file
-            entries_to_read = np.fromfile(fobj, dt, 1)[0]
-            names = list()
-            for _ in xrange(entries_to_read):
-                # index of this entry
-                idx = np.fromfile(fobj, dt, 1)[0]
-                # structure name length + string
-                name_length = np.fromfile(fobj, dt, 1)[0]
-                name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
-                names.append(name)
-                # RGBA
-                ctab[idx, :4] = np.fromfile(fobj, dt, 4)
+            ctab, names = _read_annot_ctab_new_format(fobj, -n_entries)
 
     # generate annotation values for each LUT entry
     ctab[:, [4]] = _pack_rgba(ctab[:, :4])
@@ -414,11 +402,101 @@ def read_annot(filepath, orig_ids=False):
     return labels, ctab, names
 
 
+def _read_annot_ctab_old_format(fobj, n_entries):
+    """Read in an old-style Freesurfer colour table from `fobj`.
+
+    This function is used by :func:`read_annot`.
+
+    Parameters
+    ----------
+
+    fobj : file-like
+        Open file handle to a Freesurfer `.annot` file, with seek point
+        at the beginning of the colour table data.
+    n_entries : int
+        Number of entries in the colour table.
+
+    Returns
+    -------
+
+    ctab : ndarray, shape (n_entries, 5)
+        RGBA colortable array - the last column contains all zeros.
+    names : list of str
+        The names of the labels. The length of the list is n_entries.
+    """
+    dt = _ANNOT_DT
+
+    # orig_tab string length + string
+    length = np.fromfile(fobj, dt, 1)[0]
+    orig_tab = np.fromfile(fobj, '>c', length)
+    orig_tab = orig_tab[:-1]
+    names = list()
+    ctab = np.zeros((n_entries, 5), dt)
+    for i in xrange(n_entries):
+        # structure name length + string
+        name_length = np.fromfile(fobj, dt, 1)[0]
+        name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
+        names.append(name)
+        # read RGBA for this entry
+        ctab[i, :4] = np.fromfile(fobj, dt, 4)
+
+    return ctab, names
+
+
+def _read_annot_ctab_new_format(fobj, ctab_version):
+    """Read in a new-style Freesurfer colour table from `fobj`.
+
+    This function is used by :func:`read_annot`.
+
+    Parameters
+    ----------
+
+    fobj : file-like
+        Open file handle to a Freesurfer `.annot` file, with seek point
+        at the beginning of the colour table data.
+    ctab_version : int
+        Colour table format version - must be equal to 2
+
+    Returns
+    -------
+
+    ctab : ndarray, shape (n_labels, 5)
+        RGBA colortable array - the last column contains all zeros.
+    names : list of str
+        The names of the labels. The length of the list is n_labels.
+    """
+    dt = _ANNOT_DT
+    # This code works with a file version == 2, nothing else
+    if ctab_version != 2:
+        raise Exception('Unrecognised .annot file version (%i)', ctab_version)
+    # maximum LUT index present in the file
+    max_index = np.fromfile(fobj, dt, 1)[0]
+    ctab = np.zeros((max_index, 5), dt)
+    # orig_tab string length + string
+    length = np.fromfile(fobj, dt, 1)[0]
+    np.fromfile(fobj, "|S%d" % length, 1)[0]  # Orig table path
+    # number of LUT entries present in the file
+    entries_to_read = np.fromfile(fobj, dt, 1)[0]
+    names = list()
+    for _ in xrange(entries_to_read):
+        # index of this entry
+        idx = np.fromfile(fobj, dt, 1)[0]
+        # structure name length + string
+        name_length = np.fromfile(fobj, dt, 1)[0]
+        name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
+        names.append(name)
+        # RGBA
+        ctab[idx, :4] = np.fromfile(fobj, dt, 4)
+
+    return ctab, names
+
+
 def write_annot(filepath, labels, ctab, names, fill_ctab=True):
     """Write out a Freesurfer annotation file.
 
     See:
     https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation
+    https://github.com/freesurfer/freesurfer/blob/dev/matlab/write_annotation.m
 
     Parameters
     ----------
@@ -430,14 +508,14 @@ def write_annot(filepath, labels, ctab, names, fill_ctab=True):
         RGBA + label id colortable array.
     names : list of str
         The names of the labels. The length of the list is n_labels.
-    fill_ctab : bool
+    fill_ctab : {True, False} optional
         If True, the annotation values for each vertex  are automatically
         generated. In this case, the provided `ctab` may have shape
         (n_labels, 4) or (n_labels, 5) - if the latter, the final column is
         ignored.
     """
     with open(filepath, "wb") as fobj:
-        dt = ">i4"
+        dt = _ANNOT_DT
         vnum = len(labels)
 
         def write(num, dtype=dt):
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -89,9 +89,8 @@ def _pack_rgba(rgba):
     out : ndarray, shape (n, )
         Annotation values for each colour.
     """
-    # ctab :: n x 4
     bitshifts = 2 ** np.array([[0], [8], [16], [24]], dtype=rgba.dtype)
-    return rgba.dot(bitshifts).squeeze()
+    return rgba.dot(bitshifts)
 
 
 def read_geometry(filepath, read_metadata=False, read_stamp=False):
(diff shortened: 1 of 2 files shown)


nibabel/freesurfer/io.py
```diff
@@ -318,51 +318,81 @@ def read_annot(filepath, orig_ids=False):
...
                 name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
                 names.append(name)
+                # RGBA
                 ctab[i, :4] = np.fromfile(fobj, dt, 4)
-                ctab[i, 4] = (ctab[i, 0] + ctab[i, 1] * (2 ** 8) +
+                # generate the annotation value
+                ctab[i, 4] = (ctab[i, 0] +
+                              ctab[i, 1] * (2 ** 8) +
                               ctab[i, 2] * (2 ** 16) +
                               ctab[i, 3] * (2 ** 24))
```
effigies: You can drop this calculation here. (See below.)

nibabel/freesurfer/io.py
```diff
@@ -318,51 +318,81 @@ def read_annot(filepath, orig_ids=False):
...
-                ctab[i, :4] = np.fromfile(fobj, dt, 4)
-                ctab[i, 4] = (ctab[i, 0] + ctab[i, 1] * (2 ** 8) +
-                              ctab[i, 2] * (2 ** 16))
-        ctab[:, 3] = 255
+                # RGBA
+                ctab[idx, :4] = np.fromfile(fobj, dt, 4)
+                ctab[idx, 4] = (ctab[idx, 0] +
+                                ctab[idx, 1] * (2 ** 8) +
+                                ctab[idx, 2] * (2 ** 16) +
+                                ctab[idx, 3] * (2 ** 24))
```
effigies: Given this is now showing up a couple times, maybe it should be factored into something like:

```Python
# Above
def pack_rgba(ctab):
    # ctab :: n x 4
    bitshifts = 2 ** np.array([[0], [8], [16], [24]], dtype=ctab.dtype)
    return ctab.dot(bitshifts)
```

(If you find the manual expansion more readable than the vector math, go ahead and use that.)

Now, we can drop back two indentations, out of the for loop and the `if`/`else` and operate on slices:

```Python
# Construct annotation values
ctab[:, 4] = pack_rgba(ctab[:, :4])
```

(We can drop out of the `if`/`else`, since this appears at the end of both branches.)

nibabel/freesurfer/io.py
```diff
@@ -400,6 +435,16 @@ def write_string(s):
...
 
+        # Generate annotation values for each ctab entry
+        if fill_ctab:
+            new_ctab = np.zeros((ctab.shape[0], 5), dtype=np.int32)
+            new_ctab[:, :4] = ctab[:, :4]
+            ctab = new_ctab
+            ctab[:, 4] = (ctab[:, 0] +
+                          ctab[:, 1] * (2 ** 8) +
+                          ctab[:, 2] * (2 ** 16) +
+                          ctab[:, 3] * (2 ** 24))
```
effigies: With the refactor described above, this can just be:

```Python
if fill_ctab:
    ctab = np.hstack((ctab[:, :4], pack_rgba(ctab[:, :4])))
```

Would it also make sense to add the following?

```Python
elif not np.array_equal(ctab[:, 4], pack_rgba(ctab[:, :4])):
    # Warn of invalid array, suggest using fill_ctab
```

It seems like an error case that is worth warning about.

nibabel/freesurfer/io.py
```diff
@@ -318,51 +318,81 @@ def read_annot(filepath, orig_ids=False):
...
         if n_entries > 0:
+
+            # orig_tab string length + string
             length = np.fromfile(fobj, dt, 1)[0]
             orig_tab = np.fromfile(fobj, '>c', length)
             orig_tab = orig_tab[:-1]
-
             names = list()
-            ctab = np.zeros((n_entries, 5), np.int)
+            ctab = np.zeros((n_entries, 5), np.int32)
```
effigies: Any reason not to just use `dt` instead of `np.int32`?

pauldmccarthy: No reason, will do

nibabel/freesurfer/io.py
```diff
@@ -318,51 +318,81 @@ def read_annot(filepath, orig_ids=False):
...
+            # orig_tab string length + string
             length = np.fromfile(fobj, dt, 1)[0]
             np.fromfile(fobj, "|S%d" % length, 1)[0]  # Orig table path
+            # number of entries to read from the file
             entries_to_read = np.fromfile(fobj, dt, 1)[0]
             names = list()
             for i in xrange(entries_to_read):
-                np.fromfile(fobj, dt, 1)[0]  # Structure
+                # index of this entry
+                idx = np.fromfile(fobj, dt, 1)[0]
```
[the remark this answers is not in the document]
pauldmccarthy: From what I can gather, In practice it would probably be ok to use the loop variable. But in theory the 'new-style' format allows the LUT entries to be arbitrarily ordered, and for not all to be present (i.e. indices may be skipped). Hence there are two counts in the file - the maximum LUT index that is present in the file, and the total number of entries.

The officlal matlab code is probably the best reference here:

https://github.com/freesurfer/freesurfer/blob/dev/matlab/read_annotation.m#L149 

And the notes on the freesurfer wiki:

https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation

nibabel/freesurfer/io.py
pauldmccarthy: No problemo. I suppose I was trying to minimise code changes, despite the obvious potential for refactoring.

nibabel/freesurfer/io.py
pauldmccarthy: Good point, will add the warning in.

nibabel/freesurfer/io.py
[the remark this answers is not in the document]
effigies: Sounds good. Maybe just make a quick note of this in the code. (You may have done this, and I failed to read it.)

nibabel/freesurfer/io.py
```diff
@@ -73,6 +73,27 @@ def _read_volume_info(fobj):
...
+
+    Returns
+    -------
+
+    out : ndarray, shape (n, )
+        Annotation values for each colour.
+    """
+    # ctab :: n x 4
+    bitshifts = 2 ** np.array([[0], [8], [16], [24]], dtype=rgba.dtype)
+    return rgba.dot(bitshifts).squeeze()
```
effigies: I think if you flatten this in the one case, the two cases where you have to reshape will come out a lot cleaner.

effigies: Sorry, I think that was badly worded, and badly counted, as well. If you don't flatten here, then you can cover the two cases like so:

```Python
# Setting vector - using [4] instead of [4] maintains the shape
ctab[:, [4]] = _pack_rgba(ctab[:, :4])

# Stacking matrix and column vector
ctab = np.hstack((ctab[:, :4], _pack_rgba(ctab[:, :4])))

Comment (pauldmccarthy) [reply nibabel/freesurfer/io.py]: Cool, makes the code a little cleaner...

File: nibabel/freesurfer/io.py [hunk 7]
```diff
@@ -400,6 +447,13 @@ def write_string(s):
             write(len(s))
             write(s, dtype='|S%d' % len(s))
 
+        # Generate annotation values for each ctab entry
+        if fill_ctab:
+            ctab = np.hstack((ctab[:, :4], _pack_rgba(ctab[:, :4])))
+        elif not np.array_equal(ctab[:, 4], _pack_rgba(ctab[:, :4])):
```
effigies: The shape is off here, so it's always going to evaluate `True`. Your left entry needs the square brackets around the 4. Maybe add a test with a valid and invalid annotation values, to make sure the warning is raised only in the latter case.

pauldmccarthy: Yep, the warning was always being raised. Fixed, and covered in the tests.

nibabel/freesurfer/tests/test_io.py
```diff
@@ -209,6 +209,101 @@ def test_annot():
         assert_equal(names, names2)
 
 
+def test_read_write_annot():
+
```
matthew-brett: I personally prefer to minimize blank lines, and avoid them at the top of functions.

nibabel/freesurfer/io.py
```diff
@@ -73,6 +73,26 @@ def _read_volume_info(fobj):
     return volume_info
 
 
+def _pack_rgba(rgba):
+    """Used by :meth:`read_annot` and :meth:`write_annot` to pack an RGBA
```
matthew-brett: I know this is a private function, but the first line should be standalone with further explanation after a blank line.  How about

```
""" Pack an RGBA sequence into a single integer

Used by :meth:`read_annot` and :meth:`write_annot`.
```

nibabel/freesurfer/io.py
```diff
@@ -318,51 +338,73 @@ def read_annot(filepath, orig_ids=False):
         The names of the labels. The length of the list is n_labels.
     """
     with open(filepath, "rb") as fobj:
+        # all data (apart from strings) in an .annot file is stored as
+        # big-endian int32
         dt = ">i4"
```
matthew-brett: Module level constant maybe?

nibabel/freesurfer/io.py
```diff
@@ -318,51 +338,73 @@ def read_annot(filepath, orig_ids=False):
...
                 ctab[i, :4] = np.fromfile(fobj, dt, 4)
-                ctab[i, 4] = (ctab[i, 0] + ctab[i, 1] * (2 ** 8) +
-                              ctab[i, 2] * (2 ** 16) +
-                              ctab[i, 3] * (2 ** 24))
+        # We've got a new-format .annot file
         else:
+            # file version number
             ctab_version = -n_entries
             if ctab_version != 2:
                 raise Exception('Color table version not supported')
```
matthew-brett: Sorry  - I know this is not your code, but maybe a more informative exception here.

nibabel/freesurfer/io.py
```diff
@@ -318,51 +338,73 @@ def read_annot(filepath, orig_ids=False):
...
+                # structure name length + string
                 name_length = np.fromfile(fobj, dt, 1)[0]
                 name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
                 names.append(name)
+                # read RGBA for this entry
                 ctab[i, :4] = np.fromfile(fobj, dt, 4)
-                ctab[i, 4] = (ctab[i, 0] + ctab[i, 1] * (2 ** 8) +
-                              ctab[i, 2] * (2 ** 16) +
-                              ctab[i, 3] * (2 ** 24))
+        # We've got a new-format .annot file
```
matthew-brett: This function seems a bit long - any way of splitting it into subfunctions?

nibabel/freesurfer/io.py
```diff
@@ -318,51 +338,73 @@ def read_annot(filepath, orig_ids=False):
...
-        ctab[:, 3] = 255
+                # RGBA
+                ctab[idx, :4] = np.fromfile(fobj, dt, 4)
 
-    labels = labels.astype(np.int)
+    # generate annotation values for each LUT entry
+    ctab[:, [4]] = _pack_rgba(ctab[:, :4])
+
+    # make sure names are strings, not bytes
+    names = [n.decode('ascii') for n in names]
```
matthew-brett: Names are always ascii?  Or can be latin1?

nibabel/freesurfer/io.py
```diff
@@ -388,6 +430,11 @@ def write_annot(filepath, labels, ctab, names):
         RGBA + label id colortable array.
     names : list of str
         The names of the labels. The length of the list is n_labels.
+    fill_ctab : bool
```
matthew-brett: Add `optional, as in 

```
fill_ctab : {True, False}, optional
```

The {True False} instead of bool also specifies the default.

nibabel/freesurfer/io.py
pauldmccarthy: How about `_read_annot_old_format` and `_read_annot_new_format` functions, which just contain the if/else blocks?

nibabel/freesurfer/io.py
pauldmccarthy: Not entirely sure. Latin1 is a superset of ascii, right? So it should be no harm to change to latin1.

 - https://github.com/freesurfer/freesurfer/blob/dev/matlab/read_annotation.m#L167
 - https://github.com/freesurfer/freesurfer/blob/dev/matlab/write_annotation.m#L118

effigies:
There was an issue with h5py nightly build, presumably due to Travis backlog. Resolved itself over the weekend. This is now testing clean.

I'm :+1: to merge.

@matthew-brett any further comments?

nibabel/freesurfer/io.py
effigies: Oh, forgot about this issue. Makes sense to me to be very liberal about reading and accept any byte string at all.

First, it looks like we've returning `str`/`bytes` in Python 2/3, right? So decoding to `unicode`/`str` would be an API change. Given that, I think it makes sense to continue returning `bytes` by default. If you want to add in a `name_encoding` (or similar) parameter, you could do:

```Python
if name_encoding is not None:
    names = [n.decode(name_encoding) for n in names]
```

Also, looking at those MATLAB files, there's some NUL-termination we're not handling correctly on the writing end. (Numpy correctly truncates null characters.) I'll put a comment in `write_annot` where I think we need to fix that.

nibabel/freesurfer/io.py
```diff
@@ -400,6 +525,13 @@ def write_string(s):
             write(len(s))
             write(s, dtype='|S%d' % len(s))
```
effigies: I think this needs to be:

```Python
s = s.encode() + b'\x00'
write(len(s))
write(s, dtype='|S%d' % len(s))
```

This will correctly handle `str` inputs with `NUL`-termination.

nibabel/freesurfer/io.py
pauldmccarthy: Leave the return value as str/bytes for 2/3 - fair enough. I made this change to bring the code in line with the function docstring, which promises a `list of str`(I'm guessing it was written against python 2 only, so was not handling python 3 at all).

effigies: Yes, that's almost certainly true. Probably good to update the docstring to say `bytes`, then.

nibabel/freesurfer/io.py
pauldmccarthy: Yep, you're right. The `read_annot` code works in both cases though - it seems that `np.fromfile(fobj, '|S%d' % length)` will work regardless of whether the last byte is `\00` or not, as long as the total length is correct.

nibabel/freesurfer/io.py
```diff
@@ -388,18 +504,31 @@ def write_annot(filepath, labels, ctab, names):
...
     with open(filepath, "wb") as fobj:
-        dt = ">i4"
+        dt = _ANNOT_DT
         vnum = len(labels)
 
         def write(num, dtype=dt):
             np.array([num]).astype(dtype).tofile(fobj)
 
         def write_string(s):
+            s = s.encode() + b'\00'
```
effigies: Maybe

```Python
s = (s if isinstance(s, bytes) else s.encode()) + b'\x00'
```

Also, note the `x` in `'\x00'`. `b'\00' == b'\x000'`.

pauldmccarthy: Oops .. I made sure to test on py2 before pushing, but at the cost of forgetting to test on py3!

nibabel/freesurfer/io.py
```diff
@@ -296,7 +326,20 @@ def write_morph_data(file_like, values, fnum=0):
 
 
 def read_annot(filepath, orig_ids=False):
-    """Read in a Freesurfer annotation from a .annot file.
+    """Read in a Freesurfer annotation from a `.annot` file.
```
effigies: Double back-ticks for monospace formatting. I think single is italics.

nibabel/freesurfer/io.py
```diff
@@ -73,6 +81,28 @@ def _read_volume_info(fobj):
...
 
+def _pack_rgba(rgba):
+    """Pack an RGBA sequence into a single integer.
+
+    Used by :func:`read_annot` and :func:`write_annot` to generate
+    "annotation values" for a Freesuerfer `.annot` file.
+
+    Parameters
+    ----------
+
```
effigies: Unnecessary blank line.

nibabel/freesurfer/io.py
```diff
@@ -73,6 +81,28 @@ def _read_volume_info(fobj):
...
+    rgba : ndarray, shape (n, 4)
+        RGBA colours
+
+    Returns
+    -------
+
```
effigies: Unnecessary blank line.

nibabel/freesurfer/io.py
```diff
@@ -296,7 +326,20 @@ def write_morph_data(file_like, values, fnum=0):
 
 
 def read_annot(filepath, orig_ids=False):
-    """Read in a Freesurfer annotation from a .annot file.
+    """Read in a Freesurfer annotation from a `.annot` file.
+
+    An `.annot` file contains a sequence of vertices with a label (also known
```
effigies: Double back-ticks.

nibabel/freesurfer/io.py
```diff
@@ -296,7 +326,20 @@ def write_morph_data(file_like, values, fnum=0):
...
+    as an "annotation value") associated with each vertex, and then a sequence
+    of colours corresponding to each label.
+
+    The colour table itself may be stored in either an "old-style" format, or
```
effigies: AFAICT this PR will be the first introduction of "colour" to the docs as opposed to color. For the sake of consistency, do you mind switching to the American spelling?

nibabel/freesurfer/io.py
```diff
@@ -296,7 +326,20 @@ def write_morph_data(file_like, values, fnum=0):
...
+    a "new-style" format - the :func:`_read_annot_ctab_old_format` and
```
effigies: Prefixing variables and functions with `_` indicates they're private, and I don't think the docstrings you're linking to will show up in the docs. (Also, referring to them in the docs might negate some of the point of making them private in the first place.)

Maybe, this paragraph could just be:

```
Annotation file format versions 1 and 2 are supported, corresponding to the
"old-style" and "new-style" color table layout.
```

This way, if a version 3 ever happens, the docs are less likely to become incorrect.

nibabel/freesurfer/io.py
```diff
@@ -372,11 +398,101 @@ def read_annot(filepath, orig_ids=False):
...
+        name = np.fromfile(fobj, "|S%d" % name_length, 1)[0]
+        names.append(name)
+        # RGBA
+        ctab[idx, :4] = np.fromfile(fobj, dt, 4)
+
+    return ctab, names
+
+
+def write_annot(filepath, labels, ctab, names, fill_ctab=True):
     """Write out a Freesurfer annotation file.
```
effigies: Maybe good to note that we exclusively write version 2 ("new-style") .annot files.

nibabel/freesurfer/io.py
```diff
@@ -85,17 +85,15 @@ def _pack_rgba(rgba):
     """Pack an RGBA sequence into a single integer.
 
     Used by :func:`read_annot` and :func:`write_annot` to generate
-    "annotation values" for a Freesuerfer `.annot` file.
+    "annotation values" for a Freesuerfer ``.annot`` file.
```
matthew-brett: Typo freesurfer

nibabel/freesurfer/io.py
```diff
@@ -372,11 +394,101 @@ def read_annot(filepath, orig_ids=False):
     return labels, ctab, names
 
 
-def write_annot(filepath, labels, ctab, names):
-    """Write out a Freesurfer annotation file.
+def _read_annot_ctab_old_format(fobj, n_entries):
+    """Read in an old-style Freesurfer colour table from `fobj`.
```
matthew-brett: Are we allowed to spell color like that? :)

nibabel/freesurfer/io.py
```diff
@@ -372,11 +394,101 @@ def read_annot(filepath, orig_ids=False):
...
+
+    This function is used by :func:`read_annot`.
+
+    Parameters
+    ----------
+
+    fobj : file-like
```
matthew-brett: File-like usually means it can be a filename.  I think that a filename will in fact work here, but will give the wrong result, because the position will always be 0 when doing the read.  Maybe worth a check and comment?

nibabel/freesurfer/io.py
```diff
@@ -372,11 +394,101 @@ def read_annot(filepath, orig_ids=False):
...
+
+def _read_annot_ctab_new_format(fobj, ctab_version):
+    """Read in a new-style Freesurfer colour table from `fobj`.
+
+    This function is used by :func:`read_annot`.
+
+    Parameters
+    ----------
+
+    fobj : file-like
```
matthew-brett: Comment as for reading the old version.

nibabel/freesurfer/io.py
```diff
@@ -372,11 +394,101 @@ def read_annot(filepath, orig_ids=False):
...
+
+    return ctab, names
+
+
+def write_annot(filepath, labels, ctab, names, fill_ctab=True):
+    """Write out a "new-style" Freesurfer annotation file.
 
     See:
     https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation
+    https://github.com/freesurfer/freesurfer/blob/dev/matlab/write_annotation.m
```
matthew-brett: Does this format correctly?  You might need:

```
     See:
     \* https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation
     \* https://github.com/freesurfer/freesurfer/blob/dev/matlab/write_annotation.m
```

nibabel/freesurfer/io.py
pauldmccarthy: You're right - a file name would work. I've changed the docs to specify `file`, rather than `file-like`. 

But I'm not quite sure how to check this, or whether it is necessary (given that these are private functions). I could check the seek position, but that would need both private functions to be passed the number of vertices that were read in beforehand, which I feel is getting a bit over the top. 

What do you think?

matthew-brett: Does it work though?   It looks to me as if you would pass the filename to each call, and the reading would start at position 0 each time?

It's fine to just say file-like object but not filename in the docstring, and test for the `read` attribute with an assert at the top - if that's the only interface you're using - I think it is?

pauldmccarthy: Well, when I said "a file name would work", I mean it would be accepted and probably wouldn't crash. But the data read in would be wrong because, as you say, the seek location will be 0. 

I'm not not directly using any methods on the file object - all of the reading is performed by `numpy.fromfile`. But it can't hurt to check that the file-like has a `read` method - will add it in.

matthew-brett:
All good with me now - @effigies - any further thoughts?

effigies:
I'm happy to see this go in.

PR #593, opened 2018-02-05
Stop adding extraneous metadata padding

jstutters:
If the metadata field is already aligned to a 16-byte boundary then
don't add additional padding.  This allows the user to create a
16-byte aligned extension padded with 0x20 that can be read by the [Jim
software](http://www.xinapse.com/) (which doesn't like finding 0x00 in the metadata chunk).

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -334,7 +334,8 @@ class Nifti1Extension(object):
         size = len(self._mangle(self._content))
         size += 8
         # extensions size has to be a multiple of 16 bytes
-        size += 16 - (size % 16)
+        if size % 16 != 0:
+            size += 16 - (size % 16)
         return size
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


effigies:
Good catch. Any chance of writing a test for this case?

jstutters:
@effigies I've extended test_nifti1/test_extension_basics to cover this situation.  Unfortunately coveralls is still showing that as a drop in coverage but both code paths are executed (unless I've missed something).

effigies:
The coverage checks can be a little weird at times. The main one I pay attention to is the [diff](https://codecov.io/gh/nipy/nibabel/compare/ee4ce2a646cd49ea382d24f73e58b429d4795a49...dbfb71d493b55b8c8430aa30fd5939bc11cca39f/diff).

matthew-brett:
Looks good to me too - thanks for the spot ...

PR #597, opened 2018-02-16
[Fix] buffer_size on ArraySequence

skoudoro:
Hi, 

When you initialize `ArraySequence` with a big generator on Windows,  this operation is really really slow. (6 hours for 560583 streamlines...).  Unfortunately, setting up the `buffer_size` does not have any effect. 

So the goal of this PR is to fix this issue. Now, this operation takes me 46sec.

@MarcCote, @Garyfallidis, Can you look at this PR? Thanks!

### DIFF
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -37,7 +37,7 @@ class _BuildCache(object):
         self.common_shape = common_shape
         n_in_row = reduce(mul, common_shape, 1)
         bytes_per_row = n_in_row * dtype.itemsize
-        self.rows_per_buf = self.bytes_per_buf / bytes_per_row
+        self.rows_per_buf = int(np.ceil(self.bytes_per_buf / bytes_per_row))
 
     def update_seq(self, arr_seq):
         arr_seq._offsets = np.array(self.offsets)
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -37,7 +37,7 @@ class _BuildCache(object):
         self.common_shape = common_shape
         n_in_row = reduce(mul, common_shape, 1)
         bytes_per_row = n_in_row * dtype.itemsize
-        self.rows_per_buf = bytes_per_row / self.bytes_per_buf
+        self.rows_per_buf = self.bytes_per_buf / bytes_per_row
 
     def update_seq(self, arr_seq):
         arr_seq._offsets = np.array(self.offsets)
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -37,7 +37,7 @@ class _BuildCache(object):
         self.common_shape = common_shape
         n_in_row = reduce(mul, common_shape, 1)
         bytes_per_row = n_in_row * dtype.itemsize
-        self.rows_per_buf = int(np.ceil(self.bytes_per_buf / bytes_per_row))
+        self.rows_per_buf = max(1, self.bytes_per_buf // bytes_per_row)
 
     def update_seq(self, arr_seq):
         arr_seq._offsets = np.array(self.offsets)
@@ -185,6 +185,7 @@ class ArraySequence(object):
             return
         self._build_cache.update_seq(self)
         self._build_cache = None
+        self.shrink_data()
 
     def _resize_data_to(self, n_rows, build_cache):
         """ Resize data array if required """
--- a/nibabel/streamlines/tests/test_array_sequence.py
+++ b/nibabel/streamlines/tests/test_array_sequence.py
@@ -329,4 +329,3 @@ def test_concatenate():
     new_seq = concatenate(seqs, axis=0)
     assert_true(len(new_seq), seq.common_shape[0] * len(seq))
     assert_array_equal(new_seq._data, seq._data.T.reshape((-1, 1)))
-    
--- a/nibabel/streamlines/tests/test_array_sequence.py
+++ b/nibabel/streamlines/tests/test_array_sequence.py
@@ -97,8 +97,8 @@ class TestArraySequence(unittest.TestCase):
         seq_with_buffer = ArraySequence(gen_2, buffer_size=256)
 
         # Check buffer size effect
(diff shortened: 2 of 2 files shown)


nibabel/streamlines/tests/test_array_sequence.py
```diff
@@ -91,11 +92,20 @@ def test_creating_arraysequence_from_list(self):
...
 
     def test_creating_arraysequence_from_generator(self):
-        gen = (e for e in SEQ_DATA['data'])
-        check_arr_seq(ArraySequence(gen), SEQ_DATA['data'])
+        gen_1, gen_2 = itertools.tee((e for e in SEQ_DATA['data']))
+        seq = ArraySequence(gen_1)
+        seq_with_buffer = ArraySequence(gen_2, buffer_size=256)
+
+        # Check buffer size effect
+        assert_true(seq_with_buffer.data.shape[0] > seq.data.shape[0])
```
MarcCote: Interesting, should we make sure that once creating is done we have the same `._data.shape[0]`?

MarcCote:
@matthew-brett can you confirm there is a typo in the logic when we compute the size of the buffer.
In master: 
`self.rows_per_buf = bytes_per_row / self.bytes_per_buf`
with this PR
`self.rows_per_buf = self.bytes_per_buf / bytes_per_row`

nibabel/streamlines/tests/test_array_sequence.py
skoudoro: Sorry, I do not understand what you mean here

matthew-brett:
Oops, yes, that does look like a typo.

Garyfallidis:
Looks good to me but I see that Travis is failing @MarcCote and @matthew-brett ?
Also, @matthew-brett this current bug holds the dipy release too.  We were hoping to completely move to the new streamlines API. When is the next nibabel release planned? Apologies in advance for the pressure.

nibabel/streamlines/tests/test_array_sequence.py
MarcCote: I just meant depending on the buffer_size you chose you might find yourself with a lot of "reserved" allocated space. Let say you have only one streamline in `streamlines` and do `arr_seq = ArraySequence(streamlines, buffer_size=1024)` then your `arr_seq` will be using 1024. 

I thought we might want to trim `arr_seq._data` at the end of the `extend` in order to reduce memory consumption. On the other hand, using the default `buffer_size=4` would only consume up to 4Mb which is rather small.

Garyfallidis:
Ah I see. The errors are from other nibabel functions not from this PR.

effigies:
At least the last one seems to be `test_array_sequence.test_concatenate`, which seems relevant.

I've opened an issue for the DICOM-related failures.

skoudoro:
Ok, I will look at this one @effigies. It seems to be a `buffer_size` problem.

nibabel/streamlines/tests/test_array_sequence.py
skoudoro: ok, I see, I was surprised too :-).  For sure it will be good to trim the data. We should create a new PR for that, I suppose.

nibabel/streamlines/array_sequence.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, arr_seq, common_shape, dtype):
         self.common_shape = common_shape
         n_in_row = reduce(mul, common_shape, 1)
         bytes_per_row = n_in_row * dtype.itemsize
-        self.rows_per_buf = bytes_per_row / self.bytes_per_buf
+        self.rows_per_buf = self.bytes_per_buf / bytes_per_row
```
effigies: Should `rows_per_buf` be an `int` or a `float`? If the idea is to load whole rows at a time, then you may want to use `//` to ensure that you round down.

nibabel/streamlines/tests/test_array_sequence.py
```diff
@@ -318,4 +328,4 @@ def test_concatenate():
     seqs = [seq[:, [i]] for i in range(seq.common_shape[0])]
     new_seq = concatenate(seqs, axis=0)
     assert_true(len(new_seq), seq.common_shape[0] * len(seq))
-    assert_array_equal(new_seq._data, seq._data.T.reshape((-1, 1)))
+    assert_array_equal(new_seq._data, seq._data.T.reshape((-1, 0)))
```
effigies: That's an illegal shape. What shape are you shooting for, here?

nibabel/streamlines/array_sequence.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, arr_seq, common_shape, dtype):
         self.common_shape = common_shape
         n_in_row = reduce(mul, common_shape, 1)
         bytes_per_row = n_in_row * dtype.itemsize
-        self.rows_per_buf = bytes_per_row / self.bytes_per_buf
+        self.rows_per_buf = int(np.ceil(self.bytes_per_buf / bytes_per_row))
```
matthew-brett: Do you want `ceil` or `floor` here?

skoudoro: `ceil` because I want the minimal value to be 1 for `rows_per_buf`

matthew-brett: But I guess you will have too many rows, oftentimes.  How about:

```
self.rows_per_buf = max(1, int(self.bytes_per_buf / bytes_per_row))
```

skoudoro: Thanks, I think your way is more explicit, I will do that

Garyfallidis:
Okay, I believe I was wrong DIPY's release does not need to be delayed by this fix. We can simply check for Nibabel's version and if not the one with the fix, inherit ArraySequence in our Streamlines class and overload the specific function that has the issue. !!! :)

matthew-brett:
@MarcCote - are you OK with this one now?   If so, we can merge.

MarcCote:
@matthew-brett @effigies Yes, I'm OK with all the changes. Thanks @skoudoro.

skoudoro:
Thanks all for your review

PR #599, opened 2018-02-18
MRG: refactoring for pydicom 1.0

matthew-brett:
Make tests pass against pydicom 1.0.

Refactor CI to test pydicom master, as suggested by pydicom folks.

Update minimum pydicom dependency to 0.9.9 (previous versions no longer install).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,8 @@ cache:
     - $HOME/.cache/pip
 env:
     global:
-        - DEPENDS="six numpy scipy matplotlib h5py pillow"
+        - DEPENDS="six numpy scipy matplotlib h5py pillow pydicom"
         - OPTIONAL_DEPENDS=""
-        - PYDICOM=1
         - INSTALL_TYPE="setup"
         - EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -34,29 +33,29 @@ matrix:
     # Absolute minimum dependencies
     - python: 2.7
       env:
-        - DEPENDS="numpy==1.7.1" PYDICOM=0
+        - DEPENDS="numpy==1.7.1"
```
effigies: Wasn't the idea here to make sure that we didn't break when pydicom was absent? Do we want to move to a pydicom dependency?

effigies: Wow, brain failure. I see this replaces the pydicom dependency above. Carry on.

effigies:
This looks reasonable. The AppVeyor build seems to have failed for its own reasons (I can't restart it; not sure if that's a thing in AppVeyor).

.travis.yml
matthew-brett: With this matrix entry, Pydicom isn't in the DEPENDS variable, and doesn't get installed.  On the way, the tests did in fact fail when Pydicom wasn't present, I had to fix the code to make it work again.  Have a look at the Travis-CI log to confirm this entry does not install Pydicom...

matthew-brett:
I reverted a couple of the modifications, because pydicom has fixed them already...

matthew-brett:
Thanks Chris for the review.

PR #601, opened 2018-02-19
RF: moving nib-ls to a new cmdline folder

chrispycheng:
moved nib-ls to make generic use of its functions possible for a "nib-diff" function currently in progress. Shoutout to @yarikoptic


matthew-brett:
I hate to say this, but it would be good to have a few little tests in the new module, if only as a template for adding more later.

chrispycheng:
@matthew-brett can I ask what tests you're looking for? nib-ls functionality?

matthew-brett:
Sorry - I meant tests for the not-main functions that are now in the `nibabel/cmdline` tree.

matthew-brett:
Tests failing here - I think you need to add `nibabel.cmdline` to the list of packages in `setup.py`.  You should also add `nibabel.cmdline.tests` (does it have an `__init__.py` already.  There are some extra blank lines at the top of `cmdline.__init__.py`.  Otherwise, good to go (when tests running and passing).

chrispycheng:
More tests to come for ap() and safe_get() functions

matthew-brett:
Great, it will be good to have some more tests.

There are some PEP8 fails - see https://travis-ci.org/nipy/nibabel/jobs/351289012#L1492

chrispycheng:
@matthew-brett Hi - if you have the time, could you please let me know what you think of the tests I've written? Thanks very much :-)

nibabel/cmdline/tests/test_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+class TestObject:
+    def __init__(self, test=None):
+        self.test = test
+
+    def get_test(self):
+        return self.test
+
+test = TestObject()
+test.test = 2
```
yarikoptic: please move it into `test_safe_get` so its purpose is clear and it doesn't get picked up by nose etc

matthew-brett: Yes, good idea.

matthew-brett:
Yes, I agree - we can merge and improve.  @chrispycheng - do you mind making the change Yarik suggested, and I will merge.

nibabel/cmdline/tests/test_utils.py
chrispycheng: done and ready to rumble

PR #607, opened 2018-03-08
BF: preserve error output if not valid klass in `nibabel.save()`

kaczmarj:
This PR proposes a fix for an `UnboundLocalError` in `nibabel.save`, specifically in the lines

https://github.com/nipy/nibabel/blob/e48b746af1841aed2ff862d1eea8471f35265b2c/nibabel/loadsave.py#L131-L139

The scope of variable `e` in `except Exception as e` is the `except block` in python3. Trying to raise it out of the `except` block will raise an error. I haven't been able to test this myself yet, so it's possible something else must be changed. I can add a test for this as well.

I initially came upon this error when trying to save a float16 nifti:

```python
import nibabel as nb
import numpy as np

dataobj = np.ones((10, 10, 10), dtype=np.float16)
affine = np.eye(4)
img = nb.spatialimages.SpatialImage(dataobj, affine)
nb.save(img, 'foo.nii.gz')
```

Here are minimal examples of this behavior:

```python
try:
    foo
except Exception as e:
    pass
raise e  # e is not defined
```

```python
try:
    foo
except Exception as e:
    err = e
raise err  # foo is not defined (expected)
```

Interestingly...
```python
try:
    foo
except Exception as e:
    e = e
raise e  # e is not defined
```

### DIFF
--- a/nibabel/tests/test_loadsave.py
+++ b/nibabel/tests/test_loadsave.py
@@ -10,7 +10,8 @@ import numpy as np
 from .. import (Spm99AnalyzeImage, Spm2AnalyzeImage,
                 Nifti1Pair, Nifti1Image,
                 Nifti2Pair, Nifti2Image)
-from ..loadsave import load, read_img_data
+from ..spatialimages import SpatialImage
+from ..loadsave import load, read_img_data, save
 from ..filebasedimages import ImageFileError
 from ..tmpdirs import InTemporaryDirectory, TemporaryDirectory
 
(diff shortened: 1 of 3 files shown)


matthew-brett:
Good spot.  How about adding your example above as a test?

nibabel/tests/test_loadsave.py
```diff
@@ -135,3 +136,33 @@ def test_read_img_data_nifti():
...
+def test_save():
+    with InTemporaryDirectory():
+        dataobj = np.ones((10, 10, 10), dtype=np.float16)
+        affine = np.eye(4, dtype=np.float32)
+        img = SpatialImage(dataobj, affine)
+
+        # The `save` method can raise one of many types of errors, but they are
+        # all subclasses of Exception. This attempt will fail because float16
+        # is not supported.
+        with assert_raises(Exception):
```
matthew-brett: Don't we know the error in this case though?

nibabel/tests/test_loadsave.py
```diff
@@ -135,3 +136,33 @@ def test_read_img_data_nifti():
...
+            save(img, 'foo.nii.gz')
+
+        # Test the saving of several types of images.
```
matthew-brett: I believe we have tests for these elsewhere - would you mind having a quick look?  Otherwise I will.

nibabel/tests/test_loadsave.py
kaczmarj: You're right. I will make this more specific.

nibabel/tests/test_loadsave.py
kaczmarj: I should have expected there to be tests for this. I think they are in [nibabel.tests.test_image_load_save](https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_image_load_save.py). I moved the test to that file and removed the redundant tests.

matthew-brett:
Looks good to me too - in it goes.

matthew-brett:
@kaczmarj - thanks much for the fix and for your patience.

PR #610, opened 2018-03-15
ENH: Accept pathlib.Path objects where filenames are accepted

CRiddler:
For python >= 3.6, objects representing paths now have the option to have a dunder `__fspath__` (See PEP 519 for more info) to return the string (or bytes) representation of the path. Pathlib on 3.6 includes this method so we should check for it first. Then, for pathlib objects from the py 3.4-3.5x, `__fspath__` is not available, so we dip into the py3k module to get the current python version's (2 or 3) string method (`unicode` in this case). So overall we're allowing filename to be a path object, but ensuring that it will be converted to the string/bytes representation of the path before passing it on to `image_klass.from_filename`

My only reservations is that `nib.load` lowest entry point for this type of handling, because if some one really wants to call `from_filename` from an image_klass directly, they won't have this pathlib compatibility- however I think that `nib.load` is a very common entry point, which justifies the placement.


nibabel/loadsave.py
```diff
@@ -36,6 +36,11 @@ def load(filename, **kwargs):
     img : ``SpatialImage``
        Image of guessed type
     '''
+    if hasattr(filename, '__fspath__'):
+        filename = filename.__fspath__()
+    else:
+        filename = unicode(filename)
```
effigies: I'm not sure that this is the best way to handle this. My reading is that Windows uses `unicode` paths, while Unix uses `bytes` paths. While this may work most of the time, I don't see any value to modifying the type of `filename` where we haven't, before.

As suggested by @fepegar, this may be a better way to approach this:

```Python
if isinstance(filename, getattr(os, "PathLike", ())):
    filename = os.fspath(filename)
```

As this effectively tests for whether `pathlib` exists and then allows it to handle any type conversions needed. I would be inclined to factor this out into a utility function:

```Python
def get_fname(pathlike):
    if isinstance(pathlike, getattr(os, "PathLike", ())):
        pathlike = os.fspath(pathlike)
    return pathlike
```

This could then be called in `load`, `save`, [`FileBasedImage.from_filename()`](https://github.com/nipy/nibabel/blob/19fd3f82aead1f4fe1f45ac13139acc3ad30140f/nibabel/filebasedimages.py#L257-L260) and [`FileBasedImage.to_filename()`](https://github.com/nipy/nibabel/blob/19fd3f82aead1f4fe1f45ac13139acc3ad30140f/nibabel/filebasedimages.py#L319-L334), to fully support `pathlib` objects.

effigies: Thinking a little bit more, a simpler alternative would be something like the following:

```Python
try:
    from os import fspath
except ImportError:
    def fspath(pathlike):
        return pathlike
```

This could be placed at the top of `loadsave.py` and `filebasedimages.py`, and inserted straightforwardly into the four functions/methods.

nibotmi:
:umbrella: The latest upstream changes (presumably #611) made this pull request unmergeable. Please resolve the merge conflicts.

nibabel/loadsave.py
CRiddler: I may be missing something, but python 3.4+ has the pathlib module built in, however `os.fspath` was only included in 3.6. So we will need a way to deal with pathlib.Path objects without `os.fspath` for py 3.4<3.6

effigies: I see. I didn't fully appreciate the timeline here. Could we update the above suggestion to something like the following?

```Python
try:
    from os import fspath
except ImportError:
    def fspath(pathlike):
        try:
            pathlike = pathlike.__fspath__()
        except AttributeError:
            pass
        if not isinstance(pathlike, (bytes, unicode)):
            raise TypeError("expected str, bytes or os.PathLike object, not "
                            "{}".format(type(pathlike).__name__))
        return pathlike
```

This matches the docs:

```
os.fspath = fspath(path)
    Return the file system path representation of the object.
    
    If the object is str or bytes, then allow it to pass through as-is. If the
    object defines __fspath__(), then return the result of that method. All other
    types raise a TypeError.
```

It might go best in `nibabel.py3k`.

@fepegar @matthew-brett Feel free to chime in if you have cleaner/alternate suggestions.

fepegar: As I said in #612 this might be naive, but why not converting always to `str`?
```python
path = str(pathlike)
```
This feels like the simplest solution and it's perfectly backwards compatible. See https://github.com/python/cpython/blob/3.6/Lib/pathlib.py#L701-L702

CRiddler: A few reasons to stay away from the `str(pathlike)` adaptation is justified in the `__fspath__` pep 519 https://www.python.org/dev/peps/pep-0519/#rationale. While calling `str` on pathlib.Path objects may work, calling `str` on `DirEntry` objects (as returned by `os.listdir` and `os.scandir`) does not work.

@effigies something I didn't see yesterday was that your suggestion still breaks on python 3.4<3.6 because unfortunately `os.fspath` and the dunder `.__fspath__()` were implemented in python 3.6+ so if we give it a python3.4 pathlib.Path object, it will hit the `TypeError` in your function.

In my opinion, `pandas` has done a pretty good job at resolving the pathlib issue. See https://github.com/pandas-dev/pandas/blob/53ad5c68b600f2929edd165c7b45e363c956144d/pandas/io/common.py#L121-L161

effigies: That looks reasonable to me. I'm realizing py3k is a copy from numpy, so it doesn't make sense to add to that. Perhaps, following in the spirit of #530, we can start a utils.py for this function.

Pandas is BSD, so no problems including their code, but we will need to credit them.

Note also that we'll need to translate from their `pandas.compat.text_type` to `six.

fepegar:
Why is the PR unmergeable?

matthew-brett:
It cannot be merged because there are conflicts with the master branch - see the web interface for this PR for links.

fepegar:
@CRiddler, do you think you can fix the conflicts?

CRiddler:
The conflicts have been resolved, however as stated earlier- this approach to resolving pathlib objects is not final nor foolproof. Essentially, we are checking if the `filename` is a `Pathlike` object which is only available in python 3.6+ and coercing that properly. If its not a `Pathlike` then we're coercing the object to six.unicode type. What happens when the user passes in the filename as `None`? Should we be coercing None to a string/unicode type? While the example is a corner case, it is still technically a case of an improperly handled input which can be fixed by a more in-depth filename handling procedure as discussed in the earlier comments.

effigies:
I'm setting this milestone to 3.0. Once we're only dealing with Python 3.5+, that reduces the range of behaviors we need to accommodate. If you want to try for 2.5 and support Python 2's `pathlib`, we can do that.

effigies:
Hi @CRiddler, @fepegar. By pushing off to Python3-only, I think we can be a lot dumber about this. I would say the logic can simply be:

```Python
filename = pathlib.Path(filename)
```

If we place this at the head of `load`, `save`, `to_filename` and `from_filename`, then we can just work with `Path`s internally, converting to `str` only when we hit functions that don't support them.

Sound like a reasonable approach? We would also want to add some tests where we pass in valid and invalid strings, paths, path-like objects, and non-path-likes.

fepegar:
Sounds reasonable to me.

effigies:
Hi @CRiddler, we're now requiring Python 3.5.1+, so we can resume work on this with simpler assumptions about the sort of objects we're going to be passed. I merged in master to resolve conflicts and get you going, but let me know how you want to proceed here.

CRiddler:
Hi @effigies, if we do want to convert everything into `pathlib.Path` objects and deal with those internally then a good amount of `filename_parser.py` will have to be rewritten. Do we still want to try to move forward with pathlib internally? Or should we add in the method that `pandas` uses to coerce pathlike objects to python3 string types?

effigies:
Hmm. I haven't really scoped out where all we will need to make modifications to use `Path`s internally. I figured we'd find the entry-points, and follow the logic to see where things break.

But if we want to take the Pandas approach, the entry-points are still the first step, so I don't see these as competing approaches. If anything the Pandas-like one would be a good first pass, and then we could evaluate if we want to take the plunge with internal `pathlib.Path`s. There are a number of places where we take either filenames or open filehandles, so this could end up getting messy, and might not be worth it.

As a note, it looks like Pandas has simplified their check: https://github.com/pandas-dev/pandas/blob/325dd686de1589c17731cf93b649ed5ccb5a99b4/pandas/io/common.py#L131-L160
Their minimum Python is now 3.5.3, so it seems pretty safe to follow suit. I don't think we need to copy the `expand_user` bit, though.

CRiddler:
I agree in that coercing everything to strings at the entry points will be the easiest way to go. The updated pandas method looks like it'll work perfectly for us since `__fspath__` is supported in 3.6+ and coercing `pathlib.Path` types to `str` will support any 3.5 <= 3.6 versions.

I will update the branch with a function similar to `pandas` approach (not sure how to give them credit for that though) and then add that function in the entry points you listed. We can see where to go from there? I can probably get around to this by the end of the week.

effigies:
Cool. We have external licenses in COPYING, IIRC, so that and a link to the original code (with specific revision) should suffice. Our licenses are compatible.

effigies:
@CRiddler Just a bump on this one. Let me know if there's anything I can do to help.

nibabel/filename_parser.py
```diff
@@ -14,6 +14,7 @@
 except NameError:
     basestring = str
 
+from .loadsave import _stringify_path
```
effigies: We're getting a circular import. Maybe it makes sense to define the method in this file?

effigies:
@CRiddler Just a bump on this. Let me know if there's anything I can help with. I'm going to aim to feature freeze 3.0 on October 28 (three weeks from Monday) so we can have a decent RC period.

CRiddler:
Yep! This is still on my radar, classes started for me this past week so I've been a little busy. This is the first on my to-do list once I have a bit of free time, so I should have it done over the weekend.

effigies:
I mean... also enjoy your weekend.

nibabel/filebasedimages.py
```diff
@@ -252,10 +252,12 @@ def set_filename(self, filename):
            ``.file_map`` attribute. Otherwise, the image instance will
            try and guess the other filenames from this given filename.
         '''
+        filename = _stringify_path(filename)
```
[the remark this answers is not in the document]
CRiddler: Working on this now, but had a question since `filespec_to_file_map` calls `types_filename`, do you think it would be safe to only put the guard in `types_filename` and not in `filespec_to_file_map` at all?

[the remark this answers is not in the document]
effigies: That seems reasonable.

effigies:
Hmm. Looks like your history got a bit messed up there.

Screen Shot 2019-10-16 at 16 51 58

I'd try rebasing onto fbaeacc. Should be pretty clean, as it's the merge commit that seems to be causing damage.

CRiddler:
Hopefully that fixed, I always struggle when it comes to wrestling with git history. If it needs some more reworking, I may need some hand-holding to resolve the history issue.

effigies:
Nope.

Screen Shot 2019-10-16 at 17 37 34

So the easy way to do this is just:

```Shell
git rebase -i fbaeacc
```

That will open an editor, probably `vim` or `nano`:

```
pick 6056014b update _stringify_path doc
pick b54926e3 update docstrings to accept str or os.PathLike
pick 8a34a7a0 mghformat accept pathlib for filespec_to_file_map
pick d6c96c02 tests pathlib compatible
pick cf631067 fix _stringify_path doc
```

That should be fine, so save and exit. Now you'll see:

```
Auto-merging nibabel/loadsave.py
Auto-merging nibabel/filename_parser.py
CONFLICT (content): Merge conflict in nibabel/filename_parser.py
Auto-merging nibabel/filebasedimages.py
error: could not apply b54926e3... update docstrings to accept str or os.PathLike
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply b54926e3... update docstrings to accept str or os.PathLike
```

If you run `git status`, you should see:

```
interactive rebase in progress; onto fbaeacc9
Last commands done (2 commands done):
   pick 6056014b update _stringify_path doc
   pick b54926e3 update docstrings to accept str or os.PathLike
Next commands to do (3 remaining commands):
   pick 8a34a7a0 mghformat accept pathlib for filespec_to_file_map
   pick d6c96c02 tests pathlib compatible
  (use "git rebase --edit-todo" to view and edit)
You are currently rebasing branch 'loadsave-pathlib_compat' on 'fbaeacc9'.
  (fix conflicts and then run "git rebase --continue")
  (use "git rebase --skip" to skip this patch)
  (use "git rebase --abort" to check out the original branch)

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   nibabel/loadsave.py

Unmerged paths:
  (use "git reset HEAD <file>..." to unstage)
  (use "git add <file>..." to mark resolution)

	both modified:   nibabel/filename_parser.py
```

So edit `nibabel/filename_parser.py` as you would during a merge conflict (search for `<<<<` and resolve), then:

```
git add nibabel/filename_parser.py
git rebase --continue
```

You may be dropped into an editor again, giving you a chance to modify the commit message. You can just save and exit, to leave it unchanged. Then you might see:

```
[detached HEAD 2d37fb71] update docstrings to accept str or os.PathLike
 Author: Cameron Riddell <cameronriddell1@gmail.com>
 1 file changed, 2 insertions(+), 2 deletions(-)
The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git reset'
interactive rebase in progress; onto fbaeacc9
Last commands done (5 commands done):
   pick d6c96c02 tests pathlib compatible
   pick cf631067 fix _stringify_path doc
No commands remaining.
You are currently rebasing branch 'loadsave-pathlib_compat' on 'fbaeacc9'.

nothing to commit, working tree clean
Could not apply cf631067... fix _stringify_path doc
```

In this case, there's a commit that had no effect. So you can just

```
git rebase --continue
```

When you've iterated through all of the necessary changes:

```
> git status
On branch loadsave-pathlib_compat
Your branch and 'origin/loadsave-pathlib_compat' have diverged,
and have 3 and 6 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

nothing to commit, working tree clean
```

Since your GitHub branch has a lot of old stuff in it, instead of:

```
git diff origin/loadsave-pathlib_compat
```

I would suggest just comparing to the commit you rebased against:

```
git diff fbaeacc9
```

If that looks reasonable, go ahead and force-push:

```
git push -f
```

nibabel/filename_parser.py
```diff
@@ -14,11 +14,41 @@
...
 
+def _stringify_path(filepath_or_buffer):
+    """Attempt to convert a path-like object to a string.
+
+    Parameters
+    ----------
+    filepath_or_buffer : object to be converted
+    Returns
+    -------
+    str_filepath_or_buffer : maybe a string version of the object
```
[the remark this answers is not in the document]
effigies: Bump on this comment.

nibabel/filename_parser.py
```diff
@@ -14,11 +14,41 @@
...
+    Notes
+    -----
+    Objects supporting the fspath protocol (python 3.6+) are coerced
+    according to its __fspath__ method.
+    For backwards compatibility with older pythons, pathlib.Path and
+    py.path objects are specially coerced.
```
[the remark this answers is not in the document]
effigies: Bump on this comment

effigies:
Great! Thanks for this.

fepegar:
Yay! Thank you guys for having worked on this.

PR #611, opened 2018-03-19
 detecting if the input is empty early on! 

raamana:
Apparently no one needed it so far ;)

Not sure where to write tests for it, so wanna check with you guys before taking appropriate action.


nibabel/loadsave.py
```diff
@@ -38,6 +38,10 @@ def load(filename, **kwargs):
     '''
     if not op.exists(filename):
         raise FileNotFoundError("No such file: '%s'" % filename)
+
+    if op.getsize(filename) <= 0:
+        raise FileEmptyError("Given file is empty: '%s'" % filename)
+
```
effigies: PEP8 is complaining about whitespace on this line.

nibabel/loadsave.py
```diff
@@ -38,6 +38,10 @@ def load(filename, **kwargs):
     '''
     if not op.exists(filename):
         raise FileNotFoundError("No such file: '%s'" % filename)
+
+    if op.getsize(filename) <= 0:
+        raise FileEmptyError("Given file is empty: '%s'" % filename)
```
effigies: Perhaps we can use `ImageFileError` here, rather than defining a new exception? And I think the message could be shortened to `"Empty file: '%s'" % filename`.

raamana:
I get the following, which talks about `Cannot work out file type ` which makes me think about incorrect header, not empty file.

`nibabel.filebasedimages.ImageFileError: Cannot work out file type of "/home/praamana/dev/visualqc/example_datasets/empty.nii"`

effigies:
This looks reasonable to me. I think you could quickly add a test in [`test_loadsave.py`](https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_loadsave.py). The following would exercise the new code and verify that you're getting a useful error message:

```Python
def test_empty_load():
    with InTemporaryDirectory():
        open('empty.nii', 'w').close()
        with assert_raises(ImageFileError) as cm:
            load('empty.nii')
    assert_true(cm.exception.args[0].startswith('Empty file: '))
```

nibabel/tests/test_loadsave.py
```diff
@@ -58,6 +58,14 @@ def test_file_not_found():
     assert_raises(FileNotFoundError, load, 'does_not_exist.nii.gz')
 
 
+def test_load_empty_image():
+    with InTemporaryDirectory():
+        open('empty.nii', 'w').close()
+        with assert_raises(ImageFileError) as err:
+            load('empty.nii')
```
matthew-brett: I'm surprised this works on Python 3 - I thought that the `err` variable went out of scope.

raamana: It seems to work - didn’t dig into why

effigies: From the docs:

```
Signature: unittest.TestCase.assertRaises(self, expected_exception, *args, **kwargs)
Docstring:
Fail unless an exception of class expected_exception is raised
by the callable when invoked with specified positional and
keyword arguments. If a different type of exception is
raised, it will not be caught, and the test case will be
deemed to have suffered an error, exactly as for an
unexpected exception.

If called with the callable and arguments omitted, will return a
context object used like this::

     with self.assertRaises(SomeException):
         do_something()

An optional keyword argument 'msg' can be provided when assertRaises
is used as a context object.

The context manager keeps a reference to the exception as
the 'exception' attribute. This allows you to inspect the
exception after the assertion::

    with self.assertRaises(SomeException) as cm:
        do_something()
    the_exception = cm.exception
    self.assertEqual(the_exception.error_code, 3)
File:      /anaconda3/lib/python3.6/unittest/case.py
Type:      function
```

effigies:
@matthew-brett Okay to merge?

matthew-brett:
Yes, looks good - merging - thanks.

Issue #573, opened 2017-11-05
Intent to deprecate: keep_file_open


effigies:
As of indexed_gzip [0.7.0](https://github.com/pauldmccarthy/indexed_gzip/releases/tag/v0.7.0), the default behavior of a `SafeIndexedGzipFile` is to close the underlying file between accesses. This pretty much obviates the machinery @pauldmccarthy built around the `keep_file_open` option in Opener, ArrayProxy, AnalyzeImage and MGHImage.

As we did have a release, we're going to need to deprecate the API, but I think the following scheme should be reasonably safe:

* Set minimum `indexed_gzip` version to 0.7.0
* Remove any effects of `keep_file_open` -  it only had any effect if `indexed_gzip` was installed AND `keep_file_open` was `'auto'` or `True` - and update docs to say "has no effect, will be removed vXXX"
* Any function that currently takes `keep_file_open` raises a `DeprecationWarning` if it is not `None`.
* Remove `arrayproxy.KEEP_FILE_OPEN_DEFAULT` - This one I'm not sure of. I think people would only be setting it, not reading it, but maybe it should stay just in case.
* Schedule full removal of option for v4.0

What do y'all think?

pauldmccarthy:
Hey @effigies, I don't have any strong feelings either way about this, apart from wondering whether there might be some use case (or future file IO driver) which depends on having a persistent file handle. This is not a problem for me personally though. Happy to help out if this goes ahead too!

pauldmccarthy:
I do like your `ReleasingFileHandle` idea - that would possibly make the existing code cleaner. Could this logic be built into the existing `Opener` class, and the `ArrayProxy` changed so that it persists a reference to a single `Opener`?

effigies:
That might be the cleanest way of doing it. I had a brief look, but haven't had time to think carefully about how to do that. Feel free to push ahead, and I'll try to review. Otherwise I'll poke at it as I find time.

pauldmccarthy:
Sure - I'm flying down under shortly so am quite busy, but I should have some time over the next couple of weeks.

pauldmccarthy:
Hey @effigies, I've finally gotten around to working on this - #614 -  let me know what you think!

PR #614, opened 2018-03-26
ENH:  Take advantage of IndexedGzipFile drop_handles flag

pauldmccarthy:
This is an attempt to address #573. 

 - Deprecate the `keep_file_open` parameter
 - Always use `indexed_gzip` if it is present (this can be disabled by overwriting the `nibabel.openers.HAVE_INDEXED_GZIP` flag)
 - Change `ArrayProxy` so that, if `indexed_gzip` is present, and it is given a `.gz` file, it creates and uses a single `ImageOpener`. Otherwise it creates a new `ImageOpener` on every access.

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -34,7 +34,7 @@ from .deprecated import deprecate_with_version
 from .volumeutils import array_from_file, apply_read_scaling
 from .fileslice import fileslice
 from .keywordonly import kw_only_meth
-from .openers import ImageOpener, HAVE_INDEXED_GZIP
+from . import openers
 
 
 """This flag controls whether a new file handle is created every time an image
@@ -214,7 +214,7 @@ class ArrayProxy(object):
         if hasattr(file_like, 'read') and hasattr(file_like, 'seek'):
             return False
         # don't have indexed_gzip - auto -> False
-        if keep_file_open == 'auto' and not (HAVE_INDEXED_GZIP and
+        if keep_file_open == 'auto' and not (openers.HAVE_INDEXED_GZIP and
                                              file_like.endswith('.gz')):
             return False
         return keep_file_open
@@ -263,11 +263,11 @@ class ArrayProxy(object):
         """
         if self._keep_file_open:
             if not hasattr(self, '_opener'):
-                self._opener = ImageOpener(
+                self._opener = openers.ImageOpener(
                     self.file_like, keep_open=self._keep_file_open)
             yield self._opener
         else:
-            with ImageOpener(self.file_like) as opener:
+            with openers.ImageOpener(self.file_like) as opener:
                 yield opener
 
     def get_unscaled(self):
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -52,7 +52,7 @@ If this is set to any other value, attempts to create an ``ArrayProxy`` without
 specifying the ``keep_file_open`` flag will result in a ``ValueError`` being
 raised.
 """
-KEEP_FILE_OPEN_DEFAULT = False
+KEEP_FILE_OPEN_DEFAULT = 'auto'
 
 
 class ArrayProxy(object):
@@ -186,10 +186,10 @@ class ArrayProxy(object):
         The return value is derived from these rules:
 
           - If ``file_like`` is a file(-like) object, ``False`` is returned.
-            Otherwise, ``file_like`` is assumed to be a file name.
-          - if ``file_like`` ends with ``'gz'``, and the ``indexed_gzip``
-            library is available, ``True`` is returned.
-          - Otherwise, ``False`` is returned.
+            Otherwise, ``file_like`` is assumed to be a file name
+          - If ``keep_file_open`` is ``auto``, and ``indexed_gzip`` is
+            not available, ``False`` is returned.
+          - Otherwise, the value of ``keep_file_open`` is returned unchanged.
 
         Parameters
         ----------
@@ -203,23 +203,20 @@ class ArrayProxy(object):
         -------
 
         The value of ``keep_file_open`` that will be used by this
-        ``ArrayProxy``.
+        ``ArrayProxy``, and passed through to ``ImageOpener`` instances.
         """
         if keep_file_open is None:
             keep_file_open = KEEP_FILE_OPEN_DEFAULT
-        # if keep_file_open is True/False, we do what the user wants us to do
-        if isinstance(keep_file_open, bool):
-            return keep_file_open
-        if keep_file_open != 'auto':
+        if keep_file_open not in ('auto', True, False):
             raise ValueError('keep_file_open should be one of {None, '
                              '\'auto\', True, False}')
-
         # file_like is a handle - keep_file_open is irrelevant
         if hasattr(file_like, 'read') and hasattr(file_like, 'seek'):
             return False
-        # Otherwise, if file_like is gzipped, and we have_indexed_gzip, we set
-        # keep_file_open to True, else we set it to False
-        return HAVE_INDEXED_GZIP and file_like.endswith('gz')
+        # don't have indexed_gzip - auto -> False
+        if keep_file_open == 'auto' and not HAVE_INDEXED_GZIP:
+            return False
+        return keep_file_open
 
     @property
     @deprecate_with_version('ArrayProxy.header deprecated', '2.2', '3.0')
@@ -263,12 +260,13 @@ class ArrayProxy(object):
             A newly created ``ImageOpener`` instance, or an existing one,
             which provides access to the file.
         """
-        if self._keep_file_open:
+        if bool(self._keep_file_open):
             if not hasattr(self, '_opener'):
-                self._opener = ImageOpener(self.file_like, keep_open=True)
+                self._opener = ImageOpener(
+                    self.file_like, keep_open=self._keep_file_open)
             yield self._opener
         else:
-            with ImageOpener(self.file_like, keep_open=False) as opener:
+            with ImageOpener(self.file_like) as opener:
                 yield opener
 
     def get_unscaled(self):
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -51,8 +51,12 @@ behaviour is the same as if ``keep_file_open is False``.
 If this is set to any other value, attempts to create an ``ArrayProxy`` without
 specifying the ``keep_file_open`` flag will result in a ``ValueError`` being
 raised.
+
+.. warning:: Setting this flag to a value of ``'auto'`` will become deprecated
+             behaviour in version 2.4.0. Support for ``'auto'`` will be removed
+             in version 3.0.0.
 """
-KEEP_FILE_OPEN_DEFAULT = 'auto'
+KEEP_FILE_OPEN_DEFAULT = False
 
 
 class ArrayProxy(object):
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -43,10 +43,10 @@ used for the lifetime of the ``ArrayProxy``. It should be set to one of
 ``True``, ``False``, or ``'auto'``.
 
 If ``True``, a single file handle is created and used. If ``False``, a new
-file handle is created every time the image is accessed. If ``'auto'``, and
-the optional ``indexed_gzip`` dependency is present, a single file handle is
-created and persisted. If ``indexed_gzip`` is not available, behaviour is the
-same as if ``keep_file_open is False``.
+file handle is created every time the image is accessed. For gzip files, if
+``'auto'``, and the optional ``indexed_gzip`` dependency is present, a single
+file handle is created and persisted. If ``indexed_gzip`` is not available,
+behaviour is the same as if ``keep_file_open is False``.
 
 If this is set to any other value, attempts to create an ``ArrayProxy`` without
 specifying the ``keep_file_open`` flag will result in a ``ValueError`` being
@@ -186,7 +186,7 @@ class ArrayProxy(object):
         The return value is derived from these rules:
 
           - If ``file_like`` is a file(-like) object, ``False`` is returned.
-            Otherwise, ``file_like`` is assumed to be a file name
+            Otherwise, ``file_like`` is assumed to be a file name.
           - If ``keep_file_open`` is ``auto``, and ``indexed_gzip`` is
             not available, ``False`` is returned.
           - Otherwise, the value of ``keep_file_open`` is returned unchanged.
@@ -214,7 +214,8 @@ class ArrayProxy(object):
         if hasattr(file_like, 'read') and hasattr(file_like, 'seek'):
             return False
         # don't have indexed_gzip - auto -> False
-        if keep_file_open == 'auto' and not HAVE_INDEXED_GZIP:
+        if keep_file_open == 'auto' and not (HAVE_INDEXED_GZIP and
+                                             file_like.endswith('.gz')):
             return False
         return keep_file_open
 
@@ -260,7 +261,7 @@ class ArrayProxy(object):
             A newly created ``ImageOpener`` instance, or an existing one,
             which provides access to the file.
         """
-        if bool(self._keep_file_open):
+        if self._keep_file_open:
             if not hasattr(self, '_opener'):
                 self._opener = ImageOpener(
                     self.file_like, keep_open=self._keep_file_open)
--- a/nibabel/benchmarks/bench_array_to_file.py
+++ b/nibabel/benchmarks/bench_array_to_file.py
@@ -16,6 +16,7 @@ Run this benchmark with:
 from __future__ import division, print_function
 
 import sys
+from io import BytesIO  # NOQA
 
 import numpy as np
 
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -18,16 +18,28 @@ from distutils.version import StrictVersion
 
 # is indexed_gzip present and modern?
 try:
-    from indexed_gzip import SafeIndexedGzipFile, __version__ as version
+    import indexed_gzip as igzip
+    version = igzip.__version__
 
     HAVE_INDEXED_GZIP = True
 
+    # < 0.6 - no good
     if StrictVersion(version) < StrictVersion('0.6.0'):
         warnings.warn('indexed_gzip is present, but too old '
                       '(>= 0.6.0 required): {})'.format(version))
         HAVE_INDEXED_GZIP = False
-
-    del version
+    # < 0.7 - does not support drop_handles
+    elif StrictVersion(version) < StrictVersion('0.7.0'):
+        class IndexedGzipFile(igzip.SafeIndexedGzipFile):
+            def __init__(self, *args, **kwargs):
+                kwargs.pop('drop_handles', None)
+                super(IndexedGzipFile, self).__init__(*args, **kwargs)
+    # >= 0.8 SafeIndexedGzipFile renamed to IndexedGzipFile
+    elif StrictVersion(version) < StrictVersion('0.8.0'):
+        IndexedGzipFile = igzip.SafeIndexedGzipFile
+    else:
+        IndexedGzipFile = igzip.IndexedGzipFile
+    del igzip, version
 
 except ImportError:
     HAVE_INDEXED_GZIP = False
@@ -80,9 +92,12 @@ class BufferedGzipFile(gzip.GzipFile):
 
 def _gzip_open(filename, mode='rb', compresslevel=9, keep_open=False):
 
-    # use indexed_gzip if possible for faster read access
-    if keep_open and mode == 'rb' and HAVE_INDEXED_GZIP:
-        gzip_file = SafeIndexedGzipFile(filename)
+    # use indexed_gzip if possible for faster read access.  If keep_open ==
+    # True, we tell IndexedGzipFile to keep the file handle open. Otherwise
+    # the IndexedGzipFile will close/open the file on each read.
+    if HAVE_INDEXED_GZIP and keep_open in (True, 'auto') and mode == 'rb':
+        gzip_file = IndexedGzipFile(
+            filename, drop_handles=keep_open == 'auto')
 
     # Fall-back to built-in GzipFile (wrapped with the BufferedGzipFile class
     # defined above)
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -89,9 +89,9 @@ def _gzip_open(filename, mode='rb', compresslevel=9, keep_open=False):
     # use indexed_gzip if possible for faster read access.  If keep_open ==
     # True, we tell IndexedGzipFile to keep the file handle open. Otherwise
     # the IndexedGzipFile will close/open the file on each read.
-    if HAVE_INDEXED_GZIP and keep_open in (True, 'auto') and mode == 'rb':
+    if HAVE_INDEXED_GZIP and mode == 'rb':
         gzip_file = IndexedGzipFile(
-            filename, drop_handles=keep_open == 'auto')
+            filename, drop_handles=keep_open is not True)
 
     # Fall-back to built-in GzipFile (wrapped with the BufferedGzipFile class
     # defined above)
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -90,8 +90,7 @@ def _gzip_open(filename, mode='rb', compresslevel=9, keep_open=False):
     # True, we tell IndexedGzipFile to keep the file handle open. Otherwise
     # the IndexedGzipFile will close/open the file on each read.
     if HAVE_INDEXED_GZIP and mode == 'rb':
-        gzip_file = IndexedGzipFile(
-            filename, drop_handles=keep_open is not True)
+        gzip_file = IndexedGzipFile(filename, drop_handles=not keep_open)
 
     # Fall-back to built-in GzipFile (wrapped with the BufferedGzipFile class
     # defined above)
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -23,17 +23,11 @@ try:
 
     HAVE_INDEXED_GZIP = True
 
-    # < 0.6 - no good
-    if StrictVersion(version) < StrictVersion('0.6.0'):
+    # < 0.7 - no good
+    if StrictVersion(version) < StrictVersion('0.7.0'):
         warnings.warn('indexed_gzip is present, but too old '
-                      '(>= 0.6.0 required): {})'.format(version))
+                      '(>= 0.7.0 required): {})'.format(version))
         HAVE_INDEXED_GZIP = False
-    # < 0.7 - does not support drop_handles
-    elif StrictVersion(version) < StrictVersion('0.7.0'):
-        class IndexedGzipFile(igzip.SafeIndexedGzipFile):
-            def __init__(self, *args, **kwargs):
-                kwargs.pop('drop_handles', None)
-                super(IndexedGzipFile, self).__init__(*args, **kwargs)
     # >= 0.8 SafeIndexedGzipFile renamed to IndexedGzipFile
     elif StrictVersion(version) < StrictVersion('0.8.0'):
         IndexedGzipFile = igzip.SafeIndexedGzipFile
--- a/nibabel/tests/test_openers.py
+++ b/nibabel/tests/test_openers.py
@@ -12,9 +12,10 @@ import contextlib
 from gzip import GzipFile
 from bz2 import BZ2File
 from io import BytesIO, UnsupportedOperation
+from distutils.version import StrictVersion
 
 from ..py3k import asstr, asbytes
-from ..openers import Opener, ImageOpener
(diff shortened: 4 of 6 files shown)


nibabel/arrayproxy.py
```diff
@@ -156,8 +139,9 @@ def __init__(self, file_like, spec, mmap=True, keep_file_open=None):
         # Permit any specifier that can be interpreted as a numpy dtype
         self._dtype = np.dtype(self._dtype)
         self._mmap = mmap
-        self._keep_file_open = self._should_keep_file_open(file_like,
-                                                           keep_file_open)
+        # If using indexed_gzip, we use a single ImageOpener. Otherwise, we
+        # create a new ImageOpener on each file access
+        self._persist_opener = openers.HAVE_INDEXED_GZIP
```
effigies: Main thought here is that we only want to use a single file if we have `indexed_gzip` and the file is a `.gz` file. Quick read-through suggests that the un-compressed case might not be handled correctly, but I may have just missed it.

I'll try to have a more detailed look soon.

pauldmccarthy: Aah yes, you're right. I'll change the condition to something like `HAVE_INDEXED_GZIP and isinstance(file_like, str) and file_like.endswith('.gz')`.

matthew-brett:
I hate to jump in without understanding.   But, would the flag continue to be useful in the following situation, without `indexed_gzip` installed?

```
img = nib.load('foo.nii.gz', keep_file_open=True)
for volume_no in range(img.shape[-1]):
    vol_data = img.dataobj[..., volume_no]
```

pauldmccarthy:
Hey @matthew-brett, thanks for the comment.  This PR would break that example
(i.e. `keep_file_open` would be ignored, and the file being decompressed from
its beginning on each iteration of the loop).

Based on discussions between myself and @effigies (#573), I was under the
impression that `keep_file_open` didn't have any uses outside of
`indexed_gzip`-based access.

But your example suggests otherwise - is this a real use case, and are you
suggesting that you would like to retain the ability to `keep_file_open`
(apologies for the grammar)?

matthew-brett:
i think I'm right in saying that the example above would be much faster with `keep_file_open` - is that right?

pauldmccarthy:
Yep, that's correct. So it would be useful if you need to scan through a `.gz` file sequentially.

effigies:
Yes, there's definitely a huge speed-up for keeping files open in that case, so it makes sense to me to keep the option around, given we went through the trouble of putting it in.

So at this point, I think I'd want the logic to be:

| `IndexedGzipFile` | `keep_file_open` | `.gz` | Behavior |
|---|---|---|---|
| `True` | `True` | `True` | Keep open |
| `True` | `True` | `False` | Keep open |
| `True` | `False` | `True` | Keep open |
| `True` | `False` | `False` | Re-open |
| `False` | `True` | `True` | Keep open |
| `False` | `True` | `False` | Keep open |
| `False` | `False` | `True` | Re-open |
| `False` | `False` | `False` | Re-open |

So keep open if `keep_file_open or IndexedGzipFile and .gz`. And I'm pretty sure the default behavior I'd want at this point is equivalent to the `keep_file_open == False` subset of the table, so we can still deprecate the `'auto'` and `None` options, and make them equivalent to `False`.

That said, if we're not doing a major refactor, do we want to continue to permit `indexed_gzip >= 0.6.0`, with the old behavior for `'auto'` and `None`? Unfortunately, NeuroDebian doesn't seem to be updating to 0.7+, so it looks like it might be useful to continue supporting the 0.6 series that for a while longer.

If so, then I think the logic becomes a bit annoying to put in a truth table:

```Python
if keep_file_open is None:
    keep_file_open = arrayproxy.KEEP_FILE_OPEN_DEFAULT
if keep_file_open is True:
    return True
if indexed_gzip.__version__ < 0.6:  # I treat missing as 0
    return False
if indexed_gzip.__version__ >= 0.7:
    return fname.endswith('.gz')
return keep_file_open and fname.endswith('.gz')
```

What do y'all think? (And sorry for initiating this refactor, if we're going to end up throwing most of it away.)

matthew-brett:
Sorry, I know this is annoying, but it seems to me `keep_file_open=False` leading to the file being kept open anyway (`indexed_gzip` present, `.gz` file) is a bit odd.   Can't we just change the default to 'auto'?

pauldmccarthy:
So how about we just preserve the existing behaviour, but change the default value of `keep_file_open` to `'auto'`?

matthew-brett:
Yes, that's my suggestion - but I'm afraid I didn't follow the previous discussion.  Does that suit everyone?

matthew-brett:
Specifically, I think we should change the default value in `nibabel.arrayproxy.KEEP_FILE_OPEN_DEFAULT` - and leave `None` as the default value in the keyword arguments.  That way, if anyone is customizing the default behavior there, they will see no change in the default.

effigies:
The reason to "keep open" for 0.7.0+ is that indexed gzip now closes the
actual filehandle when not performing file operations. My opinion is that
as long as the filehandle gets closed, it doesn't matter whether we use a
persistent Opener object. That's just the means to the end when we're
directly holding the filehandle, or something that acts equivalently.

pauldmccarthy:
@effigies I see your logic - for `indexed_gzip` >= 0.7, the `keep_file_open`
flag is not actually controlling whether the file handle is kept open, but
rather whether a single `Opener` (and hence an `IndexedGzipFile` object) is
kept alive across accesses.

So semantically the name `keep_file_open` is a bit inaccurate (and this is why
in the current state of the PR, I changed the internal wording in `ArrayProxy`
to `_persist_opener`, rather than `_keep_file_open`).


But in the interest of minimising both confusion and the need to refactor the
code, I'm sort of inclined to go with @matthew-brett's suggestion, and keep
the code more or less as-is. So the `keep_file_open` flag will simply dictate
whether an `ArrayProxy` uses one `Opener` across accesses, or one for each
access.

Happy to be overruled though!

effigies:
_persist_opener is a hidden variable, so it isn't what's relevant to user
expectations, and most will not know what an Opener is or that the normal
behavior is to throw it away. keep_file_open is a good description of our
goal and its impact on OS resource limits.

The whole point of the exercise, to my mind, is to give users the advantage
of indexed_gzip for free when doing so doesn't change OS resource
consumption, which is to say, for version 0.7+.

I think for v0.6 (did we decide whether to keep supporting it?), the old
default should be maintained, for all of the reasons we set it to False.
And if we consider 0.7's internal closing to be in the correct spirit, then
False remains the best default across versions.

pauldmccarthy:
> I think for v0.6 (did we decide whether to keep supporting it?)

Probably safe to, given neurodebian.


What about the following:

1. Clarify the semantics of the `keep_file_open` parameter and the
   `KEEP_FILE_OPEN_DEFAULT` constant:

- `True`: A single `Opener` and file object is used across accesses,
  regardless of file type
  
- `False`: An `Opener` and file object is created for each access,
  regardless of file type
  
- `'auto'`: If file is `.gz` and `indexed_gzip` is present, as for
  `True`. Otherwise as for `False`.


2. Set `KEEP_FILE_OPEN_DEFAULT` depending on the presence of, and version of
   `indexed_gzip`:

`indexed_gzip` | `KEEP_FILE_OPEN_DEFAULT`
---------------|-------------------------
Not installed, or version < 0.7 | `False` 
Version >= 0.7 | `'auto'`


3. (purely cosmetic) Move the `KEEP_FILE_OPEN_DEFAULT` constant into the
   `openers` module, and initialise it in the same place that
   `HAVE_INDEXED_GZIP` is initialised. This just has the advantage that we
   wouldn't need to test for `indexed_gzip` presence/version in both `openers`
   and `ArrayProxy`.

effigies:
That seems fine, if that's the consensus. I just don't think anybody cares about Openers who doesn't hack on nibabel, and changing the value of a constant based on dependency versions feels less user friendly than fiddling with implementation details.

But I'm okay with your approach. Just making sure we are all deciding based on a full picture.

But I don't think we can move the constant, since anybody modifying it will expect it to continue working.

pauldmccarthy:
Sigh, if only there was a nicer way than [this](http://www.dr-josiah.com/2013/12/properties-on-python-modules.html) to use `@property` at the module level ... because we could then easily deprecate `arrayproxy.KEEP_FILE_OPEN_DEFAULT`.

matthew-brett:
Sorry I'm on my phone.   Are you saying that indexed_gzip will close the open file we pass ?.  So the flag says keep open but in fact it gets closed?

pauldmccarthy:
@matthew-brett - yes, as of 0.7.0, an `IndexedGzipFile` will, by default, close/re-open the file handle on each access.  Just like what  `nibabel` does by default. Hence the complexity of this conversation :)

matthew-brett:
Sorry to be ignorant here, but do we not have a way of using the final result of `keep_file_open` to tell `indexed_gzip` whether to keep the file open?

Is the point that, there is no advantage to keeping the file open with 0.7?   If so, can't we deal with that by making `auto` change its behavior for 0.7?

effigies:
In 0.7, the file is always closed between actions. There is no benefit
whatsoever to keeping it open, and thus no way to pass the option was
provided. In effect, it's doing what ArrayProxy is doing when
keep_file_open is False, but at a lower level. The only effect of closing
an IndexedGzipFile in 0.7 is to throw out the index.

Thus my reasoning is that nobody ever wants to throw out the index except
to close filehandles, which is the straightforward meaning of k
eep_file_open=False. If we're able to give them the index and close
filehandles, why bother killing the index over the implementation detail of
persistent Openers?

Again, I'm okay with changing the auto constant, if that's the consensus,
but it feels less user friendly to me.

We could also switch Opener to persistent, pass the keep open parameter to
it, and have it handle the opening and closing, rather than ArrayProxy, if
that's cleaner.

effigies:
Oh, actually, my mistake. `IndexedGzipFile(..., drop_handles=False)` is possible (and, of course, passing an open filehandle will not drop it, since there's no way to reopen), but we ran benchmarks, and there is no penalty to closing the file, because the index is so helpful. Thus `drop_handles=True` is the default. I would not object to passing `IndexedGzipFile(..., drop_handles=keep_file_open is not True)`. Our tests indicate would have negligible performance benefits, but it would be very true to user intent.

For reference, though it's not exactly light reading: https://github.com/pauldmccarthy/indexed_gzip/compare/v0.6.1...v0.7.0

pauldmccarthy:
Hey @effigie and @matthew-brett, sorry for the delay, and apologies if this is too late for 2.3.0.  

Here is my new take on how `nibabel` should behave. If `keep_file_open in (True, False)`, the behaviour is what you would intuitively expect from the name, regardless of whether `gzip` or `indexed_gzip` is used.  If `indexed_gzip` is present, and `keep_file_open is True`, the `drop_handles` feature  of `indexed_gzip` is not used (i.e. the file handle is kept open).

What do you guys think?

| `HAVE_INDEXED_GZIP` | `keep_file_open` | Library used   | ArrayProxy persists opener | `indexed_gzip.drop_handles` |
| ------------------- | ---------------- | -------------- | -------------------------- | --------------------------- |
| `False`             | `False`          | `gzip`         | No                         | na                          |
| `False`             | `'auto'`         | `gzip`         | No                         | na                          |
| `False`             | `True`           | `gzip`         | Yes                        | na                          |
| `True`              | `False`          | `gzip`         | No                         | na                          |
| `True`              | `'auto'`         | `indexed_gzip` | Yes                        | `True`                      |
| `True`              | `True`           | `indexed_gzip` | Yes                        | `False`                     |

effigies:
You're 5 hours ahead of me. As far as I'm concerned, this could still go in. So just to clarify this case:

| `HAVE_INDEXED_GZIP` | `keep_file_open` | Library used   | ArrayProxy persists opener | `indexed_gzip.drop_handles` |
| ------------------- | ---------------- | -------------- | -------------------------- | --------------------------- |
| `True`              | `False`          | `gzip`         | No                         | na                          |

We're now using `keep_file_open == False` as a way of disabling `indexed_gzip`. That is, if we want to drop filehandles, setting `indexed_gzip.drop_handles=True` would be sufficient to do everything but drop the actual `Opener` object, which IMO is an implementation detail with whose garbage collection users are not going to be concerned.

Anyway, this seems fine, especially if `'auto'` is now default. I just suspect we'll need to reference this thread at some point to explain why `keep_file_open` is also changing the library used to open compressed files.

nibabel/arrayproxy.py
```diff
@@ -186,10 +186,10 @@ def _should_keep_file_open(self, file_like, keep_file_open):
         The return value is derived from these rules:
 
           - If ``file_like`` is a file(-like) object, ``False`` is returned.
-            Otherwise, ``file_like`` is assumed to be a file name.
-          - if ``file_like`` ends with ``'gz'``, and the ``indexed_gzip``
-            library is available, ``True`` is returned.
-          - Otherwise, ``False`` is returned.
+            Otherwise, ``file_like`` is assumed to be a file name
```
effigies: For consistency, I would restore this period (`.`).

nibabel/arrayproxy.py
```diff
@@ -203,23 +203,20 @@ def _should_keep_file_open(self, file_like, keep_file_open):
...
         # file_like is a handle - keep_file_open is irrelevant
         if hasattr(file_like, 'read') and hasattr(file_like, 'seek'):
             return False
-        # Otherwise, if file_like is gzipped, and we have_indexed_gzip, we set
-        # keep_file_open to True, else we set it to False
-        return HAVE_INDEXED_GZIP and file_like.endswith('gz')
+        # don't have indexed_gzip - auto -> False
+        if keep_file_open == 'auto' and not HAVE_INDEXED_GZIP:
+            return False
+        return keep_file_open
```
effigies: We're dropping consideration of `keep_file_open == 'auto' and not file_like.endswith('gz')`. In this case, we will keep uncompressed files open, unnecessarily.

I would suggest making this final line:

```Python
return keep_file_open and file_like.endswith('gz')
```

nibabel/arrayproxy.py
```diff
@@ -263,12 +260,13 @@ def _get_fileobj(self):
             A newly created ``ImageOpener`` instance, or an existing one,
             which provides access to the file.
         """
-        if self._keep_file_open:
+        if bool(self._keep_file_open):
```
effigies: `bool(x)` is implicit in `if x`. I would revert this.

nibabel/openers.py
```diff
@@ -18,16 +18,28 @@
...
                       '(>= 0.6.0 required): {})'.format(version))
         HAVE_INDEXED_GZIP = False
-
-    del version
+    # < 0.7 - does not support drop_handles
+    elif StrictVersion(version) < StrictVersion('0.7.0'):
+        class IndexedGzipFile(igzip.SafeIndexedGzipFile):
+            def __init__(self, *args, **kwargs):
+                kwargs.pop('drop_handles', None)
+                super(IndexedGzipFile, self).__init__(*args, **kwargs)
```
effigies: I'm uncomfortable with this, as we're now making the default non-closing `IndexedGzipFile`s. The reason we previously had `KEEP_FILE_OPEN_DEFAULT = False` was because users could unexpectedly end up with a bunch of open filehandles by installing `indexed_gzip`. So that will be the case if users have `indexed_gzip.__version__ < 0.7.0`.

I think our only options are to either make 0.7 the minimum version (in line 27), or to make the value of `KEEP_FILE_OPEN_DEFAULT` depend on the version of `indexed_gzip`.

matthew-brett:
I'm really sorry to be slow on this one - but I thought we had found in https://github.com/nipy/nibabel/pull/614#issuecomment-377783817 that it was possible to pass the keep_file_open flag via `IndexedGzipFile(..., drop_handles=False)`.  If we are doing that, can we not still use `indexed_gzip` for .gz files, if available?

nibabel/arrayproxy.py
pauldmccarthy: Aah - I left out the `endswith('.gz')` test intentionally, because if the user says `keep_file_open`, we should do just that, regardless of the file type, right?

But I've added the test back in, as `keep_file_open == 'auto'` and `not file_like.endswith('.gz')` should revert to `keep_file_open == False`. Commits coming in a few minutes!

effigies: > because if the user says keep_file_open, we should do just that, regardless of the file type, right?

Right.

> But I've added the test back in, as keep_file_open == 'auto' and not file_like.endswith('.gz') should revert to keep_file_open == False

Yes. That's what I should have said. Good catch.

pauldmccarthy:
Hey @matthew-brett, yes we could. I added in this "revert to gzip" behaviour because there is a performance penalty to using `indexed_gzip` for one-off reads - `gzip` performs better in this case, because it doesn't need to build the index.

So I thought it would be good to have the option to bypass `indexed_gzip` if necessary. 

Although I guess in this case, an alternative would be to override the `openers.HAVE_INDEXED_GZIP` flag. If you're happy with this, I'll change the code to always use `indexed_gzip` if `HAVE_INDEXED_GZIP is True`.

matthew-brett:
A continued apology for my lack of understanding.  I thought from https://github.com/nipy/nibabel/pull/614#issuecomment-377783817 that not holding open the file didn't have much of a performance penalty?

pauldmccarthy:
No, that's a slightly different scenario - when using `indexed_gzip`, there is no performance penalty to setting `drop_handles=True` when compared to `drop_handles=False`.

But `gzip` and `indexed_gzip` perform better or wose than each other depending on the use case:

 - For *one-off* reads (e.g. you're only interested in the 100th volume), `gzip` will be faster than `indexed_gzip`
 - For continued random-access reads, `indexed_gzip` will be faster. 

Does that make sense?

nibabel/openers.py
pauldmccarthy: So ... given neurodebian I guess the latter option is the way t o go ...

effigies: Well, we can always just do the same as for < 0.6, and warn. It's not that people using Debian can't upgrade packages with `pip`.

pauldmccarthy: If you're happy with that, I am!

effigies: That would be my preference, just because it's simpler, and makes constants a little more constant.

But we should also go ahead and argue for NeuroDebian updating its package.

effigies:
I went ahead and ran the test above (with 100 instead of 1000 loops).

With `HAVE_INDEXED_GZIP==True`:
Time:  22.333674907684326

With `HAVE_INDEXED_GZIP==False`:
Time:  21.305696725845337

Building the index has a ~5% overhead in this case, and works out to ~.01s / image.

nibabel/tests/test_openers.py
```diff
@@ -132,14 +132,18 @@ def test_Opener_gzip_type():
...
-            (True,  {'mode' : 'rb', 'keep_open' : True},  MockIndexedGzipFile),
-            (True,  {'mode' : 'rb', 'keep_open' : False}, GzipFile),
-            (True,  {'mode' : 'wb', 'keep_open' : True},  GzipFile),
-            (True,  {'mode' : 'wb', 'keep_open' : False}, GzipFile),
+            (False, {'mode' : 'rb', 'keep_open' : True},   GzipFile),
+            (False, {'mode' : 'rb', 'keep_open' : False},  GzipFile),
+            (False, {'mode' : 'rb', 'keep_open' : 'auto'}, GzipFile),
+            (False, {'mode' : 'wb', 'keep_open' : True},   GzipFile),
+            (False, {'mode' : 'wb', 'keep_open' : False},  GzipFile),
+            (False, {'mode' : 'rb', 'keep_open' : 'auto'}, GzipFile),
```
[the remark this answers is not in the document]
pauldmccarthy: Just making sure that `IndexedGzipFile` doesn't get used when writing (because it doesn't support writing).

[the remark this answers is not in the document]
effigies: I meant that `'rb'` should be `'wb'`, right?

nibabel/benchmarks/bench_array_to_file.py
```diff
@@ -16,6 +16,7 @@
 from __future__ import division, print_function
 
 import sys
+from io import BytesIO  # NOQA
```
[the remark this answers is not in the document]
pauldmccarthy: Style warning without it - `BytesIO` only gets used in `eval`ed strings, so the style checker doesn't pick it up as being used.

nibabel/tests/test_openers.py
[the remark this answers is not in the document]
pauldmccarthy: Oh, I must be cross-eyed :)

pauldmccarthy:
Although I think that relative performance of `indexed_gzip` will degrade as the image size increases. I'll start a larger benchmark running now ...

effigies:
@yarikoptic While we have your attention over around these parts... Any chance of getting an update to [python-indexed-gzip](http://neuro.debian.net/pkgs/python-indexed-gzip.html)/[python3-indexed-gzip](http://neuro.debian.net/pkgs/python3-indexed-gzip.html)?

pauldmccarthy:
Hmm, I was sure that I had observed a performance degradation with `indexed_gizp`, but am struggling to reproduce it - results for a slightly modified benchmark (on a `(91, 109, 91, 825) float32` image, 1.3GB compressed):

```python
import numpy as np
import nibabel as nb
from time import time
from unittest import mock

fname = '/Users/paulmc/Data/bigdata.nii.gz'
nvols = nb.load(fname).shape[-1]
vols = [int(v) for v in np.random.randint(0, nvols, 100)]

def getvols():
    i = nb.load(fname)
    v1 = i.dataobj[..., vols[0]]
    # print(i.dataobj._keep_file_open)
    # print(getattr(i.dataobj, '_opener', mock.MagicMock()).fobj)
    return [nb.load(fname).dataobj[..., v] for v in vols[1:]]

nb.openers.HAVE_INDEXED_GZIP = True
tic = time()
getvols()
toc = time()
print("Time: ", toc - tic)

nb.openers.HAVE_INDEXED_GZIP = False
tic = time()
getvols()
toc = time()
print("Time: ", toc - tic)
```
```
Time:  625.4601578712463
Time:  601.442715883255
```

effigies:
Okay, so it's still only a 4% overhead in this large file case.

I say for the sake of truth in advertising, we should go ahead and use `indexed_gzip` and just drop handles if `keep_file_open==False`. The advantages to a persistent index in non-single-shot cases seem worth this fairly rare penalty.

pauldmccarthy:
Updated to always use `indexed_gzip` if present. And if we're always using `indexed_gzip`, should we get rid of the `'auto'` option too? e.g. deprecate it and make it equivalent to `False`.

Then the `ArrayProxy` would use `if HAVE_INDEXED_GZIP or keep_file_open` to determine whether to persist a single `ImageOpener`.

If we're all happy with this, I will try and finish this PR off tonight (UK time).

effigies:
I think that makes sense, as the default was already `False`, and should simplify the logic. I might wait a version before actively producing a warning for `'auto'`, though. `DeprecationWarning`s can be pretty noisy for users when an app developer wants to support multiple versions of a library. (Willing to be overruled for an immediate `DeprecationWarning`.)

matthew-brett:
Just working through the logic, 'auto' means keep_file_open when indexed_gzip is present.  But, with your investigations, you concluded that keeping the file open didn't actually have much effect on performance.   So 'auto' won't have much effect on performance?

I guess there may be situations where it does?   NFS mounts?  Slow hard disks?

effigies:
> 'auto' means keep_file_open when indexed_gzip

What it means is, if we update the above table to the following:


| `HAVE_INDEXED_GZIP` | `keep_file_open` | Library used   | ArrayProxy persists opener | `indexed_gzip.drop_handles` |
| ------------------- | ---------------- | -------------- | -------------------------- | --------------------------- |
| `False`             | `False`          | `gzip`         | No                         | na                          |
| `False`             | `'auto'`         | `gzip`         | No                         | na                          |
| `False`             | `True`           | `gzip`         | Yes                        | na                          |
| `True`              | `False`          | `indexed_gzip`         | Yes                         | `True`                          |
| `True`              | `'auto'`         | `indexed_gzip` | Yes                        | `True`                      |
| `True`              | `True`           | `indexed_gzip` | Yes                        | `False`                     |

Then in all cases `'auto'` is equivalent to `False`.

effigies:
(The fourth row is the one we've been discussing, for reference.)

matthew-brett:
Right - but dropping the file handles for `auto` is new, right?   Why not retain the file handles for `auto`?

matthew-brett:
Sorry again for not following, but am I right in thinking that `auto` with `indexed_gzip` available, used to cause `keep_file_open` to be set to True?   If that logic does not change, then that will now cause `drop_handles` to be False?

effigies:
No worries.

> [Am] I right in thinking that `auto` with `indexed_gzip` available, used to cause `keep_file_open` to be set to True?

Yes.

> If that logic does not change, then that will now cause drop_handles to be False?

Ah, that logic is changing. The issue here is that we've been conflating `Opener`s with file handles, because when `indexed_gzip` did not have `drop_handles` (equivalent to `drop_handles = False`), they were equivalent. However, we did see, when adding the `drop_handles` parameter, that it had no performance impact, and thus set the default to `True`. Therefore, *without change in nibabel*, the behavior of `'auto'` has become to drop file handles when possible.

The benefit of `keep_file_open=False` has been to preserve the old nibabel behavior of dropping file handles opportunistically to avoid hitting quotas. `'auto'` now meets that criterion, and we've been discussing whether to make `keep_file_open=False` drop an `Opener` (and thus ignore `indexed_gzip`) or use `indexed_gzip` with `drop_handles=True`.

To address your earlier questions:

> But, with your investigations, you concluded that keeping the file open didn't actually have much effect on performance.

We've found two things.

1) Dropping file-handles within `indexed_gzip` had no effect on performance. Agreed that this was local, and we did not try over NFS or similar. However, this argues for users to set `keep_file_open=True`, not `False` or `'auto'`, so that decision should be unaffected by the `'auto'`/`False` collapse.

2) There is a 4%-5% overhead for a single-shot, random-access volume selection with `indexed_gzip` over `gzip`. However, disabling `indexed_gzip` where `keep_file_open=False` does not seem like an intuitive use of the parameter, and would reflect history more than describe the behavior of the library.

matthew-brett:
Right - so I agree that disabling `indexed_zip` for keep_file_open=False is a bad idea, but I'm wondering whether we should preserve the behavior of `auto`, causing file handles to be preserved, on the basis that it may sometimes make a difference to performance.  Anyone using `auto` will already have had the chance of running out of filehandles.  It's pretty easy to explain that `auto` is equivalent to `keep_file_open=True` when `indexed_gzip` is available, but we can also say that it will likely have little impact on performance.

effigies:
I'll start by saying: That's fine. If that makes sense to people, okay.

That said, I don't think it's very intuitive that `'auto'` would come to mean "keep files open for `indexed_gzip` but not `gzip`". In every case where there's a performance benefit for `indexed_gzip`, there will be one for `gzip` (e.g., if closing the file leads to higher latencies over the network). So we would be changing behavior based on the Python environment for historical reasons that were pragmatic and are now speculative.

But again, it's fine, especially if documented as you suggest. I'll always use `False` (or `True` in cases where keeping a FH open is a known benefit), so I have no personal stake in `'auto'`.

matthew-brett:
Yes, I see the argument (that keeping files open = True is always faster regardless of whether you have indexed_gzip.   How about warning in the docstring about deprecation this release, and deprecating next release, as you suggested a bit further up?

effigies:
>  - if `keep_file_open == 'auto' and HAVE_INDEXED_GZIP`, we should keep files open (i.e. `drop_handles=False`)
>  - if `keep_file_open == 'auto' and not HAVE_INDEXED_GZIP`, we should drop file handles (by way of the `ArrayProxy` dropping `ImageOpener` instances)

I understood Matthew to mean we shouldn't change the behavior of `'auto'`, so we would end up back to `'auto' == False`, but again, I don't care here. Matthew, feel free to clarify your preference. Otherwise, I'm okay with whichever Paul chooses.

>  - Add to the docs that `keep_file_open = 'auto'` will be deprecated in 2.4, and removed in ... 3.0?

Sounds good to me.

> I guess I'll also change the default behaviour (dictated by `arrayproxy.KEEP_FILE_OPEN_DEFAULT`) back to `False`, so that the default behaviour is to drop file handles.

Yes, please.

matthew-brett:
Paul - yes - that's what I meant - keep the file open, and don't drop file handles, for `auto` and `indexed_gzip`.  The docstring from current master has:

```
    If ``'auto'``, and the optional ``indexed_gzip`` dependency is
    present, a single file handle is created and persisted. If
    ``indexed_gzip`` is not available, behaviour is the same as if
    ``keep_file_open is False``
```

... so I thought (keep file open, don't drop handles, if indexed_gzip present), was as close as possible to that.  Chris - did I get that wrong?

effigies:
Nope, you're right. I'm :+1: to merge. Any final issues?

matthew-brett:
Sorry - yes - go ahead - please do merge.

effigies:
Thanks, Matthew.

pauldmccarthy:
Thanks guys!

PR #615, opened 2018-03-30
RF: moving more cmdline things to the cmdline folder

chrispycheng:
also added .DS_Store to the .gitignore file

- [x] nib-nifti-dx
- [x] nib-dicomfs
- [x] parrec2nii
- [x] tests for all the cmdline tools to run happily with `--help` 
       - required a fix for PY3 in nib-dicomfs (which otherwise was not tested anyhow).
         It remains not tested on CI since would need `fuse` module which is not required per se


yarikoptic:
me still needs to figure out how to manage it to avoid freaking out crashing on `import fuse` during --with-doctest ...

effigies:
@yarikoptic @chrispycheng Can you merge master, to make sure any errors are resulting from your code?

yarikoptic:
@effigies done! and fixed up one test

yarikoptic:
any idea why I do not see travis report? ;-)

effigies:
Tests are passing. Will merge this PM, barring any outcry.

effigies:
I don't really know, but the pre-merge commit is passing.

yarikoptic:
as for coverage -- now, that the scripts bodies are inside the `nibabel`, it reflects coverage better.  Main testing is done though by executing them externally, so coverage does not cover those runs.  Eventually should refactor tests to just import and call within the same process to gain coverage.
I will introduce one more change to hopefully trigger the travis and appveyor, where it probably would fail due to that fuse/doctesting issue I have mentioned.

yarikoptic:
I wonder why appveyor no longer runs for this PR?  Is it running for the others?

yarikoptic:
@effigies I am afraid that if you merge it as is, then it might cause appveyor and whatknows failures down the road.  Better be safe than sorry ;-)

effigies:
Yeah, something's happened to AppVeyor, and it's affecting the release branch, too: https://ci.appveyor.com/project/nipy/nibabel

But I don't have any permissions there, so not much I can do about it. Do you have permissions? Or do we need to bother Matthew?

yarikoptic:
hm , https://help.appveyor.com/discussions/problems/11287-the-build-phase-is-set-to-msbuild-mode-default-but-no-visual-studio-project-or-solution-files-were-found#comment_44538892 suggests that we should revoke and readd somehow... but since it also happens to release branch, it is probably not me or @chrispycheng ... ?  let me try on a dummy new PR to see if somehow relates to the original author of the PR

yarikoptic:
I deauth and reauthed appveyor for myself and it started to run... so here itcomes!

effigies:
@yarikoptic I assume you saw AppVeyor is failing? Windows doesn't like `os.getuid()` found in `nib-dicomfs`.

effigies:
@yarikoptic My change resolved the test failure. If you want to make sure that you're okay with it, you can go ahead and merge.

yarikoptic:
@effigies thank you! I guess the battle field will show if we didn't miss anything ;-)

yarikoptic:
Doh, either merge it or I will do when get to the laptop. Silly Travis policy to not allow merging on the phone when any tests fails

effigies:
Sounds good. rel/2.3.0 is synced to the latest master. We may still see #614 in, but having enough time to test is more important. If we freeze on Monday 9am EDT, would that be enough time for you to test and get a release before OHBM?

PR #617, opened 2018-04-05
NF nib-diff

chrispycheng:
Initial commits are duplicated with PR: RF nib-ls move which has since been merged

Example:

```shell
(dev) Chriss-MacBook:20170901 chrischeng$ nib-diff -t spmT_0001.nii.gz bold_1slice.nii.gz
Field      spmT_0001.nii.gz                             bold_1slice.nii.gz                           
dim        [3, 121, 145, 121, 1, 1, 1, 1]               [4, 40, 20, 1, 121, 1, 1, 1]                 
datatype   16                                           4                                            
bitpix     32                                           16                                           
pixdim     [-1.0, 1.5, 1.5, 1.5, 1.0, 1.0, 1.0, 1.0]    [-1.0, 3.0999999, 3.75, 3.75, 2.5, 0.0, 0.0, 0.0]
cal_max    0.0                                          2623.0                                       
descrip    b'SPM{T_[1066.0]} - contrast 1: AA-GG'       b'FSL5.0'                                    
qform_code 2                                            1                                            
sform_code 2                                            1                                            
qoffset_x  90.0                                         60.44999694824219                            
qoffset_y  -126.0                                       -35.625                                      
qoffset_z  -72.0                                        0.0                                          
srow_x     [-1.5, 0.0, 0.0, 90.0]                       [-3.0999999, 0.0, 0.0, 60.449997]            
srow_y     [0.0, 1.5, 0.0, -126.0]                      [0.0, 3.75, 0.0, -35.625]                    
srow_z     [0.0, 0.0, 1.5, -72.0]                       [0.0, 0.0, 3.75, 0.0] 
```

TODOs:
- [x] robust tests

Future diff (to be done in separate PR)
- allow for numeric differences in the values ( of header fields, or data)


nibabel/cmdline/utils.py
```diff
@@ -10,7 +10,6 @@
 Helper utilities to be used in cmdline applications
 """
 
-
```
yarikoptic: avoid modifying files which are not pertinent to suggested new functionality

nibabel/tests/test_diff.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@given(st.data())
+def test_diff_values_float(data):
+    x = data.draw(st.just(0), label='x')
+    y = data.draw(st.floats(min_value = 1e8), label='y')
+    z = data.draw(st.floats(max_value = -1e8), label='z')
+
+    assert diff_values(x, x) is False
+    assert diff_values(x, y) == True
```
yarikoptic: no need for "== True" since `assert` would cast the value anyways into bool.  So you could just say `assert not something_to_be_expected_to_beFalse` and `assert something_to_be_expected_to_beTrue`

nibabel/tests/test_scripts.py
```diff
@@ -65,6 +65,58 @@ def check_nib_ls_example4d(opts=[], hdrs_str="", other_str=""):
...
 
+
+def check_nib_diff_examples(opts=[], hdrs_str="", other_str=""):
+    # test nib-diff script
+    fnames = [pjoin(DATA_PATH, f)
+              for f in ('example4d.nii.gz', 'standard.nii.gz')]
+    fnames2 = [pjoin(DATA_PATH, f)
+              for f in ('example4d.nii.gz', 'example4d.nii.gz')]
+    code, stdout, stderr = run_command(['nib-diff'] + fnames, check_code=False)
+    assert_equal(stdout, "Field      " + pjoin(DATA_PATH, 'example4d.nii.gz') + pjoin(DATA_PATH, 'standard.nii.gz')
```
yarikoptic: since paths might be super long, as I have emailed, what if you print only filenames if they differ?

chrispycheng:
Running nose tests locally works fine but because of a quirk of the way files are organized or whatever Travis CI build isn't completing. Otherwise the nib-diff function is completely functional. Will work on getting Travis CI to jive

nibotmi:
:umbrella: The latest upstream changes (presumably #561) made this pull request unmergeable. Please resolve the merge conflicts.

appveyor.yml
```diff
@@ -21,6 +21,7 @@ install:
 
   # Install the dependencies of the project.
   - pip install numpy scipy matplotlib nose h5py mock
+  - pip install hypothesis
   - pip install pydicom
```
effigies: Might as well combine all three of these lines. Let pip do the dependency resolution in one shot.

nibabel/tests/test_scripts.py
```diff
@@ -67,6 +68,98 @@ def check_nib_ls_example4d(opts=[], hdrs_str="", other_str=""):
...
+                                                                                         "                           0."
+                                                                                         "0                            "
+                                                                                         "              " + "\n" +
+                 "qoffset_z  -7.248798370361328                           0.0                                          "
+                 + "\n" + "srow_x     [-2.0, 6.7147157e-19, 9.0810245e-18, 117.8551][1.0, 0.0, 0.0, 0.0]               "
+                          "          " + "\n" + "srow_y     [-6.7147157e-19, 1.9737115, -0.35552824, -35.722942][0.0, 3"
+                                                ".0, 0.0, 0.0]                         " + "\n" +
+                 "srow_z     [8.2554809e-18, 0.32320762, 2.1710818, -7.2487984][0.0, 0.0, 2.0, 0.0]                    "
+                 "     " + "\n" + "DATA: These files are different.\nChecksum   b0abbc49                    "
+                 + "                 0a2576dd")
```
effigies: Could you clean up the white space on this? Also, string literals can be placed next to each other without `+`.

nibabel/tests/test_scripts.py
```diff
@@ -67,6 +68,98 @@ def check_nib_ls_example4d(opts=[], hdrs_str="", other_str=""):
...
+
+    code, stdout, stderr = run_command(['nib-diff'] + fnames2, check_code=False)
+    assert_equal(stdout, "DATA: These files are identical!")
+
+
+
```
effigies: One too many lines. (This might be a bad merge on my part.)

chrispycheng:
So the current issues with Travis CI and AppVeyor output are that for some reason they get a different output than my computer. I'll list them as follows:

- Byte output on my computer is printed as b'[insert contents here]' whereas on Travis CI and AppVeyor it outputs simply as [insert contents here]
- Scientific notation has 5 fewer digits of precision on Travis CI and AppVeyor compared to my computer
- Numbers in general have 4 fewer digits of precision on Travis CI and AppVeyor compared to my computer

So I don't really know how to address these issues but moving forward I'll be implementing a new test which takes advantage of my reorganization of the way I get diff headers. See below for the latest commit and what I did. It'll basically be testing that the correct headers were found for the differences between the two files. I imagine that another test will be needed to verify the accuracy of the values found for those headers, but I'm going to take this one step at a time.

chrispycheng:
Travis should be clear after this commit. AppVeyor has problems relating to its Python environments - some tests that weren't touched by this pull request started crapping out. Otherwise, aside from maybe a few more tests, hopefully this function is ready to be pulled into NiBabel.

nibabel/cmdline/tests/test_utils.py
```diff
@@ -5,13 +5,16 @@
...
+from nose.tools import assert_equal
 
+import nibabel as nib
+import numpy as np
 from nibabel.cmdline.utils import *
+from nibabel.cmdline.diff import get_headers_diff
+from os.path import (dirname, join as pjoin, abspath)
+
+
+DATA_PATH = abspath(pjoin(dirname(__file__), '../../tests/data'))
```
yarikoptic: there is already one defined, just use
```python
from nibabel.testing import data_path
```
note that in above '/' within the path might be *nix specific and might (?) not work on Windows.  That is why we use all the `path.join`

nibabel/cmdline/tests/test_utils.py
```diff
@@ -42,3 +45,60 @@ def get_test(self):
...
+    assert_equal(actual_difference["sform_code"], expected_difference["sform_code"])
+    assert_equal(actual_difference["quatern_b"], expected_difference["quatern_b"])
+    assert_equal(actual_difference["quatern_c"], expected_difference["quatern_c"])
+    assert_equal(actual_difference["quatern_d"], expected_difference["quatern_d"])
+    assert_equal(actual_difference["qoffset_x"], expected_difference["qoffset_x"])
+    assert_equal(actual_difference["qoffset_y"], expected_difference["qoffset_y"])
+    assert_equal(actual_difference["qoffset_z"], expected_difference["qoffset_z"])
+    np.testing.assert_array_equal(actual_difference["srow_x"], expected_difference["srow_x"])
+    np.testing.assert_array_equal(actual_difference["srow_y"], expected_difference["srow_y"])
+    np.testing.assert_array_equal(actual_difference["srow_z"], expected_difference["srow_z"])
```
yarikoptic: although I like when tests are "explicit" like this, I hate code duplication more ;-)  and [apparently](https://stackoverflow.com/a/28848365/1265472) (thanks google) we could just use  `np.testing.assert_array_equal(actual_difference, expected_difference)`

nibabel/tests/test_scripts.py
```diff
@@ -67,6 +67,14 @@ def check_nib_ls_example4d(opts=[], hdrs_str="", other_str=""):
     assert_equal(fname, stdout[:len(fname)])
     assert_re_in(expected_re, stdout[len(fname):])
 
+
+def check_nib_diff_examples():
+    fnames2 = [pjoin(DATA_PATH, f)
+              for f in ('example4d.nii.gz', 'example4d.nii.gz')]
+    code, stdout, stderr = run_command(['nib-diff'] + fnames2, check_code=False)
+    assert_equal(stdout, "These files are identical.")
+
```
yarikoptic: what about comparing different files now?  at least a basic test using those nibabel/tests/data/spmT_0001*.nii.gz ?

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,185 @@
...
+
+    if data_diff:
+        diff['DATA(md5)'] = data_diff
+
+    if diff:
+        print("These files are different.")
+        print("{:<15}".format('Field'), end="")
+
+        for f in files:
+            output = ""
```
raamana: Is `output` simply the file name or something more complicated? If not, using `os.path.basename` might be easier, readable and generic across OSes to obtain the filename?

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,185 @@
...
+                else:
+                    output += f[i]
+                i += 1
+
+            print("{:<55}".format(output), end="")
+
+        print()
+
+        for key, value in diff.items():
+            print("{:<15}".format(key), end="")
```
raamana: I would suggest capturing string lengths (15, 55) into variables, as hardcoding them is not a good idea, and having them in variables allows us to adapt for different environments. For example, for terminals with smaller widths etc..

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,185 @@
...
+
+            for item in value:
+                item_str = str(item)
+                # Value might start/end with some invisible spacing characters so we
+                # would "condition" it on both ends a bit
+                item_str = re.sub('^[ \t]+', '<', item_str)
```
raamana: how is this different from using `str.strip()`?

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,185 @@
...
+                i += 1
+
+            print("{:<55}".format(output), end="")
+
+        print()
+
+        for key, value in diff.items():
+            print("{:<15}".format(key), end="")
+
+            for item in value:
```
raamana: Another idea that might be interesting to consider is whether layout differences vertically, which can scale better with number of images. Although horizontal layout looks nice for 2/3 images, which can be the default, but for a 10 images, it might be easier print each differing field into a separate block of values from all images.

nibabel/cmdline/diff.py
yarikoptic: And eventually the table Formatter which was used by nib-ls could be used to make it all dynamic without hardcoring much... And then eventually pyout by @kyleam could be used for the most flexible tabulator ;-)
But we thought here to get it to some Bible state first and then improve upon

nibabel/cmdline/diff.py
yarikoptic: Place those indicators only if anything was replaced at any end

nibabel/cmdline/diff.py
yarikoptic: Indeed, but most frequent case is actually two files and many fields differing

nibabel/tests/test_scripts.py
chrispycheng: Don't you think that testing that get_headers_diff works is enough? My concern is that testing text output here has traditionally really messed up with Travis and AppVeyor because their Python environments for some reason print fewer significant figures than my computer does. What do you think?

yarikoptic: - in principle, you don't have to check output for 1-to-1 matching, you could check for the important segments to be there in the output (that fields you expect are listed, etc)
- ideally we should have output consistent across machines.
- as I've outlined in a separate email - if all the printing is moved to another function, then may be it should be tested in a dedicated to that function test. Here we just would need to check that script is indeed running and exiting as expected when we give the identical or different files, with only minimal testing of output (to verify that it is there)

chrispycheng: DONE

chrispycheng:
OK so a (-0.004%) decrease in coverage is probably not going to mean the end of the world. nib-diff in its current stage is at a good starting point for functionality in NiBabel and I think is ready for merge. @yarikoptic  Moving forwards, there are several functionalities to implement that could improve it:

- Layout across different terminals
- TODOs listed in the code itself
- Table formatting

... and more!

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,201 @@
...
+        values = [header.get(field) for header in file_headers]  # get corresponding value
+
+        # if these values are different, store them in a dictionary
+        if are_values_different(*values):
+            difference[field] = values
+
+    return difference
+
+
+def get_data_md5sums(files):
```
yarikoptic: I know that you would hate continued review @chrispycheng but, by seeing the function name I got confused why below it returns an empty list if there is only one unique value.  So, please 
- rename to e.g. `get_data_diff`
- add a docstring describing the output like you nicely did for the above `get_header_diff`

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,201 @@
...
+        except TypeError:
+            pass
+
+        if type(value0) != type(value):  # if types are different, then we consider them different
+            return True
+        elif isinstance(value0, np.ndarray):
+            if np.any(value0 != value):  # special test for ndarray
+                return True
+            else:
+                return False
```
yarikoptic: you can just `return np.any(...)` here. no need for all the return True/False dance since `np.any` returns a single bool here

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,201 @@
...
+
+def get_data_md5sums(files):
+
+    md5sums = [
+        hashlib.md5(np.ascontiguousarray(nib.load(f).get_data(), dtype=np.float32)).hexdigest()
+        for f in files
+    ]
+
+    if len(set(md5sums)) == 1:
+        return []
```
yarikoptic: that might be one contributor to your .000?% coverage miss (do you have codecov extension to the browser installed to see what lines aren't covered?).  Apparently there is no test which verifies that you do get empty list in output whenever two files have the same data? you could make a dedicated test for this function and feed it

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if nibabel.cmdline.utils.verbose_level < 3:
+        # suppress nibabel format-compliance warnings
+        nib.imageglobals.logger.level = 50
+
+    assert len(files) >= 2, "Please enter at least two files"
+
+    file_headers = [nib.load(f).header for f in files]
+
+    if opts:  # will almost always have a header field
```
yarikoptic: I think you don't need this `if opts` (so no `else` is needed either) since they always will be there!

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,201 @@
...
+        # signals "all fields"
+        if opts.header_fields == 'all':
+            # TODO: header fields might vary across file types, thus prior sensing would be needed
+            header_fields = file_headers[0].keys()
+        else:
+            header_fields = opts.header_fields.split(',')
```
yarikoptic: apparently has no test case to test this!

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if data_diff:
+        diff['DATA(md5)'] = data_diff
+
+    if diff:
+        out.write(display_diff(files, diff))
+        raise SystemExit(1)
+
+    else:
+        out.write("These files are identical.\n")
```
yarikoptic: and again no test to test this?

nibabel/cmdline/diff.py
chrispycheng: How do you test an intermediary if/else statement within a function?

chrispycheng:
I'd like to state for the record that the AppVeyor fail was because of some other file! It wasn't me!

nibabel/cmdline/diff.py
chrispycheng: In my latest commit I added the simple/integration test. I'm not sure how I would go about the advanced logic test though?

nibabel/cmdline/diff.py
chrispycheng: So that would be the command parser itself? I don't think any other NiBabel function has a test for that, maybe that could be a separate PR?

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,207 @@
...
+        Returns
+        -------
+        list
+          np.array: md5 values of respective files
+        """
+
+    md5sums = [
+        hashlib.md5(np.ascontiguousarray(nib.load(f).get_data(), dtype=np.float32)).hexdigest()
+        for f in files
+    ]
```
effigies: Is there a reason you're using MD5 and not something more collision-resistant such as SHA256?

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,207 @@
...
+
+    diff = get_headers_diff(file_headers, header_fields)
+    data_diff = get_data_diff(files)
+
+    if data_diff:
+        diff['DATA(md5)'] = data_diff
+
+    if diff:
+        out.write(display_diff(files, diff))
+        raise SystemExit(1)
```
effigies: Is this now preferred to `sys.exit` (or just returning a return code and having the entry point exit with that code)?

nibabel/cmdline/diff.py
```diff
@@ -0,0 +1,298 @@
...
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""
+Quick summary of the differences among a set of neuroimaging files
+"""
+from __future__ import division, print_function, absolute_import
+
+import re
+import sys
+from collections import OrderedDict
+from optparse import OptionParser, Option
```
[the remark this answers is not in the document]
yarikoptic: yeah... we should convert all the cmdline tools which still use optparse in some one PR ;)

nibabel/cmdline/diff.py
yarikoptic: I guess it is where python violated its own Zen:
```ipython
In [4]: import this
The Zen of Python, by Tim Peters
...
There should be one-- and preferably only one --obvious way to do it.
...

In [2]: sys.exit?
Docstring:
exit([status])

Exit the interpreter by raising SystemExit(status).
```
so seems to be exactly the same thing...  but it is easier to explain that "we will raise exception and then test in the test that it was raised" ;)

effigies:
Okay. Since nothing's likely to depend on two separate versions of `nib-diff`, if we ever decide to use a stronger hash, that shouldn't cause problems.

:+1: to merge.

yarikoptic:
thanks! Let's merge then!
FWIW -- checksum there was just a proxy to visualize which files differ in data -- since nib-diff could consume more than 2 files, there is no easy way (besides adding a full matrix of pairs of files) to show which are actually different

PR #618, opened 2018-04-11
BF: be resilient to optional module non-ImportError exceptions upon import

yarikoptic:
ATM h5py is giving a grief while trying to work on a filesystem without
utf-8 encoding support in filenames.

### DIFF
--- a/nibabel/tests/test_optpkg.py
+++ b/nibabel/tests/test_optpkg.py
@@ -1,6 +1,7 @@
 """ Testing optpkg module
 """
 
+import mock
 import types
 import sys
 from distutils.version import LooseVersion
@@ -36,6 +37,12 @@ def test_basic():
     assert_good('os.path')
     # We never have package _not_a_package
     assert_bad('_not_a_package')
+    def raise_Exception(*args, **kwargs):
+        raise Exception(
+            "non ImportError could be thrown by some malfunctioning module "
+            "upon import, and optional_package should catch it too")
+    with mock.patch('__builtin__.__import__', side_effect=raise_Exception):
+        assert_bad('nottriedbefore')
 
 
 def test_versions():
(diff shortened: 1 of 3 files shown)


nibabel/tests/test_optpkg.py
```diff
@@ -36,6 +38,12 @@ def test_basic():
     assert_good('os.path')
     # We never have package _not_a_package
     assert_bad('_not_a_package')
+    def raise_Exception(*args, **kwargs):
+        raise Exception(
+            "non ImportError could be thrown by some malfunctioning module "
+            "upon import, and optional_package should catch it too")
+    with mock.patch.object(builtins, '__import__', side_effect=raise_Exception):
```
effigies: In Python 3, `builtins is __builtin__`, and in Python 2 you're setting `builtins = __builtin__`. Could you just import `__builtin__` and patch that here? I'm not sure what we're getting by adding this to `py3k`.

yarikoptic: - `In Python 3, builtins is __builtin__` .  I am confused with what you mean ;):
```shell
$> python3 -c 'import builtins'      
$> python3 -c 'import __builtin__'
ModuleNotFoundError: No module named '__builtin__'
```
- `py3k` is a module to establish the compatibility layer and feels intended to do so that code using it stays "python3" like.  With this change then any code which needs to use `builtins` module in either of pythons would do so with python3 naming.  I didn't want to breed python 2-vs-3 code right here in the test

effigies: Ah, perhaps this is an Ipython thing.

```Python
In [1]: __builtin__
Out[1]: <module 'builtins' (built-in)>
```

But:

```Shell
$ python -c 'import builtins; print(builtins is __builtin__)'
NameError: name '__builtin__' is not defined
```

yarikoptic:
ok - merging since no disagreement ;-)

PR #621, opened 2018-04-19
Better handle pathological filenames

### LINKED ISSUE #620: Cannot work out file type of ".nii"
While this is a corner case, nibabel should properly load `.nii.` instead of raising an exception.
```
>>> import nibabel
>>> nibabel.load('.nii')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/nibabel/loadsave.py", line 49, in load
    filename)
nibabel.filebasedimages.ImageFileError: Cannot work out file type of ".nii"
```

DimitriPapadopoulos:
_os.path.splitext()_ does not provide the behavior we need to properly
recognize edge cases such as `.nii` as image files.

Fixes #620

### DIFF
--- a/nibabel/filename_parser.py
+++ b/nibabel/filename_parser.py
@@ -264,9 +264,14 @@ def splitext_addext(filename,
     for ext in addexts:
         if endswith(filename, ext):
             extpos = -len(ext)
-            addext = filename[extpos:]
-            filename = filename[:extpos]
+            filename, addext = filename[:extpos], filename[extpos:]
             break
     else:
         addext = ''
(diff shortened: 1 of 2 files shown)


yarikoptic:
Makes world a better place so merging (looks kosher on cursory review), Thanks!

lanranranM:
Hello, I'm using nib.load(file_name) to access a ".nii.gz" file, and the function raise ImageFileError and I found that nib.filebasedimages.FileBasedImage.path_maybe_image('.nii') still gives me a (False, None). How can I solve the problem? Thanks

PR #625, opened 2018-04-27
[WIP] ENH: Relax default threshold for negative w2 in fillpositive

faustomilletari:
I have added the condition `and not np.isclose(w2, 0, rtol=1e-05, atol=1e-05)` because I had some problems to load data as the w2 value was a very very very small negative value close to zero (precision error).

This fixed it for me. Does it make sense?

### DIFF
--- a/nibabel/quaternions.py
+++ b/nibabel/quaternions.py
@@ -96,7 +96,8 @@ def fillpositive(xyz, w2_thresh=None):
     # Calculate w
     w2 = 1.0 - np.dot(xyz, xyz)
     if w2 < 0:
-        if w2 < w2_thresh:
+        # if w2 is negative and is not very very close to zero
+        if w2 < w2_thresh and not np.isclose(w2, 0, rtol=1e-05, atol=1e-05):
             raise ValueError('w2 should be positive, but is %e' % w2)
         w = 0
     else:


nibabel/quaternions.py
```diff
@@ -96,7 +96,8 @@ def fillpositive(xyz, w2_thresh=None):
     # Calculate w
     w2 = 1.0 - np.dot(xyz, xyz)
     if w2 < 0:
-        if w2 < w2_thresh:
+        # if w2 is negative and is not very very close to zero
+        if w2 < w2_thresh and not np.isclose(w2, 0, rtol=1e-05, atol=1e-05):
```
effigies: I think this is equivalent to setting `w2_thresh = 1e-05`.

faustomilletari:
The thing is that it could be also a very small NEGATIVE number... 

Fausto Milletarì

effigies:
Ah, sorry. I meant -1e-05.

faustomilletari:
Yeah,

You are totally right... 

So it's a problem of how small this constant should be in order to work with any data. I never had this issue before, today it has happened to me so I tried a quick fix. Do you think there is a way to fix this once and for all?


Fausto

effigies:
More, it's a problem of how small it should be to distinguish between a legitimate (if error-inducing) negative number and a rounding error. This is beyond my experience to determine, and clearly the original author thought that it ought to be on the order of standard epsilon.

Is there a possibility that your file was incorrectly written, and could be regenerated with a better qform?

In the short term, it looks like you can deal with this (albeit somewhat awkwardly) by modifying the quaternion threshold in the `Nifti1Header`:

```Python
import nibabel as nb
img = nb.load(fname)
img.header.quaternion_threshold = -1e-05
...
```

To do this more generally:

```Python
import nibabel as nb
nb.Nifti1Header.quaternion_threshold = -1e-05
...
```

If you're willing to accept the loss of precision, you can update your qform as follows:

```Python
import nibabel as nb
img = nb.load(in_fname)
img.header.quaternion_threshold = -1e-05
img2 = nb.Nifti1Image(img.get_data(), img.affine, img.header)
img2.set_qform(img.get_qform(), img.header._structarr['qform_code'])
img2.set_sform(img.get_sform(), img.header._structarr['sform_code'])
img2.to_filename(out_fname)
```

@matthew-brett Looks like you introduced this allowance in cfa87993d5bd9b58d1f59e5d9dd0e37a8833e5cc. What do you think about relaxing the threshold?

faustomilletari:
The file I have tried to read comes from a public dataset. This is the miccai brain atlas segmentation challenge data. I don't know which of the volumes caused that, but it's the only occurrence I even had of something like this. 

Thanks for your help!


Fausto

effigies:
@faustomilletari I'm disinclined at this point to relax this threshold as it seems to produce a fairly significant change in affine. Please reopen if you want to argue this...

PR #635, opened 2018-06-01
bu - just a TEST

yarikoptic:


yarikoptic:
hm, no appveyor for my own PR as well

PR #637, opened 2018-06-05
MRG: use get_fdata in docs

matthew-brett:
We will soonish deprecate get_data, so remove from the docs now, so as
not to trip up new users.

### DIFF
--- a/doc/source/nibabel_images.rst
+++ b/doc/source/nibabel_images.rst
@@ -220,21 +220,38 @@ False
 Getting the image data the easy way
 ===================================
 
-For either type of image (array or proxy) you can always get the data with
-the :meth:`get_data() <nibabel.spatialimages.SpatialImage.get_data>` method.
+For either type of image (array or proxy) you can always get the data with the
+:meth:`get_fdata() <nibabel.spatialimages.SpatialImage.get_fdata>` method.
 
-For the array image, ``get_data()`` just returns the data array:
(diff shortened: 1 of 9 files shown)


doc/source/nibabel_images.rst
```diff
@@ -220,21 +220,38 @@ False
...
+
+>>> farray_img = nib.Nifti1Image(image_data.astype(np.float64), affine)
+>>> farray_data = farray_img.get_fdata()
+>>> farray_data.dtype == np.dtype(np.float64)
+True
+
+There was no cast, so the array returned is exactly the array attached to the image:
+
+>>> image_data is array_img.dataobj
+False
```
effigies: I might be misreading, but from the above sentence, I'd think this ought to be:

```Python
>>> farray_data is farray_img.dataobj
True
```

matthew-brett: Good spot - you're right, but then this test fails, revealing the bug I think I have fixed in https://github.com/nipy/nibabel/pull/638

matthew-brett:
This PR needs to be merged after https://github.com/nipy/nibabel/pull/638 (with fixes suggested above).

effigies:
These stochastic AppVeyor failures seem to be cropping up again. Are you okay if I just go ahead and merge?

matthew-brett:
Yes, go for it ...

Issue #492, opened 2016-10-25
Images are deleted after Bus error


fepegar:
I have written a small function in IPython to delete the origin information from an image:
`In [1]: def removeOrigin(imPath):`
`     ...:     nii=nib.load(imPath)`
`     ...:     affine = nii.affine`
`     ...:     affine[:3, 3] = 0`
`     ...:     data = nii.get_data()`
`     ...:     new = nib.Nifti1Pair(data, affine)`
`     ...:     nib.save(new, imPath)`

My images are in Nifti1Pair format and very small (500 kB - 1 MB).

When I call my function, IPython crashes, the Terminal says `Bus error` and something very bad happens: my image data is gone!! The .hdr/.imgs file contain now 0 bytes, so I just lost my data.

I've tried outside the function and it happens the same when I try to overwrite the original image, i.e.:
`nib.save(new, imPath)` -> Bus error
`nib.save(new, imPath.replace('.hdr', '_2.hdr'))` -> Ok

effigies:
Hi, sorry about the delay.

My guess is that the image is an `mmap`, so that when you begin writing, you''re truncating the file, and thus losing the data you want to write.

You could try `data = nii.get_data().copy()` to make sure the data is loaded into memory.

fepegar:
Better late than never.

I don't have the technical background to fully understand your solution, but I'm sure it works. However, shouldn't nibabel be able to prevent this from hapenning?

effigies:
Perhaps. Certainly worth a discussion. I do believe (though I haven't tested) that #466 or something similar would resolve the problem by unlinking the input file before writing to that name.

It should also be possible to verify that the target filename is the same as the source `mmap`, but there  might be weirdness there I'm not immediately seeing.

As an aside, I would say that, for reproducibility and documentation reasons, I'd be very cautious about overwriting files. In this case, performing your action twice is the same as performing it once, so at least there's no danger to forgetting you've already done it. But in general, I'd strongly recommend setting up your workflows such that there is a clear flow of `(input file, operation, output file)`.

fepegar:
Sure, one has to be cautious when overwriting. But if everything I want is, say, change the `origin` of my 2GB nifti, maybe I wouldn't want two copies of my file...

effigies:
Fair enough. Well, in the short term, I'd say use `nib.load(fname, mmap=False)`, which won't require a manual copy command.

fepegar:
Duly noted, thanks!

[...]

PR #638, opened 2018-06-07
MRG: fix bug in get_fdata return value

matthew-brett:
We document that get_fdata returns the contained array, from an array image, if
it is of the desired type, but in fact we were always making a copy.

This revealed from updating the doctests in the documentation.

Refactor the image API tests to be a bit more readable (a tiny bit), fix the
bug, and add tests.

### DIFF
--- a/nibabel/dataobj_images.py
+++ b/nibabel/dataobj_images.py
@@ -344,7 +344,7 @@ class DataobjImage(FileBasedImage):
         if self._fdata_cache is not None:
             if self._fdata_cache.dtype.type == dtype.type:
                 return self._fdata_cache
-        data = np.asanyarray(self._dataobj).astype(dtype)
+        data = np.asanyarray(self._dataobj).astype(dtype, copy=False)
         if caching == 'fill':
             self._fdata_cache = data
         return data
--- a/nibabel/tests/test_image_api.py
+++ b/nibabel/tests/test_image_api.py
@@ -324,12 +324,6 @@ class DataInterfaceMixin(GetSetDtypeMixin):
     def _check_array_interface(self, imaker, meth_name):
         for caching in (None, 'fill', 'unchanged'):
             self._check_array_caching(imaker, meth_name, caching)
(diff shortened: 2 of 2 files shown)


effigies:
Looks like Windows+3.4 in particular doesn't like this. From a quick glance, it's unclear why these issues are raising here and not in other cases.

effigies:
Also, as a point of information, you've previously said that `get_fdata()` should always do a copy: https://github.com/nipy/nibabel/issues/492#issuecomment-383884386

matthew-brett:
Yeah - that Appveyor error is very odd.  Here's what I believe to be the same code, passing: https://ci.appveyor.com/project/matthew-brett/nibabel/build/1.0.17  . I couldn't replicate the test error on the Appveyor machines, attached via remote desktop.

For the always copy comment, what I think I meant was, it always copies if it's loaded from disk.   The previous tests seemed to imply that I had planned to return the internal array object only if it was already the given datatype - it's just that they weren't testing that, because they were not testing creating array images with float64 data.  If you think that's the wrong thing to do, we can revisit.

The returned array won't behave like a copy in all circumstances, because of the caching.  So, although always returning a copy might be more consistent, it does soak up a lot of extra memory, and the API is already not consistent with respect to copies, because of the caching.

effigies:
Oh, gotcha. That makes sense.

I'm rerunning the AppVeyor build to see if it does replicate. Maybe it was just a weird cache.

effigies:
It passed. Who knows what was up with that. I'm :+1: to merge.

effigies:
@yarikoptic I assume `get_fdata` isn't going to affect the tests you're currently running... Any objections to merging this and the doc fixes into `rel/2.3.0`?

effigies:
@matthew-brett This LGTM. Did you want to get this and #637 in before the 2.3.0 release?

matthew-brett:
Yes, please - I would love to get these in before the release.

PR #641, opened 2018-06-25
ENH: Define Cifti2 Axes describing the rows/columns of the Cifti2 data

MichielCottaar:
This pull request is designed to add semantic understanding of the Cifti2 file format to nibabel. This is essentially a port from https://github.com/MichielCottaar/cifti as recently suggested to me by @demianw . 

In short, the Cifti2 file format allows for the storage off matrices (or tensors) with a header describing what each row/column means. This meaning can be assigning to every row/column either a voxel or vertex (BrainModel), a group of voxels and/or vertices (Parcel), simply assigning a name (Scalar), a name and a color map (Label), or a starting point and step (Series). All these different options are stored in the same XML structure and hence have the same class representation in the current Cifti2Header file.

This pull request defines a new cifti2_axes.Axis class, which is a superclass of all the options defined above (BrainModel, Parcel, Scalar, Label, and Series). These new class all support:
1. they can be read from Cifti2 header
2. they can be used to write a Cifti2 header
3. they can be easily generated from scratch (either using the __init__ method or using some from_* class methods)
4. they can be easily introspected
5. they support operations like slicing and can be concatenated  through the '+' operator

There is quite a lot going on here. I would welcome any feedback on the current feature set, the API, and whether people think it is a good idea to include this in nibabel at all.


demianw:
I think we can also merge this [utils.py](https://github.com/demianw/nibabel/blob/cifti2-tools/nibabel/cifti2/utils.py) from my branch which allows for a better manipulation of indices. What do you think?

MichielCottaar:
Thanks, Demian. Your MatrixIndexMap seems quite similar to my Axis class, although rather than using sub-classing to support the various Cifti2 axes types you handle the various types through an if-statement in the __init__. So I would view it as an alternative, simpler implementation for if people think that my implementation is too complex.

demianw:
You are right! I didn't look in detail your code yet. I will go over the interfaces more than the implementation and check their simplicity. You code looks very nice at first sight!

effigies:
Hi @MichielCottaar @demianw, what's the status of this? I'm inclined to try to release soon, so if this isn't too far away, I can put it in the queue.

demianw:
Hey! I meant to do a code review months ago and I was buried in work. I expect to be able to review by nov 1st the latest. Does that sound good?

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1224 @@
...
+
+
+class Axis(object):
+    """
+    Generic object describing the rows or columns of a CIFTI vector/matrix
+
+    Attributes
+    ----------
+    arr : np.ndarray
+        (N, ) typed array with the actual information on each row/column
```
demianw: @MichielCottaar  This documentation is a bit unclear. This corresponds more to an abstract class than to a generic object, which, at least to me, makes me think of Generic as in type-parameterised objects.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1224 @@
...
+        """
+        Creates a new BrainModel axis defining the vertices and voxels represented by each row/column
+
+        Parameters
+        ----------
+        arr : np.ndarray
+            (N, ) structured array with for every element a tuple with 3 elements:
+            - vertex index (-1 for voxels)
+            - 3 voxel indices (-1 for vertices)
+            - string (name of brain structure)
```
demianw: @MichielCottaar I think that passing the voxel/vertex/structure elements separated as in the attributes presented on line 130 might be an easier to understand interface. The typed arrays with `-1` harcoded for an ignored value seems suboptimal as interface design. @effigies , you have more the hand on what's the interface style in nibabel, what do you think?

Alternatively, this constructor could be hidden and just creating the objects from the factory methods, such as `from_mask`, which have very comfy and clear interfaces, might be a good option.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1224 @@
...
+        Creates a new BrainModel axis defining the vertices and voxels represented by each row/column
+
+        Parameters
+        ----------
+        arr : np.ndarray
+            (N, ) structured array with for every element a tuple with 3 elements:
+            - string (name of parcel)
+            - (M, 3) int array with the M voxel indices in the parcel
+            - Dict[String -> (K, ) int array] mapping surface brain structure names to vertex indices
+        affine : np.ndarray
```
demianw: same general comment as with the BrainModel class. I thin that @effigies should advise us on wether there's a preferred type of representation for axis elements than the typed array.

MichielCottaar:
I just rebased onto master and adjusted the documentation according to your first suggestion.

I agree that the typed array is not the most natural interface. The main reason for doing this was that I was able to move some of the logic to the Axis class. However, in practice I ended up overriding most of these methods anyway. So perhaps I should just use the Axis class as a convenient class for type checking and to define the interface. I could even make it an abstract base class (https://docs.python.org/3.7/library/abc.html) to make this explicit.

demianw:
Great reactivity!
I think that the Axis-classes could just hide the constructor and work through the factory functions which seem quite easy to use.  This just needs updating the documentation of the class to indicate the *recommended* way of building them and leave the internal representation hidden.

Agreed that if finally you don't use anything of the Axis class, the ABC scenario is good design.

> I just rebased onto master and adjusted the documentation according to your first suggestion.
> 
> I agree that the typed array is not the most natural interface. The main reason for doing this was that I was able to move some of the logic to the Axis class. However, in practice I ended up overriding most of these methods anyway. So perhaps I should just use the Axis class as a convenient class for type checking and to define the interface. I could even make it an abstract base class (https://docs.python.org/3.7/library/abc.html) to make this explicit.

MichielCottaar:
I have removed the underlying typed arrays as I agree that they were mainly confusing. The constructors are now more useful. Next step is to check why the style check in travis is failing.

effigies:
@MichielCottaar Might be worth merging master, to make sure it's not something unrelated that we fixed earlier.

nibabel/cifti2/cifti2_axes.py
MichielCottaar: The typed array has been removed and the constructors updated

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+        Parameters
+        ----------
+        mim : cifti2.Cifti2MatrixIndicesMap
+
+        Returns
+        -------
+        BrainModel
+        """
+        nbm = np.sum([bm.index_count for bm in mim.brain_models])
+        voxel = -np.ones((nbm, 3))
```
effigies: Given that these are going to be cast to `int`, later:

```suggestion
        voxel = np.full((nbm, 3), fill_value=-1, dtype=int)
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+        vertex = -np.ones(nbm)
```
effigies:
```suggestion
        vertex = np.full(nbm, fill_value=-1, dtype=int)
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+            name.extend([bm.brain_structure] * bm.index_count)
+            if is_surface:
+                vertex[bm.index_offset: index_end] = bm.vertex_indices
+                nvertices[bm.brain_structure] = bm.surface_number_of_vertices
+            else:
+                voxel[bm.index_offset: index_end, :] = bm.voxel_indices_ijk
+                if affine is None:
+                    shape = mim.volume.volume_dimensions
+                    affine = mim.volume.transformation_matrix_voxel_indices_ijk_to_xyz.matrix
+                else:
```
effigies: You're checking the same `mim.volume.volume_dimensions` and `mim.volume.transformation_matrix_voxel_indices_ijk_to_xyz.matrix` every loop through. I don't think these are going to change on iterations over `mim.brain_models` (unless the API is weirder than I think). You can probably just drop this `else` block altogether.

I'm assuming that the logic here is actually to avoid setting `affine` and `shape` unless there is a volume model in the `mim.brain_models`, and then using common volume information. If it's possible that there's differing volume information, I assume it will be within the `bm` object, not `mim`.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+                    mim.volume = cifti2.Cifti2Volume(self.volume_shape, affine)
+            cifti_bm = cifti2.Cifti2BrainModel(
+                    to_slice.start, len(bm),
+                    'CIFTI_MODEL_TYPE_SURFACE' if is_surface else 'CIFTI_MODEL_TYPE_VOXELS',
+                    name, nvertex, voxels, vertices
+            )
+            mim.append(cifti_bm)
+        return mim
+
+    def iter_structures(self, ):
```
effigies:
```suggestion
    def iter_structures(self):
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+            proposed_name = 'CIFTI_STRUCTURE_%s' % structure.upper()
+        else:
+            proposed_name = 'CIFTI_STRUCTURE_%s_%s' % (structure.upper(), orientation.upper())
+        if proposed_name not in cifti2.CIFTI_BRAIN_STRUCTURES:
+            raise ValueError('%s was interpreted as %s, which is not a valid CIFTI brain structure'
+                             % (name, proposed_name))
+        return proposed_name
+
+    @property
+    def is_surface(self, ):
```
effigies:
```suggestion
    def is_surface(self):
```

This pattern appears a lot. Is there a reason for it? Feel free to argue over it, but if you're okay with removing the `, ` from all such instances, I'd prefer it.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+        Affine of the volumetric image in which the greyordinate voxels were defined
+        """
+        return self._affine
+
+    @affine.setter
+    def affine(self, value):
+        if value is not None:
+            value = np.asarray(value)
+            if value.shape != (4, 4):
+                raise ValueError('Affine transformation should be a 4x4 array')
```
effigies: There are a lot of validation checks like this that don't get exercised in the tests. Could you add them?

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+            return False
+        return (
+                ((self.affine is None and other.affine is None) or
+                 (abs(self.affine - other.affine).max() < 1e-8 and
+                  self.volume_shape == other.volume_shape)) and
+                (self.nvertices == other.nvertices) and
+                (self.name == other.name).all() and
+                (self.voxel[~self.is_surface] == other.voxel[~other.is_surface]).all() and
+                (self.vertex[~self.is_surface] == other.vertex[~other.is_surface]).all()
+        )
```
effigies: This is a tough one to do line-by-line suggestions, so I'll do it old-style:

```Python
        return (
            (self.affine is None or
             np.allclose(self.affine, other.affine) and
             self.volume_shape == other.volume_shape) and
            self.nvertices == other.nvertices and
            np.array_equal(self.name, other.name) and
            np.array_equal(self.voxel[~self.is_surface], other.voxel[~other.is_surface]) and
            np.array_equal(self.vertex[~self.is_surface], other.vertex[~other.is_surface])
        )
```

I dropped `other.affine is None`, because the `xor` case is already covered on L477. If you want to use a specific threshold for comparing affines, you can add the `atol` and `rtol` parameters to `np.allclose`.

Also, is `~obj.is_surface` the correct selector for `obj.vertex` as well as `obj.voxel`?

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+        Parameters
+        ----------
+        other : BrainModel
+            brain model to be appended to the current one
+
+        Returns
+        -------
+        BrainModel
+        """
+        if isinstance(other, BrainModel):
```
effigies: For guards like these that cover the whole body, I would suggest inverting and returning the error case, and then de-indenting the rest.

```suggestion
        if not isinstance(other, BrainModel):
            return NotImplemented
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+            if self.affine is None:
+                affine, shape = other.affine, other.volume_shape
+            else:
+                affine, shape = self.affine, self.volume_shape
+                if other.affine is not None and (
+                        (other.affine != affine).all() or
+                        other.volume_shape != shape
+                ):
+                    raise ValueError("Trying to concatenate two BrainModels defined " +
+                                     "in a different brain volume")
```
effigies: Can simplify these 10 lines:

```Python
affine, shape = other.affine, other.volume_shape
if self.affine is not None and (
        not np.allclose(affine, self.affine) or
        self.volume_shape != shape):
    raise ValueError(...)
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+            nvertices = dict(self.nvertices)
+            for name, value in other.nvertices.items():
+                if name in nvertices.keys() and nvertices[name] != value:
+                    raise ValueError("Trying to concatenate two BrainModels with inconsistent " +
+                                     "number of vertices for %s" % name)
+                nvertices[name] = value
+            return type(self)(
```
effigies: I think we generally use the `__class__` approach for methods like these:

```suggestion
            return self.__class__(
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+
+        Returns
+        -------
+        tuple with 3 elements
+        - boolean, which is True if it is a surface element
+        - vertex index if it is a surface element, otherwise array with 3 voxel indices
+        - structure.BrainStructure object describing the brain structure the element was taken from
+        """
+        is_surface = self.name[index] in self.nvertices.keys()
+        name = 'vertex' if is_surface else 'voxel'
```
effigies: Rather than using `getattr` below, I'd suggest using:

```suggestion
        struct = self.vertex if is_surface else self.voxel
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+        return is_surface, getattr(self, name)[index], self.name[index]
```
effigies: And then
```suggestion
        return is_surface, struct[index], self.name[index]
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1330 @@
...
+        - vertex index if it is a surface element, otherwise array with 3 voxel indices
+        - structure.BrainStructure object describing the brain structure the element was taken from
+
+        Otherwise returns a new BrainModel
+        """
+        if isinstance(item, integer_types):
+            return self.get_element(item)
+        if isinstance(item, string_types):
+            raise IndexError("Can not index an Axis with a string (except for Parcels)")
+        return type(self)(self.name[item], self.voxel[item], self.vertex[item],
```
effigies:
```suggestion
        return self.__class__(self.name[item], self.voxel[item], self.vertex[item],
```

nibabel/cifti2/cifti2_axes.py
MichielCottaar: If the other BrainModel has an affine or volume_shape of None (which is expected if it does not contain any voxels), this change will lead to an error. 

The existing code ensures two things:
1. If only one of the BrainModels has volumetric data (i.e. the affine is not set to None), the local variables affine and volume_shape will be set to that BrainModel.
2. If both BrainModels have volumetric data, they are checked for consistency of the affine and volume_shape.

nibabel/cifti2/cifti2_axes.py
MichielCottaar: I agree that there is no reason for the `else` block. I will get rid of it

nibabel/cifti2/cifti2_axes.py
effigies: Ah, right. Sorry about that.

effigies:
Ah. Looks like `numpy.full` didn't exist until numpy 1.8. I'm a little inclined to update the minimum version to 1.8. It's only 2 months younger than 1.7.1. I'll shoot you a quick PR.

MichielCottaar:
I could also go back to -np.ones if you prefer. The arrays in this case are only 1-dimensional, so there should be no memory problems (unless people have more than tens of millions of voxels/vertices in a BrainModel).

effigies:
You can if you want, but we've been sitting on 1.7.1 for 2 years. I think it's a slow enough walk up the version chain. I'm inclined to go ahead and increase the version, and if we somehow get somebody who can't upgrade past 1.7.1 complaining, the next patch release can revert and replace `full` with a compatibility function like:

```Python
def np_full(shape, fill_value, dtype=None, order='C'):
    if dtype is None:
        dtype = np.array(fill_value).dtype
    ret = np.empty(shape, dtype=dtype, order=order)
    ret.fill(fill_value)
    return ret
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1340 @@
...
+        return self._unit
+
+    @unit.setter
+    def unit(self, value):
+        if value.upper() not in ("SECOND", "HERTZ", "METER", "RADIAN"):
+            raise ValueError("Series unit should be one of " +
+                             "('second', 'hertz', 'meter', or 'radian'")
+        self._unit = value.upper()
+
+    def extend(self, other_axis):
```
effigies: The expectation with a method named `extend` is in-place extension, as in lists. If possible, I'd prefer not to violate these expectations. Could this implementation simply be renamed to `__add__`?

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1340 @@
...
+        if isinstance(item, slice):
+            step = 1 if item.step is None else item.step
+            idx_start = ((self.size - 1 if step < 0 else 0)
+                         if item.start is None else
+                         (item.start if item.start >= 0 else self.size + item.start))
+            idx_end = ((-1 if step < 0 else self.size)
+                       if item.stop is None else
+                       (item.stop if item.stop >= 0 else self.size + item.stop))
+            if idx_start > self.size:
+                idx_start = self.size - 1
```
effigies: In lists:

```Python
>>> a = list(range(5))
>>> a[6:]
[]
```

By the current logic, this should be equivalent to `a[4:]`, or `[4]`. So I think if we want consistency with lists:

```suggestion
                idx_start = self.size
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1340 @@
...
+        index : int
+            Indexes the row/column of interest
+
+        Returns
+        -------
+        float
+        """
+        if index < 0:
+            index = self.size + index
+        if index >= self.size:
```
effigies:
```suggestion
        if index >= self.size or index < 0:
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1340 @@
...
+            raise IndexError("index %i is out of range for Series with size %i" %
+                             (index, self.size))
```
effigies: If you're going to report the index, I'd suggest reporting the originally passed index, so that negative indices don't get transformed before being reported. Otherwise if you have a series with 9 elements and ask for -1, you'll get "index -2 is out of range for Series with size 9".

effigies:
Actually, if you just want to look through the [CodeCov report](https://codecov.io/gh/nipy/nibabel/pull/641/diff), that highlights what behavior is still untested.

demianw:
Hey @MichielCottaar  @effigies! This is great work. I was swamped this past two weeks but I can review this on Wednesday (or maybe sat morning)

effigies:
Thanks @demianw. If you do have time on Saturday, I'd appreciate it. I'm hoping to release on Monday, and this would be a nice addition. No worries if you can't. I'm okay with pushing off a few days.

nibabel/cifti2/cifti2_axes.py
MichielCottaar: These lines are meant to deal with the following situation:
>>> a = list(range(5))
>>> a[6::-1]
[4, 3, 2, 1, 0]

In that case the `idx_start` is `self.size - 1`. I will alter the if-statement to by `if idx_start > self.size and step < 0` to clarify this.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1340 @@
...
+        nvertices = {}
+        for idx_parcel, (parcel_name, bm) in enumerate(named_brain_models):
+            all_names.append(parcel_name)
+
+            voxels = bm.voxel[~bm.is_surface]
+            if voxels.shape[0] != 0:
+                if affine is None:
+                    affine = bm.affine
+                    volume_shape = bm.volume_shape
+                else:
```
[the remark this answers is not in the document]
MichielCottaar: Fixed

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1340 @@
...
+        label : np.ndarray
+            single dictionary or (N, ) object array with dictionaries mapping
+            from integers to (name, (R, G, B, A)), where name is a string and R, G, B, and A are
+            floats between 0 and 1 giving the colour and alpha (i.e., transparency)
+        meta :  np.ndarray
+            (N, ) object array with a dictionary of metadata for each row/column
+        """
+        self.name = np.asanyarray(name, dtype='U')
+        if isinstance(label, dict):
+            label = [label] * self.name.size
```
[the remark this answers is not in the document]
MichielCottaar: Fixed and added a test for it as well

nibabel/cifti2/cifti2_axes.py
MichielCottaar: Done

MichielCottaar:
I added @effigies comments, merged with master, and increased the test coverage (mainly by testing the __eq__ methods in much more detail). The tests will still fail as the numpy version is still 1.7.1 instead of 1.8...

effigies:
@MichielCottaar This looks great. Thanks so much for all your work, and I'll wait on @demianw's final review.

Apologies for the nit-pick, but could I ask you to format your docstrings according to [numpy conventions](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)? The main thing I see is things like `Tuple[cifti2_axes.Axis]`, which would translate to something like

```Rst
tuple of :class:`.cifti2_axes.Axis`
```

I know this has been a lot of back-and-forth, so if you like, we can merge it as is and I can do a big pass over the docs before the release.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1322 @@
...
+            return cls.from_surface(np.where(mask != 0)[0], mask.size, name=name)
+        elif mask.ndim == 3:
+            voxels = np.array(np.where(mask != 0)).T
+            return cls(name, voxel=voxels, affine=affine, volume_shape=mask.shape)
+        else:
+            raise ValueError("Mask should be either 1-dimensional (for surfaces) or "
+                             "3-dimensional (for volumes), not %i-dimensional" % mask.ndim)
+
+    @classmethod
+    def from_surface(cls, vertices, nvertex, name='Cortex'):
```
satra: if it's left as a default, perhaps we use `Unknown`. i don't think the library should make this choice.

effigies:
@satra I agree that the documentation sorely needs updating. The focus for a bit has been more getting support for the file formats in, and little thought has been given for supporting users outside the downstream projects that often contributed the code. It would be good to try to make a concerted push to update the documentation before the 3.0 release at latest; I'd be interested to hear any strategies people have for dividing the work and organizing the work.

@MichielCottaar If you have the spare cycles to write some documentation in the short term (possibly just porting over https://github.com/MichielCottaar/cifti/blob/master/README.md into our docs structure?), I would be happy to push off release for a week (or two at most). In my experience, the post-release period has a lull where a lot of good intentions can fade, so if we can harness some momentum now, it's worth the delay.

nibabel/cifti2/cifti2_axes.py
MichielCottaar: I've set it to Other (which is the closest equivalant to Unknown allowed by the CIFTI spec)

MichielCottaar:
I should have time to port the existing README file and update the language/import statements. Would this involve creating a new rst file in doc/source explaining the module and add it to manual.rst (so that is shows up in https://nipy.org/nibabel/manual.html#manual)? Or would it make more sense to add it to the cifti2/__init__.py file (so that it shows up on the https://nipy.org/nibabel/reference/nibabel.cifti2.html#module-nibabel.cifti2)?

Also, appveyor seems to throw some errors in the Minc2 codebase (potentially related to the update to numpy 1.8). Could somebody look into that?

demianw:
Great! I’ll review this tomorrow

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1430 @@
...
+        Returns
+        -------
+        tuple with 2 elements
+        - unicode name of the row/column
+        - dictionary with the element metadata
+        """
+        return self.name[index], self.meta[index]
+
+
+class Label(Axis):
```
demianw: `LabelAxis` might be a slightly more clear name. `Label` might be confusing even if it's within a `cifti2_axis` module. What do you think @effigies ?

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1430 @@
...
+            meta = [{} for _ in range(self.name.size)]
+        self.meta = np.asanyarray(meta, dtype='object')
+
+        for check_name in ('name', 'meta', 'label'):
+            if getattr(self, check_name).shape != (self.size, ):
+                raise ValueError("Input {} has incorrect shape ({}) for Label axis".format(
+                        check_name, getattr(self, check_name).shape))
+
+    @classmethod
+    def from_mapping(cls, mim):
```
demianw: I'd rather use `index_mapping` instead of `mapping`. This is because `mapping` makes me think of the `Mapping` collection and I try to use it with a `dict` object or similar.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1430 @@
...
+    Parameters
+    ----------
+    axes : iterable of :py:class:`Axis` objects
+        one or more axes describing each dimension in turn
+
+    Returns
+    -------
+    cifti2.Cifti2Header
+    """
+    axes = list(axes)
```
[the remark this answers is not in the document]
MichielCottaar: I index it in a bit (axes[:dim] & axes.index(ax)) and the tee object does not work for that. I did make it a tuple as it does not need to be altered, which should be slightly faster

nibabel/cifti2/cifti2_axes.py
MichielCottaar: done

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1430 @@
...
+
+        Parameters
+        ----------
+        index : int
+            Indexes the row/column of interest
+
+        Returns
+        -------
+        tuple with 3 elements
+        - boolean, which is True if it is a surface element
```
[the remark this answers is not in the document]
MichielCottaar: Returning one of the 'CIFTI_MODEL_TYPE' makes sense to me. I think distinguishing these two strings should be easy and declaritive enough (e.g., if 'SURFACE' in model_type) without adding an additional function or method.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1430 @@
...
+            proposed_name = 'CIFTI_STRUCTURE_%s' % structure.upper()
+        else:
+            proposed_name = 'CIFTI_STRUCTURE_%s_%s' % (structure.upper(), orientation.upper())
+        if proposed_name not in cifti2.CIFTI_BRAIN_STRUCTURES:
+            raise ValueError('%s was interpreted as %s, which is not a valid CIFTI brain structure'
+                             % (name, proposed_name))
+        return proposed_name
+
+    @property
+    def is_surface(self):
```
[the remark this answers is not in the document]
MichielCottaar: If we are going to rename these, I would go with `surface_mask` and `voxel_mask` (or `volume_mask`), but I'm not really convinced that we need both as they are just the negation of each other.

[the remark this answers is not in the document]
demianw: I think the cost of having both is negligible and it will enable more declarative programming

nibabel/cifti2/cifti2_axes.py
MichielCottaar: I would either add Axis to all of the names (`Scalar` could be unclear as well) or to none for consistency, however as we are always within the `cifti2` environment I think it should be clear enough. @effigies, more thoughts on this would be welcome.

demianw: I agree that it’s all or none. I think that often the module naming gets a bit lost  in python’s flexible import system. In nibabel we have naming like `Nifti2Image` when it could just be `Image` within a module `nitfi2` so this issue is considered elsewhere. I agree that the last call should come from @effigies

effigies: Setting aside the question of interpretability when imported as a bare name, which would you find a better description of the object in question: `cifti2.Label` or `cifti2.LabelAxis` (and so on for each `Axis` subclass)? We should consider (our estimate of) the intuition of someone roughly familiar with the CIFTI-2 format but not our data structures. Note we also have `cifti2.Cifti2Label` objects for the XML nodes.

From my perspective as one who's vaguely acquainted with CIFTI-2, and more via code reviews than actual manipulation of files, I'm a little inclined toward `(BrainModel|Parcels|Scalar|Label|Series)Axis`, but I don't feel very strongly about it.

You asked for a last call, so I'd say let's go with @demianw, but if you feel strongly the other way, @MichielCottaar, please feel free to argue further.

I'm going to go through hopefully a last round of review now.

nibabel/cifti2/__init__.py
```diff
@@ -6,14 +6,15 @@
 #   copyright and license terms.
 #
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
-"""CIfTI format IO
+"""CIfTI2 format IO
```
effigies: Silly pedantic thing, but CIFTI (like GIFTI) doesn't have a lowercase `f`. Would you mind doing a find-and-replace for this? (The `Cifti2` that refers to the prefix of, e.g. `Cifti2Image` or `Cifti2BrainModel` doesn't need updating. Just the format itself.)

```suggestion
"""CIFTI-2 format IO
```

nibabel/cifti2/cifti2.py
```diff
@@ -1268,6 +1265,40 @@ def get_index_map(self, index):
...
+        Generates the Cifti2 axis for a given dimension
+
+        Parameters
+        ----------
+        index : int
+            Dimension for which we want to obtain the mapping.
+
+        Returns
+        -------
+        axis : cifti2_axes.Axis
```
effigies:
```suggestion
        axis : :class:`.cifti2_axes.Axis`
```

nibabel/cifti2/cifti2.py
```diff
@@ -1268,6 +1265,40 @@ def get_index_map(self, index):
...
+        return cifti2_axes.from_index_mapping(self.matrix.get_index_map(index))
+
+    @classmethod
+    def from_axes(cls, axes):
+        '''
+        Creates a new Cifti2 header based on the Cifti2 axes
+
+        Parameters
+        ----------
+        axes : Tuple[cifti2_axes.Axis]
```
effigies:
```suggestion
        axes : tuple of :class`.cifti2_axes.Axis`
```

nibabel/cifti2/cifti2.py
```diff
@@ -1297,15 +1328,19 @@ def __init__(self,
             Object containing image data.  It should be some object that
             returns an array from ``np.asanyarray``.  It should have a
             ``shape`` attribute or property.
-        header : Cifti2Header instance
+        header : Cifti2Header instance or Sequence[cifti2_axes.Axis]
```
effigies:
```suggestion
        header : Cifti2Header instance or sequence of :class:`cifti2_axes.Axis`
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+A :class:`Parcels` axis can be created from a sequence of :class:`BrainModel` axes using
+:meth:`Parcels.from_brain_models`.
+
+Examples
+--------
+We can create brain models covering the left cortex and left thalamus using:
+
+>>> from nibabel import cifti2
+>>> bm_cortex = cifti2.BrainModel.from_mask(cortex_mask,
+...                                         brain_structure='cortex_left') # doctest: +SKIP
```
effigies: Are you able to generate these (or dummy) masks easily? Since we run doctests on docs like this, it would help to ensure that the docs are providing usable suggestions. If not, we can add this to the list of things to look into when we do a big sweep through the docs.

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+import abc
+
+
+def from_index_mapping(mim):
+    """
+    Parses the MatrixIndicesMap to find the appropriate CIfTI2 axis describing the rows or columns
+
+    Parameters
+    ----------
+    mim : cifti2.Cifti2MatrixIndicesMap
```
effigies:
```suggestion
    mim : :class:`.cifti2.Cifti2MatrixIndicesMap`
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+
+    Returns
+    -------
+    subtype of Axis
```
effigies:
```suggestion
    subclass of :class:`Axis`
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+    Converts the axes describing the rows/columns of a CIfTI2 vector/matrix to a Cifti2Header
+
+    Parameters
+    ----------
+    axes : iterable of :py:class:`Axis` objects
+        one or more axes describing each dimension in turn
+
+    Returns
+    -------
+    cifti2.Cifti2Header
```
effigies:
```suggestion
    :class:`.cifti2.Cifti2Header`
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+            return False
+        if xor(self.affine is None, other.affine is None):
+            return False
+        return (
+                (self.affine is None or
+                 np.allclose(self.affine, other.affine) and
+                 self.volume_shape == other.volume_shape) and
+                self.nvertices == other.nvertices and
+                np.array_equal(self.name, other.name) and
+                np.array_equal(self.voxel[~self.surface_mask], other.voxel[~other.surface_mask]) and
```
effigies: Now I'm looking at this, shouldn't the rest of the non-voxel entries be equal, as well:
```suggestion
                np.array_equal(self.voxel, other.voxel) and
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+                np.array_equal(self.vertex[self.surface_mask], other.vertex[other.surface_mask])
```
effigies: And here:

```suggestion
                np.array_equal(self.vertex, other.vertex)
```

nibabel/cifti2/cifti2_axes.py
```diff
@@ -0,0 +1,1441 @@
...
+        affine = None
+        volume_shape = None
+        all_names = []
+        all_voxels = np.zeros(nparcels, dtype='object')
+        all_vertices = np.zeros(nparcels, dtype='object')
+        nvertices = {}
+        for idx_parcel, (parcel_name, bm) in enumerate(named_brain_models):
+            all_names.append(parcel_name)
+
+            voxels = bm.voxel[~bm.surface_mask]
```
effigies: Might be more intuitive to switch from `~x.surface_mask` to `x.volume_mask`, now that it exists:

```suggestion
            voxels = bm.voxel[bm.volume_mask]
```

nibabel/cifti2/cifti2_axes.py
MichielCottaar: I created some small, fake masks. I guess a better example would load some atlas and select a single ROI to include in the BrainModelAxis, but I am not aware of one easily availble

nibabel/cifti2/cifti2_axes.py
MichielCottaar: I'm not sure, for greyordinates on the surface the voxel values are meaningful. So a BrainModel where they have been set to -1 (my default) or some other value are referreing to the same greyordinates. And similarly for volumetric greyordinates are the same if they have the same voxel indices, irrespetive of their vertex indices.

effigies: Okay.

MichielCottaar:
Any ideas about the appveyor errors, @effigies ? They only happened in python 3.4 for the 32-bit system and seem unrelated with the changes I made in the code.

effigies:
@MichielCottaar Those errors are stochastic but persistent. I'm currently not worrying about them. Thanks for this.

@demianw Any final comments?

demianw:
Nope! Great job @michielcottaar!

effigies:
Thanks for your patience.

MichielCottaar:
Thanks for all your great feedback.

demianw:
Yay!!!!!

satra:
awesome job @MichielCottaar .  this will help us move things forward. i asked @effigies if we can start doing some interactive tutorials for nibabel that would serve as both static tutorials but also something people can immediately start playing with online. we can use the nipype tutorials as an example: https://miykael.github.io/nipype_tutorial/

PR #644, opened 2018-07-02
ENH: Add SerializableImage class with to/from_bytes methods

### LINKED ISSUE #639: FileBasedImage.serialize()
Serializing an image to a `bytes` object via `BytesIO` has come up multiple times.

For reference, the following can be done (works for NIFTI; I'd have to check if it works for other images):

```Python
from io import BytesIO
import nibabel as nb

# Create an image `img`

bio = BytesIO()
file_map = {'image': nb.FileHolder(fileobj=bio), 'header': nb.FileHolder(fileobj=bio)}
img.to_file_map(file_map)
```

Would it be worth adding a `.serialize()` method (or similar) that returns a `bytes` serialization, rather than forcing users to go through this somewhat convoluted approach?

effigies:
Closes #639.

Implements to/from_bytes for single-image variants of NIfTI-1, NIfTI-2, GIFTI and CIFTI2 images, as well as MGH images.

Created an abstract class deriving from `FileBasedImage` that can be used in other images, if it becomes desirable, but didn't want to spend a ton of time nailing it down for image types I have an, at best, passing familiarity with.

### DIFF
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -511,3 +511,9 @@ class FileBasedImage(object):
         if sniff is None or len(sniff[0]) < klass._meta_sniff_len:
             return False, sniff
         return klass.header_class.may_contain_header(sniff[0]), sniff
+
+    def serialize(self):
+        bio = io.BytesIO()
+        file_map = self.make_file_map({'image': bio, 'header': bio})
+        self.to_file_map(file_map)
+        return bio.getvalue()
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -561,7 +561,10 @@ class SerializableImage(FileBasedImage):
 
     Images that consist of separate header and data files (e.g., Analyze
     images) currently do not support this interface.
+    For multi-file images, ``to_bytes()`` and ``from_bytes()`` must be
+    overridden, and any encoding details should be documented.
     '''
+
     @classmethod
     def from_bytes(klass, bytestring):
         """ Construct image from a byte string
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -513,6 +513,18 @@ class FileBasedImage(object):
         return klass.header_class.may_contain_header(sniff[0]), sniff
 
     def serialize(self):
+        """ Return a ``bytes`` object with the contents of the file that would
+        be written if the image were saved.
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        bytes
+            Serialized image
+        """
         bio = io.BytesIO()
         file_map = self.make_file_map({'image': bio, 'header': bio})
         self.to_file_map(file_map)
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -261,21 +261,6 @@ class FileBasedImage(object):
         return klass.from_file_map(file_map)
 
     @classmethod
-    def from_bytes(klass, bstring):
-        """ Construct image from a byte string
-
-        Class method
-
-        Parameters
-        ----------
-        bstring : bytes
-            Byte string containing the on-disk representation of an image
-        """
-        bio = io.BytesIO(bstring)
-        file_map = self.make_file_map({'image': bio, 'header': bio})
-        return klass.from_file_map(file_map)
-
-    @classmethod
     def from_file_map(klass, file_map):
         raise NotImplementedError
 
@@ -348,24 +333,6 @@ class FileBasedImage(object):
         '''
         self.file_map = self.filespec_to_file_map(filename)
         self.to_file_map()
-
-    def to_bytes(self):
-        """ Return a ``bytes`` object with the contents of the file that would
-        be written if the image were saved.
-
-        Parameters
-        ----------
-        None
-
-        Returns
-        -------
-        bytes
-            Serialized image
-        """
-        bio = io.BytesIO()
-        file_map = self.make_file_map({'image': bio, 'header': bio})
-        self.to_file_map(file_map)
-        return bio.getvalue()
 
     @deprecate_with_version('to_filespec method is deprecated.\n'
                             'Please use the "to_filename" method instead.',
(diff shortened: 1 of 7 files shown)


Shotgunosine:
to_bytes and from_bytes sounds good to me. Easier for a novice to interpret
as well.

On Tue, Jul 31, 2018 at 12:21 PM, Chris Markiewicz <notifications@github.com

effigies:
I think this is ready for a review, if anybody has some time. Pretty compact, in the end. I may not have used the testing machinery to its full potential, but hopefully it isn't too grotesque.

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,79 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+
+    methods:
+
+       * .to_bytes() - serialize image to byte string
+
+    classmethods:
+
+       * from_bytes(bytestring) - make instance by deserializing a byte string
+
+    The following properties should hold:
```
matthew-brett: How do you want this to look in the rendered doc?  What happens with these indented lists?

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,79 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+
+      * ``klass.from_bytes(bstr).to_bytes() == bstr``
```
matthew-brett: Is that true in general for round-tripping?  For example, what happens with nifti images with image scaling?  I think the scale factors may get recalculated.

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,79 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+      * if ``img = orig.__class__.from_bytes(orig.to_bytes())``, then
+        ``img.header == orig.header`` and ``img.get_data() == orig.get_data()``
+
+    Further, for images that are single files on disk, the following methods of loading
+    the image must be equivalent:
+
+        img = klass.from_filename(fname)
```
matthew-brett: Make this ``.. code:: python`` block?

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,79 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+            img = klass.from_bytes(fobj.read())
+
+    And the following methods of saving a file must be equivalent:
+
+        img.to_filename(fname)
+
+        with open(fname, 'wb') as fobj:
+            fobj.write(img.to_bytes())
+
+    Images that consist of separate header and data files will generally
```
matthew-brett: I'm not sure what this means - maybe give an example, of an Analyze image?

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,79 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+
+        Class method
+
+        Parameters
+        ----------
+        bstring : bytes
+            Byte string containing the on-disk representation of an image
+        """
+        bio = io.BytesIO(bytestring)
+        file_map = klass.make_file_map({'image': bio, 'header': bio})
```
matthew-brett: Ouch - this is a bit frightening.   I guess we now have to worry about the order in which the header and image get written, within the image classes?

nibabel/gifti/gifti.py
```diff
@@ -724,6 +724,9 @@ def to_xml(self, enc='utf-8'):
 <!DOCTYPE GIFTI SYSTEM "http://www.nitrc.org/frs/download.php/115/gifti.dtd">
 """ + xml.XmlSerializable.to_xml(self, enc)
 
+    # Avoid the indirection of going through to_file_map
+    to_bytes = to_xml
```
matthew-brett: Just checking - `to_xml` does return a byte string?

nibabel/tests/test_image_api.py
```diff
@@ -451,6 +452,73 @@ def validate_affine_deprecated(self, imaker, params):
...
+
+        Abstract classes may have undefined equality, in which case test for
+        same type
+        """
+        not_implemented = False
+        header_eq = True
+        try:
+            header_eq = header_a == header_b
+        except NotImplementedError:
+            header_eq = type(header_a) == type(header_b)
```
matthew-brett: Should we require a header equality check being defined?  This doesn't seem like a good test.

effigies: This is a hack to compare `FileBasedHeader`s. I can put in a specific carve-out for that type, instead.

nibabel/gifti/gifti.py
effigies: Yup. I checked.

nibabel/filebasedimages.py
effigies: Yes. At present this works for single-file images, where they are already passed a single file object in both fields. I would expect that, if we extend to multi-file images, it would be the responsibility of the implementor to ensure that the serialization is consistent and can be de-serialized.

I'm a little hesitant to pre-engineer for those cases, as I don't have the time or need to work out those problems. As long as we're not providing this function for multi-image files, I'm not sure it's a problem.

effigies: To be clear, if it looks like the decisions I've made now are going to actively make it harder to support those cases in the future than some other implementation, I'm happy to change it.

nibabel/filebasedimages.py
effigies: Fair point. I guess what we want is load-level idempotence, to the same level of precision as save/load cycles.:

```Python
img_a = klass.from_bytes(bstr)
img_b = klass.from_bytes(img_a.to_bytes())

np.allclose(img_a.get_fdata(), img_b.get_fdata())
```

Would the following be too strong, as well?

```Python
img_a.to_bytes() == img_b.to_bytes()
```

nibabel/filebasedimages.py
effigies: I actually haven't built the docs here. I'll fiddle with this.

effigies:
@matthew-brett  I've addressed (or responded) to your comments, if you want to have another look.

effigies:
I've updated this PR to do the following:

1) The mixin methods `to_bytes`/`from_bytes` now check whether the class's `files_types` attribute indicates a multi-file image, and raises a `NotImplementedError`. This can be overridden by a multi-image class, but ensures we don't accidentally make promises we can't keep.
2) `Nifti1Pair` was marked serializable, when it should have been `Nifti1Image`. This has been corrected.
3) The header equality check in the API tests is now more specifically targeted to the one corner case, i.e., the abstract `FileBasedHeader` class.

@matthew-brett I believe this has addressed all of your comments. Please let me know if you have any further concerns.

I would like to get this in before the final Python 2 feature release (2.5.0) if possible.

If anybody else has some spare cycles, I'd appreciate a review on this one.

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,89 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+
+    .. code:: python
+
+        img.to_filename(fname)
+
+        with open(fname, 'wb') as fobj:
+            fobj.write(img.to_bytes())
+
+    Images that consist of separate header and data files (e.g., Analyze
+    images) currently do not support this interface.
```
matthew-brett: Maybe say something about the fact that image types can override this class method to implement it, for multi-file images.

nibabel/filebasedimages.py
```diff
@@ -511,3 +512,89 @@ def path_maybe_image(klass, filename, sniff=None, sniff_max=1024):
...
+    '''
+    @classmethod
```
matthew-brett: PEP8 doesn't need a newline here?

nibabel/tests/test_volumeutils.py
```diff
@@ -700,7 +700,7 @@ def test_a2f_non_numeric():
     # Some versions of numpy can cast structured types to float, others not
     try:
         arr.astype(float)
-    except ValueError:
+    except (TypeError, ValueError):
```
matthew-brett: Why do you need this change in this PR?

effigies: Ah, nope. Rebased onto the wrong branch.

effigies:
Thanks for the review, Matthew.

PR #647, opened 2018-07-12
BF: deterministic order of slice_time deduction, warning if multiple match

yarikoptic:
Initially hit while trying to make dcmstack python3 compatible: https://github.com/moloney/dcmstack/pull/61
What I found that dcmstack test fails due to slice_order not matching original one (4) on some runs under Python 3.  But the issue is generic -- reliance on order of keys in dict, and ignoring possible multiple "matches".  It was triggered by dcmstack since its test image just has 2 slices so multiple slice orders would match.

I've looked into the logic in nibabel, and my proposed solution is just **a** solution:
- I think ideally `Recoder` class which stores those mappings starts to use OrderedDict to maintain order in which choices are specified, and then code should explore in that order. I was scared to touch `Recorder` class since it relies on `self.__dict__` assignments
- so I just sorted choices for the slice orders based on their expanded names (since probably original order is gone by then) and in reverse so sequentials go first
- I issue a warning now that multiple choices match.  I am not sure how good that is, i.e. may be the warning is not really "due" in some cases? but imho it is also bad to just choose "some" without a clear idea which one is the "correct one"

I would be interested to hear what others think about this issue, and how likely it affected anyone's real data

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -121,36 +121,40 @@ Sphinx 0.6 doesn't work properly.
   OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
   DAMAGE.
 
-Ordereddict
+OrderedSet
 -----------
 
-In ``nibabel/externals/ordereddict.py``
+In ``nibabel/externals/oset.py``
 
-Copied from: https://pypi.python.org/packages/source/o/ordereddict/ordereddict-1.1.tar.gz#md5=a0ed854ee442051b249bfad0f638bbec
+Copied from: https://files.pythonhosted.org/packages/d6/b1/a49498c699a3fda5d635cc1fa222ffc686ea3b5d04b84a3166c4cab0c57b/oset-0.1.3.tar.gz
 
 ::
 
-  Copyright (c) 2009 Raymond Hettinger
+    Copyright (c) 2009, Raymond Hettinger, and others All rights reserved.
 
-  Permission is hereby granted, free of charge, to any person
-  obtaining a copy of this software and associated documentation files
-  (the "Software"), to deal in the Software without restriction,
-  including without limitation the rights to use, copy, modify, merge,
-  publish, distribute, sublicense, and/or sell copies of the Software,
-  and to permit persons to whom the Software is furnished to do so,
-  subject to the following conditions:
+    Package structured based on the one developed to odict Copyright (c) 2010, BlueDynamics Alliance, Austria
 
-      The above copyright notice and this permission notice shall be
-      included in all copies or substantial portions of the Software.
+    - Redistributions of source code must retain the above copyright notice, this
+      list of conditions and the following disclaimer.
 
-      THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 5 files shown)


yarikoptic:
hm, appveyor failures on python 3.5 seems to not be related

matthew-brett:
Would you mind adding a test that triggers the failure?

yarikoptic:
added some basic test for guaranteeing that it is always the same, see 244c8cfa73c0dd10d32eb616657c4b0503b8efc3

yarikoptic:
@matthew-brett @effigies - but what do you think in general, either my solution would suffice or it should be more thorough, e.g. relying on the original order of those labels in the structure?  we better fix it once instead of making output inconsistent across versions if we keep fixing it up incrementally ;-)

effigies:
My inclination would be to set a canonical ordering, rather than depend on sorting by name, and insertion order is the most intuitive way to impose this order. This preference is due to the general move in Python to ordered dictionaries, as well as lexicographical order being an artificial (and potentially annoying) constraint.

Not a strongly-held opinion though. Happy to be argued out.

yarikoptic:
Thank you @effigies ! Do you see an easy generic way to make those Recorders to use ordered dict without duplicating storage etc?

effigies:
I see a couple options:

1) Change the default `map_maker` to `OrderedDict`:

https://github.com/nipy/nibabel/blob/2a127ccdba26f63079e870a1c18b99e2271b1bde/nibabel/volumeutils.py#L81

2) Change the specific `map_maker for `slice_order_codes`:

https://github.com/nipy/nibabel/blob/2a127ccdba26f63079e870a1c18b99e2271b1bde/nibabel/nifti1.py#L143-L150

Becomes:

```Python
slice_order_codes = Recoder((  # code, label
    (0, 'unknown'),
    (1, 'sequential increasing', 'seq inc'),
    (2, 'sequential decreasing', 'seq dec'),
    (3, 'alternating increasing', 'alt inc'),
    (4, 'alternating decreasing', 'alt dec'),
    (5, 'alternating increasing 2', 'alt inc 2'),
    (6, 'alternating decreasing 2', 'alt dec 2')),
    fields=('code', 'label'), map_maker=OrderedDict)
```

Not sure if this resolves your duplicated storage concern, as I'm not quite sure what that concern is. These seem like pretty small dictionaries, anyway.

yarikoptic:
Thank you @effigies ! went 1. route, had to add OrderedSet construct, seems to work and IMHO should be "better" ;-)  Hopefully noone relied on doctesting output to be a `set` ;-)

effigies:
Pre-release tests are seg-faulting, but otherwise this LGTM.

effigies:
May be getting bitten by xianyi/OpenBLAS#1641. May be worth reporting upstream, probably numpy, but I haven't dug any further yet.

effigies:
It seems that in numpy/numpy#11551 there were OpenBLAS-based build issues, but it was decided that this wasn't something that could be fixed in numpy, is that right @matthew-brett?

I'm inclined to go ahead and merge, and accept failing `--pre` tests for now.

effigies:
Okay, looks like the broken wheels were rebuilt. I'm :+1: for merge.

@matthew-brett, any further comments?

effigies:
Thanks @yarikoptic.

Issue #650, opened 2018-07-23
Test failing on Python 3.7


sanjayankur31:
```
======================================================================
FAIL: nibabel.tests.test_testing.test_clear_and_catch_warnings
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/fedora/ankursinha/nibabel-virt/lib64/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/fedora/ankursinha/nibabel/nibabel/tests/test_testing.py", line 89, in test_clear_and_catch_warnings
    assert_warn_len_equal(my_mod, 1)
  File "/home/fedora/ankursinha/nibabel/nibabel/tests/test_testing.py", line 72, in assert_warn_len_equal
    assert_equal(len(mod_warns), 2)  # including 'version'
AssertionError: 1 != 2

----------------------------------------------------------------------
```

Tried on a fresh python 3.7 virtual env too and got the same.

effigies:
Closed by #651.

PR #651, opened 2018-07-23
CI: Test on Python 3.7 on Travis and AppVeyor

effigies:
Hopefully will reproduce #650.


effigies:
There doesn't currently seem to be any way to test on Python 3.7 on Travis. The suggestions from travis-ci/travix-ci#9815 fail.

matthew-brett:
Maybe install pyenv ?  I think that has Python 3.7 available.

effigies:
It looks like pyenv is pre-installed, but the docs aren't clear on how to use that to install and use a specific version. Can do some digging, or is this something you're familiar enough with to do quickly>

grlee77:
You could also try specifying python as `3.7-dev`. That uses a beta version of 3.7, which may be good enough until 3.7 is available on Travis? That is what we have for now in PyWavelets/pywt#390.

effigies:
@grlee77 Looks like that was enough to reproduce the issue. Thanks.

effigies:
Okay, I've had a look at this, and I don't understand warning registries enough to be sure I'd be resolving it correctly. In 3.7, this code:

https://github.com/nipy/nibabel/blob/bb8c6fa79e5ac8430d25b66a70bd1175ed67d673/nibabel/tests/test_testing.py#L86-L88

Produces `my_mod.__warningregistry__ == {'version': 14}`. So the test is kind of correctly failing, in that our expectations are violated. But I don't know how to respond.

effigies:
The `warnings.simplefilter('ignore')` lines were preventing warnings from being added to registries. It appears that this is a change in 3.7, where the registry is not filled in these cases, though whether that's considered a fixed bug or changed behavior is unclear to me. In any event, it doesn't seem relevant to our tests, so I removed it.

effigies:
@matthew-brett Care to review?

effigies:
When this goes in, I think it would be good to cut a bug-fix release.

effigies:
If there's no objection, I'll merge Monday. It would be good to get 3.7 testing for all new PRs.

.travis.yml
```diff
@@ -22,11 +21,15 @@ env:
         - EXTRA_PIP_FLAGS="--find-links=$EXTRA_WHEELS"
         - PRE_PIP_FLAGS="--pre $EXTRA_PIP_FLAGS --find-links $PRE_WHEELS"
 python:
-    - 3.4
+    - 2.7
```
matthew-brett: Why the extra 2.7 test?

effigies: I think it looked like we were missing the `INSTALL_TYPE="setup"` from the `INSTALL_TYPE` block. But I guess all of them are hitting that, so I'll drop it.

matthew-brett:
Great - thanks - in it goes.

PR #652, opened 2018-07-24
FIX: Interpret empty string in SliceThickness DICOM field as missing (default to 1)

mattcieslak:
I was running into an issue where sometimes the value for SliceThickness was an empty string instead of None. I'm not sure if this is the right place to add the change, though.


effigies:
Hi, sorry for the slow response. Is there any chance of providing a test DICOM file that will hit this code path?

effigies:
Hi @mattcieslak, just wanted to see if you have a test file that exhibits this property? Synthetic or anonymized would both work fine.

mattcieslak:
I'll check to see if I can find this again, I think it might have been a non-image dicom that was causing the trouble.

effigies:
Hi @mattcieslak, sorry to be a nuisance, but I'm going to try to release next Monday, so it'd be great to get this in ASAP.

mattcieslak:
Found the file and am checking with the study owners what the sharing policy is. I'm thinking this might just be a corrupt dicom because the rest of the files from that series have valid numbers for slice thickness

effigies:
Awesome. Thanks for your patience on this. (Failing tests are the usual AppVeyor weirdness.)

PR #655, opened 2018-08-01
Speed up reading of CIFTI's: replace genfromtxt with loadtxt

MichielCottaar:
This is much faster (it reduced the running time for the cifti2 tests from 51 seconds to 15 seconds on my computer). The main disadvantage of loadtxt is that it can not handle missing values, but there should be no missing values.


effigies:
LGTM. Small change, so merging now.

PR #661, opened 2018-09-14
RF+ENH: nib-diff - allow to specify absolute and/or relative maximal diff to tolerate

yarikoptic:
So now it should be possible to get an idea on how much data in the given files differs:

    $> nib-diff --ma 0.000001 --mr .001 ./tests-run/output/./sub-1_T1w_5mm_noise_corrected.nii.gz /tmp/sub-1_T1w_5mm_noise_corrected.nii.gz
    These files are different.
    Field          1:sub-1_T1w_5mm_noise_corrected.nii.gz                   2:sub-1_T1w_5mm_noise_corrected.nii.gz
    DATA(md5)      65df09c06b236342eaf7e2fe57aabf55                       3c6e9069e6e054e714f2894419848df0
    DATA(diff 1:)  -                                                      abs: 7.6293945e-06, rel: 0.002224694

and could be applied to >2 files as well

TODOs
- [ ] seek feedback regarding naming of parameters (e.g. `data_max_abs_diff`) - they are kinda mouthful but wanted to be specific happen we add some later on dedicated to header etc
- [ ] seek feedback regarding output formatting
- [ ] seek feedback regarding "logic" - ATM specification of --ma would also affect estimation of relative difference.  I did it on purpose so we could easily avoid divisions in tiny numbers, which could reach high values.  But may be that is undesired?
- [x] add tests

attn @chrispycheng


chrispycheng:
For the diff function itself:

- FIX test_utils on line np.testing.assert_equal(main(test_names, StringIO()), expected_difference) which spits out:
ValueError: operands could not be broadcast together with shapes (4,5,7) (128,96,24,2) 
- FIX test_display_diff on line assert_equal(display_diff(bogus_names, dict_values), expected_output) which is different obviously because the output was changed
- FIX test_scripts which anticipates "Field" as opposed to the new header in the table output, line 75

chrispycheng:
1. Parameter name lengths aren't a problem - can't be more abbreviated than -ma, anyways. I reckon it'll be nice to be able to see what -ma stands for in the code, but I don't expect anyone to use those longer option names.
2. Output looks fine overall but maybe we want a bit of spacing in the table headers with file names between the colon and subsequent value. This way we can minimize confusion between the numbering of each file and each file name itself.

chrispycheng:
Coverage is handled and Travis passed and the only reason AppVeyor isn't cruising is because it's glitched out with some NiBabel stuff unrelated to this pull request.

nibabel/cmdline/diff.py
```diff
@@ -125,6 +140,75 @@ def get_data_diff(files):
...
+    max_abs: float, optional
+      Maximal absolute difference to tolerate.
+    max_rel: float, optional
+      Maximal relative (`abs(diff)/mean(diff)`) difference to tolerate.
+      If `max_abs` is specified, then those data points with lesser than that
+      absolute difference, are not considered for relative difference testing
+
+    Returns
+    -------
+    TODO
```
yarikoptic: Would you be so kind to ad address this one as well?

nibabel/cmdline/tests/test_utils.py
```diff
@@ -11,7 +11,7 @@
 import nibabel as nib
 import numpy as np
 from nibabel.cmdline.utils import *
-from nibabel.cmdline.diff import get_headers_diff, display_diff, main, get_data_diff
```
yarikoptic: One of us has managed to make this file executable, please undo:
capture _2018-09-27-20-06-42
If you like a challenge - undo by rewriting that original commit. Workflow:
- fix, commit
- `git rebase -i BADCOMMIT^` where you reposition fixing commit after the one to fix, and give it `s` status to squash them into one
- `git push -f` since now you rewritten a commit

nibabel/tests/test_scripts.py
```diff
@@ -72,11 +72,14 @@ def check_nib_diff_examples():
     fnames = [pjoin(DATA_PATH, f)
```
yarikoptic: The same here about permissions

nibabel/tests/test_scripts.py
```diff
@@ -72,11 +72,14 @@ def check_nib_diff_examples():
...
-    checked_fields = ["Field", "regular", "dim_info", "dim", "datatype", "bitpix", "pixdim", "slice_end",
+    checked_fields = ["Field/File", "regular", "dim_info", "dim", "datatype", "bitpix", "pixdim", "slice_end",
                       "xyzt_units", "cal_max", "descrip", "qform_code", "sform_code", "quatern_b",
                       "quatern_c", "quatern_d", "qoffset_x", "qoffset_y", "qoffset_z", "srow_x",
-                      "srow_y", "srow_z", "DATA(md5)"]
+                      "srow_y", "srow_z", "DATA(md5)", "DATA(diff 1:)"]
     for item in checked_fields:
+        if item not in stdout:
+            print(item)
+            print(stdout)
```
yarikoptic: Some Gods dislike such printouts (although I am not sure if that want left by me here :-)). This print will get lost since when you run all tests at once, errors details reported at the end whenever print happened long before. Add a `msg` to your assert below providing what you want us to see when it fails

nibabel/tests/test_scripts.py
chrispycheng: Please clarify?

chrispycheng: Jk I got it

chrispycheng:
AppVeyor's fail here is fake news, it's tripping up again

nibabel/cmdline/diff.py
```diff
@@ -101,8 +116,8 @@ def get_headers_diff(file_headers, names=None):
     return difference
 
 
-def get_data_diff(files):
-    """Get difference between md5 values
+def get_data_md5_diff(files):
```
effigies: Just future-proofing: How about naming it `get_data_hash_diff`? (Understood that the hash will be MD5 for the foreseeable future.)

nibabel/cmdline/diff.py
```diff
@@ -125,6 +140,76 @@ def get_data_diff(files):
     return md5sums
 
 
+def get_data_diff(files, max_abs=0, max_rel=0):
+    """Get difference between data
+
+    Parameters
+    ----------
+    files: list of (str or ndarray)
+      If list of strings is provided -- they must be existing file names
```
effigies: `ndarray` I assume means a data block equivalent to one loaded with `nib.load().get_fdata()` or similar?

nibabel/cmdline/diff.py
```diff
@@ -125,6 +140,76 @@ def get_data_diff(files):
...
+      If `max_abs` is specified, then those data points with lesser than that
+      absolute difference, are not considered for relative difference testing
+
+    Returns
+    -------
+    OrderedDict
+        str: absolute and relative differences of each file, given as float
+    """
+    # we are doomed to keep them in RAM now
+    data = [f if isinstance(f, np.ndarray) else nib.load(f).get_data() for f in files]
```
effigies: Given that `get_data()` is on its way out, I would suggest using either `dataobj.get_unscaled()` or `get_fdata(np.float32)` here, depending on whether you're interested in on-disk values or (more likely) scaled values.

nibabel/cmdline/diff.py
```diff
@@ -125,6 +140,76 @@ def get_data_diff(files):
...
+      Maximal absolute difference to tolerate.
+    max_rel: float, optional
+      Maximal relative (`abs(diff)/mean(diff)`) difference to tolerate.
+      If `max_abs` is specified, then those data points with lesser than that
+      absolute difference, are not considered for relative difference testing
+
+    Returns
+    -------
+    OrderedDict
+        str: absolute and relative differences of each file, given as float
```
effigies: I don't really understand the shape of the output, given this. Are the values 2-tuples, or a list of N-1 2-tuples? And what is the absolute diff for a file? Presumably the (max/mean/median) of the voxelwise absolute diffs. It would help to be explicit here.

nibabel/cmdline/diff.py
```diff
@@ -39,6 +39,21 @@ def get_opt_parser():
...
+               type=float,
+               default=0.0,
+               help="Maximal absolute difference in data between files to tolerate."),
+
+        Option("--mr", "--data-max-rel-diff",
+               dest="data_max_rel_diff",
+               type=float,
+               default=0.0,
+               help="Maximal relative difference in data between files to tolerate."
+                    " If also --data-max-abs-diff specified, only the data points "
```
effigies: `If --data-max-abs-diff is also specified`

nibabel/cmdline/diff.py
```diff
@@ -125,6 +140,80 @@ def get_data_diff(files):
...
+    -------
+    diffs: OrderedDict
+        An ordered dict with a record per each file which has differences with other files subsequent detected.
+        Each record is a list of difference records, one per each file pair. Each difference record is an Ordered
+        Dict with possible keys 'abs' or 'rel' showing maximal absolute or relative differences in the file 
+        or record ('CMP': 'incompat') if file shapes are incompatible.
+    """
+    
+    # we are doomed to keep them in RAM now
+    data = [f if isinstance(f, np.ndarray) else nib.load(f).get_fdata() for f in files]
```
effigies: Note that `get_fdata()` returns float64 arrays by default. If you're hoping to keep things moderately compact, you could use `get_fdata(dtype=np.float32)`. The precision loss should not affect equivalent files, and should be in the noise for any plausible MRI data.

effigies:
@chrispycheng @yarikoptic I just went ahead and fixed the style issue. I'm happy to merge if you're all set on this one.

yarikoptic:
in local conversion with @chrispycheng I have questioned the change in https://github.com/nipy/nibabel/pull/661/commits/034c2768dd9fd8628f2236c739eeda50e125f69b "hardcoding" the data type to float32.  I think no type conversion should be done so if files are of different data type, and thus possibly of different values (e.g. 1/3-rd would be different in float32 and float64), we could see that.  so may be that change should be reverted?

effigies:
`get_fdata()` is *always* `np.float64` unless otherwise specified.

Do you want to compare on-disk values (`dataobj.get_unscaled()` or effective values (`get_fdata()`)? `get_data` does give you sort of that sweet spot of "maybe it's one, maybe it's the other", but it's also supposed to be deprecated as of Apr 2018...

yarikoptic:
>  get_fdata() is always np.float64 unless otherwise specified.

`get_unscaled` probably would be "too low level". 
We are typically interested in the effective values.  With all the scaling etc we indeed can't rely on original data types and I would have stick to the default (float64) as providing the best "fidelity" as decided by nibabel in general to be the default data type to be returned.  May be eventually `get_fdata` would get somehow smarter and not downcast if data type is e.g. float128.  So the less of casting we do here, the better imho

effigies:
> I would have stick to the default (float64) as providing the best "fidelity" as decided by nibabel in general to be the default data type to be returned

Okay. Just making sure that's what you wanted. If diffing BOLD series, that could get expensive quickly.

yarikoptic:
> > I would have stick to the default (float64) as providing the best "fidelity" as decided by nibabel in general to be the default data type to be returned
> 
> Okay. Just making sure that's what you wanted. If diffing BOLD series, that could get expensive quickly.

indeed... but if we decide to provide help for those, we should just add  another parameter (`dtype`) so user could specify explicitly.  Could be done in a separate PR.

yarikoptic:
Hi @effigies, need your guidance here with appveyor.
python3.4 environment - the beast fails to install hypothesis from "source tarball" failing with 
```
  Downloading https://files.pythonhosted.org/packages/b0/58/5def5924eb6068f50855dfb74f9fe230a6437b7d104cd593c8d04daef400/hypothesis-3.74.1.tar.gz (177kB)
    Complete output from command python setup.py egg_info:
    C:\Users\appveyor\AppData\Local\Temp\1\pip-install-a58k4iic\hypothesis\setup.py:39: UserWarning: This version of setuptools is too old to correctly store conditional dependencies in binary wheels.  For more info, see:  https://hynek.me/articles/conditional-python-dependencies/
      'This version of setuptools is too old to correctly store '
    c:\python34\lib\distutils\dist.py:260: UserWarning: Unknown distribution option: 'python_requires'
      warnings.warn(msg)
    error in hypothesis setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected version spec in enum34; python_version=="2.7" at ; python_version=="2.7"
```
It is a known issue to the hypothesis people (https://github.com/HypothesisWorks/hypothesis/issues/1091) which they decided just to ignore altogether.  The funny part is that there is a whl for hypothesis for python3, and some times it gets installed (just fine) instead of trying to get it installed from the source tarball.  That is when appveyor doesn't fail.
So I wonder what  do you think we should we do about it?  I do not think we should merge into master as is to not cause those sporadic failures on appveyor

effigies:
My best guess from reading this is we need to update setuptools in appveyor. I'll look around a little.

effigies:
Watch this build: https://ci.appveyor.com/project/nipy/nibabel/build/1.0.501

effigies:
Feel free to cherry-pick 716b1c6d7b5312aae699a7431d4aa2b78c7a62b0 (on `upstream/appveyor/upgrade_setuptools`). At least it doesn't seem to cause problems, and the setuptools is woefully outdated. That said, it hasn't tried to build hypothesis from source. It found the wheels in both the 32-bit and 64-bit jobs.

effigies:
Pushed the Appveyor update directly to this branch, as it at least causes no harm. Please let me know what the status is on this PR.

chrispycheng:
@effigies seems like still no dice with appveyor

effigies:
Those AppVeyor bugs are hitting every PR and master. Nothing to do with hypothesis that I can see.

yarikoptic:
"those" is a spectrum here ;-)  Some were due to hypothesis, some not.

yarikoptic:
AVSD - AppVeyor Spectrum Disorder

yarikoptic:
eh, this one is a base now for 2 other PRs (#672 and #678), besides that commit to update setuptools which didn't help. @effigies - what do you prefer?  
- close this here and have #678 replace it
- or merge this one and then the others (#678 I think is ready)

effigies:
Let's close this in favor of #678. I'll update the name over there.

PR #666, opened 2018-09-27
FIX: Annotation labels are packed RGB, not RGBA

### LINKED ISSUE #649: Nibabel 2.3 freesurfer.read_annot() breaks backwards compatibility
Some .annot files have become unreadable since nibabel 2.3, with the error below:

```
In [22]: nibabel.freesurfer.read_annot('rh.fmri-tmean-25.annot')
IndexError                                Traceback (most recent call last)
<ipython-input-22-c6989d89b74f> in <module>()
----> 1 nibabel.freesurfer.read_annot('rh.fmri-tmean-25.annot')

/Users/christian/Code/nibabel/nibabel/freesurfer/io.pyc in read_annot(filepath, orig_ids)
    391         mask = labels != 0
    392         labels[~mask] = -1
--> 393         labels[mask] = ord[np.searchsorted(ctab[ord, -1], labels[mask])]
    394     return labels, ctab, names
    395 

IndexError: index 1 is out of bounds for axis 1 with size 1

In [23]: 
```

I think I traced the problem to a change in the way the label ids (`ctab[:, 4]`) are calulated:  now it is 
```
    bitshifts = 2 ** np.array([[0], [8], [16], [24]], dtype=rgba.dtype)
    return rgba.dot(bitshifts)
```
previously it used to be the equivalent of 
```
    bitshifts = 2 ** np.array([[0], [8], [16]], dtype=rgba.dtype)
    return rgba[:, :3].dot(bitshifts)
```
So older files have different label IDs ... Not sure what the proper way to fix this is, revert the use of alpha in label IDs? but that would probably break compatibility with files saved with nibabel 2.3...

Here is a file that breaks: https://drive.google.com/file/d/11w6KPjfQdD0AliNxE_YP4pQNweX3Xgvk/view?usp=sharing for files with multiple labels it does not break, but I suspect label order might be shifted.

effigies:
Contrary to FreeSurfer documentation, FreeSurfer code uses only the RGB values of colors to identify to annotation labels, ignoring the transparency value.

It is worth noting that FreeSurfer stores a *transparency* value, which is 255 - alpha. I'm somewhat inclined to continue returning raw transparency values, as we are not returning labeled data structures, but if people feel strongly that we should return alpha values, I'm okay with that, as well.

Fixes #649.

### DIFF
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -333,9 +333,13 @@ def read_annot(filepath, orig_ids=False):
     Annotation file format versions 1 and 2 are supported, corresponding to
     the "old-style" and "new-style" color table layout.
 
+    Note that the output color table ``ctab`` is in RGBT form, where T
+    (transparency) is 255 - alpha.
+
     See:
      * https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotation
      * https://github.com/freesurfer/freesurfer/blob/dev/matlab/read_annotation.m
+     * https://github.com/freesurfer/freesurfer/blob/8b88b34/utils/colortab.c
 
     Parameters
     ----------
@@ -352,7 +356,7 @@ def read_annot(filepath, orig_ids=False):
         Annotation id at each vertex. If a vertex does not belong
         to any label and orig_ids=False, its id will be set to -1.
     ctab : ndarray, shape (n_labels, 5)
-        RGBA + label id colortable array.
+        RGBT + label id colortable array.
     names : list of str (python 2), list of bytes (python 3)
         The names of the labels. The length of the list is n_labels.
     """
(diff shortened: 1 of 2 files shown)


effigies:
@pauldmccarthy Are you comfortable with these changes?

effigies:
Going by the response on the [mailing list](https://mail.nmr.mgh.harvard.edu/pipermail/freesurfer/2018-September/058742.html), Doug at least agrees with this interpretation.

@larsoner says it works for them over in PySurfer in https://github.com/nipy/nibabel/issues/649#issuecomment-425486813.

Barring objection, plan to merge this fix on Monday.

effigies:
freesurfer/freesurfer#541 has been merged, so let's get on the same page.

pauldmccarthy:
@effigies sorry, was away from laptop all of last week - all looks good to me, thanks again.

PR #674, opened 2018-10-04
ENH: Add ndim to ArrayProxy and DataobjImage

### LINKED ISSUE #595: DataobjImage doesn't proxy ndim
`DataobjImage` currently [proxies `shape`](https://github.com/nipy/nibabel/blob/6eacaf556c31c62f698ac07b4112245842d21f4e/nibabel/dataobj_images.py#L391-L393) but not [`ndim`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.ndim.html) even though various docs refer to `ndim(s)`: 
* http://nipy.org/nibabel/reference/nibabel.nifti1.html#nibabel.nifti1.Nifti1Header.set_data_shape
* http://nipy.org/nibabel/reference/nibabel.nifti2.html#nibabel.nifti2.Nifti2Header.set_data_shape

Might be useful to proxy `ndim` through.

effigies:
Closes #595.

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -233,6 +233,10 @@ class ArrayProxy(object):
         return self._shape
 
     @property
+    def ndim(self):
+        return len(self.shape)
+
+    @property
     def dtype(self):
         return self._dtype
 
--- a/nibabel/dataobj_images.py
+++ b/nibabel/dataobj_images.py
@@ -28,8 +28,8 @@ class DataobjImage(FileBasedImage):
         ----------
         dataobj : object
            Object containg image data.  It should be some object that retuns an
-           array from ``np.asanyarray``.  It should have a ``shape`` attribute
-           or property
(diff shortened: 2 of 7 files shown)


effigies:
@cancan101 Would you like to test that this provides the interface(s) you were asking for?

effigies:
@matthew-brett Up for a quick review? This is a pretty small change.

matthew-brett:
Thanks - yes - it looks fine.  The only question is whether it's really worth adding the code for the benefit of going from `ndim = dataobj.ndim` to `ndim = len(dataobj.shape)`.    What do you think?

effigies:
I agree it's trivial. The advantage of the addition to ArrayProxy is that now ndim exists regardless of how the image is created. It's just a least surprise sort of thing. Passing it though to the image itself is less obviously simplifying, but what #595 originally asked about, so I went ahead and included it.

matthew-brett:
Then I'm fine with it, if you are ...

PR #675, opened 2018-10-04
BF(workaround): skip test_diff if hypothesis is not importable

yarikoptic:
Apparently hypothesis  is quite a tricky beast.  It might either fail to install
on elderly systems from source tarball, or even if it is installable, apparently
might rely on some coverage API which is present only in more recent versions
and thus fail upon import.

All those are of relevance for older systems and only for testing.

I know that e.g. #661 would add tests in the `test_diff.py`  which would not even need hypothesis, but I do not see easy way to make it work since hypothesis's given is used as a decorator.

### DIFF
--- a/nibabel/tests/test_diff.py
+++ b/nibabel/tests/test_diff.py
@@ -7,9 +7,13 @@ from __future__ import division, print_function, absolute_import
 from os.path import (dirname, join as pjoin, abspath)
 import numpy as np
 
-from hypothesis import given
-import hypothesis.strategies as st
-
+try:
+    import hypothesis
+    from hypothesis import given
+    import hypothesis.strategies as st
+except Exception as exc:
+    from nose import SkipTest
+    raise SkipTest("Failed to import hypothesis or its components: %s" % exc)
 
 DATA_PATH = abspath(pjoin(dirname(__file__), 'data'))
 


yarikoptic:
worse comes to worse, I could actually just keep this patch in the debian pkg so it would help to avoid some of those on older systems... may be that would actually be better?

yarikoptic:
Or given all the appveyor difficulties we should just rip appveyor out for now and limit ourselves to just some set of cases?

nibabel/tests/test_diff.py
```diff
@@ -7,9 +7,13 @@
...
-from hypothesis import given
-import hypothesis.strategies as st
-
+try:
+    import hypothesis
+    from hypothesis import given
+    import hypothesis.strategies as st
+except Exception as exc:
+    from nose import SkipTest
+    raise SkipTest("Failed to import hypothesis or its components: %s" % exc)
```
effigies: To avoid necessarily skipping the entire file, what about making this `except` statement:

```Python
except ImportError:
    class st(object):
        @classmethod
        def data(klass):
            return None

    def given(obj):
        def decorator(func):
            def null_test():
                return None
        return decorator
```

Maybe raise a warning, as well.

yarikoptic:
#678 prunes use of hypothesis altogether

PR #676, opened 2018-10-05
WIP: Enh appveyor debug

yarikoptic:


yarikoptic:
FWIW @effigies - numpy is the same but the dtype is a different beast:
https://ci.appveyor.com/project/nipy/nibabel/builds/19350533/job/89d24ohp5edv1uyv#L831
```
mnc.get_data_dtype().type  ID: 1935635128, __module__: numpy, id(sys[__module__]): 53534272
              tp['dtype']  ID: 1935634912, __module__: numpy, id(sys[__module__]): 53534272
```

nibotmi:
:umbrella: The latest upstream changes (presumably #678) made this pull request unmergeable. Please resolve the merge conflicts.

effigies:
Just as an update, still happening intermittently on master: https://ci.appveyor.com/project/nipy/nibabel/builds/19462742

yarikoptic:
I think our best bet is to try to bisect it to when it started to happen/commit which lead to it (if it is nibabel issue and now external library/python). When did you say it started -- June'ish?

effigies:
June 8 was the first I could find digging through AppVeyor history. It was on #638.

I found nothing in AppVeyor's logs that indicate a change to 3.4 or 3.5 around that time. I haven't looked at numpy releases. Perhaps we should start by pinning older versions of numpy?

yarikoptic:
so the code before #638 merge does or doesn't fail if we run it in current appveyor setup? ;-)

effigies:
I tried reverting only #638 (see #677) and the errors persisted, but I haven't tried reverting to dfd113c531ad0dbb301db919b8573ad12ae0bb56.

effigies:
This effort is dead. The bug seems to be happening less lately, so hopefully whatever was going on over at AppVeyor is resolved. I think if this recurs, we should think about moving to Azure Pipelines for Windows tests.

PR #678, opened 2018-10-06
RF+BF: Add tolerances and data types for nib-diff, remove hypothesis dependency

yarikoptic:
As pointed by a few added tests, comparison wasn't doing proper job, especially in the light of comparing NaN values.  For the purpose of this function we would like to assume that two NaNs are equal - we are comparing different headers so bringing the difference were both have NaNs is not desired.

Attn @chrispycheng  - please review the changes.  Since there were no tests pointing specifically what was the intention from the code comparing for NaNs, I had difficult time figuring out what we wanted to say.


nibabel/cmdline/diff.py
```diff
@@ -45,25 +45,41 @@ def get_opt_parser():
...
+            if value0_nans is not None:
+                value_nans = np.isnan(value)
+                if np.any(value0_nans != value_nans):
+                    return True
+                if np.any(value0[np.logical_not(value0_nans)]
+                          != value[np.logical_not(value0_nans)]):
+                    return True
+            elif np.any(value0 != value):
+                return True
+        elif value0 is np.NaN:
```
effigies: Numpy [prefers](https://docs.scipy.org/doc/numpy/reference/constants.html#numpy.NaN) you to use `np.nan`. I don't know if this means they plan to remove the synonyms in the future or if it's just a style thing. I'd go with their recommendation, though.

Also appears in the test file.

yarikoptic:
Thank you @effigies - pushed 1996eadcb85269447523863c542fe14c33e8c470 .

effigies:
@chrispycheng Are you happy with this PR?

yarikoptic:
Finally tests are green!

yarikoptic:
@chrispycheng please look at https://codecov.io/gh/nipy/nibabel/pull/678/src/cmdline/diff.py and add those missing tests so we get all cases covered (ignore the uncovered reraising the exception case)

chrispycheng:
@effigies I'm going to sort out the coverage Yarik mentioned and then we should be good to go

chrispycheng:
@yarikoptic @effigies Changes are fired up and ready to go with Travis passing and increased coverage. AppVeyor is causing trouble again but not due to anything from this commit.

effigies:
@yarikoptic If you're happy with this, want to merge?

yarikoptic:
Ok, let's do it! thanks @effigies and @chrispycheng

PR #685, opened 2018-10-22
FIX: Read .PAR files corresponding to ADC maps

grlee77:
Normally, for any diffusion scan, the data will have multiple volumes (at least two b values must be acquired). However, it is possible to export ADC maps which seem to retain the same `general_info` section from the original scan's header, but actually only contain a single processed volume. This violates the assumptions that were made in `get_bvals_bvecs` and causes `parrec2nii` to fail when converting .PAR/.REC data corresponding to ADC maps.

This PR fixes this by returning (None, None) for `bvals` and `bvecs` when there is only a single volume present in the file.

A small (anonymous) .PAR header and test case have been included.

### DIFF
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -784,7 +784,13 @@ class PARRECHeader(SpatialHeader):
         if self.general_info['diffusion'] == 0:
             return None, None
         reorder = self.get_sorted_slice_indices()
-        n_slices, n_vols = self.get_data_shape()[-2:]
+        if len(self.get_data_shape()) == 3:
+            # Any original diffusion scans will have >=2 volumes. However, a
+            # single dynamic is possible for a post-processed difusion volume
+            # such as an ADC map. The b-values are unavailble in this case.
+            return None, None
+        else:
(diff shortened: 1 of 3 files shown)


nibabel/parrec.py
```diff
@@ -784,7 +784,13 @@ def get_bvals_bvecs(self):
         if self.general_info['diffusion'] == 0:
             return None, None
         reorder = self.get_sorted_slice_indices()
-        n_slices, n_vols = self.get_data_shape()[-2:]
+        if len(self.get_data_shape()) == 3:
+            # Any original diffusion scans will have >=2 volumes. However, a
+            # single dynamic is possible for a post-processed difusion volume
```
effigies: Missing f in diffusion

nibabel/parrec.py
```diff
@@ -784,7 +784,13 @@ def get_bvals_bvecs(self):
...
+            # such as an ADC map. The b-values are unavailble in this case.
```
effigies: unavail*a*ble

Issue #665, opened 2018-09-26
[WIN] Stochastic bug: Minc2 data may load as np.longdouble in h5py < 2.10


effigies:
These AppVeyor failures are strange and stochastic. The failures are consistent:

```
======================================================================
ERROR: autogenerated test from validate_data_deprecated
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\python34\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "c:\python34\lib\site-packages\nibabel\tests\test_api_validators.py", line 23, in meth
    validator(self, imaker, params)
  File "c:\python34\lib\site-packages\nibabel\tests\test_image_api.py", line 380, in validate_data_deprecated
    assert_data_similar(img._data, params)
  File "c:\python34\lib\site-packages\nibabel\tests\test_helpers.py", line 80, in assert_data_similar
    real_arr = np.asarray(arr)
  File "c:\python34\lib\site-packages\numpy\core\numeric.py", line 501, in asarray
    return array(a, dtype, copy=False, order=order)
  File "c:\python34\lib\site-packages\nibabel\minc1.py", line 261, in __array__
    return self.minc_file.get_scaled_data()
  File "c:\python34\lib\site-packages\nibabel\minc2.py", line 139, in get_scaled_data
    return self._normalize(raw_data, sliceobj)
  File "c:\python34\lib\site-packages\nibabel\minc1.py", line 189, in _normalize
    dmin, dmax = self._get_valid_range()
  File "c:\python34\lib\site-packages\nibabel\minc2.py", line 98, in _get_valid_range
    info = np.iinfo(ddt.type)
  File "c:\python34\lib\site-packages\numpy\core\getlimits.py", line 516, in __init__
    raise ValueError("Invalid integer data type.")
ValueError: Invalid integer data type.
======================================================================
ERROR: nibabel.tests.test_minc2.TestMinc2File.test_array_proxy_slicing
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\python34\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "c:\python34\lib\site-packages\nibabel\tests\test_minc1.py", line 200, in test_array_proxy_slicing
    arr = img.get_data()
  File "c:\python34\lib\site-packages\nibabel\dataobj_images.py", line 202, in get_data
    data = np.asanyarray(self._dataobj)
[...]

PR #686, opened 2018-10-23
MNT: Remove deprecated numpy.asscalar

effigies:
I think this should fix the error in the prerelease tests: https://travis-ci.org/nipy/nibabel/jobs/444811573

The semantics of `ndarray.item()` have not changed since before 1.7.1, so there is no need to update the minimum numpy version.


grlee77:
Looks good to me. The failures on Appveyor for Python 3.4 64-bit do not look related to the PR.

effigies:
Yeah, those are #665-related. Thanks for the review.

Issue #669, opened 2018-10-03
Warn on affine/header mismatch at write time


effigies:
My understanding of the `SpatialImage` API is that `image.affine` is authoritative, and that it is written into the header prior to writing to file. This can lead to surprises for users modifying header attributes that correspond to the affine. For example, zooms may be used to construct an affine (MGHImage/NIfTI with no valid sform) or be unrelated (see #619 for further discussion).

`image.update_header()` is called at image construction and writing. I would suggest that we warn if it causes changes to the header at write time. (At construction time might occasionally be useful, but it will cause a lot of noise.)

Sub-issue identified in #619.

Related: #668

effigies:
Relatedly: The documentation for `SpatialImage` should be updated to make clear that `image.affine` is authoritative and may clobber some header fields.

PR #688, opened 2018-11-02
Ensure affine attribute in sync with header after init.

takhs91:
Hi,
I made the following changes that ensure that the affine attribute is in sync with the header after initializing a Nifti Image.

Those two may differ because the affine provided may contain values that need float64 to store, while they have to be casted to float32 for the nifti1 header and thus some values are slightly changed.

I actually encountered this while using `as_closest_canonical` (which actually creates a new instance) on images and then expecting the affine I had in memory to be the same as the one that will be written and later read from file.

I found some related issues such as http://nipy.org/nibabel/devel/modified_images.html and 
https://github.com/nipy/nibabel/issues/669 but they deal with the more general issue of memory/file mismatch that may happen while modifying things on the instance.

This is a more fundamental step that ensures the affine attribute just after init is the same as the one that will be written excluding later modifications


takhs91:
Any thoughts on that?

effigies:
Thanks for this. Looking at it, I have a couple things I want to consider about the order of operations. I'll try to sort out my thoughts in the next day or two. In the meantime, would you mind merging master? Also, want to have a look at the tests and see what might be causing the failure?

takhs91:
Thanks for pointing out the failing tests.

The tests that fail are:

* nibabel.tests.test_brikhead.TestAFNIImage.test_load
* nibabel.tests.test_minc2.TestMinc2File.test_load
* nibabel.tests.test_processing.test_resample_to_output
* nibabel.tests.test_spaces.test_vox2out_vox

Since my changes alter the affine to float32 precision on init at NIfti1 images it seems logical for some tests comparing affines to fail. 

I see two options here. 

1. Replacing Nifti1 in those tests with Nifti2. With its higher precision affine the test will have to pass
2. Keep the Nifti1 in those tests and alter the tolerance at the equality comparisons

effigies:
Okay. So now I'm less sure that this is a good approach, as this will be a pretty major violation of the historical expectation for images in memory to not fully synchronize until written to disk.

We could consider an alternative method for `FileBasedImage`s:

```Python
class FileBasedImage:
    def as_synchronized(self):
        """Generate a copy of the image as if it had been saved to disk and reloaded"""
        ...
```

This would allow the programmer to explicitly choose when the constraints of the format should be enforced.

takhs91:
That is an interesting proposition. I believe it a has a bigger implementation scope though (correct me if wrong). I was kind of aware of this expectation (read about it in other issues mostly) and also learned about it the hard way. I think it needs to be better communicated in the documentation. Except it is and I missed that of course.

effigies:
At least for `Nifti1Image`s (and other single-file images), I can quickly write you a synchronizer that will keep everything in memory:

```Python
import io

def synchronize_image(img):
    bio = io.BytesIO()
    file_map = img.make_file_map({'image': bio, 'header': bio})
    img.to_file_map(file_map)
    return img.__class__.from_file_map(file_map)
```

So if you just need this in your own code for now, you can use that, and we can work toward a good nibabel-level solution on a longer time frame.

As to the documentation, do you have any suggestions for making these features more obvious?

takhs91:
Hi I didn't have time to look mote into that yet. I will come back to this asap

PR #692, opened 2018-11-10
BF: fix repetition time in PARREC header

Roosted7:
While I was working on #683, I noticed that the scaling in the repetition time header field is wrong.
It looks like the last value of `get_zooms()` should report the repetition time in milliseconds and not in seconds.

Ofcourse, I also fixed the tests.

### DIFF
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -897,11 +897,11 @@ class PARRECHeader(SpatialHeader):
         zooms[:2] = self._get_unique_image_prop('pixel spacing')
         slice_thickness = self._get_unique_image_prop('slice thickness')
         zooms[2] = slice_thickness + slice_gap
-        # If 4D dynamic scan, convert time from milliseconds to seconds
+        # If 4D dynamic scan, save repetition time in miliseconds
         if len(zooms) > 3 and self.general_info['dyn_scan']:
             if len(self.general_info['repetition_time']) > 1:
                 warnings.warn("multiple TRs found in .PAR file")
-            zooms[3] = self.general_info['repetition_time'][0] / 1000.
+            zooms[3] = self.general_info['repetition_time'][0]
         return zooms
 
     def get_affine(self, origin='scanner'):
--- a/nibabel/tests/test_parrec.py
+++ b/nibabel/tests/test_parrec.py
@@ -144,7 +144,7 @@ EXAMPLE_IMAGES = [
         # We disagree with Philips about the right affine, for the moment, so
         # use our own affine as determined from a previous load in nibabel
         affine=AN_OLD_AFFINE,
-        zooms=(3.75, 3.75, 8.0, 2.0),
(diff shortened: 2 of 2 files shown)


effigies:
Please merge master to fix style checks.

Roosted7:
@effigies Sure! I rebased my pr on master

effigies:
Thanks. Just out of curiosity, why use milliseconds over seconds? Simply to transparently report the values in the file, or was there some other bug? I ask because a lot of code assumes that `get_zooms()[3]` is in seconds, regardless of image type.

effigies:
fMRIPrep was not written to expect TRs to be in milliseconds, and in fact ignores the NIfTI file TRs when possible, preferring the `RepetitionTime` metadata in the JSON sidecars (which is required by BIDS to be in seconds, FWIW).

Is it possible that nibabel is outputting NIfTIs with TRs in seconds but reporting millisecond units? That could cause problems.

Roosted7:
JSON sidecarts are not required for fMRIPrep to run right?

That would explain these problems indeed, but is there a field in the NifTI header that sets the units for the Tr?

effigies:
fMRIPrep assumes a valid BIDS dataset. While not all mandatory metadata is accessed, we do attempt to access some of it, including `RepetitionTime`, so if it runs, you must have that.

The xyzt_units field of the NIfTI header can includes the bitwise AND of spatial and temporal units, with codes described here:

https://github.com/nipy/nibabel/blob/ad26878d7dd817e0bef7cd3ccfdd3f95b9694657/nibabel/nifti1.py#L130-L141

A code of 10 indicates spatial units of `mm` and temporal units of `sec`, and most software will assume those if either code is missing, and many programs will assume those regardless of the actual units specified (I recall seeing a peripheral FSL program that did not check units and got tripped up by ms TR within the last two years).

To quickly view the shape and units of a file you have (if nibabel is installed):

```
nib-ls -H dim,pixdim,xyzt_units file.nii.gz
```

effigies:
Hi @Roosted7, just checking in on what your thoughts are on this? Are you still having issues with some software expecting milliseconds? Could you provide a test where a file is written with bad metadata?

epongpipat:
I'm not sure if this is fixed yet, but I'm also having this issue. The TR number is being converted to sec, but the time scale is still in msec. When I run `nib-ls -H xyzt_units` I get the 18 rather than 10 for the xyzt_units.

effigies:
@epongpipat Sorry about the extreme delay, but thanks for your comment. I had a poke around `parrec2nii`, and found that what seems to be happening is that the zooms are in seconds, but there is a method to enable NIfTI conversion here:

https://github.com/nipy/nibabel/blob/7a7a913785060535af68fa9d904cd76724af5afe/nibabel/parrec.py#L766-L780

This reports `msec` instead of `sec`. I proposed an alternative fix in #931.

effigies:
Closing for now.

PR #695, opened 2018-11-27
STY: Flake8 3.6+ cleanup

### LINKED ISSUE #687: flake8 3.6 breaks with our configuration
A number of errors are suddenly appearing. It seems pycodestyle has introduced new errors and flake8 has introduced new syntax for updating the ignore list as opposed to replacing it.

- https://gitlab.com/pycqa/flake8/issues/463
- https://gitlab.com/pycqa/flake8/issues/466

If somebody has time, going through and seeing which are things worth conforming to, ignoring in individual files, and ignoring in the whole project would probably be a good idea. I don't really want to have to constrain style checkers.

The short term solution is probably just to pin to flake8<3.6.

effigies:
Fixes #687.


effigies:
Minor cleanups, so I'll merge tomorrow unless someone wants to review sooner.

PR #700, opened 2018-12-10
FIX: Numpy pre-release accommodations

effigies:
This has turned into something of an omnibus PR, which does the following things:

* Updates the setup script to use `setuptools` and take advantage of conditional dependencies
* Require `bz2file` for Python 2.7 only
* Add `Opener.readinto` method (via #702)
* Use `bytearray` and `filelike.readinto` in numerous places to preserve zero-copy construction of numpy arrays
* Remove deprecated `numpy.fromstring` (via #702)

Closes #702.

----

## Original

It looks like we've been taking advantage of numpy letting us declare a memory block writeable even when the underlying object is ostensibly immutable. This has allowed us to avoid copies for objects that we knew wouldn't be accessed through their original interfaces, but they decided this was a bug and closed this loophole in numpy/numpy#11739.

Don't know what else to do but make the copies they think we should be making, but perhaps others have more numpy arcana to share.

Fixes #697.

Issue #697, opened 2018-12-03
Numpy pre-release breaks setting some arrays writable


effigies:
Most likely culprit is numpy/numpy#11739, which was merged 7 days ago.

Example failures:

```
ERROR: test_load_simple_file (nibabel.streamlines.tests.test_streamlines.TestLoadSave)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/tests/test_streamlines.py", line 152, in test_load_simple_file
    lazy_load=lazy_load)
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/__init__.py", line 96, in load
    return tractogram_file.load(fileobj, lazy_load=lazy_load)
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/tck.py", line 154, in load
    streamlines = ArraySequence(tck_reader)
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/array_sequence.py", line 93, in __init__
    self.extend(iterable)
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/array_sequence.py", line 239, in extend
    for e in elements:
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/tck.py", line 436, in _read
    pts.flags.writeable = True
ValueError: cannot set WRITEABLE flag to True of this array
```

```
ERROR: test_load_complex_file_in_big_endian (nibabel.streamlines.tests.test_trk.TestTRK)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/tests/test_trk.py", line 194, in test_load_complex_file_in_big_endian
    'complex_trk_big_endian_fname', endian='>')
  File "/home/travis/build/nipy/nibabel/venv/lib/python2.7/site-packages/nibabel/streamlines/tests/test_trk.py", line 109, in trk_with_bytes
    trk_struct.flags.writeable = True
ValueError: cannot set WRITEABLE flag to True of this array
```

```
ERROR: test_big_offset_exts (nibabel.tests.test_analyze.TestAnalyzeImage)
[...]

PR #702, opened 2018-12-17
RF: Circumvents a deprecation warning from `np.fromstring`

arokem:
I am seeing this warning:

```
/Users/arokem/.virtualenvs/afq/lib/python3.7/site-packages/nibabel/streamlines/trk.py:562: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
  header_rec = np.fromstring(string=header_str, dtype=header_2_dtype)
```

I believe this PR fixes that.

### DIFF
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -209,6 +209,9 @@ class Opener(object):
     def read(self, *args, **kwargs):
         return self.fobj.read(*args, **kwargs)
 
+    def readinto(self, *args, **kwargs):
+        return self.fobj.readinto(*args, **kwargs)
+
     def write(self, *args, **kwargs):
         return self.fobj.write(*args, **kwargs)
 
--- a/nibabel/streamlines/trk.py
+++ b/nibabel/streamlines/trk.py
@@ -557,10 +557,10 @@ class TrkFile(TractogramFile):
 
         with Opener(fileobj) as f:
 
-            # Read the header in one block.
-            header_str = f.read(header_2_dtype.itemsize)
-            header_rec = np.frombuffer(buffer=bytearray(header_str),
-                                       dtype=header_2_dtype)
+            # Read the header into a bytearray.
+            header_buf = bytearray(header_2_dtype.itemsize)
+            n_read = f.readinto(header_buf)
+            header_rec = np.frombuffer(buffer=header_buf, dtype=header_2_dtype)
             # Check endianness
             endianness = native_code
             if header_rec['hdr_size'] != TrkFile.HEADER_SIZE:
--- a/nibabel/streamlines/trk.py
+++ b/nibabel/streamlines/trk.py
@@ -559,8 +559,8 @@ class TrkFile(TractogramFile):
 
             # Read the header in one block.
             header_str = f.read(header_2_dtype.itemsize)
-            header_rec = np.frombuffer(buffer=header_str, dtype=header_2_dtype)
-            header_rec.setflags(write=1)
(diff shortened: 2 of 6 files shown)


effigies:
Thanks. Looks like `frombuffer()` has the same API back at least to our minimum numpy, so this seems reasonable. I think there's a chance you'll have trouble setting the write flag on the pre-release numpy, so we might not be done yet.

Any chance you could swap out `numpy.fromstring()` in the [rest of the codebase](https://github.com/nipy/nibabel/search?q=fromstring&unscoped_q=fromstring), while you're at it?

arokem:
> Any chance you could swap out numpy.fromstring() in the rest of the
> codebase
> <https://github.com/nipy/nibabel/search?q=fromstring&unscoped_q=fromstring>,
> while you're at it?
>

Yep. Will do.

arokem:
As anticipated, pre-release numpy doesn't allow setting this array to writeable. Do you know if there is any fix to this?

effigies:
You'll need to `readinto` some read-write object, such as a `bytearray`. See discussions in #697, #700.

arokem:
OK. I believe that remaining appearances of `fromstring` are not of the deprecated kind.

nibabel/streamlines/trk.py
```diff
@@ -559,8 +559,8 @@ def _read_header(fileobj):
 
             # Read the header in one block.
             header_str = f.read(header_2_dtype.itemsize)
-            header_rec = np.fromstring(string=header_str, dtype=header_2_dtype)
-
+            header_rec = np.frombuffer(buffer=bytearray(header_str),
```
effigies: This still makes a copy from `header_str` to `bytearray(header_str)`. Instead, we can create a bytearray and read directly into it:

```Python
            header_buf = bytearray(header_2_dtype.itemsize)
            n_read = f.readinto(header_buf)
            # Maybe if n_read != header_2_dtype.itemsize: raise HeaderError(...)
            header_rec = np.frombuffer(buffer=header_buf, dtype=header_2_dtype)
```

arokem:
Oh, scratch that. Now it does :-)

nibabel/openers.py
```diff
@@ -209,6 +209,9 @@ def fileno(self):
     def read(self, *args, **kwargs):
         return self.fobj.read(*args, **kwargs)
 
+    def readinto(self, *args, **kwargs):
+        return self.fobj.readinto(*args, **kwargs)
+
```
effigies: Oh, nice. I hadn't realized we didn't already have this...

arokem:
Any suggestions on how to test that? I got a bit lost in test_openers

effigies:
I think this function is as close as we come to fully testing the API for equivalence:

https://github.com/nipy/nibabel/blob/a960b92b48f46517fe458b3a675206d71fb8cfea/nibabel/tests/test_openers.py#L198-L213

arokem:
Looks like we have similar issues in other places as well. Let me see if I can use similar approaches in these other places as well...

effigies:
Awesome, thanks. In #700, I'll be fixing the `volumeutils` piece to use `bz2file`, which implements `readinto` and should take care of that one, but the TCK and TRK ones could use your attention.

arokem:
OK. I'll look into those.

arokem:
Sorry: I have not been able to resolve these other spots and might not have time to work on this any time soon. If it's OK, I would suggest merging this one, and then keep working on the other ones on another PR.

arokem:
That's great. Thanks!

PR #706, opened 2018-12-25
BUG: Fix deprecation warning on use of `np.fromstring`.

jhlegarreta:
Fix deprecation warning on use of `np.fromstring`. Fixes:
```
DeprecationWarning: The binary mode of fromstring is deprecated, as it
behaves surprisingly on unicode inputs. Use frombuffer instead
```

observed in projects using Nibabel, e.g. DIPY:
https://travis-ci.org/nipy/dipy/jobs/471947136


effigies:
Thanks for this, but it's already being addressed in #700/#702.

PR #719, opened 2019-01-23
Don't refcheck on this resize, to avoid errors

arokem:
I stumbled into this in the course of my work on [pyAFQ](http://yeatmanlab.github.io/pyAFQ/). 

For testing, part of the code reads data from a trk file. This seems to cause some issues in this method, but only on 3.7: https://travis-ci.org/arokem/pyAFQ/jobs/483151878. Unfortunately, I was not able to reproduce these errors locally on my machine...

After reading about what this might be, and implementing the proposed change, these tests are now passing on the Travis build for that branch, though. 

Any objections to this?

### DIFF
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -199,7 +199,8 @@ class ArraySequence(object):
             self._data.resize(new_shape)
 
     def shrink_data(self):
-        self._data.resize((self._get_next_offset(),) + self.common_shape)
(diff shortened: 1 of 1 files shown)


matthew-brett:
I defer to @MarcCote on this one ...

MarcCote:
After looking at the resize documentation, I'd say let's merge this. It should be fine.

effigies:
Thanks for the review.

PR #720, opened 2019-01-27
FIX: Accept any valid delimiters/EOF markers in TCK files

soichih:
According to mrtrix/tck specification.

```
The binary track data themselves are stored as triplets of floating-point values (at this stage in 32 bit floating-point format), one per vertex along the track. Tracks are separated using a triplet of NaN values. Finally, a triplet of Inf values is used to indicate the end of the file.
```
> https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#tracks-file-format-tck

The current implementation of streamlines/tck.py / _read() expects that these delimiters are in specific binary format; whatever numpy.nan and numpy.inf happens to be). However, according to IEEE standard, NaN/Inf can have many different binary representations. 

```
EEE 754 NaNs are represented with the exponent field filled with ones (like infinity values), and some non-zero number in the significand (to make them distinct from infinity values); 
```

> https://en.wikipedia.org/wiki/NaN

Therefore, streamlines generated by other programming languages (such as Matlab) can store these NaN and Inf with a slightly different binary format, and nibabel.streamlines.tck fails to detect these delimieters which results in the following error

> https://github.com/brainlife/app-convert-tck-to-trk/issues/1#issue-403473788

I have re-implemented _read() so that it will actually look at each values to see if they are NaN values to determine if the values are stream line delimiters. My code runs around 30% slower (4.7sec v.s. 3.6sec for loading 500k streamlines) but I believe proper checking of NaN/Inf are unavoidable so that this code can load streamlines generated by other programming languages (or numpy might change the definition for numpy.nan and numpy.inf in the future)


matthew-brett:
Thanks for this.

Can you add a test, maybe with a currently failing small trk file?

Can you think of a way of speeding up again?  Can we first check what the binary format is, then use that in subsequent reads from the same file?

soichih:
@matthew-brett Yes, I can try adding a unit test. Currently though.. "make test" takes a really long time.. (it's stuck on 99% CPU usage for >30 min when I try it) Is it normal? 

> Can we first check what the binary format is, then use that in subsequent reads from the same file?

Maybe.. but I think it's a bit risky..  tck file could be generated by concatenating different .tck files from different sources, so depending on the implementation, it could end up with having a mix of different binary formats within a single file (I can think of a case where someone tries to by-pass the same obj. buffer that's loaded from a file and appended it to the new aggregated file)

Using a proper float32 NaN/Inf function is slower, but loading 600K fiber would only take a few seconds so I wouldn't worry too much about optimizing this. I think it's best to do this in the right way (and simpler); looking for a specific binary pattern for NaN/Inf or any floating decimal values for that matter is a bit iffy. 

I've attached a sample .tck file that fails to load with the current implementation.

[matlab.zip](https://github.com/nipy/nibabel/files/2804154/matlab.zip)

nibabel/streamlines/tck.py
```diff
@@ -401,46 +396,36 @@ def _read(cls, fileobj, header, buffer_size=4):
...
+                    end = delims[i]
+                    stream = np.append(leftover, coords[begin:end], axis=0)
+                    leftover = np.empty([0,3])
+                    yield stream
+                    n_streams += 1
+
+                    begin = end+1 #skip the delimiter
+
+                # the rest gets appended to the leftover
+                leftover = np.append(leftover, coords[begin:], axis=0)
```
effigies: I don't know if this is going to be a major contributor to the slowdown, but we can reduce the number of appends to <=1 per batch instead of 1 per streamline + 1 per batch:

```Python
delims = np.where(np.all(np.isnan(coords), axis=1))[0]

if leftover.size:
    coords = np.vstack((leftover, coords))
    delims += leftover.shape[0]
    leftover = np.empty((0, 3), dtype='<f4')

begin = 0
for delim in delims:
    yield coords[begin:delim]
    n_streams += 1
    begin = delim + 1

leftover = coords[begin:]

File: nibabel/streamlines/tck.py [hunk 2]
```diff
@@ -401,46 +396,36 @@ def _read(cls, fileobj, header, buffer_size=4):
             f.seek(header["_offset_data"], os.SEEK_SET)
 
             eof = False
-            buffs = []
             n_streams = 0
-
+            leftover = np.empty([0,3])
```
effigies: This should be the same `dtype` as `coords`, below, otherwise things will get upcast to `float64`:

```suggestion
            leftover = np.empty((0, 3), dtype='<f4')
```

soichih:
@effigies Thanks for the suggestion. I've applied a few of your suggestions and now my code runs about 5% faster compared to the current production version (instead of slower). I assumed that numpy.append with an empty array is a no-op, but I guess I was wrong.

I've kept most of my original code, as your version doesn't handle the EOF delimiter checking correctly. I think there is a way to make it work..  but I thought mine is a bit simpler. I still can't get the unittest to work, but I've tested my code with various input files I had and made sure that it loads the correct number of fibers/streams.

matthew-brett:
For some reason I can't get the Travis-CI page with the test failures - can you describe them?

Can you add a small test file written by Matlab to confirm this fixes the read?

effigies:
The test failures occur because numpy expects a real file object in `fromfile`, which fails if we have any stream that isn't that. I will try a `bytesarray` fix...

nibabel/streamlines/tck.py
```diff
@@ -411,37 +407,36 @@ def _read(cls, fileobj, header, buffer_size=4):
...
+                # Recover leftovers, which can't have delimiters in them
+                if leftover.size:
+                    delims += leftover.shape[0]
+                    coords = np.vstack((leftover, coords))
+
+                begin = 0
+                for delim in delims:
+                    pts = coords[begin:delim]
+                    if pts.size:
+                        yield coords[begin:delim]
```
effigies:
```suggestion
                        yield pts
```

matthew-brett:
@effigies - sorry - I'm completely overwhelmed with work at the moment, will be until Wednesday or so.  @MarcCote - do you mind having a look?

effigies:
Thanks for the heads up Matthew.

MarcCote:
Thanks for pinging me. I'm not explicitly watching Nibabel repo anymore but I'm happy to help whenever I can. I'll have a look tonight.

nibabel/streamlines/tck.py
```diff
@@ -411,37 +407,36 @@ def _read(cls, fileobj, header, buffer_size=4):
...
+                        n_streams += 1
+                    begin = delim + 1
+
+                # The rest gets appended to the leftover
```
MarcCote:
```suggestion
                # The rest becomes the new leftover.
```

effigies:
Thanks, @MarcCote. @soichih If you have a couple minutes to apply the two suggestions, we can go ahead and merge ASAP.

soichih:
@effigies Did I do it right? I believe I've applied @MarcCote 's suggestions.

nibabel/streamlines/tck.py
effigies: Could you apply this suggestion, as well?

soichih: done!

effigies:
Great! Thanks!

PR #726, opened 2019-02-13
ENH: Add a missing white space between words in string literal.

jhlegarreta:
Add a missing white space between words in a string literal split in two
lines.


jhlegarreta:
@effigies Not sure why coveralls reports a coverage decrease. This is just a change in a string message.

effigies:
Coverage can be weird, and I suspect the situation here is the fact that coveralls doesn't aggregate across multiple tests, as codecov does. I'm removing it anyway in #732, so no worries.

Thanks for this.

PR #727, opened 2019-02-20
ENH: Support for Philips DICOMs w/ derived volume

### LINKED ISSUE #693: Exclude trace volume from Philips DICOM?
Originally brought up in nipy/heudiconv#275 by @bmelo

Version: `nibabel-2.3.1`

Loading up a Philips diffusion DICOM (with trace volume) results in an incorrect image shape with an extra? dimension (96, 96, 60, 2, 33).

```python
In [1]: from nibabel.nicom.dicomwrappers import wrapper_from_file

In [2]: mw = wrapper_from_file('philips-dwi.dcm', force=True)

In [3]: mw.image_shape
WrapperError                              Traceback (most recent call last)
<ipython-input-3-4963e90423f3> in <module>()
----> 1 mw.image_shape

/home/mathias/miniconda2/envs/dev3/lib/python3.6/site-packages/nibabel/onetime.py in __get__(self, obj, type)
    139         # Errors in the following line are errors in setting a
    140         # OneTimeProperty
--> 141         val = self.getter(obj)
    142 
    143         setattr(obj, self.name, val)

/home/mathias/miniconda2/envs/dev3/lib/python3.6/site-packages/nibabel/nicom/dicomwrappers.py in image_shape(self)
    535         n_vols = np.prod(shape[3:])
    536         if n_frames != n_vols * shape[2]:
--> 537             raise WrapperError("Calculated shape does not match number of "
    538                                "frames.")
    539         return tuple(shape)

WrapperError: Calculated shape does not match number of frames.
```

mgxd:
Closes #693 

Currently, if a Philips DICOM includes a derived volume, `MultiframeWrapper` blows up due to incorrect `image_shape`. This PR adds the following:
* if DICOM has MR Diffusion Sequence (0018, 9117), iter through frames and discard any with type derived
* ~~hidden attribute `_nframes` to track if `NumberOfFrames` has changed~~
* 4D diffusion DICOM with tracer volume for testing (thanks to @bmelo)

I also noticed `get_data()` output is incorrect, but haven't addressed that (any ideas regarding that would be appreciated!) My main motivation for this was to fix https://github.com/nipy/heudiconv/issues/275

Relevant material https://github.com/rordenlab/dcm2niix/blob/d04ba75c8b098240d0efef67379947ff75bc1d36/Philips/README.md#derived-parametric-maps-stored-with-raw-diffusion-data


nibabel/pydicom_compat.py
```diff
@@ -40,6 +40,7 @@
     read_file = pydicom.read_file
 
 if have_dicom:
+    from pydicom.sequence import Sequence
```
effigies: Does this work with `pydicom < 1.0`? Importing doesn't respect renamed modules, so I suspect this is going to need to go up in the `else` blocks (L35-L40).

nibabel/nicom/dicomwrappers.py
```diff
@@ -461,10 +461,23 @@ def __init__(self, dcm_data):
...
         self.frames = dcm_data.get('PerFrameFunctionalGroupsSequence')
+        self._nframes = self.get('NumberOfFrames')
         try:
             self.frames[0]
         except TypeError:
             raise WrapperError("PerFrameFunctionalGroupsSequence is empty.")
+        try:
+            # DWI image where derived isotropic, ADC or trace volume
+            # was appended to the series
+            if self.frames[0].get([0x18, 0x9117], None):
```
effigies: Looks like `None` is the default anyway.

```suggestion
            if self.frames[0].get([0x18, 0x9117]):
```

nibabel/nicom/dicomwrappers.py
```diff
@@ -461,10 +461,23 @@ def __init__(self, dcm_data):
...
+                self.frames = Sequence(
+                    frame for frame in self.frames if
+                    frame.get([0x18, 0x9117])[0].get([0x18, 0x9075]).value
+                    != 'ISOTROPIC'
+                    )
+                self._nframes = len(self.frames)
+        except AttributeError:
+            pass
```
effigies: This is a pretty broad `try`/`except` block. Can the expected failures be narrowed down at all?

Is there a possibility of an `IndexError` with `frame.get([0x18, 0x9117])[0]`? If so, is that an indicator of corrupt data where we should raise a `WrapperError`?

nibabel/nicom/dicomwrappers.py
```diff
@@ -528,12 +540,15 @@ def image_shape(self):
         # Store frame indices
         self._frame_indices = frame_indices
         if n_dim < 4:  # 3D volume
-            return rows, cols, n_frames
+            return rows, cols, self._nframes
         # More than 3 dimensions
         ns_unique = [len(np.unique(row)) for row in self._frame_indices.T]
+        if len(ns_unique) == 3:
+            # derived volume is included
+            ns_unique.pop(1)
```
effigies: I don't really follow this guard. Can you elaborate?

mgxd: I'm not sure this addresses the problem properly / gracefully, but I'm unsure how to proceed. I've come across a derived image included in the diffusion DICOM, which is currently not properly handled. I thought this meant 5D data wasn't yet handled, but in fact there is a test specifically for this introduced in 3f04e3c60acd7d28204d0c81d4294cb6e7397cfe (maybe @matthew-brett can share some wisdom)
https://github.com/nipy/nibabel/blob/ad0b13f24a7b91a7be9ce70a46bb79901b4b5f48/nibabel/nicom/tests/test_dicomwrappers.py#L656-L682

bmelo:
I tested the last commit with the same dataset I was working on when I opened the issue nipy/heudiconv#275. Now I can convert our images to BIDS without problems.

@mgxd, thank you so much!

effigies:
@mgxd Is this still a WIP, or should I review?

mgxd:
ready for a review

nibabel/nicom/dicomwrappers.py
```diff
@@ -502,8 +503,31 @@ def image_shape(self):
...
+            try:
+                self.frames = Sequence(
+                    frame for frame in self.frames if
+                    frame.MRDiffusionSequence[0].DiffusionDirectionality
+                    != 'ISOTROPIC'
+                    )
+                if n_frames != len(self.frames):
+                    warnings.warn("Derived images found and removed")
+                    n_frames = len(self.frames)
+                    has_derived = True
```
effigies: Can you move this `if` block to an `else` block after `except AttributeError`?

nibabel/nicom/dicomwrappers.py
```diff
@@ -522,6 +546,9 @@ def image_shape(self):
         if stackid_tag in dim_seq:
             stackid_dim_idx = dim_seq.index(stackid_tag)
             frame_indices = np.delete(frame_indices, stackid_dim_idx, axis=1)
+        if has_derived:
+            # derived volume is included
+            frame_indices = np.delete(frame_indices, 1, axis=1)
```
effigies: It is always the second entry?

effigies:
@mgxd If you have a few minutes, I think we can probably finish this up today. Just a question and a minor cleanup.

nibabel/nicom/dicomwrappers.py
mgxd: Based off my limited testing yes - however we should definitely keep an eye on this

effigies: Can we use either a derived index or make an assertion that will fail loudly when the assumption is violated?

effigies:
LGTM. Merging.

effigies:
Ugh, saw after merging that your test file is 40MB. Had to force-push master back to pre-merge state. Can you find/produce a smaller version of the file, and open a new PR?

effigies:
@mgxd Just a bump on this. I believe we were looking for a (possibly truncated) file with a compressed data frame.

effigies:
Hi @mgxd, any news here?

mgxd:
I'll see if I can get to it this weekend

effigies:
Hi @mgxd. Sorry to take so long to think of this, but what if we just create a `nibabel-data` submodule? I can do it, if you let me know the license.

mgxd:
That sounds like a good idea.

@bmelo any restriction with the provided DICOM?

bmelo:
No, you can use it without restrictions.

effigies:
Okay, I published under CC0: https://github.com/effigies/nitest-dicom

@bmelo, please let me know if you have any preference for being credited.

bmelo:
I have no preference. Thank you!

effigies:
Cool. Feel free to open a PR to add your affiliation, or any provenance of the file here: https://github.com/effigies/nitest-dicom/blob/master/CREDITS

Thanks for providing this test file.

PR #728, opened 2019-02-26
FIX: Accommodate Python 3 in DICOM slice sorting

reddigari:
Looks like `list.sort()` in Python 3 requires a `key` instead of `cmp` parameter, which has to be given as a keyword argument.

The function that uses them (`nicom.dicomreaders.slices_to_series()`) doesn't have a test, and I wasn't sure whether there was any appropriate test data to use in one, so I didn't write one. I'd be happy to if someone has thoughts on what data to use.


effigies:
Thanks for this. Can you provide a minimal reproduction of the issue you ran into? Given that, it might be easier to write a test.

reddigari:
I can't provide the DICOMs I was using, but any should do:
```Python
import glob
from nibabel.nicom import dicomreaders

fnames = glob.glob("*.dcm")
wrappers = [dicomreaders.wrapper_from_file(f) for f in fnames]
series = dicomreaders.slices_to_series(wrappers)
```
Output before commit:
```
  File nibabel_test.py, line 9, in <module>
    series = dicomreaders.slices_to_series(wrappers)
  File /home/brainspec/Tools/nibabel/nibabel/nicom/dicomreaders.py, line 149, in slices_to_series
    vol_list.sort(_slice_sorter)
TypeError: must use keyword argument for key function
```

effigies:
Cool. There are some `.dcm` files in [nibabel/tests/data/](https://github.com/nipy/nibabel/tree/master/nibabel/tests/data), if you'd like to put together a test. That file looks pretty under-covered.

reddigari:
Apologies for the gross derivation of the test data path, all the other tests in that suite use data in `nicom/tests/data`, so I had to use relative paths back up to `nibabel/tests/data`.

Not sure why coverage didn't change; I can't find any other tests that call `slices_to_series`.

effigies:
So it looks like there are actually separate DICOMs in [`nibabel/nicom/tests/data`](https://github.com/nipy/nibabel/tree/master/nibabel/nicom/tests/data), and this directory is already available through the `IO_DATA_PATH` variable.

As to the coverage, it [did update](https://codecov.io/gh/nipy/nibabel/pull/728/diff), but because AppVeyor is failing, CodeCov isn't updating the comment.

The AppVeyor failure looks unrelated, so don't worry about that.

reddigari:
It was my understanding that this function is meant to take single-slice dicoms from one or more series, separate them into series, and then sort the slices (apparently only along the z-dimension). The two test files I used in [`nibabel/tests/data`](https://github.com/nipy/nibabel/tree/master/nibabel/tests/data) are from the same series, so I thought they'd be the better choice.

Additionally, one of the two dicoms in [`nibabel/nicom/tests/data`](https://github.com/nipy/nibabel/tree/master/nibabel/nicom/tests/data) is a multiframe dicom, so it's less clear to me how the slice sorter should handle it.

Let me know your thoughts on how/whether to proceed. Perhaps we can copy the two files I used into the nicom test data directory, use all the .dcm files (except maybe the multiframe one), and add a test to make sure it separates the series?

reddigari:
We could duplicate one of the slices, and append it to the list of wrappers. That should register as a duplicated z-value, and call `_third_pass()`. We can probably fidget with its `instance_number` attribute to fully cover the function.

I might be able to tackle that some time this weekend, if you think it's the right move.

effigies:
On the whole more coverage is better, but it should be a valid DICOM. I'll confess to knowing little about the format, so I don't know if what you describe is normal or something that will technically work but isn't actually valid. Another thing to consider is looking through PyDICOM's [test data](https://github.com/pydicom/pydicom/tree/master/pydicom/data/test_files) directory for something that triggers this code.

effigies:
I'm going to go ahead and merge. If you get a chance to add a new test file, it would be highly appreciated.

PR #736, opened 2019-03-01
ENH: Add NIFTI_XFORM_TEMPLATE_OTHER xform code

### LINKED ISSUE #730: New NIFTI_XFORM_ code under consideration
There is an open [discussion](https://www.nitrc.org/forum/forum.php?thread_id=10029&forum_id=1942) on NITRC for a `NIFTI_XFORM` code 5 that will indicate alignment to an unspecified template. This use case has largely been relegated to `NIFTI_XFORM_ALIGNED_ANAT` (2) up to now.

I'm overall in support of this move, and it should be a very minimal effort to support these files. More importantly, if it happens, nibabel should be able to handle these files gracefully, so we should move quickly to adopt. This is particularly urgent in that we should have at least one Python 2-compatible release with this support, so that users on legacy operating systems that do not make Python 3 easy to get will be able to work with these files with a simple `pip install`.

Please contribute to the discussion on NITRC if you have any thoughts relevant to the standard, and here if you have any thoughts relevant to nibabel in particular.

effigies:
This is an initial implementation of the proposed xform code 5. The test might be redundant, but I didn't immediately see anything that would exercise this addition, so it's just a simple verification that any defined code can be set without raising errors and retrieved reliably.

I used the macro name I proposed, but am content to revert to `NIFTI_XFORM_GEN_STANDARD`, if that's the consensus.

Closes #730.

cc @mrneont @afni-rickr @satra 

Ref: [NITRC thread 10029 - Proposed (minor) update for NIFTI sform values](https://www.nitrc.org/forum/forum.php?thread_id=10029&forum_id=1942)

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -125,7 +125,9 @@ xform_codes = Recoder((  # code, label, niistring
     (1, 'scanner', "NIFTI_XFORM_SCANNER_ANAT"),
     (2, 'aligned', "NIFTI_XFORM_ALIGNED_ANAT"),
     (3, 'talairach', "NIFTI_XFORM_TALAIRACH"),
-    (4, 'mni', "NIFTI_XFORM_MNI_152")), fields=('code', 'label', 'niistring'))
+    (4, 'mni', "NIFTI_XFORM_MNI_152"),
+    (5, 'template', "NIFTI_XFORM_TEMPLATE_OTHER"),
+    ), fields=('code', 'label', 'niistring'))
 
 # unit codes
 unit_codes = Recoder((  # code, label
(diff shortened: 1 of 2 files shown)


effigies:
Added to the official `nifti1.h` in https://github.com/NIFTI-Imaging/nifti_clib/commit/984304164a72816fca80e2448d83f23e8232b5a4.

effigies:
Now that this is appearing in the official repository, so we have macro name confirmation, I'd like to go ahead and get this in for the 2.4 release. Anybody up for a review?

effigies:
Review is looking unlikely. Here we go.

PR #738, opened 2019-03-08
RF: Centralize import of decorators from numpy.testing

yarikoptic:
I kept receiving messages that imports from numpy.testing.decorators are
deprecated (since 1.5 IIRC), so I found that some places still import them
in the test files instead of centraly from nibabel.testing which already
provides imports adaptor.  So I RFed to avoid such imports, and moved a
stub for @slow there too

### DIFF
--- a/nibabel/testing/__init__.py
+++ b/nibabel/testing/__init__.py
@@ -21,8 +21,18 @@ from numpy.testing import assert_array_equal
 try:
     from numpy.testing import dec
     skipif = dec.skipif
+    slow = dec.slow
 except ImportError:
     from numpy.testing.decorators import skipif
+
+    # Recent (1.2) versions of numpy have this decorator
+    try:
+        from numpy.testing.decorators import slow
+    except ImportError:
+        def slow(t):
+            t.slow = True
+            return t
+
 # Allow failed import of nose if not now running tests
 try:
     from nose.tools import (assert_equal, assert_not_equal,
--- a/nibabel/testing/__init__.py
+++ b/nibabel/testing/__init__.py
@@ -23,15 +23,7 @@ try:
     skipif = dec.skipif
     slow = dec.slow
 except ImportError:
-    from numpy.testing.decorators import skipif
-
-    # Recent (1.2) versions of numpy have this decorator
-    try:
-        from numpy.testing.decorators import slow
-    except ImportError:
-        def slow(t):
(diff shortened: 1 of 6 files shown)


nibabel/testing/__init__.py
```diff
@@ -21,8 +21,18 @@
 try:
     from numpy.testing import dec
     skipif = dec.skipif
+    slow = dec.slow
 except ImportError:
     from numpy.testing.decorators import skipif
+
+    # Recent (1.2) versions of numpy have this decorator
```
effigies: We have a minimum version of 1.7.1. I don't think this is guard is necessary anymore.

effigies:
Oh, the test failure is just the usual AppVeyor nonsense. I thought there was a problem with a recent version of matplotlib. Might be thinking of another project.

nibabel/testing/__init__.py
```diff
@@ -21,8 +21,10 @@
 try:
     from numpy.testing import dec
     skipif = dec.skipif
+    slow = dec.slow
 except ImportError:
-    from numpy.testing.decorators import skipif
+    from numpy.testing.decorators import (skipif, slow)
```
effigies: Something's weird with the testing because this is never getting hit, even on the numpy==1.7.1 build. See 2007.5, here:

https://codecov.io/gh/nipy/nibabel/commit/0c36551c7398dcfb132fe4ce37b115c23d183b72/build

I know it's not really relevant to this PR, but any thoughts? I checked the [travis log](https://api.travis-ci.org/v3/job/503684117/log.txt), and see no indication that anything other than 1.7.1 was installed.

yarikoptic: No immediate idea, I am away from the laptop ATM, but if I were with it I would remove try/except and see how it fails

effigies: Ah, it's because `numpy.testing.dec` has existed [forever](https://github.com/numpy/numpy/blob/v1.7.1/numpy/testing/__init__.py#L10). It's just an alias of an imported module, and the module it's aliasing has changed from `numpy.testing.decorators` to [`numpy.testing._private.decorators`](https://github.com/numpy/numpy/blob/master/numpy/testing/__init__.py#L13).

How about we just drop the `try`/`except` bit, and load `dec`, `skipif`, and `slow`?

PR #739, opened 2019-03-12
BF: allow for BrainModels or Parcels  to contain a single vertex

MichielCottaar:
Fixes an edge case in the reading of CIFTI files.
Tests for this edge case have been added for both the BrainModel and Parcel axes.

### DIFF
--- a/nibabel/cifti2/parse_cifti2.py
+++ b/nibabel/cifti2/parse_cifti2.py
@@ -517,7 +517,7 @@ class Cifti2Parser(xml.XmlParser):
             # conversion to numpy array
             c = BytesIO(data.strip().encode('utf-8'))
             vertices = self.struct_state[-1]
-            vertices.extend(np.atleast_1d(np.loadtxt(c, dtype=np.int)))
+            vertices.extend(np.loadtxt(c, dtype=np.int, ndmin=1))
             c.close()
 
         elif self.write_to == 'VoxelIndices':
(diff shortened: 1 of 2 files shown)


nibabel/cifti2/parse_cifti2.py
```diff
@@ -517,7 +517,7 @@ def flush_chardata(self):
             # conversion to numpy array
             c = BytesIO(data.strip().encode('utf-8'))
             vertices = self.struct_state[-1]
-            vertices.extend(np.loadtxt(c, dtype=np.int))
+            vertices.extend(np.atleast_1d(np.loadtxt(c, dtype=np.int)))
```
effigies:
```suggestion
            vertices.extend(np.loadtxt(c, dtype=np.int, ndmin=1))
```

nibabel/cifti2/parse_cifti2.py
```diff
@@ -531,7 +531,7 @@ def flush_chardata(self):
             # conversion to numpy array
             c = BytesIO(data.strip().encode('utf-8'))
             index = self.struct_state[-1]
-            index.extend(np.loadtxt(c, dtype=np.int))
+            index.extend(np.atleast_1d(np.loadtxt(c, dtype=np.int)))
```
effigies:
```suggestion
            index.extend(np.loadtxt(c, dtype=np.int, ndmin=1))
```

MichielCottaar:
Well spotted, that indeed has the same effect (sorry, I only noticed that I could simply have accepted your suggested changes after pushing a commit based on your update)

effigies:
LGTM. Thanks.

PR #741, opened 2019-03-17
FIX: Correctly reorient dim_info when reorienting NIfTI images

### LINKED ISSUE #716: nifti dim_info transformation after orientation
Running the test
```python
import numpy as np
import nibabel as nib

data = np.arange(24).reshape((2,3,4))
img = nib.Nifti1Image(data, np.eye(4))
img.header.set_dim_info(1, 0, 2)

ornt = np.array([[2, 1], [0, 1], [1, 1]])
img_ornt = img.as_reoriented(ornt)
print(ornt[:,0])
print(img.shape)
print(img_ornt.shape)
print(img.header.get_dim_info())
print(img_ornt.header.get_dim_info())
```
we get
```python
[2 0 1]
(2, 3, 4)
(3, 4, 2)
(1, 0, 2)
(1, 2, 0)
```
although (if I'm not wrong) the correct output should look like
```python
[2 0 1]
(2, 3, 4)
(3, 4, 2)
(1, 0, 2)
(0, 2, 1)
```
(note the difference in the last line).

Shouldn't the following
https://github.com/nipy/nibabel/blob/5ab94146d569dc905bfbbb5b7a4ae4b190bbd133/nibabel/nifti1.py#L2023
be `new_dim[idx] = ornt[value, 0]`?

[In which case, it would be better to ensure that value has an integer data type; however, `io_orienatation` returns a float array.]

effigies:
Starting with a test to demonstrate the failure. Will push a fix shortly.

cc @constracti

Fixes #716.

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -2014,13 +2014,9 @@ class Nifti1Pair(analyze.AnalyzeImage):
             return img
 
         # Also apply the transform to the dim_info fields
-        new_dim = list(img.header.get_dim_info())
-        for idx, value in enumerate(new_dim):
-            # For each value, leave as None if it was that way,
-            # otherwise check where we have mapped it to
-            if value is None:
-                continue
-            new_dim[idx] = ornt[value, 0]
+        new_dim = [
+            None if orig_dim is None else int(ornt[orig_dim, 0])
+            for orig_dim in img.header.get_dim_info()]
 
         img.header.set_dim_info(*new_dim)
 
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -2020,7 +2020,7 @@ class Nifti1Pair(analyze.AnalyzeImage):
             # otherwise check where we have mapped it to
             if value is None:
                 continue
-            new_dim[idx] = np.where(ornt[:, 0] == idx)[0]
+            new_dim[idx] = ornt[value, 0]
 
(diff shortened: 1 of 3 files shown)


effigies:
Anybody up for a review on this?

effigies:
@markhymers This was originally your contribution. Would you care to check my logic, and make sure that I'm not breaking things?

effigies:
Thanks for the review.

I thought the confusion was that the freq/phase/slice ordering will only sometimes rotate in the same way as the shapes, but it turns out that orientations also didn't work how I thought... The issue was that I keep thinking of the orientation produced by axcodes2ornt as how to get from RAS to a given axis code, but it's exactly the reverse. So my consistency check was off, which was why your fix didn't seem to work. By instead cycling through orientations and getting the axis codes from the affine, I can now verify your proposed fix was correct.

effigies:
@constracti I found 46476ccf812f3b624ed3e4210ad1b1cbf5216d8c in your history, so I went ahead and cherry-picked it and got rid of my flailing about for a wrong solution. I did still tack on the refactor to a list comprehension and casting to `int()`, as orientation matrices may not be integer-typed. If you want to let me know your preferred name for publication, your [ORCID](https://orcid.org/) and your affiliation, I'll update the mailmap and Zenodo file as needed.

effigies:
Any more comments?

effigies:
Thanks for the contribution and feedback.

PR #696, opened 2018-11-27
CI: Move AppVeyor to Miniconda instead of native Python installations

effigies:
Fixes #665.

Closes #676.
Closes #680.

PR #680, opened 2018-10-13
DEBUG: Revert to pre-error trees and enable AppVeyor debugging

effigies:

PR #745, opened 2019-03-21
CI: Virtualenvs in AppVeyor

effigies:
This place is not a place of honor.

This is an experiment along the lines of #676, #677, #680 #696.


effigies:
Well, tests pass, so it at least won't cause harm. Perhaps the issue was somehow two copies of numpy somehow getting loaded and compared against each other. Merging, in hopes that the stochastic failures will go away.

effigies:
Abandon all hope. https://ci.appveyor.com/project/nipy/nibabel/builds/23292543/job/0ypumy06fg2dsjan

PR #749, opened 2019-04-23
 DOC: Update Sphinx config to support recent Sphinx/numpydoc

effigies:
Builds are breaking on master due to the numpydoc 0.9.0 release and https://github.com/numpy/numpydoc/issues/210. We have an old constraint keeping us at Sphinx < 1.6 (added a4a667df060c666adbb2f2bca7c797a48a6179ed), which is exactly the new minimum version required by numpydoc. I've tested with removing the constraint and LaTeX support seems to have been restored.

Additional changes:

* The `html_sidebar` options have been updated, and the old options deprecated at 2.0, so the new config/template maintains the current sidebar.
* The autodoc module now triggers a `mock` import, so I just added it to the dependencies.

PR #750, opened 2019-04-24
FIX: Ensure loaded GIFTI files expose writable data arrays

### LINKED ISSUE #746: Cannot set values in GIFTI data array loaded from file
Dear numpy experts, 

I am running fmriprep on singularity with a version 3.01 image

and my whole 76 subjects dataset run in parallel are crashing on our HPC and give this error message: 

"
Node: 
fmriprep_wf.single_subject_109_wf.func_preproc_ses_1_task_restingstate_wf.bold_surf_wf.medial_nans
Working directory: /mnt/data/loic2/work/fmriprep_wf/single_subject_109_wf/func_preproc_ses_1_task_restingstate_wf/bold_surf_wf/_hemi_rh/medial_nans

Node inputs:

in_file = ['/mnt/data/loic2/work/fmriprep_wf/single_subject_109_wf/func_preproc_ses_1_task_restingstate_wf/bold_surf_wf/_hemi_rh/sampler/mapflow/_sampler0/rh.fsaverage5.gii']
subjects_dir = <undefined>
target_subject = ['fsaverage5']

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 473, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 1253, in _run_interface
    self.config['execution']['stop_on_first_crash'])))
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 1128, in _collate_results
    for i, nresult, err in nodes:
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/utils.py", line 99, in nodelist_runner
    result = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 473, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 557, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 637, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 369, in run
    runtime = self._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/freesurfer.py", line 328, in _run_interface
    newpath=runtime.cwd)
  File "/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/freesurfer.py", line 463, in medial_wall_to_nan
    darray.data[medial] = np.nan
ValueError: assignment destination is read-only
"
Would you know how to fix it? 
It looks like there was a modification of numpy, isn't it?

Thanks in advance for your help!

effigies:
Fixes #746.

### DIFF
--- a/nibabel/gifti/parse_gifti_fast.py
+++ b/nibabel/gifti/parse_gifti_fast.py
@@ -33,42 +33,40 @@ class GiftiParseError(ExpatError):
 
 def read_data_block(encoding, endian, ordering, datatype, shape, data):
     """ Tries to unzip, decode, parse the funny string data """
-    ord = array_index_order_codes.npcode[ordering]
     enclabel = gifti_encoding_codes.label[encoding]
+    dtype = data_type_codes.type[datatype]
     if enclabel == 'ASCII':
         # GIFTI_ENCODING_ASCII
         c = StringIO(data)
-        da = np.loadtxt(c)
-        da = da.astype(data_type_codes.type[datatype])
+        da = np.loadtxt(c, dtype=dtype)
         return da  # independent of the endianness
-
-    elif enclabel == 'B64BIN':
-        # GIFTI_ENCODING_B64BIN
-        dec = base64.b64decode(data.encode('ascii'))
-        dt = data_type_codes.type[datatype]
-        sh = tuple(shape)
-        newarr = np.frombuffer(dec, dtype=dt)
-        if len(newarr.shape) != len(sh):
-            newarr = newarr.reshape(sh, order=ord)
-
-    elif enclabel == 'B64GZ':
-        # GIFTI_ENCODING_B64GZ
-        # convert to bytes array for python 3.2
-        # http://www.diveintopython3.net/strings.html#byte-arrays
-        dec = base64.b64decode(data.encode('ascii'))
-        zdec = zlib.decompress(dec)
-        dt = data_type_codes.type[datatype]
-        sh = tuple(shape)
-        newarr = np.frombuffer(zdec, dtype=dt)
-        if len(newarr.shape) != len(sh):
-            newarr = newarr.reshape(sh, order=ord)
 
     elif enclabel == 'External':
         # GIFTI_ENCODING_EXTBIN
         raise NotImplementedError("In what format are the external files?")
 
-    else:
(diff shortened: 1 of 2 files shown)


effigies:
So I dug a bit into our options for finding ways of translating from B64BIN/B64GZ to a bytearray with fewer copies, but at least the out-of-the-box attempts like `BytesIO` + `GzipFile` will make copies internally and seem unlikely to save us much.

On the plus side, data arrays will rarely get larger than a point set (up to 163842x3 float64s) or triangle mesh (~330000x3 int32s), which will each be less than 4MB, so one or two unnecessary copies are unlikely to cause an out-of-memory error, and the XML parsing will probably be more expensive.

That said, if anybody knows some clever tricks, I'm happy to hear them.

This is ready for review. Travis failures will be fixed with #749.

pauldmccarthy:
I tried and failed to fix this issue with `indexed_gzip`, but I'm wrapping the `IndexedGzipFile` class with an `io.BufferedReader` which, regardless of what I do in the `IndexedGzipFile` class, will always return a `bytes` object :(

However, I personally don't need the data to be writeable - the only reason I was bitten by this is because `pyopengl` currently (and unnecessarily) requires writeable arrays.  Personally, I'd be inclined to just return read-only data - if the caller needs the data to be writeable, they can explicitly take on the burden of creating the copy.

But as you said, these arrays aren't very big most of the time, so copying is unlikely to be a bottleneck.

effigies:
@pauldmccarthy Pretty much all of our image types expose writable data arrays, so I'm inclined to call their being read-only a bug.

effigies:
Any further comments/reviews?

PR #753, opened 2019-04-25
MRG: Safer warning registry manipulation when checking for overflows

### LINKED ISSUE #616: Bug: get an IndexError when I should be getting a ValueError
This line shouldn't use `.pop()` as there's no guarantee the element being popped is there. This fails to throw the ValueError and instead throws an IndexError which is not caught by my application which puts `nib.load` in a try except using this error `nib.filebasedimages.ImageFileError`. Removing the line fixes the issue but perhaps there's a better solution. https://github.com/nipy/nibabel/blob/949d466ee9c1d21a85ccb03a03f438aa01595782/nibabel/volumeutils.py#L1356

effigies:
Following @matthew-brett's suggestion in https://github.com/nipy/nibabel/issues/616#issuecomment-379192456.

This approach unconditionally prepends the filter, which hopefully reduces the odds that removing it at the end will interfere with any other threads attempts to modify the registry. In particular, it should be re-entrant, so multiple threads calling this same function should not interfere with each other unless they happen to race on modifying the list. We could add a lock around the insertion/removal to avoid self-interference, though as noted in the other thread, no other libraries will respect it.

Looking at Python 3.4+, adding filters to the filter list is followed by a call to `warnings._filters_mutated()`, so I'm adding that behavior in, even though it's not part of the API. When Python 2 is dropped, the `getattr` part can be removed, though perhaps it's safer to do this for private functions anyway.

Sorry it took so long to get back to this. I completely forgot about that discussion.

@tomvars In case you're still having this issue, could you verify that this resolves it for you?

Closes #616.

### DIFF
--- a/nibabel/tests/test_volumeutils.py
+++ b/nibabel/tests/test_volumeutils.py
@@ -1256,7 +1256,7 @@ def test__ftype4scaled_finite_warningfilters():
     # 32MiB reliably produces the error on my machine; use 128 for safety
     shape = (1024, 1024, 32)
     tst_arr = np.zeros(shape, dtype=np.float32)
-    # Ensure that an overflow will happen
+    # Ensure that an overflow will happen for < float64
     tst_arr[0, 0, 0] = np.finfo(np.float32).max
     tst_arr[-1, -1, -1] = np.finfo(np.float32).min
     go = threading.Event()
@@ -1274,7 +1274,7 @@ def test__ftype4scaled_finite_warningfilters():
         def run(self):
             go.wait()
             try:
-                # Use float16 to buy us two failures
+                # Use float16 to ensure two failures and increase time in function
                 _ftype4scaled_finite(tst_arr, 2.0, 1.0, default=np.float16)
             except Exception as e:
                 err.append(e)
--- a/nibabel/tests/test_volumeutils.py
+++ b/nibabel/tests/test_volumeutils.py
@@ -1253,12 +1253,12 @@ def test_array_from_file_overflow():
 def test__ftype4scaled_finite_warningfilters():
     # This test checks our ability to properly manage the thread-unsafe
     # warnings filter list.
-    # 32MiB reliably produces the error on my machine; use 128 for safety
-    shape = (1024, 1024, 32)
-    tst_arr = np.zeros(shape, dtype=np.float32)
+
+    # _ftype4scaled_finite always operates on one-or-two element arrays
     # Ensure that an overflow will happen for < float64
-    tst_arr[0, 0, 0] = np.finfo(np.float32).max
-    tst_arr[-1, -1, -1] = np.finfo(np.float32).min
+    finfo = np.finfo(np.float32)
+    tst_arr = np.array((finfo.min, finfo.max), dtype=np.float32)
+
     go = threading.Event()
     stop = threading.Event()
     err = []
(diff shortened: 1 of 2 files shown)


effigies:
Wrote a test that obliterates the warning filters while we are in this section. It reliably produces errors on my machine that are resolved by this fix. (The warnings appear in the test output, but we expect sub-optimal behavior with badly behaving threads.)

CI for tests without fix:

* [AppVeyor](https://ci.appveyor.com/project/nipy/nibabel/builds/24149700)
* [Travis](https://travis-ci.org/nipy/nibabel/builds/525486820)

Reviews welcome.

effigies:
Ah. Realized I forgot to finish this comment to explain the new approach:

----

Thinking more about this, I decided to go in a different direction. Since we have no hope of being thread-safe, we can use context managers with very small scopes to modify the warning filter list and revert it for us, rather than manually inserting and removing and making hidden API calls.

Considering the cases of two racing threads, A using this function and B in other ways modifying the filter list:

* A enters first, then
  * B modifies, executes and reverts before A exits: Expected behavior
  * B modifies before A exits and never reverts: B's changes are lost
  * B modifies before A exits and reverts after A exits: A's changes are restored
* A enters after B modifies
  * B never reverts: Expected behavior
  * A exits before B reverts: Expected behavior
  * B reverts and then A exits: A's changes are lost, and B's are restored

This last case permits a failure to detect an overflow via catching the warning raised as an exception, so the isfinite check is left in.

In the interleaved cases (A modifies, B modifies, A reverts, B reverts; and reversed) spurious filters will be left in the list. Not much getting around it. Our biggest threat really is B leaving an `'error'` case in that bites apps later. It's possible we could try removing afterward anyway to head off that threat.

effigies:
Anybody up for a review?

nibabel/volumeutils.py
```diff
@@ -1334,26 +1334,37 @@ def _ftype4scaled_finite(tst_arr, slope, inter, direction='read',
...
+    overflow_filter = ('error', '.*overflow.*', RuntimeWarning)
+    for ftype in OK_FLOATS[def_ind:]:
+        tst_trans = tst_arr.copy()
+        slope = slope.astype(ftype)
+        inter = inter.astype(ftype)
+        try:
             if direction == 'read':  # as in reading of image from disk
                 if slope != 1.0:
-                    tst_trans = tst_trans * slope
+                    # Keep warning contexts small to reduce the odds of a race
```
matthew-brett: Is the repetition really necessary?  The test array is always two elements, so the whole thing adds up to (2 multiplies 2 adds) or  (2 subtracts and 2 divides), with an if statement.  This is pretty damn fast:

```
[ins] In [1]: import numpy as np                                                                                       
[ins] In [2]: ii = np.iinfo(np.int64)                                                                                  
[ins] In [3]: tst_arr = np.array([ii.min, ii.max], dtype=np.int64)    
[nav] In [6]: def do_mult(tst_arr): 
         ...:     if True: 
         ...:         res = tst_arr - 1 
         ...:         res = tst_arr / 0.1 
         ...:                                                                                                          
[ins] In [7]: timeit do_mult(tst_arr)                                                                                  
1.79 µs ± 25 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Presumably this is a slight overestimate, with the function call overhead.  Compare to:

```
[ins] In [8]: timeit tst_arr / 0.1                                                                                     
988 ns ± 22.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)                              
```

So - I think the time window reduction is very small.

nibabel/tests/test_volumeutils.py
```diff
@@ -1245,3 +1248,42 @@ def read(self, n_bytes):
                  'Expected {0} bytes, got {1} bytes from {2}\n'
                  ' - could the file be damaged?'.format(
                      11390625000000000000, 0, 'object'))
+
+
+def test__ftype4scaled_finite_warningfilters():
```
matthew-brett: Could you explain a little more why you need the big array here?

effigies: This was to force the job to take long enough I could reliably produce a race that resulted in an error condition.

nibabel/volumeutils.py
effigies: Okay, I see that I failed to follow all the logic and realize this is only ever used on a 2-element array. I'll wrap the entire branch.

effigies:
@matthew-brett I switched to a more realistic test array, and looped the function call to increase the odds of a race.

Let me know what you think.

effigies:
@matthew-brett Any chance of another look?

matthew-brett:
Looks fine - merging - thanks.

PR #755, opened 2019-04-26
MAINT: Effect FutureWarnings, including deprecations

### LINKED ISSUE #524: Transition FutureWarnings older than X versions to Deprecations?
Continuing on from the `ArrayProxy.header` deprecation in #521, it'd probably be good to move `FutureWarning`s from <=2.0 (2.1?) to `DeprecationWarning`s, more generally.

Not sure how to use the versioned deprecation machinery to deprecate modules (looking at `nibabel.minc`).

effigies:
We've been threatening changes for a while, so we can go ahead and effect some.

I'll have another look through for `FutureWarnings` to make sure they're appropriate.

The newish deprecation machinery transitions to an error on use when the `until` version is hit, which means we should keep those in a bit longer to ensure people see the error messages. Some of the things I've moved to that machinery promised removal at 3.0, so I'll also make sure we have tests that will fail if they still exist in the release branch.

Closes #524.


effigies:
Only pre-test failing (#768), so this is ready for review.

I think `nibabel.trackvis` are the changes most warranting careful attention.

I've also added a test file to schedule removal of modules and other API elements. I'd considered trying something a bit more automatic, e.g., `deprecate_with_version` producing appropriate tests, but don't have the energy for that at the moment. A single place to record intent to deprecate I think is sufficient for now.

Issue #756, opened 2019-05-01
Cifti2Image.from_filename does not accept mmap option


effigies:
Analyze-based images (including `Nifti2Image`) permit a `mmap` option for `from_filename` and `from_filemap`. `Cifti2Image` should allow the same. This should be an easy fix, simply adding the option, passing it through to `Nifti2Image`, and updating the docstrings.

I will address this today and target a 2.4.1 release. Please chime in if you have concerns.

effigies:
Note that this was discovered in poldracklab/niworkflows#350 by @oesteban.

effigies:
For `nibabel.load` to accept this option generally, we'll need to either allow it as an option to all `FileBasedImage`s, even if they don't support `mmap`ing, or have some way of pruning non-applicable kwargs before calling `image_klass.from_filename`.

PR #759, opened 2019-05-18
RF: Move mmap/keep_file_open parameters to DataobjImage.from_file*

effigies:
In this PR I propose to move the `mmap` and `keep_file_open` parameters to `Image.from_filename` and `Image.from_file_map` class methods to all `DataobjImage`. This overall cleans up a lot of replication across different subclasses, and applies naturally to the idea of images with a data object.

This will help with #756, and bring pretty much all images except GIFTI into the `mmap` fold.

These are applied on a best-effort basis, already, so I'm not overly worried about ECAT and Minc images not currently modifying behavior. Does this bother anybody?

Starting with a test to establish the failures.

### DIFF
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -988,47 +988,6 @@ class AnalyzeImage(SpatialImage):
                            'file_map': copy_file_map(file_map)}
         return img
 
-    @classmethod
-    @kw_only_meth(1)
-    def from_filename(klass, filename, mmap=True, keep_file_open=None):
-        '''Class method to create image from filename `filename`
-
-        .. deprecated:: 2.4.1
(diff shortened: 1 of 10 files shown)


pauldmccarthy:
This makes a load of sense, and is probably what we should have done with `keep_file_open` in the first place! I wasn't (and am still not) familiar enough with the class hierarchy to realise that `keep_file_open` could be pushed upwards.

effigies:
Pre-emptively merged #761 to handle conflicts. Can strip it out if we change our mind about versions.

I'll give this one two more days. If anybody else wants to review but doesn't have time before Wednesday, just let me know and I'll hold off.

effigies:
Thanks for having a look, @pauldmccarthy.

oesteban:
Nice! Thanks!

PR #761, opened 2019-05-18
MAINT: Deprecate keep_file_open == "auto" (scheduled for 2.4.0)

effigies:
I missed this in the 2.4.0 release. Our options are either to effect the deprecation for 2.4.1 or wait until 2.5.0. We've been warning in the docs for a while, so I'd rather move on to DeprecationWarning sooner than later to catch the few people who might be using the old behavior. I don't feel strongly, though, so if anybody objects, I'll delay to 2.5.

I'm suggesting that we keep the old behavior and warn, while removing descriptions of the old behavior, as there's no reason to prefer it to `False`.

@pauldmccarthy Would you care to have a look through to make sure this looks like the deprecation you envisioned?

### DIFF
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -935,7 +935,11 @@ class AnalyzeImage(SpatialImage):
     @classmethod
     @kw_only_meth(1)
     def from_file_map(klass, file_map, mmap=True, keep_file_open=None):
-        '''class method to create image from mapping in `file_map ``
+        ''' Class method to create image from mapping in ``file_map``
+
+        .. deprecated:: 2.4.1
+            ``keep_file_open='auto'`` is redundant with `False` and has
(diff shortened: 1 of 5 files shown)


pauldmccarthy:
Hey @effigies - sorry - I should have remembered to clean this up in time for 2.4.0. It looks perfect to me, thanks for sorting it out.

effigies:
Part of this was my fault for not really respecting the `maint/` branches and releasing bug fix releases off of master. I have some hopes of adding processes and tests that will do a better job of warning us about things like this in 2.5.

effigies:
Thanks for the review, btw.

I'll give this another day for objections to including in the 2.4.1 release.

effigies:
Thanks for the review, @pauldmccarthy.

PR #763, opened 2019-05-24
FIX: Correctly write .annot files with duplicate labels

### LINKED ISSUE #535: nibabel.freesurfer.write_annot successfully writes files that cannot be read by freesurfer
I will edit this issue when I can come up with a good example of this separate from my data, but the headline is that the `write_annot` function can write (without erroring) a file that Freesurfer is unable to read.

In the current case, `nib.freesurfer.read_annot` is also unable to read the written file. However, I have also experienced that `nib.freesurfer.read_annot` will read the file but it will be incorrect.

Edit: and I have now written an annotation that Freesurfer can read but nibabel cannot.

effigies:
Starting with a test, thanks to @rnemec-ng. Will follow-up with a fix proposed by @rnemec-ng when a CI cycle finishes.

Fixes #535.

### DIFF
--- a/nibabel/freesurfer/io.py
+++ b/nibabel/freesurfer/io.py
@@ -561,7 +561,7 @@ def write_annot(filepath, labels, ctab, names, fill_ctab=True):
         write(-2)
 
         # maxstruc
-        write(np.max(labels) + 1)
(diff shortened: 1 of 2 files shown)


effigies:
Thanks for the review.

PR #764, opened 2019-05-25
MAINT: Move fully to setuptools

### LINKED ISSUE #717: scripts have wrong interpreter after "conda install nibabel" in Windows
Was trying to reproduce https://github.com/datalad/datalad/issues/3116 on a freshly installed Window$ machine. Installed miniconda, created "datalad-dev" conda environment.
In regular "Anaconda Prompt" which seems to be running pure Windows cmd prompt, doesn't find/run nib-ls at all (regular entrypoint mechanism installed apps like datalad seems to work fine, but somehow it just ended up in /usr/bin/datalad using regular /usr/bin/python for interpreter -- yet to figure things out).  If I start bash within it, then I get
```
/mnt/c/Temp/datasets.datalad.org/openneuro/ds000114$ nib-ls
-bash: /mnt/c/Users/XXX/AppData/Local/Continuum/miniconda3/envs/datalad-dev/Scripts/nib-ls: C:/Users/XXX/AppData/Local/Continuum/miniconda3/envs/datalad-dev/bin/python: bad interpreter: No such file or directory
```
well -- first of all the problem is that path is windows path so not sure if this bash could comprehend it.  Then, the python is actually available not under /bin but /mnt/c/Users/XXX/AppData/Local/Continuum/miniconda3/envs/datalad-dev/python.exe

But even if I correct shebang to be a correct windows path -- it would not run in bash, and remains (well, there is no .exe etc I guess) not found by windows cmd.  But even if I fix the path to be a full path to python.exe POSIX style I can't run it in that bash
```
/mnt/c/Temp/datasets.datalad.org/openneuro/ds000114$ nib-ls
C:\Users\XXX\AppData\Local\Continuum\miniconda3\envs\datalad-dev\python.exe: can't open file '/mnt/c/Users/XXX/AppData/Local/Continuum/miniconda3/envs/datalad-dev/Scripts/nib-ls': [Errno 2] No such file or directory
```
@chrisfilo - do you have recollection on how nibabel was installed and what you have in your shebang for it?

effigies:
Dug through nisext finally, and I agree with https://github.com/nipy/nibabel/issues/717#issuecomment-472726035 that we can probably drop most of that from our setup.py. While I'm at it, I'm moving us to a `setup.cfg`-based setup.

No rush on this one. Would be nice to move by 2.5 or 3.0.

Closes #717.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,5 @@
 [metadata]
 name = nibabel
-version = attr: nibabel.__version__
 url = https://nipy.org/nibabel
 download_url = https://github.com/nipy/nibabel
 author = nibabel developers
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,79 @@
+[metadata]
+name = nibabel
+version = attr: nibabel.__version__
+url = https://nipy.org/nibabel
+download_url = https://github.com/nipy/nibabel
+author = nibabel developers
+author_email = neuroimaging@python.org
+maintainer = Chris Markiewicz
+maintainer_email = neuroimaging@python.org
+classifiers =
+    Development Status :: 4 - Beta
+    Environment :: Console
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: Scientific/Engineering
+license = MIT License
+description = Access a multitude of neuroimaging data formats
+long_description = file:README.rst
+long_description_content_type = text/x-rst; charset=UTF-8
+platforms = OS Independent
+provides =
+    nibabel
+    nisext
+
+[options]
+python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*
+install_requires =
+    numpy >=1.8
+    six >=1.3
+    bz2file ; python_version < "3.0"
+tests_require = nose
+test_suite = nose.collector
+zip_safe = False
+packages = find:
+include_package_data = True
+
+[options.extras_require]
+dicom =
+    dicom >=0.9.9
+doc =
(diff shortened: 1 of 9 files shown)


effigies:
If we are inclined to move to [versioneer](https://pypi.org/project/versioneer/), then we can drop the nisext usage altogether.

effigies:
This is all set. This sets a minimum setuptools requirement of 30.3.0, due to moving most of the metadata to [`setup.cfg`](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files). Any objections?

I've pushed off versioneer mostly because of the `upstream/2.5.0.dev` tag. I'm a little inclined to move to that after the 2.5.0 tag, which will get us off of nisext entirely and reduce the fiddliness of releasing. We can discuss that later, as well.

setup.cfg
```diff
@@ -0,0 +1,78 @@
...
+provides =
+    nibabel
+    nisext
+
+[options]
+python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*
+install_requires =
+    numpy >=1.8
+    six >=1.3
+    bz2file ; python_version < "3.0"
```
yarikoptic: Do you by any chance remember from what setuptools version this extended setup.cfg is supported? I just fear possible troubles on older systems

effigies: 30.3.0

effigies: In general, will it be a problem for older systems to upgrade setuptools? That version had a `python_requires='>=2.6,!=3.0.*,!=3.1.*,!=3.2.*'` constraint, which is a superset of our own support level.

effigies: @yarikoptic Sorry to bug you, but I just want to check whether that's sufficiently old, or you still think this could produce problems... I can revert to a pre-30.3-style setuptools config, if this is a serious concern.

yarikoptic:
Current debian stable stretch has 33. Oldstable Jessie only elderly 5 and we backported nibabel for it before http://neuro.debian.net/pkgs/python-nibabel.html but I think it is ok to stop carrying about it looking forward

effigies:
Okay. If somebody complains after 2.5 comes out, we can release a bugfix. Thanks for checking into this.

PR #774, opened 2019-07-04
ENH: Check CIFTI-2 data shape matches shape described by header

MichielCottaar:
Checks whether the shape of the data matches the shape expected based on the CIFTI-2 header
- raises an error when trying to write out a mismatched CIFTI image to disk
- raises a warning when creating a mismatched CIFTI file
I've added tests for both of these behaviors (and fixed many old tests that produced mismatched CIFTI images).

In the end I did not adopt the framework of _get_checks as this seemed to be only implemented for subclasses of WrapStruct, which are the header objects. However, this is a test that should be run on the Image class (as it requires knowing the shape of the dataobj).

### DIFF
--- a/nibabel/cifti2/cifti2.py
+++ b/nibabel/cifti2/cifti2.py
@@ -30,6 +30,7 @@ from ..dataobj_images import DataobjImage
 from ..nifti2 import Nifti2Image, Nifti2Header
 from ..arrayproxy import reshape_dataobj
 from ..keywordonly import kw_only_meth
+from warnings import warn
 
 
 def _float_01(val):
@@ -1374,11 +1375,18 @@ class Cifti2Image(DataobjImage):
         super(Cifti2Image, self).__init__(dataobj, header=header,
                                           extra=extra, file_map=file_map)
         self._nifti_header = Nifti2Header.from_header(nifti_header)
+
         # if NIfTI header not specified, get data type from input array
         if nifti_header is None:
             if hasattr(dataobj, 'dtype'):
                 self._nifti_header.set_data_dtype(dataobj.dtype)
         self.update_headers()
+
+        if self._nifti_header.get_data_shape() != self.header.matrix.get_data_shape():
+            warn("Dataobj shape {} does not match shape expected from CIFTI-2 header {}".format(
+                self._dataobj.shape, self.header.matrix.get_data_shape()
+            ))
+
 
     @property
     def nifti_header(self):
--- a/nibabel/cifti2/cifti2.py
+++ b/nibabel/cifti2/cifti2.py
@@ -1463,9 +1463,10 @@ class Cifti2Image(DataobjImage):
         extension = Cifti2Extension(content=self.header.to_xml())
         header.extensions.append(extension)
         if header.get_data_shape() != self.header.matrix.get_data_shape():
-            raise ValueError("Dataobj shape {} does not match shape expected from CIFTI-2 header {}".format(
-                self._dataobj.shape, self.header.matrix.get_data_shape()
-            ))
+            raise ValueError(
+                "Dataobj shape {} does not match shape expected from CIFTI-2 header {}".format(
+                    self._dataobj.shape, self.header.matrix.get_data_shape()
+                ))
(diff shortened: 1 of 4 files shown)


MichielCottaar:
I do think a `ConsistencyWarning` rather than a `UserWarning` would work well here, if you decide to implement one.

nibabel/cifti2/cifti2.py
```diff
@@ -1209,6 +1210,36 @@ def _to_xml_element(self):
...
+        return cifti2_axes.from_index_mapping(self.get_index_map(index))
+
+    def get_data_shape(self):
+        """
+        Returns data shape expected based on the CIFTI-2 header
+        """
+        from . import cifti2_axes
+        if len(self.mapped_indices) == 0:
+            return ()
+        base_shape = [-1 for _ in range(max(self.mapped_indices) + 1)]
```
effigies: Are all `-1`s guaranteed to be replaced?

Also:

```suggestion
        base_shape = [-1] * (max(self.mapped_indices) + 1)
```

nibabel/cifti2/cifti2.py
```diff
@@ -1345,12 +1375,18 @@ def __init__(self,
...
                                           extra=extra, file_map=file_map)
         self._nifti_header = Nifti2Header.from_header(nifti_header)
+
         # if NIfTI header not specified, get data type from input array
         if nifti_header is None:
             if hasattr(dataobj, 'dtype'):
                 self._nifti_header.set_data_dtype(dataobj.dtype)
         self.update_headers()
 
+        if self._nifti_header.get_data_shape() != self.header.matrix.get_data_shape():
```
effigies: We should compare the same objects we report:

```suggestion
        if self._dataobj.shape != self.header.matrix.get_data_shape():
```

nibabel/cifti2/cifti2.py
```diff
@@ -1426,6 +1462,11 @@ def to_file_map(self, file_map=None):
         header = self._nifti_header
         extension = Cifti2Extension(content=self.header.to_xml())
         header.extensions.append(extension)
+        if header.get_data_shape() != self.header.matrix.get_data_shape():
```
effigies:
```suggestion
        if self._dataobj.shape != self.header.matrix.get_data_shape():
```

effigies:
Also, I doubt I'm going to have time to think through and implement new warning classes before the 2.5 release. Perhaps we can revisit afterwards.

nibabel/cifti2/cifti2.py
MichielCottaar: No, the CIFTI-2 header could leave certain dimensions undefined. Similarly to headers which define the wrong array shape, this should not happen. The reason I set this to -1 is that this would automatically cause the shape check to fail (leading to a warning when loading or creating such a file and an error when trying to write it). 

Now that I think about this again, I should have set it to np.nan (which should be a bit more self-expanatory) and document this behaviour in the docstring. 

Alternatively, we could raise an error in the method if there are undefined dimensions. After all there is no valid shape. However, that would prevent people from loading files for which not all dimensions have been defined in the CIFTI-2 header.

MichielCottaar:
I'm happy to either have this pull request to remain open until you have time to decide on the new warning system. We could also just put it in with the default python warning and later replace that with a custom warning system. I have no strong preference either way.

nibabel/cifti2/cifti2.py
```diff
@@ -1209,6 +1210,38 @@ def _to_xml_element(self):
...
+        axis : :class:`.cifti2_axes.Axis`
+        '''
+        from . import cifti2_axes
+        return cifti2_axes.from_index_mapping(self.get_index_map(index))
+
+    def get_data_shape(self):
+        """
+        Returns data shape expected based on the CIFTI-2 header
+
+        Any dimensions omitted in the CIFIT-2 header will be given a default size of None.
```
effigies:
```suggestion
        Any dimensions omitted in the CIFTI-2 header will be given a default size of None.
```

effigies:
LGTM. Thanks.

PR #783, opened 2019-07-30
CI: Move from AppVeyor to Azure Pipelines

effigies:
Starting by adapting scikit learn's configs and we'll see where things go. Azure seems to be much more than pipelines, so hopefully I can figure this out...


effigies:
@skoudoro In case you're looking for as simple a model for moving to Azure as possible... Also, I've registered the nipy org for Azure, so hopefully I can make sure you have access, too.

effigies:
Note that #665 is alive and well in here.

matthew-brett:
Nice to see that #665 is still keeping in shape ... :)

effigies:
I think its stochasticity is an effect of `dict` ordering, which is why it only shows up in Python < 3.6, so I guess if we just wait for one more end-of-life the tests will be fixed.

skoudoro:
Nice 👍 , Can you add me to the nipy org for Azure @effigies? I will add DIPY project this week or next one.

effigies:
@skoudoro I'll try to figure that out. Ping me again if I haven't in a couple days.

effigies:
@skoudoro I tried to add you to the nipy organization on Azure (also @satra, @matthew-brett and @arokem, because I could think of their emails without looking them up). It is not very intuitive.

Here are the docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#organizations

Let me know if that works out for you, or if you need me to try to bump up your permissions. I'm not sure how to make you all admins, but I made you "stakeholders", which might be roughly equivalent.

PR #786, opened 2019-08-05
MAINT: Manage versioning with versioneer

effigies:
This is a follow-up to #764. Now that we have a tag post-[upstream/2.5.0.dev](https://github.com/nipy/nibabel/releases/tag/upstream%2F2.5.0.dev), Versioneer should be able to construct valid PEP-440-style version strings from any revision, and we can drop our custom githash handling code and nisext.

I don't think this should have negative impacts, but I'm going to leave this one open for a week or two in case anybody foresees problems with this move.


effigies:
This ended up leading to a sort of chicken-egg problem with long_description. Because versions are no longer set statically, `nibabel/info.py` is no longer `exec`-able, and we can't move `__version__` into `__init__.py` because of the `cmp_pkg_version` function. So I ended up just moving the canonical text of `long_description` into a `long_description.rst` file in the root, and updating `refresh_readme.py` accordingly.

effigies:
One remaining question is whether to move from the custom version comparison code to [`packaging.version.Version`](https://packaging.pypa.io/en/latest/version/#packaging.version.Version), which is the reference implementation of PEP440. Our comparison is not PEP440-compatible, as it assumes that any post-string means a pre-release version.

It would be a new dependency, but fairly lightweight and pushes correctness concerns upstream.

effigies:
No major outcry, and we're testing `git archive` now, so I think all of the old behavior is tested and has a decent fallback option. I don't think I'll bother trying to port this to 2.5.x, unless anybody feels strongly about consistent versioning for all active branches.

PR #787, opened 2019-08-06
MAINT: Require Python >= 3.5, purge Python 2 compatibility

effigies:
Summary of changes:

* Minimum Python: 3.5.1 (skipping 3.5.0 lets us drop the [gzip buffer hack](https://github.com/nipy/nibabel/blob/cee6bf3e9a028012728b9a292a9a3a5d8f52f87d/nibabel/openers.py#L50-L88))
* Minimum numpy: 1.12.0 (`assert_warns` does not work properly with multiple warnings being emitted until then; we've been somehow missing this edge case by checking minimum versions on Py2 only)
* Drop `nibabel.externals.six` altogether, along with all imports from `six`
* Switch from `nibabel.py3k` to `numpy.compat.py3k`. They don't seem to be dropping it when dropping Python 2 support so it seems safe to rely on it.
* A few hacks found with `grep -E -rIi 'py(thon)? ?[23]' nibabel | grep '\.py:'` were removed.
* Compatibility in `pkg_info` will be handled in #786.

---

* [x] Update tests
* [x] Purge `six`
* [x] Purge `py3k`
* [x] Purge `from __future__`
* [x] Check for any `PY2`/`PY3` tags indicating compatibility hacks
* [x] Update requirements


doc/source/installation.rst
```diff
@@ -81,14 +81,10 @@ is for you.
...
-    nibabel/info.py
-    requirements.txt
-    .travis.yml
-
-*  Python_ 2.7, or >= 3.4
-*  NumPy_ 1.8 or greater
-*  Six_ 1.3 or greater
+.. check these against setup.cfg
+
+*  Python_ 3.5 or greater
```
effigies:
```suggestion
*  Python_ 3.5.1 or greater
```

effigies:
Taking the plunge.

PR #790, opened 2019-08-13
Diagnose stochastic numpy dtypes

effigies:
Summary of current state:

A `PYTHONHASHSEED=283137131` will reliably produce the `AssertionError: <class 'numpy.float64'> != <class 'numpy.float64'>` error, corroborating my suspicion that insertion-ordered dictionaries are the difference between 3.4 and 3.5 and 3.6+. This is not affected by x86 vs x64 architecture.

Possibly useful, a `PYTHONHASHSEED` that did not produce the error: 1706868626

I don't have a Windows box to develop on, so if somebody wants to explore the numpy dtypes structures under both conditions, we might be able to create a simple regression test for numpy.

cc @yarikoptic @matthew-brett 

---

Now that we have Azure, we can throw a bit more resources at tracking down #665.

Following up on https://github.com/nipy/nibabel/issues/665#issuecomment-516490678.


effigies:
Looks like this was actually h5py/h5py#1051, fixed in h5py/h5py#1134, just released a couple days ago.

I don't think it was ever at risk of producing bad data, although it did fail cache checks. Do you suppose we ought to require h5py >= 2.10?

PR #791, opened 2019-08-14
CI: Test installing from git archive

effigies:
Related to #786. It looks like Versioneer doesn't handle archives that don't include tags very well, so I'm verifying the usefulness of the current setup, and we'll need to find a way to make this work with Versioneer, if it's worth moving to Versioneer at all at this point.


effigies:
Does anybody actually do this? We have the machinery though, so I figure we should test it until we remove it.

Issue #789, opened 2019-08-08
Propose gifti data API


htwangtw:
I am working with HCP Conte69 surface data in resolution that's not supported in the official HCP pipeline. I mostly work with surf.gii, shape.gii, func.gii and cifti format. I will focus on gifti in this issue as cifti is a different file format. I found the nibabel gifti function a bit difficult to use for read/write the actual data. I would like to propose a new API aiming at more intuitive access to the data and save new output. 

The most important information for me as a user is
1. the intents of the darrys
2. size of the collection of darray (currently supported through GiftiImage.numDA)
3. resolution (ie. mesh face and coordinate numbers)
4. access data as numpy arrays rather than going through all GiftiImage.darrays to concatenate the information (similar to the output from nilearn.surface.load_surface_mesh) 

## Header-like method  to access information of in point 1 to 3
The current `GiftiImage.print_summary()` method will give you all the metadata, but the only useful information is the metadata in the intents. Currently, you have to access the darrays first to get more useful information.
`GiftiImage.darrays[0].print_summary()`
Some 
I would like to have a header that can access the information about the intents.
Something like this:
```
GiftiImage.header 
GiftiImage.header.intents - what are the intents of the data array
GiftiImage.header.meta - for the metadata
```

## Access data  
For `.func.gii` and other metric files, it would be nice to concatenate all darrys for timeseires data through `get_data()`. 
For any kind of mesh (`surf.gii`), the method will return a tuple of corrdinates and faces.
The preliminary idea is as follow:
```
get_data(intent_code=None):
    timeseries = vstack(da.data for da in get_array_by_intent)
    coords = get_arrays(pointset).data
    faces = get_arrays(triangles).data
    return (coords, faces)
    if None:
        tuple(da.data for da in self.darrays)
```

## Creating gifti files
Currently I replace the data in exisitng gifti files to create new image. It would be helpful if I can just pass a dictionary with the intents as keys and the data as the entry to create a new file.  
For surface mesh (`surf.gii`), the user can pass a tuple of (coords, faces) and the funciton will translate that into data arrays. For timeseries data or other metric data (`func.gii`, `shape.gii`, and so on), we can have a np array of the size time x faces and parse each timepoint as a darray to fit the gifti format.

effigies:
Hi @htwangtw, thanks for starting this conversation. I definitely agree that GIFTI could stand a friendlier API.

[...]

PR #793, opened 2019-08-18
NF: GiftiImage method agg_data to return usable data arrays

htwangtw:
Here's the first attempt of `agg_data`related to issue #789 
I looked into other official file types and found most of them consist of one or more `DataArray`s of the same intent type. The three exceptions are 
1. `surf.gii` with two `DataArray` of different intent code (pointset, triangle)
2. `coord.gii` with one `DataArray` set to pointset 
3. `topo.gii` with `DataArray` set to triangle. 

This makes the implementation of random wild intent types slightly easier. If there is more than one data array and all intents are the same, `agg_data` will return a stacked array of all values in `darrays`. Otherwise it returns a tuple of array(s). 

Let me know what you think.


nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,43 @@ def get_arrays_from_intent(self, intent):
...
+
+        darrays = self.darrays if intent_code is None else self.get_arrays_from_intent(intent_code)
+        all_data = tuple(da.data for da in darrays)
+        all_intent = tuple(intent_codes.niistring[da.intent] for da in darrays)
+
+        # Gifti files allows usually one or more data array of the same intent code
+        # surf.gii is a special case of having two data array of different intent code
+
+        if (self.numDA > 1 and all(el == all_intent[0] for el in all_intent)):
+            return np.column_stack(all_data)
```
effigies: I'm not sure about this one. It's not clear to me that:

1) We can assume multiple arrays of the same intent will be stackable. The shapes might not work out. For example, suppose I have a GIFTI with multiple pointsets of different lengths. (I don't know why you'd want this, but it's definitely not ruled out by the standard.)
2) A user will prefer this stacking for any arbitrary intent. For instance, if I have a two-parameter statistical distribution like `NIFTI_INTENT_NORMAL`, I might get alternating columns `(mu, sigma, mu, sigma, ...)`. In that case, perhaps concatenating on a new axis would be preferred. The nice thing about time series is that the spec is pretty clear that each time step is a 1-D array. And if we see `NIFTI_INTENT_NONE`, all bets are off.

I'm hesitant to make concatenation contingent on the shapes working out, as the output type will then be an even more complex function of the input data than we're already proposing. I see a few options:

1) Try to `np.column_stack`, and if it fails, raise an error saying `agg_data` failed, so please munge the data arrays yourself.
2) `np.column_stack` for time series only (I didn't see any others that seem obvious candidates), `tuple` for everything else. This should be pretty safe.
3) Take option (2) by default, but allow it to be parameterized with an `aggregator` parameter:
    ```Python
    def agg_data(self, intent_code=None, aggregator=None):
        if aggregator is None:
            if intent_code == 'NIFTI_INTENT_TIME_SERIES':
                aggregator = np.column_stack
            else:
                aggregator = tuple
    ```
4. Require an aggregator parameter, using `tuple` by default. This would make time series a less special case, but would make the return type much more predictable.

I'm inclined toward (2), with an option to move to (3) if people actually want control over that. Not sure they will, as they can always do `np.column_stack(img.agg_data())`. WDYT?

htwangtw: I wrote something but for got to hit reply.....

Anyway, the TLDR of my original comment was, these are great suggestions! I didn't think about those weird but totally legit use of gifti format. I agreed to go for 2 for now and move to 3. I think we will need more people to use these functions to know what is common for users. 
Option 1 is just cruel. Option 4 is going to be confusing for users that don't really know the data structure.

nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,45 @@ def get_arrays_from_intent(self, intent):
         it = intent_codes.code[intent]
         return [x for x in self.darrays if x.intent == it]
 
+    def agg_data(self, intent_code=None):
+        """
+        Retrun a numpy arrary of aggregated GiftiDataArray of the same intent code
+        or
+        Retrun GiftiDataArray in tuples for surface files
+
```
effigies: It would be good to add some doctests that demonstrate how to use this. If we include a time series and a surface file, we could follow the examples of `get_fdata()`:

https://github.com/nipy/nibabel/blob/22fe8c20de0c92456a279627a70b22c3fb84dd16/nibabel/dataobj_images.py#L287-L342

I would show aggregating data:

* without intent code
* with matching intent codes
* with mismatching intent codes (should return `()`)
* with tuple intent codes

nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,45 @@ def get_arrays_from_intent(self, intent):
...
+        Parameters
+        ----------
+        intent_code : None, string, integer or tuple of string, optional
```
effigies:
```suggestion
        intent_code : None, string, integer or tuple of strings or integers, optional
```

nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,45 @@ def get_arrays_from_intent(self, intent):
...
+            Intent code, or string describing code.
```
effigies:
```suggestion
            code(s) specifying nifti intent
```

nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,45 @@ def get_arrays_from_intent(self, intent):
...
+            Accept tuple that contains multiple intents to specify the order.
```
effigies: I think if we expand the docstring above, we can leave this line out.

```suggestion
```

nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,45 @@ def get_arrays_from_intent(self, intent):
...
+
+        Returns
+        -------
+        all_data : tuple of ndarray or ndarray
```
effigies: Return values aren't named.

```suggestion
        tuple of ndarrays or ndarray
```

nibabel/gifti/gifti.py
```diff
@@ -680,6 +680,45 @@ def get_arrays_from_intent(self, intent):
...
+        # Gifti files allows usually one or more data array of the same intent code
+        # surf.gii is a special case of having two data array of different intent code
+
+        if self.numDA > 1 and len(all_intent) == 1:
+            if all_intent== 'NIFTI_INTENT_TIME_SERIES':  # stack when the gifti is a timeseries
+                return np.column_stack(all_data)
+            else:
+                return all_data
+        else:
+            return all_data
```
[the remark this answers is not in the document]
htwangtw: This is great! Cheers!

effigies:
@htwangtw Just a note that I'm going to aim to feature freeze 3.0 on October 28 (three weeks from Monday) so we can have a decent RC period. I think we're probably going to need one or two more rounds of review to get this one in.

effigies:
Also, can you merge/rebase master? Your branch is pretty far behind at this point.

htwangtw:
Thanks for the tips! I have rebased the branch to the latest master.
This is the first time I do a rebase so please let me know if there's anything not right.

effigies:
Looks like the merge went fine. I don't think you rebased, but that's okay. It was an either/or. (If you did rebase, you're going to need to force push, `git push -f`. Otherwise you're likely to run into merge conflicts next time you make a change.)

htwangtw:
Ah I see
I have now fixed the format, I will continue to work on the docs

htwangtw:
@effigies I am happy with the current state of the documentation. Let me know what you think and what can I help next.

effigies:
Hi @htwangtw. Sorry, I'm a little busy right now, so I don't have time to do a very detailed review. The tests are failing, though, which is going to need to be fixed, so I'll try to give you some pointers. If you click the "Details" of the failing tests, you should be able to see exactly what's going wrong (you will need to click through to the failing job on Travis or "View more details" on Azure). Often the same test will fail in all of them, but occasionally there are differences as some environments behave differently.

From a quick glance, a lot of it looks like it's formatting issues, still; your expected outputs just look different from the actual outputs, probably mostly through whitespace. Doctests are extremely finicky, and require every character to be right. You can use the following command to check without running all the tests:

```
nosetests --with-doctest nibabel/gifti
```

(You might need to `pip install -r dev-requirements.txt` in your environment to make sure you have all of the tools for testing.)

The Windows checks indicate there's also an issue with the checks for the test data, related to the use of `/` instead of `\\` in Windows. This should be fixable with `os.path.sep`, `os.path.join` or `pathlib.Path`.

It looks like the [coverage](https://codecov.io/gh/nipy/nibabel/pull/793/diff) is in pretty good shape, so I suspect when we get the doctests all cleaned up, this will be almost ready.

Please feel free to ask questions.

htwangtw:
No worries! This information was enough for me to figure out most of the fixes.
The Windows tests and the Python 3.5 related tests are still failing because of the docstrings. 

For windows, the failure is related to `int32`. 
```python3
>>> surf_gii_img.agg_data('triangle')
array([0, 1, 2], dtype=int32)
```
In Windows, the same command doesn't print out the `dtype=int32`.
```python3
>>> surf_gii_img.agg_data('triangle')
array([0, 1, 2])
```
This issue doesn't happen to`float32`. See the failed report [here](https://github.com/nipy/nibabel/pull/793/checks?check_run_id=262016851):
```
File "c:\hostedtoolcache\windows\python\3.5.4\x86\lib\site-packages\nibabel\gifti\gifti.py", line 775, in nibabel.gifti.gifti.GiftiImage.agg_data
Failed example:
    surf_gii_img.agg_data(('pointset', 'triangle'))
Expected:
    (array([[-16.07201 , -66.187515,  21.266994],
           [-16.705893, -66.05434 ,  21.232786],
           [-17.61435 , -65.40164 ,  21.071466]], dtype=float32), array([0, 1, 2], dtype=int32))
Got:
    (array([[-16.07201 , -66.187515,  21.266994],
           [-16.705893, -66.05434 ,  21.232786],
           [-17.61435 , -65.40164 ,  21.071466]], dtype=float32), array([0, 1, 2]))
```
If I removed `dtype=int32`, the test should fail on Linux. I looked up the existing documentations with similar output but still ain't quite sure about the solution.  
The Pyton3.5 failures are related to the decimal points in the docstrings. Python 3.5 returns less decimal points than other versons it seems. I have looked around a bit for solutions but was not very sucessful.
Please advice on solutions.

Minnor issue - The `.func.gii` test file has a really long name. I renamed it to `task.func.gii`. I hope this is an informative enough name.

effigies:
At least some of this (the changing precision) is the numpy 1.14 changes to printing. It's not actually Python 3.5, but that we test minimum requirements on Python 3.5. Until we ratchet forward to numpy 1.14, we need to roll back the defaults here:

https://github.com/nipy/nibabel/blob/debf173903b949f21ff57369f99388768263a7bc/nibabel/__init__.py#L39-L59

I think if we add `'floatmode': None` to `_test_states` and set the print option to `'fixed'`, that should resolve those issues. You'll need to update your outputs to match the old style, as well.

See:

* https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.set_printoptions.html
* https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.set_printoptions.html

For the Windows issue, I'll need to think a bit. This looks like it might need a deep hack involving [`typelessdata`](https://github.com/numpy/numpy/blob/b89caf6dd1341032fbffbe0888097a018037af00/numpy/core/arrayprint.py#L1313-L1317). If you have some time to dig into this (will be easier with a Windows machine...), feel free.

effigies:
Well that sure had unintended consequences...

Perhaps let's undo that last one, and move the difficult to display demonstrations to a test instead of in the doctest.

htwangtw:
Yep I have realised that after I pushed the commit.... 
What do you mean by moving the docstring to a test? Do you have an example?

effigies:
I just meant that we can exercise some of the behavior in [`nibabel/gifti/tests/test_gifti.py`](https://github.com/nipy/nibabel/blob/master/nibabel/gifti/tests/test_gifti.py) instead of in the doctest. This is probably the easiest way to deal with the `np.int32` showing up in Linux but not Windows.

Another option is printing things with [numpy.array2string](https://docs.scipy.org/doc/numpy/reference/generated/numpy.array2string.html) when necessary, which gives you more control over the format.

There are also some strategies for handling variable output here: https://docs.python.org/3/library/doctest.html#warnings

effigies:
This is looking great! We have full coverage of this method, which is a good place to be, and we cover a couple different types of images. I think I want to restore a couple examples in the docstring, if a little less detail to avoid the earlier issues. I can submit a PR to this branch based on your tests.

I think it's a good time to invite wider comment on the API. Does anybody have concerns? In particular, if we can think of ways to induce unintuitive output, it would be good to add to the test battery. I'm not sure we need to overly concern ourselves with truly pathological inputs, but sensible inputs producing surprising outputs is worth thinking about.

htwangtw:
This is great! We might just need to comment out the examples so the docstest won't fail. Let me know if there's a better idea.

One of the most common problems with gifti metadata assigns arbitrary intent code like `NIFTI_NORMAL` or something like that for timeseries image. Workbench can load and display those just fine despite the mistake. Any other suggestions are welcome!

effigies:
I've opened htwangtw#2 with some suggestions for updating the docstring. Feel free to merge that as-is or build on it.

I don't have any further concerns at this point, and I'm not seeing any general outcry, so I think once we finalize the docstring, this will be ready to merge.

> One of the most common problems with gifti metadata assigns arbitrary intent code like `NIFTI_NORMAL` or something like that for timeseries image.

I'm not sure we can reasonably hope to detect this case. I suspect workbench is able to do it because it's loading it *as* a time series and ignoring the intent codes altogether. Perhaps we can think of an API for easily modifying intent codes *en masse*, but that should probably be a separate PR. (We could also add warnings when loading the extension doesn't match the intents.)

htwangtw:
> I'm not sure we can reasonably hope to detect this case. I suspect workbench is able to do it because it's loading it as a time series and ignoring the intent codes altogether. 
I never really investigate how workbench under the hood for those cases. My guess is the same as yours. After thinking about it a bit, including special cases will just add more complexity of the code.

> Perhaps we can think of an API for easily modifying intent codes en masse, but that should probably be a separate PR. (We could also add warnings when loading the extension doesn't match the intents.)
Good point. Adding warning is a good idea - maybe list out the intent codes in the file when showing the warning as well? I agree this should be a separate PR.

effigies:
In it goes! Thanks very much for this.

Feel free to open a new issue if/when you want to start thinking about other methods.

PR #796, opened 2019-08-23
ENH: deprecate Wrapper.get_affine - use affine property

yarikoptic:
To stay inline with the regular interfaces

If there are maintenance releases to happen with python2 support -- this should probably cherry picked there or I could rebase -- should it be maint/2.5.x?

### DIFF
--- a/nibabel/nicom/dicomwrappers.py
+++ b/nibabel/nicom/dicomwrappers.py
@@ -96,7 +96,7 @@ class Wrapper(object):
 
     Methods:
 
-    * get_affine()
+    * get_affine() (deprecated, use affine property instead)
     * get_data()
     * get_pixel_array()
     * is_same_series(other)
@@ -105,6 +105,7 @@ class Wrapper(object):
 
     Attributes and things that look like attributes:
 
+    * affine : (4, 4) array
     * dcm_data : object
     * image_shape : tuple
     * image_orient_patient : (3,2) array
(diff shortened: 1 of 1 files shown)


effigies:
Easiest is to rebase on the maintenance branch. We can still merge that branch into master pretty easily. Can move to a backport approach once that gets painful.

yarikoptic:
> Easiest is to rebase on the maintenance branch.

maint/2.5.x ?

yarikoptic:
ok, tests against master passed. rebased, force pushed and now will change the base in the PR to maint/2.5.x

effigies:
Looks like CI got confused about the changing target branch. Bumped.

effigies:
Thanks, @yarikoptic. Would you mind updating the docstring to note that `affine` is now a property and remove `get_affine`?

On the whole, it looks like the class could use some conforming to the `SpatialImage` API, but I don't want to expand the scope of this PR. I'll open an issue.

yarikoptic:
Sorry @effigies for not reacting, thanks for pushing the tune up!

effigies:
The AppVeyor failures are fixed in #804, so the easiest fix is to merge that and merge or rebase `maint/2.5.x`.

yarikoptic:
please proceed any way you see best fit ;)

yarikoptic:
Sure, why not? ;-)

PR #798, opened 2019-08-29
BF: Fix hardcoded maximum number of CSA tags

hbraunDSP:
It looks like a hardcoded limit of 128 CSA tags was left over in one place after raising it to 1000 everywhere else.  This fixes that oversight.

### DIFF
--- a/nibabel/nicom/csareader.py
+++ b/nibabel/nicom/csareader.py
@@ -100,7 +100,7 @@ def read(csa_str):
     csa_dict['n_tags'], csa_dict['check'] = up_str.unpack('2I')
     if not 0 < csa_dict['n_tags'] <= MAX_CSA_ITEMS:
         raise CSAReadError('Number of tags `t` should be '
-                           '0 < t <= %s'%MAX_CSA_ITEMS)
+                           '0 < t <= %s' % MAX_CSA_ITEMS)
     for tag_no in range(csa_dict['n_tags']):
         name, vm, vr, syngodt, n_items, last3 = \
             up_str.unpack('64si4s3i')
--- a/nibabel/nicom/csareader.py
+++ b/nibabel/nicom/csareader.py
@@ -98,9 +98,9 @@ def read(csa_str):
         hdr_type = 1
     csa_dict['type'] = hdr_type
     csa_dict['n_tags'], csa_dict['check'] = up_str.unpack('2I')
-    if not 0 < csa_dict['n_tags'] <= 128:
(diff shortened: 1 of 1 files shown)


effigies:
This looks reasonable to me, but I don't know enough about the format. @matthew-brett Do you know if this 128 limit was supposed to be format imposed, or was just considered a reasonable upper bound at the time?

matthew-brett:
Yes, just reasonable upper bound.  We were just building the parser, and wanted to make sure it didn't run away with us, if the format was not as we expected.

effigies:
Great. @hbraunDSP is there any chance you have a test file you'd be able to share with us to add to our test suite?

hbraunDSP:
I don't have a test file but I'll definitely look into writing one.

effigies:
To be clear, I meant a DICOM file that we can add to our battery that will cause the existing tests to hit this code path, not necessarily writing a new test.

nibabel/nicom/csareader.py
```diff
@@ -98,9 +98,9 @@ def read(csa_str):
         hdr_type = 1
     csa_dict['type'] = hdr_type
     csa_dict['n_tags'], csa_dict['check'] = up_str.unpack('2I')
-    if not 0 < csa_dict['n_tags'] <= 128:
+    if not 0 < csa_dict['n_tags'] <= MAX_CSA_ITEMS:
         raise CSAReadError('Number of tags `t` should be '
-                           '0 < t <= 128')
+                           '0 < t <= %s' % MAX_CSA_ITEMS)
```
effigies:
```suggestion
                           '0 < t <= %d' % MAX_CSA_ITEMS)
```

hbraunDSP:
I do have a file that generated an error and prompted me to look into this.  As soon as I can confirm it doesn't contain any PHI or other sensitive data I'll add it.

effigies:
I'm realizing that we're not actually going to cover the missing lines codecov is complaining about. The pass condition was already being hit, and we're making it less likely to hit the error condition. So at best it would be a check that we're no longer failing on files with 129-1000 tags, which is not a very clear gain.

Sorry for the trouble. I'll merge after tests pass.

hbraunDSP:
No problem.  I might still make a header with exactly `MAX_CSA_ITEMS=1000` tags just to be thorough, but thanks for your help!

effigies:
If you do want to submit a test file, please send it to https://github.com/effigies/nitest-dicom/.

PR #799, opened 2019-09-01
RF: compare data types using .dtype.type to avoid effect of endianness

yarikoptic:
Rationale is that in a typical case user is interested to check that file
contains correct data.  Comparison with effect of endianness would probably
be more of interest to developers.  May be someone later would add a flag
to nib-diff?


effigies:
Looks good. Thanks, Yarik.

Issue #729, opened 2019-02-26
Selecting on `data_per_streamline`


arokem:
I am trying to use the `data_per_streamline` attribute to select streamlines from within a `Tractogram` object. I am running into the following behavior. 

I initialize a very simple tractogram object. The first streamline has the `data_per_streamline` `foo` key set to `0` and the other two streamlines have this set to `1`:

```
import numpy as np 
import nibabel.streamlines as sl 
  
tgram = sl.Tractogram([np.array([[0, 0, 0], [1, 1, 1]]), 
                       np.array([[2, 2, 2], [3, 3, 3], [4, 4, 4]]), 
                       np.array([[5, 5, 5], [6, 6, 6]])], 
                       data_per_streamline={'foo': [[0],[1],[1]]}) 
````                                                                        

I can find the right streamlines using `np.where`:

```
idx = np.where(tgram.data_per_streamline['foo'] == 1)[0] 
idx
array([1, 2])
```
It even looks like it's doing the right thing when I use this for indexing:

```
tgram[idx].streamlines                                                   
ArraySequence([array([[2, 2, 2],
       [3, 3, 3],
       [4, 4, 4]]), array([[5, 5, 5],
       [6, 6, 6]])])
```

But under the hood:

```
tgram[idx].streamlines._data                                             
array([[0, 0, 0],
       [1, 1, 1],
       [2, 2, 2],
       [3, 3, 3],
       [4, 4, 4],
       [5, 5, 5],
       [6, 6, 6]])
```

Why is it still holding onto all of the data? Is there some way to drop the data for the first (not selected) streamline?

**Comment by arokem** [comment]
In case you are wondering, also: 

```
tgram.streamlines[idx]._data   
array([[0, 0, 0],
       [1, 1, 1],
       [2, 2, 2],
       [3, 3, 3],
       [4, 4, 4],
       [5, 5, 5],
       [6, 6, 6]])
```

**Comment by arokem** [comment]
Surely, I am not supposed to do the following? 

```
new_tgram = sl.Tractogram((tgram.streamlines._data[tgram.streamlines._offsets[i]:tgram.streamlines._offsets[i]+tgram.streamlines._offsets[i]+1] for i in idx),
                           data_per_streamline={k: [tgram.data_per_streamline[k][i] for i in idx] for k in tgram.data_per_streamline.keys()})
```

Though it does seem to have the expected behavior (at least in this case).

effigies:
[...]

PR #811, opened 2019-09-16
Fixing ArraySequence functionalities

MarcCote:
This PR aims to address the issues related to exposing ArraySequence's internal data (e.g. as discussed in #729 and https://github.com/nipy/dipy/pull/1956 - [this post exactly](https://github.com/nipy/dipy/pull/1956#discussion_r321023963), and my original answer to those issues can be found [here](https://github.com/nipy/nibabel/issues/729#issuecomment-467959221)).

There are three parts to this PR (I've made separate commits for convenience).

1. `ArraySequence.data` will now return a copy of the data. This way, altering the data of sliced/indexed tractogram will not affect the original data.
2. `ArraySequence` now supports some Python operators to modify its internal data (e.g. +=, -=, ...). Also, assigning/overwriting sequences can be done using `arr_seq[idx] = arr_seq2` (provided the shape are compatible).
3. Fixing a bug in `Tractogram.apply_affine` where calling that function on a sliced/indexed `Tractogram` object would result in modifying all streamlines (i.e. even those not selected).

If you think it is missing some Python operators that might be useful let me know.

@skoudoro 
NB: I haven't looked extensively but this PR would affect Dipy a bit. For instance, this part (https://github.com/nipy/dipy/blob/master/dipy/tracking/streamlinespeed.pyx#L104-L115) should have been using `._data` like the other functions in that file. Also, some of @frheault's `StatefulTractogram` code might be affected.

### DIFF
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -1,6 +1,5 @@
 
 import numbers
-import warnings
 from operator import mul
 from functools import reduce
 
@@ -156,7 +155,9 @@ class ArraySequence(object):
                             '3.0', '4.0')
     def data(self):
         """ Elements in this array sequence. """
-        return self.get_data()
+        view = self._data.view()
+        view.setflags(write=False)
+        return view
 
     def get_data(self):
         """ Returns a *copy* of the elements in this array sequence.
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -6,6 +6,9 @@ from functools import reduce
 
 import numpy as np
 
+from nibabel.deprecated import deprecate_with_version
+
+
 MEGABYTE = 1024 * 1024
 
 
@@ -148,16 +151,15 @@ class ArraySequence(object):
         return np.sum(self._lengths)
 
     @property
+    @deprecate_with_version("'ArraySequence.data' property is deprecated.\n"
+                            "Please use the 'ArraySequence.get_data()' method instead",
+                            '3.0', '4.0')
     def data(self):
         """ Elements in this array sequence. """
-        warnings.warn("The 'ArraySequence.data' property has been deprecated"
-                      " in favor of 'ArraySequence.get_data()'.",
-                      DeprecationWarning,
-                      stacklevel=2)
         return self.get_data()
 
     def get_data(self):
-        """ Returns a copy of the elements in this array sequence.
+        """ Returns a *copy* of the elements in this array sequence.
 
         Notes
         -----
@@ -384,7 +386,7 @@ class ArraySequence(object):
             seq._lengths = self._lengths[off_idx]
             return seq
 
-        if isinstance(off_idx, list) or is_ndarray_of_int_or_bool(off_idx):
+        if isinstance(off_idx, (list, range)) or is_ndarray_of_int_or_bool(off_idx):
             # Fancy indexing
             seq._offsets = self._offsets[off_idx]
             seq._lengths = self._lengths[off_idx]
@@ -425,7 +427,7 @@ class ArraySequence(object):
             offsets = self._offsets[off_idx]
             lengths = self._lengths[off_idx]
 
-        elif isinstance(off_idx, list) or is_ndarray_of_int_or_bool(off_idx):
+        elif isinstance(off_idx, (list, range)) or is_ndarray_of_int_or_bool(off_idx):
             # Fancy indexing
             offsets = self._offsets[off_idx]
             lengths = self._lengths[off_idx]
@@ -434,12 +436,25 @@ class ArraySequence(object):
             raise TypeError("Index must be either an int, a slice, a list of int"
                             " or a ndarray of bool! Not " + str(type(idx)))
 
-        if len(lengths) != elements.total_nb_rows:
-            msg = "Trying to set {} sequences with {} sequences."
-            raise TypeError(msg.format(len(lengths), elements.total_nb_rows))
+        if is_array_sequence(elements):
+            if len(lengths) != len(elements):
+                msg = "Trying to set {} sequences with {} sequences."
+                raise ValueError(msg.format(len(lengths), len(elements)))
 
-        for o1, l1, o2, l2 in zip(offsets, lengths, elements._offsets, elements._lengths):
-            data[o1:o1 + l1] = elements._data[o2:o2 + l2]
+            if sum(lengths) != elements.total_nb_rows:
+                msg = "Trying to set {} points with {} points."
+                raise ValueError(msg.format(sum(lengths), elements.total_nb_rows))
+
+            for o1, l1, o2, l2 in zip(offsets, lengths, elements._offsets, elements._lengths):
+                data[o1:o1 + l1] = elements._data[o2:o2 + l2]
+
+        elif isinstance(elements, numbers.Number):
+            for o1, l1 in zip(offsets, lengths):
+                data[o1:o1 + l1] = elements
+
+        else:  # Try to iterate over it.
+            for o1, l1, element in zip(offsets, lengths, elements):
+                data[o1:o1 + l1] = element
 
     def _op(self, op, value=None, inplace=False):
         """ Applies some operator to this arraysequence.
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -329,6 +329,54 @@ class ArraySequence(object):
         raise TypeError("Index must be either an int, a slice, a list of int"
                         " or a ndarray of bool! Not " + str(type(idx)))
 
+    def __setitem__(self, idx, elements):
+        """ Set sequence(s) through standard or advanced numpy indexing.
+
+        Parameters
+        ----------
+        idx : int or slice or list or ndarray
+            If int, index of the element to retrieve.
+            If slice, use slicing to retrieve elements.
+            If list, indices of the elements to retrieve.
+            If ndarray with dtype int, indices of the elements to retrieve.
+            If ndarray with dtype bool, only retrieve selected elements.
+        elements: ndarray or :class:`ArraySequence`
+            Data that will overwrite selected sequences.
+            If `idx` is an int, `elements` is expected to be a ndarray.
+            Otherwise, `elements` is expected a :class:`ArraySequence` object.
+        """
+        if isinstance(idx, (numbers.Integral, np.integer)):
+            start = self._offsets[idx]
+            self._data[start:start + self._lengths[idx]] = elements
+            return
+
+        if isinstance(idx, tuple):
+            off_idx = idx[0]
+            data = self._data.__getitem__((slice(None),) + idx[1:])
+        else:
+            off_idx = idx
+            data = self._data
+
+        if isinstance(off_idx, slice):  # Standard list slicing
+            offsets = self._offsets[off_idx]
+            lengths = self._lengths[off_idx]
+
+        elif isinstance(off_idx, list) or is_ndarray_of_int_or_bool(off_idx):
+            # Fancy indexing
+            offsets = self._offsets[off_idx]
+            lengths = self._lengths[off_idx]
+
+        else:
+            raise TypeError("Index must be either an int, a slice, a list of int"
+                            " or a ndarray of bool! Not " + str(type(idx)))
+
+        if len(lengths) != elements.total_nb_rows:
+            msg = "Trying to set {} sequences with {} sequences."
+            raise TypeError(msg.format(len(lengths), elements.total_nb_rows))
+
+        for o1, l1, o2, l2 in zip(offsets, lengths, elements._offsets, elements._lengths):
+            data[o1:o1 + l1] = elements._data[o2:o2 + l2]
+
     def __iter__(self):
         if len(self._lengths) != len(self._offsets):
             raise ValueError("ArraySequence object corrupted:"
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -1,5 +1,6 @@
 
 import numbers
+import warnings
 from operator import mul
 from functools import reduce
 
@@ -147,7 +148,22 @@ class ArraySequence(object):
     @property
     def data(self):
         """ Elements in this array sequence. """
-        return self._data
+        warnings.warn("The 'ArraySequence.data' property has been deprecated"
+                      " in favor of 'ArraySequence.get_data()'.",
+                      DeprecationWarning,
+                      stacklevel=2)
+        return self.get_data()
+
+    def get_data(self):
+        """ Returns a copy of the elements in this array sequence.
+
+        Notes
+        -----
+        To modify the data on this array sequence, one can use
+        in-place mathematical operators (e.g., `seq += ...`) or the use
+        assignment operator (i.e, `seq[...] = value`).
+        """
+        return self.copy()._data
 
     def _check_shape(self, arrseq):
         """ Check whether this array sequence is compatible with another. """
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -53,6 +53,35 @@ class _BuildCache(object):
         arr_seq._lengths = np.array(self.lengths)
 
 
+def _define_operators(cls):
+    """ Decorator which adds support for some Python operators. """
+    def _wrap(cls, op, name=None, inplace=False, unary=False):
+        name = name or op
+        if unary:
+            setattr(cls, name, lambda self: self._op(op))
+        else:
+            setattr(cls, name,
+                    lambda self, value: self._op(op, value, inplace=inplace))
+
+    for op in ["__iadd__", "__isub__", "__imul__", "__idiv__",
+               "__ifloordiv__", "__itruediv__", "__ior__"]:
+        _wrap(cls, op, inplace=True)
+
+    for op in ["__add__", "__sub__", "__mul__", "__div__",
+               "__floordiv__", "__truediv__", "__or__"]:
+        op_ = "__i{}__".format(op.strip("_"))
+        _wrap(cls, op_, name=op)
+
+    for op in ["__eq__", "__ne__", "__lt__", "__le__", "__gt__", "__ge__"]:
+        _wrap(cls, op)
+
+    for op in ["__neg__"]:
+        _wrap(cls, op, unary=True)
+
+    return cls
+
+
+@_define_operators
 class ArraySequence(object):
     """ Sequence of ndarrays having variable first dimension sizes.
 
@@ -119,6 +148,23 @@ class ArraySequence(object):
     def data(self):
         """ Elements in this array sequence. """
         return self._data
+
+    def _check_shape(self, arrseq):
+        """ Check whether this array sequence is compatible with another. """
+        msg = "cannot perform operation - array sequences have different"
+        if len(self._lengths) != len(arrseq._lengths):
+            msg += " lengths: {} vs. {}."
+            raise ValueError(msg.format(len(self._lengths), len(arrseq._lengths)))
+
+        if self.total_nb_rows != arrseq.total_nb_rows:
+            msg += " amount of data: {} vs. {}."
+            raise ValueError(msg.format(self.total_nb_rows, arrseq.total_nb_rows))
+
+        if self.common_shape != arrseq.common_shape:
+            msg += " common shape: {} vs. {}."
+            raise ValueError(msg.format(self.common_shape, arrseq.common_shape))
+
+        return True
 
     def _get_next_offset(self):
         """ Offset in ``self._data`` at which to write next rowelement """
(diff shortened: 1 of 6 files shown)


effigies:
Hi @MarcCote, thanks for this. I'll try to review today or tomorrow. To be clear, are you targeting next week's 2.5.1 release, or November's 3.0 release (release candidate at the end of October)?

If you're going for 2.5.1, could you rebase onto the `maint/2.5.x` branch?

effigies:
Other reviewers are of course very welcome... As this will affect @nipy/team-dipy the most, your input would be particularly appreciated.

skoudoro:
Thank you for this nice PR @MarcCote, will review and test it as soon as possible. 

I would suggest targeting Nibabel 3.0 Release and you can remove your check on Python version like on [this line](https://github.com/nipy/nibabel/pull/811/files#diff-a288dc3e6dc0e694269c7212d1933d1bR349)

ppoulin91:
I think the only "risky" thing left is when a Tractogram is created with an ArraySequence view, which could be externally modified (or the tractogram could modify a part of the original ArraySequence). However, I think it is reasonable to let the user manage this.
It might be good to at least pop a warning when creating a Tractogram with a view, what do you think?

MarcCote:
@skoudoro @effigies it makes more sense to target Nibabel 3.0.
@ppoulin91 yes we can do that. What about the case where someone sliced a Tractogram, then applies some transform to it? Should we also detect and warns the user? Maybe we can set the default value for in-place to False when applying a transform.

effigies:
@MarcCote Sounds good. I'll push off my review until after the 2.5.1 release, then, just because I'm a bit slammed.

@skoudoro How would you feel about being the primary reviewer on this one? I suspect some back-and-forth with Dipy use cases will be the most productive way to refine this. When you're satisfied I can come back for a final round.

ppoulin91:
@MarcCote I think if we want to stick close to numpy, we shouldn't warn when slicing. Once a Tractogram is created, the user is responsible for it.

nibabel/streamlines/array_sequence.py
```diff
@@ -118,7 +148,39 @@ def total_nb_rows(self):
...
     def data(self):
         """ Elements in this array sequence. """
-        return self._data
+        warnings.warn("The 'ArraySequence.data' property has been deprecated"
+                      " in favor of 'ArraySequence.get_data()'.",
+                      DeprecationWarning,
+                      stacklevel=2)
+        return self.get_data()
+
+    def get_data(self):
```
effigies: Are there cases where someone might want read-only access to the original data, without performing a copy?

nibabel/streamlines/tractogram.py
```diff
@@ -432,11 +432,8 @@ def apply_affine(self, affine, lazy=False):
         if np.all(affine == np.eye(4)):
             return self  # No transformation.
 
-        BUFFER_SIZE = 10000000  # About 128 Mb since pts shape is 3.
-        for start in range(0, len(self.streamlines.data), BUFFER_SIZE):
-            end = start + BUFFER_SIZE
-            pts = self.streamlines._data[start:end]
-            self.streamlines.data[start:end] = apply_affine(affine, pts)
+        for i in range(len(self.streamlines)):
+            self.streamlines[i] = apply_affine(affine, self.streamlines[i])
```
effigies: Does this not cause a significant slowdown, to only update one streamline at a time?

nibabel/streamlines/tests/test_tractogram.py
```diff
@@ -688,6 +688,22 @@ def test_tractogram_apply_affine(self):
...
 
+        # Applying the affine to a tractogram that has been indexed or sliced
+        # shouldn't affect the remaining streamlines.
+        tractogram = DATA['tractogram'].copy()
+        transformed_tractogram = tractogram[::2].apply_affine(affine)
+        assert_true(transformed_tractogram is not tractogram)
+        check_tractogram(tractogram[::2],
+                         streamlines=[s*scaling for s in DATA['streamlines'][::2]],
+                         data_per_streamline=DATA['tractogram'].data_per_streamline[::2],
+                         data_per_point=DATA['tractogram'].data_per_point[::2])
```
effigies: This feels very strange, but to be clear, is the following the intended semantics?

```Python
sliced_tractogram = tractogram[::2]  # A "view" tractogram
transformed_tractogram = sliced_tractogram.apply_affine(affine)
assert transformed_tractogram is sliced_tractogram
assert np.array_equal(tractogram[::2].data, sliced_tractogram.data)
```

nibabel/streamlines/tests/test_array_sequence.py
```diff
@@ -277,6 +277,128 @@ def test_arraysequence_getitem(self):
...
+        seq = SEQ_DATA['seq'] * 0
+        for i, e in enumerate(SEQ_DATA['seq']):
+            seq[i] = e
+
+        check_arr_seq(seq, SEQ_DATA['seq'])
+
+        if sys.version_info < (3,):
+            seq = ArraySequence(SEQ_DATA['seq'] * 0)
+            for i, e in enumerate(SEQ_DATA['seq']):
+                seq[long(i)] = e
```
effigies: Nibabel 3.0 will be Python 3-only. You can drop this check.

nibabel/streamlines/array_sequence.py
```diff
@@ -53,6 +54,35 @@ def update_seq(self, arr_seq):
         arr_seq._lengths = np.array(self.lengths)
 
 
+def _define_operators(cls):
```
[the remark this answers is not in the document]
MarcCote: I'm not sure how I can do it with an abstract class exactly. Do you mean defining all the methods one by one in the abstract, then make `ArraySequence` subclass it? If that's the case, I might as well as defining them directly in `ArraySequence`. Also, I was trying to keep the number of new lines to a minimum.

nibabel/streamlines/array_sequence.py
MarcCote: Since it's not always possible to return a view on *just* the data belonging to a sliced/indexed `ArraySequence`, I decided to always make it return a copy.

I know it is bad practice but I'm expecting Dipy's methods to deal directly with `._data` using `._offsets` and `._lengths`.

That said, maybe there exists a solution but I fail to see it. I'm happy to consider alternatives.

nibabel/streamlines/tractogram.py
MarcCote: It didn't show up when running `benchmark_streamlines.py` but to be fair that benchmark might not be realistic. I don't have any large tractogram file I could use to benchmark it. Do you know where I could easily get one? (or @skoudoro maybe knows ?).

nibabel/streamlines/tests/test_tractogram.py
MarcCote: Kinda but `Tractogram` doesn't have a `.data` property. `LazyTractogram` does have one but it returns a generator over TractogramItem, not a ndarray. In the test, I'm also looking at `data_per_point` which is a dictionary of `ArraySequence`s.

nibabel/streamlines/tests/test_array_sequence.py
MarcCote: Will do.

MarcCote:
@effigies thanks for the review. I'll see what I can do about writing up a small tutorial to showcase the API. I should have done way before :P.

nibabel/streamlines/array_sequence.py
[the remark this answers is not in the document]
effigies: Yeah, I guess if it's only ever one class, that's one thing. Perhaps we can change them to proper methods and give them reasonable docstrings.

I've looked around at [numbers](https://docs.python.org/3/library/numbers.html) and numpy to see if there was a base class we could use, but doesn't look like it.

effigies:
@MarcCote What's the status here?

I'll be feature freezing 3.0.x when the release candidate comes out. 3.1.0 should be February or March next year. I'm okay delaying the release candidate if we have a short-ish time frame (<2 weeks) to get this in.

I don't have the bandwidth to guide this one in in that time, so if you want to aim for 3.0, could we find a volunteer from Dipy to be the primary reviewer?

skoudoro:
Really sorry for the delay @effigies, it is on my to-do list, and I will spend time tomorrow to do it.

nibabel/streamlines/tractogram.py
skoudoro: I supposed you already downloaded one. You can look in your home folder:  `.dipy/bundle_atlas_hcp842/Atlas_80_Bundles/whole_brain/whole_brain_MNI.trk`

nibabel/streamlines/array_sequence.py
skoudoro: Agree, I do no think that a lot of users might want a read-only access to the original data

effigies:
Hi guys, what's the status here? Does this feel like something that will be ready to merge in the next week or so?

MarcCote:
I'm working on it right now. I'll ping @skoudoro when I addressed all of the points raised.

nibabel/streamlines/array_sequence.py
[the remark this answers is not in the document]
MarcCote: @effigies So, I'm going to remove the decorator in favor of explicitly listing all the operators. Sounds good?

nibabel/streamlines/array_sequence.py
```diff
@@ -329,6 +391,85 @@ def __getitem__(self, idx):
...
+        seq = self if inplace else self.copy()
+
+        if is_array_sequence(value) and seq._check_shape(value):
+            for o1, l1, o2, l2 in zip(seq._offsets, seq._lengths, value._offsets, value._lengths):
+                seq._data[o1:o1 + l1] = getattr(seq._data[o1:o1 + l1], op)(value._data[o2:o2 + l2])
+
+        else:
+            args = [] if value is None else [value]  # Dealing with unary and binary ops.
+            for o1, l1 in zip(seq._offsets, seq._lengths):
+                seq._data[o1:o1 + l1] = getattr(seq._data[o1:o1 + l1], op)(*args)
```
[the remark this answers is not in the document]
MarcCote: Good catch.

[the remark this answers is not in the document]
MarcCote: Good catch.

nibabel/streamlines/array_sequence.py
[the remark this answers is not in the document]
effigies: That's fine with me.

To be clear, though, what I was suggesting in my last xcomment was replacing the `lambda` with a full function, so `lambda self: self.op(op)` would become `def fn(self): return self._op(op)`, and then you could add a docstring with `fn.__doc__ = ...` before assigning it.

But whatever's the least work is fine.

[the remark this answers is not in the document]
skoudoro: +1 for using `def fn(self): return self._op(op)`.  Not a big fan of lambda, but it is a personal preference...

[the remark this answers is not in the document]
MarcCote: Oh, gotcha. That makes sense.

nibabel/streamlines/array_sequence.py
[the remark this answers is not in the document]
MarcCote: @skoudoro the dtype now changes depending on the arithmetic operation and the provided value. Regarding division by 0, it will produce NaN (i.e. it follows Numpy's convention).

nibabel/streamlines/array_sequence.py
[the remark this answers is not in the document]
MarcCote: @effigies done. Here's what it looks like now.
```
NAME
    array_sequence

CLASSES
    builtins.object
        ArraySequence
    
    class ArraySequence(builtins.object)
     |  ArraySequence(iterable=None, buffer_size=4)
     |  
     |  Sequence of ndarrays having variable first dimension sizes.
     |  
     |  This is a container that can store multiple ndarrays where each ndarray
     |  might have a different first dimension size but a *common* size for the
     |  remaining dimensions.
     |  
     |  More generally, an instance of :class:`ArraySequence` of length $N$ is
     |  composed of $N$ ndarrays of shape $(d_1, d_2, ... d_D)$ where $d_1$
     |  can vary in length between arrays but $(d_2, ..., d_D)$ have to be the
     |  same for every ndarray.
     |  
     |  Methods defined here:
     |  
     |  __abs__(self)
     |      abs(self)
     |  
     |  __add__(self, value)
     |      Return self+value.
     |  
     |  __and__(self, value)
     |      Return self&value.
     |  
     |  __eq__(self, value)
     |      Return self==value.
     |  
     |  __floordiv__(self, value)
     |      Return self//value.
     |  
     |  __ge__(self, value)
     |      Return self>=value.
     |  
     |  __getitem__(self, idx)
     |      Get sequence(s) through standard or advanced numpy indexing.
     |      
     |      Parameters
     |      ----------
     |      idx : int or slice or list or ndarray
     |          If int, index of the element to retrieve.
     |          If slice, use slicing to retrieve elements.
     |          If list, indices of the elements to retrieve.
     |          If ndarray with dtype int, indices of the elements to retrieve.
     |          If ndarray with dtype bool, only retrieve selected elements.
     |      
     |      Returns

```

MarcCote:
Also, it seems I didn't finish completely the `__setitem__` method! I'll finish it tomorrow.

nibabel/streamlines/array_sequence.py
```diff
@@ -53,6 +54,37 @@ def update_seq(self, arr_seq):
...
+        def fn_unary_op(self):
+            return self._op(op)
+
+        def fn_binary_op(self, value):
+            return self._op(op, value, inplace=inplace)
+
+        setattr(cls, op, fn_unary_op if unary else fn_binary_op)
+        fn = getattr(cls, op)
+        fn.__name__ = op
+        fn.__doc__ = getattr(np.ndarray, op).__doc__
```
skoudoro: Beautiful! I love this `wrap` function 😄

MarcCote:
I just finished the `__setitem__`. @skoudoro your example should work now. Thanks for catching that :D.

nibabel/streamlines/array_sequence.py
MarcCote: Yes! It turned out nicely.

MarcCote:
I just realized that indexing a column is counter-intuitive. I'm planning on forcing the use of a full slice (':') for the second dimension (i.e. the points dimension). What do you think?

```python
from nibabel.streamlines import ArraySequence

seq = ArraySequence([[(1, 2, 3)],
                     [(4, 5, 6), (7, 8, 9)],
                     [(10, 11, 12), (13, 14, 15), (16, 17, 18)]
                    ])

# Before
seq[:, 2]  # Retrieving the third coordinates of every point of every sequence. 
> ArraySequence([array([3]), array([6, 9]), array([12, 15, 18])])

# After. We force to use full slice on the second dimension.
seq[:, 2]
> TypeError: Only full slice (':') is allowed for the second dimension.

seq[:, :, 2]  # Dimensions: 1) sequences, 2) points, 3) common shape data.
> ArraySequence([array([3]), array([6, 9]), array([12, 15, 18])])
```

skoudoro:
ok, I am not personally again that, but I would like the @ppoulin91 and @frheault opinions.

My only concern is I would prefer that you force the full slicing all the time or never.  I do not like the idea to have it only for a specific case.

MarcCote:
> ok, I am not personally again that, but I would like the @ppoulin91 and @frheault opinions.
> 
> My only concern is I would prefer that you force the full slicing all the time or never. I do not like the idea to have it only for a specific case.

It would always be forced with my next commit. The example I showed was "before" and "after" the change.

nibabel/streamlines/array_sequence.py
```diff
@@ -118,7 +150,39 @@ def total_nb_rows(self):
     @property
     def data(self):
         """ Elements in this array sequence. """
-        return self._data
+        warnings.warn("The 'ArraySequence.data' property has been deprecated"
+                      " in favor of 'ArraySequence.get_data()'.",
+                      DeprecationWarning,
+                      stacklevel=2)
```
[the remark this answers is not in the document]
MarcCote: Sweet.

matthew-brett:
Just a note - I think it's a bad idea to have a property that returns a copy - it breaks the rules in the [property manifesto](https://github.com/nipy/nibabel/wiki/property_manifesto).

If you're going to return a copy, I think you should use a method instead, and deprecate `.data`.

effigies:
Good point. We should keep the old behavior as we deprecate.

MarcCote:
@effigies @matthew-brett so we should do the following?
```python
    @property
    @deprecate_with_version("'ArraySequence.data' property is deprecated.\n"
                            "Please use the 'ArraySequence.get_data()' method instead",
                            '3.0', '4.0')
    def data(self):
        """ Elements in this array sequence. """
        return self._data
```
instead of `self.copy()._data`

effigies:
That's what I was thinking.

There might be room in the property manifesto rules to make it a read-only view, which could achieve some of your goal without performing a copy behind people's backs. Just one man's opinion...

MarcCote:
I like that idea. Something like that?
```python
view = self._data
view.setflags(write=False)
return view
```

effigies:
```python
view = self._data.view()
view.setflags(write=False)
return view
```

MarcCote:
Oops. There you go.

effigies:
@skoudoro How's this looking on your end?

skoudoro:
Overall, looks good to me! But I need to do some tests one last time so I will come back to you by tomorrow evening.  Is that ok?

effigies:
Sounds good.

skoudoro:
Hi @MarcCote, 

We are really close! There is a small issue. Below the code to reproduce it:

```
from nibabel.streamlines import ArraySequence

seq = ArraySequence([[(1, 2, 3)],
                     [(4, 5, 6), (7, 8, 9)],
                     [(10, 11, 12), (13, 14, 15), (16, 17, 18)]
                    ])

# This works fine
seq = seq \ seq

# This case is failing. see below the error in details
seq /= seq
```

here the error
```
TypeError                                 Traceback (most recent call last)
<ipython-input-29-089af863d745> in <module>
----> 1 seq /= seq

~\Devel\nibabel\nibabel\streamlines\array_sequence.py in fn_binary_op(self, value)
     65
     66         def fn_binary_op(self, value):
---> 67             return self._op(op, value, inplace=inplace)
     68
     69         setattr(cls, op, fn_unary_op if unary else fn_binary_op)

~\Devel\nibabel\nibabel\streamlines\array_sequence.py in _op(self, op, value, inplace)
    485             # Change seq.dtype to match the operation resulting type.
    486             o0, l0, o1, l1, o2, l2 = next(elements)
--> 487             tmp = getattr(self._data[o1:o1 + l1], op)(value._data[o2:o2 + l2])
    488             seq._data = seq._data.astype(tmp.dtype)
    489             seq._data[o0:o0 + l0] = tmp

TypeError: No loop matching the specified signature and casting
```

MarcCote:
@skoudoro This is also true with numpy. Since your ArraySequence is composed of integers, you can't do a true division in-place. Any suggestions on how I should properly deal with that?

```python
In [1]: import numpy as np                                                      

In [2]: A = np.array([(10, 11, 12), (13, 14, 15), (16, 17, 18)])                

In [3]: A /= A                                                                  
TypeError                                 Traceback (most recent call last)
<ipython-input-4-68e406011332> in <module>
----> 1 A /= A

TypeError: No loop matching the specified signature and casting was found for ufunc true_divide
```

skoudoro:
Make sense. You can let it as it is.  Thank you @marccote and @effigies.

  +1 for merging

effigies:
@skoudoro Thanks for reviewing. I'll have another look later today.

@matthew-brett Any further concerns?

effigies:
Oh, @ppoulin91, I forgot to tag you. Do you have any further comments?

nibabel/streamlines/array_sequence.py
```diff
@@ -5,6 +5,9 @@
 
 import numpy as np
 
+from nibabel.deprecated import deprecate_with_version
```
effigies:
```suggestion
from ..deprecated import deprecate_with_version
```

nibabel/streamlines/array_sequence.py
```diff
@@ -5,6 +5,9 @@
...
+
```
effigies:
```suggestion
```

nibabel/streamlines/tractogram.py
effigies: And you checked that this was fine?

ppoulin91:
I had a quick check with @frheault, we didn't follow everything in the last weeks, but everything looks good to us!

nibabel/streamlines/tractogram.py
MarcCote: It's a bit slower but at least it doesn't have side-effects.
Benchmark: loading the `whole_brain_MNI.trk` five times.
```
Old: Loaded 144,678 streamlines 5 times in  16.57
New: Loaded 144,678 streamlines 5 times in  19.61
```

MarcCote:
@effigies I can move the last commit to a separate PR if needed. That commit is about speeding up the loading of TRK files.

```
# Loading 5,000 streamlines:
nibabel.trackvis.read:     1.10
nibabel.streamlines:       8.3
nibabel.streamlines (new): 7.8

# Loading 5,000 streamlines with scalars:
nibabel.trackvis.read:      1.75
nibabel.streamlines:       30.80
nibabel.streamlines (new):  8.83

# Loading 144,678 streamlines (whole_brain_MNI.trk):
nibabel.trackvis.read:      2.05  (456.26 MB)
nibabel.streamlines:       10.60  (178.06 MB)
nibabel.streamlines (new):  3.71  (178.06 MB)
```

effigies:
Alright, thanks for the contribution, @MarcCote, and the reviews, @skoudoro, @ppoulin91, @frheault!

In it goes.

PR #815, opened 2019-09-20
ENH: Add a utility to calculate obliquity of affines

oesteban:
This implementation mimics the implementation of AFNI, and will be useful in #656 .

### DIFF
--- a/nibabel/affines.py
+++ b/nibabel/affines.py
@@ -2,8 +2,6 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """ Utility routines for working with points and affine transforms
 """
-
-from math import acos, pi as PI
 import numpy as np
 
 from functools import reduce
@@ -299,17 +297,27 @@ def voxel_sizes(affine):
     return np.sqrt(np.sum(top_left ** 2, axis=0))
 
 
-def obliquity(affine, degrees=False):
+def obliquity(affine):
     r"""
-    Estimate the obliquity an affine's axes represent, in degrees from plumb.
+    Estimate the *obliquity* an affine's axes represent.
 
+    The term *obliquity* is defined here as the rotation of those axes with
+    respect to the cardinal axes.
     This implementation is inspired by `AFNI's implementation
     <https://github.com/afni/afni/blob/b6a9f7a21c1f3231ff09efbd861f8975ad48e525/src/thd_coords.c#L660-L698>`_
 
+    Parameters
+    ----------
+    affine : 2D array-like
+        Affine transformation array.  Usually shape (4, 4), but can be any 2D
+        array.
+
+    Returns
+    -------
+    angles : 1D array-like
+        The *obliquity* of each axis with respect to the cardinal axes, in radians.
+
     """
     vs = voxel_sizes(affine)
-    fig_merit = (affine[:-1, :-1] / vs[np.newaxis, ...]).max(axis=1).min()
-    radians = abs(acos(fig_merit))
-    if not degrees:
-        return radians
-    return radians * 180 / PI
+    best_cosines = np.abs((affine[:-1, :-1] / vs).max(axis=1))
+    return np.arccos(best_cosines)
--- a/nibabel/affines.py
+++ b/nibabel/affines.py
@@ -321,5 +321,5 @@ def obliquity(affine):
 
     """
     vs = voxel_sizes(affine)
-    best_cosines = np.abs((affine[:-1, :-1] / vs).max(axis=1))
+    best_cosines = np.abs(affine[:-1, :-1] / vs).max(axis=1)
     return np.arccos(best_cosines)
--- a/nibabel/affines.py
+++ b/nibabel/affines.py
@@ -3,6 +3,7 @@
 """ Utility routines for working with points and affine transforms
 """
 
+from math import acos, pi as PI
 import numpy as np
 
 from functools import reduce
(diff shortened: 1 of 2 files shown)


matthew-brett:
Thanks for this.  Some quick thoughts:

* the test should go into a test function, and it needs some more tests (for not-oblique, for example).
* if the doctest stays, can you find something more obvious as an example?  The exponential notation values are hard to read.
* Can you explain the algorithm?   I was expecting you to take the dot products of the columns of  unit-length, something like this:

```python
R = affine[:-1, :-1] / vs
col_cosines = R @ R.T
upper_cosines = col_cosines[np.triu_indices(3, 1)]
angles = np.arccos(upper_cosines) 
np.any(np.abs(angles - np.pi / 2) > thresh)
```

oesteban:
> * the test should go into a test function, and it needs some more tests (for not-oblique, for example).

Sure, the latest commit should address this.



> * if the doctest stays, can you find something more obvious as an example?  The exponential notation values are hard to read.

The doctest did not stay, but now I'm generating the matrix in a more understandable way (first, generate an aligned matrix; second explicitly rotate it).

> Can you explain the algorithm? I was expecting you to take the dot products of the columns of unit-length, something like this:

TBH I just transferred AFNI's implementation because it behaves differently when the affines are not alligned when applying transforms, so I needed to identify this situation for AFNI.

Now looking at the code, I believe AFNI's implementation is pretty close to your suggestion since it first projects each column of the affine to the canonical axes, then finds the projection for each axis (i.e. that `.max(axis=1)`) and finally it computes the arccos only on the smallest of these projected vectors (which will give the largest angle). Your implementation would be calculating the obliquity of the three axes.

I don't mind either way. However, I'd be interested in having an implementation parallel to AFNI's to be able to predict it's behavior. In other words, I can check whether your implementation is equivalent for the largest angle. If not, then rename the function to afni_obliquity or something that notes this particular.

oesteban:
Comparing implementations

```Python
import numpy as np
from math import acos, pi as PI
from nibabel.eulerangles import euler2mat
from nibabel.affines import from_matvec, voxel_sizes
aligned = np.diag([2.0, 2.0, 2.3, 1.0]) 
aligned[:-1, -1] = [-10, -10, -7] 
R = from_matvec(euler2mat(x=0.09, y=0.001, z=0.001), [0.0, 0.0, 0.0]) 
oblique = R.dot(aligned)
vs = voxel_sizes(oblique)
O = oblique[:-1, :-1] / vs[np.newaxis, ...]
fig_merit = O.max(axis=1).min()
afni = abs(acos(fig_merit) * 180 / PI)
col_cosines = O @ O.T
upper_cosines = col_cosines[np.triu_indices(3, 1)]
angles = np.arccos(upper_cosines)
angles_deg = np.abs(angles - np.pi / 2)
```

```
>>> afni                                                                                                                                                                                                                                                                                                                                                                                                      
5.156994888317267
>>> angles_deg
array([0., 0., 0.])
```

I suspect your code is actually checking the orthogonality of the axes.

effigies:
From a quick glance it looks like they're using the inf-norm of the cosines instead of the 2-norm, so perhaps you could parametrize the norm?

oesteban:
In https://github.com/nipy/nibabel/pull/815/files#diff-fadc4e2f32f9100f1b5ce2384d00238fR311,

the `affine[:-1, :-1] / vs[np.newaxis, ...]` is implicitly using the L2-norm for normalization (the L2-norm is calculated in `voxel_sizes`).
I think the inf-norm is just used to project on axes.

matthew-brett:
Oh - sorry - yes - my code is definitely checking the orthogonality of axes, and that's because I thought that's what you wanted to test.  So now I realize I need to ask - what does 'obliquity' mean, exactly?   Some rotation that puts the axes out of alignment with any permutation of the RAS+ axes?   In that case `R = affine[:-1, :-1] / vs` gives the cosines of the angles of the axes with respect to the three canonical axes, right (`np.eye(3) @ R`)?   In which case, `M = np.abs(R).max(axis=1)` (don't we need the `abs`?) gives the cosine of the angle corresponding to the closest match of any column of R to `np.eye(3)`,  for each of the three canonical axes (rows of `np.eye(3)`).   And then the `.min` checks for the canonical axis with the worst match to the axes of R (furthest from 1).  Then the algorithm is basically correct, and could also be written:

```
cosines = oblique[:-1, :-1] / vs
best_cosines = np.abs(cosines).max(axis=1)
```

and so on.

I'd rather stick with radians if possible, we're not using degrees anywhere else, I think.

By the way, I believe the the `[np.newaxis, ...]` in `affine[:-1, :-1] / vs[np.newaxis, ...]` is redundant.

oesteban:
> Oh - sorry - yes - my code is definitely checking the orthogonality of axes, and that's because I thought that's what you wanted to test. So now I realize I need to ask - what does 'obliquity' mean, exactly? Some rotation that puts the axes out of alignment with any permutation of the RAS+ axes?

Yes, in AFNI's terms they have an explanation (where they mix up the concept of "obliquity" with the orientation of NIfTI images) [here](https://sscc.nimh.nih.gov/sscc/dglen/Obliquity).

> I'd rather stick with radians if possible, we're not using degrees anywhere else, I think.

I'll add a `degrees` keyword argument, defaulting to False, so that we can directly compare to AFNI (after all, the initiative to add this is a result of a need to predict its behavior).


> By the way, I believe the the `[np.newaxis, ...]` in `affine[:-1, :-1] / vs[np.newaxis, ...]` is redundant.

I preferred to be explicit to make sure the columns were normalized. But yes, both are equivalent. Beyond explicitness, I have no preference - let me know what aligns best with existing code.

matthew-brett:
Do you agree with me about `abs` above?

If you prefer to make the broadcasting rule explicit with `vs[np.newaxis, ...]`, then the eliipsis is confusing - the combination of the not-strictly-necessary new axis, and the ellipsis, looks like you are thinking of more than one dimension.    I'd prefer `vs[None, :]` personally (or just `vs`).

The `degrees` keyword seems messy to me, can't we just convert to degrees outside the function to test against AFNI?  As in `obliquity(affine) * 180 / pi == my_afni_result` ?

Can you put more explanation in the docstring aboutt what 'obliquity' is?

oesteban:
@matthew-brett, did the last commit address all your concerns?

matthew-brett:
Yes - thanks for your patience.

Can you also add a link to https://sscc.nimh.nih.gov/sscc/dglen/Obliquity in the docstring?

Is it worth checking this calculation for non-orthogonal axes?    It will say 'oblique', but I suppose that is what is intended?

oesteban:
> Is it worth checking this calculation for non-orthogonal axes? It will say 'oblique', but I suppose that is what is intended?

Like adding a test were just one of the axes is not orthogonal to the plane where the other two are?.

nibabel/affines.py
```diff
@@ -296,3 +295,31 @@ def voxel_sizes(affine):
...
+        array.
+
+    Returns
+    -------
+    angles : 1D array-like
+        The *obliquity* of each axis with respect to the cardinal axes, in radians.
+
+    """
+    vs = voxel_sizes(affine)
+    best_cosines = np.abs((affine[:-1, :-1] / vs).max(axis=1))
```
matthew-brett: Oh, sorry, I meant:

```
best_cosines = np.abs(affine[:-1, :-1] / vs).max(axis=1)
```

Don't you then need to take the min?    As in:

```
return np.arccos(best_cosines.min())
```

?  This will give the arccos of the largest deviation ...  I think.

matthew-brett: Sorry - by 'largest deviation' I mean the one with the angle furthest from zero.

oesteban: Now the function returns the obliquity of the three axes, if that makes sense - up to the user to pick the most deviated.

matthew-brett: What do you think about `best_cosines = np.abs(affine[:-1, :-1] / vs).max(axis=1)` ?

oesteban: Good catch, thanks!

oesteban:
Thanks for your patience, anything else I should address here?

matthew-brett:
OK with me - Chris - OK with you?

effigies:
I didn't follow this very closely, but I have no objections.

PR #821, opened 2019-10-06
FIX: Re-import externals/netcdf.py from scipy

effigies:
Fixes deprecation error currently appearing in the [pre-release Travis build](https://travis-ci.com/nipy/nibabel/jobs/242476908).

Targeting 2.5.x first, since some Python 2 compatibility was removed in 3.0-dev. Once this is fixed, I'll use this diff to port to 3.0-dev.

This ends up overwriting incremental fixes put in netcdf.py since the last sync (#179), most notably #493. @rdvincent If you have any files you'd like to test that aren't in our battery, I would appreciate it to make sure we're not breaking MINC1.

@matthew-brett Do you remember if we have any fixes/hacks where we specifically chose to diverge from scipy?

### DIFF
--- a/nibabel/externals/netcdf.py
+++ b/nibabel/externals/netcdf.py
@@ -221,8 +221,8 @@ class netcdf_file(object):
     directly to memory-mapped data on disk:
 
         >>> data = time[:]
-        >>> data.base.base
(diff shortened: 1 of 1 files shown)


effigies:
Tests seem to pass, and the coverage drops seem unrelated.

Issue #832, opened 2019-11-07
ArrayProxy always casts image data to float64


effigies:
Consider a BOLD series (obtainable with git-annex [here](https://github.com/neuroscout-datasets/LearningTemporalStructure/blob/master/fmriprep/sub-01/func/sub-01_task-movie_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz)) saved in a NIfTI-1 with dtype `int16`, `scl_slope == 0.061591435` and `scl_inter == 1839.8748`. Both `scl_slope` and `scl_inter` are `np.float32`:

https://github.com/nipy/nibabel/blob/ddf26830eebdb11470009c3a04047377f29a4732/nibabel/nifti1.py#L49-L50

The ArrayProxy casts these data to `float64`:

```Python
>>> img.dataobj[0, 0, 0, 0].dtype
dtype('float64')
```

Note the following:

```Python
>>> type(img.dataobj.slope)
float
>>> type(img.dataobj.inter)
float
```

This is due to the cast in nifti1.py:

https://github.com/nipy/nibabel/blob/ddf26830eebdb11470009c3a04047377f29a4732/nibabel/nifti1.py#L1164-L1167

If I use:

```Python
>>> unscaled = img.dataobj.get_unscaled()
>>> slope = np.float32(img.dataobj.slope)
>>> inter = np.float32(img.dataobj.inter)
>>> nb.volumeutils.apply_read_scaling(unscaled, slope, inter).dtype
dtype('float32')
```

So the precision of the data available permits us to safely return this data block as float32 (`apply_read_scaling` would bump to float64 if it were needed), but instead we always return float64 data arrays.

This means that `img.get_fdata(dtype=np.float32)` will first load a 64-bit array and then cast it down to 32-bit.

My expectation would generally be to use the smallest data type as the default and scale up at need or request. While I do see a potential argument for using float64 across all image types (e.g., NIfTI-2 and thus CIFTI-2 have 64-bit scales and intercepts) as safe and consistent, that feels to me appropriate for a high-level interface like `get_fdata()`, rather than the low-level load of `np.array(img.dataobj)`.

matthew-brett:
Yes, I noticed this a while ago.  I didn't mean for that to happen, but when I noticed, I didn't do anything about it because I was worried that this might cause subtle changes in the analysis results.  What do you think?

effigies:
Just to get a sense of the potential error, and taking the current slope/inter as representative:

```Python
[...]

PR #833, opened 2019-11-08
NF: Enable data scaling within the target dtype

effigies:
Following some initial discussion in #832, this PR allows for ArrayProxies to scale data objects within a target dtype, rather than always defaulting to float64, which will prevent memory spikes and speed up results for `get_fdata(dtype=np.float32)`.

This obviously has consequences for accumulating rounding errors. The old behavior can be achieved with `get_fdata().astype(np.float32)`, although the caching properties are changed.

This involves a refactor that uniformizes access to unscaled and scaled data using file slices, and `get_scaled()`, `get_unscaled()` and `__array__` will now call `fileslice` with the `()` slicer. With some informal benchmarking, `img.dataobj[()]` is at least as fast as `np.asanyarray(img.dataobj)`:

```Python
In [60]: %timeit _ = np.array(img.dataobj)                                                          
7.38 s ± 169 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [62]: %timeit _ = img.dataobj[:]                                                                 
5.89 s ± 376 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [63]: %timeit _ = img.dataobj[()]                                                                
5.76 s ± 159 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [67]: %timeit _ = np.asanyarray(img.dataobj)                                                     
6.48 s ± 233 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

One question is whether to have `Nifti1Header.get_slope_inter()` return `np.float32`s or continue returning `float`s. By continuing to return `float`, `get_data()` will continue to return `float64` arrays when scale factors are applied.

This is a potentially consequential API change, so I'd appreciate wide comments from @nipy/team-nibabel, and any other interested parties.

Tests on the way.

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -362,11 +362,8 @@ class ArrayProxy(object):
         use_dtype = scl_slope.dtype if dtype is None else dtype
         slope = scl_slope.astype(use_dtype)
         inter = scl_inter.astype(use_dtype)
-        # Read array
-        raw_data = self._get_unscaled(slicer=slicer)
-        # Upcast as necessary for big slopes, intercepts
-        scaled = apply_read_scaling(raw_data, slope, inter)
-        del raw_data
+        # Read array and upcast as necessary for big slopes, intercepts
+        scaled = apply_read_scaling(self._get_unscaled(slicer=slicer), slope, inter)
         if dtype is not None:
             scaled = scaled.astype(np.promote_types(scaled.dtype, dtype), copy=False)
         return scaled
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -359,14 +359,17 @@ class ArrayProxy(object):
         # Ensure scale factors have dtypes
         scl_slope = np.asanyarray(self._slope)
         scl_inter = np.asanyarray(self._inter)
-        if dtype is None:
-            dtype = scl_slope.dtype
-        slope = scl_slope.astype(dtype)
-        inter = scl_inter.astype(dtype)
+        use_dtype = scl_slope.dtype if dtype is None else dtype
+        slope = scl_slope.astype(use_dtype)
+        inter = scl_inter.astype(use_dtype)
         # Read array
         raw_data = self._get_unscaled(slicer=slicer)
         # Upcast as necessary for big slopes, intercepts
-        return apply_read_scaling(raw_data, slope, inter)
+        scaled = apply_read_scaling(raw_data, slope, inter)
+        del raw_data
+        if dtype is not None:
+            scaled = scaled.astype(np.promote_types(scaled.dtype, dtype), copy=False)
+        return scaled
 
     def get_unscaled(self):
         """ Read data from file
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -33,7 +33,7 @@ import numpy as np
 
 from .deprecated import deprecate_with_version
 from .volumeutils import array_from_file, apply_read_scaling
-from .fileslice import fileslice
+from .fileslice import fileslice, canonical_slicers
 from .keywordonly import kw_only_meth
 from . import openers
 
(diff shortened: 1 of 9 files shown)


nibabel/arrayproxy.py
```diff
@@ -336,36 +336,68 @@ def _get_fileobj(self):
...
 
         This is an optional part of the proxy API
         """
-        with self._get_fileobj() as fileobj, self._lock:
-            raw_data = array_from_file(self._shape,
-                                       self._dtype,
-                                       fileobj,
-                                       offset=self._offset,
-                                       order=self.order,
-                                       mmap=self._mmap)
```
effigies: Oh right. This kills the mmap. :-/

effigies:
The existing failing tests ended up taking up most of my day, so I'll think about an appropriate set of new tests on the train on Sunday.

The `allclose` checks might cover most of what we need, but I suspect we'll want to really exercise the bounds where a proxy gets pushed to higher precision.

nibabel/arrayproxy.py
```diff
@@ -336,36 +336,80 @@ def _get_fileobj(self):
...
-    def get_unscaled(self):
-        """ Read of data from file
-
-        This is an optional part of the proxy API
-        """
-        with self._get_fileobj() as fileobj, self._lock:
-            raw_data = array_from_file(self._shape,
+    def _get_unscaled(self, slicer):
+        if canonical_slicers(slicer, self._shape, False) == \
+                canonical_slicers((), self._shape, False):
```
effigies: This is a heavy check for whether we can just grab the whole file, which ensures we get an `mmap` if possible.

Might break some people using `dataobj[:]` to always retrieve an in-memory copy.

nibabel/tests/test_proxy_api.py
```diff
@@ -131,6 +131,22 @@ def validate_asarray(self, pmaker, params):
...
+    def validate_get_scaled(self, pmaker, params):
+        # Check proxy returns expected array from asarray
+        prox, fio, hdr = pmaker()
+        out = prox.get_scaled()
+        assert_array_equal(out, params['arr_out'])
+        assert_dt_equal(out.dtype, params['dtype_out'])
+        # Shape matches expected shape
+        assert_equal(out.shape, params['shape'])
+
+        for dtype in np.sctypes['float']:
```
effigies: This could easily be extended with non-float types, and the checks would still be valid. Not sure if that's a case we want to specifically support...

effigies:
This is ready for a review.

nibabel/arrayproxy.py
```diff
@@ -336,36 +336,80 @@ def _get_fileobj(self):
...
+        scl_slope = np.asanyarray(self._slope)
+        scl_inter = np.asanyarray(self._inter)
+        use_dtype = scl_slope.dtype if dtype is None else dtype
+        slope = scl_slope.astype(use_dtype)
+        inter = scl_inter.astype(use_dtype)
+        # Read array
+        raw_data = self._get_unscaled(slicer=slicer)
+        # Upcast as necessary for big slopes, intercepts
+        scaled = apply_read_scaling(raw_data, slope, inter)
+        del raw_data
```
rmarkello: Any reason we need to explicitly keep `raw_data` around and manually delete it? Would this work:

```python
scaled = apply_read_scaling(self._get_unscaled(slicer=slicer), slope, inter)
```

nibabel/parrec.py
```diff
@@ -633,38 +633,53 @@ def dtype(self):
...
         # Slice scaling to give output shape
-        return raw_data * slopes[slicer] + inters[slicer]
+        return raw_data * slopes[slicer].astype(final_type) + inters[slicer].astype(final_type)
+
+
+    def get_unscaled(self):
+        return self._get_unscaled(slicer=())
+
+    def get_scaled(self, dtype=None):
+        return self._get_scaled(dtype=dtype, slicer=())
```
rmarkello: Add docs from `ArrayProxy` since this isn't (yet?) a subclass.

nibabel/tests/test_image_api.py
```diff
@@ -314,18 +314,18 @@ def _check_proxy_interface(self, imaker, meth_name):
         # New data dtype, no caching, doesn't use or alter cache
         fdata_new_dt = img.get_fdata(caching='unchanged', dtype='f4')
         # We get back the original read, not the modified cache
-        assert_array_equal(fdata_new_dt, proxy_data.astype('f4'))
+        assert_allclose(fdata_new_dt, proxy_data.astype('f4'), rtol=1e-05, atol=1e-08)
```
rmarkello: Maybe you can update the comment to reflect your rationale for changing this?

nibabel/tests/test_proxy_api.py
rmarkello: Any reason not to check this? Just to avoid the possibility of a silent fail.

effigies:
Thanks for the review, @rmarkello.

I'll plan to merge on Friday if there are no further comments.

effigies:
Alright, here we go...

Issue #842, opened 2019-11-18
parrec/minc1 don't subclass ArrayProxy


rmarkello:
While looking at #833 I noticed that neither `PARRECArrayProxy` nor `MincImageArrayProxy` subclass `ArrayProxy` despite there being considerable overlap in their functionality and structure. It seems like it would be reasonable to modify both classes such that they directly inherit from `ArrayProxy` to reduce duplication of some methods / doc-strings and make them a bit easier to keep in-line with `ArrayProxy` functionality. (Though I hope there's not a whole lot more changes coming to `ArrayProxy` in the near future.)

To be fair, based on my understanding neither `parrec` nor `minc1` are widely used formats anymore, so this would be a relatively low-impact change for what could be a headache of a PR. Still, figured I'd bring it up for discussion and see what people thought. I'd be happy to try my hand at this in the coming weeks, if there's interest.

matthew-brett:
I seem to remember I made a deliberate decision not to sub-class.  The array proxy was always meant to be an API, so it made sense for different classes to implement the API.  On the other hand it's perfectly possible that led to some code-duplication.  If there is some, that would be cleaned up by a PR, it would be worth giving a shot.  Maybe just do the basic stuff to see what it looks like?

effigies:
There are optional parts of the API, so perhaps it would make more sense to abstract out the mandatory parts into a(n abstract?) base class?

Some scale and some don't, so that might be a sensible branch in the hierarchy.

I suspect a large part of the duplication was added in #833, so this may only just now be starting to be worth it.

PR #844, opened 2019-11-19
FIX: Accept dtype parameter to ArrayProxy.__array__

### LINKED ISSUE #843: ArrayProxy.__array__ should support dtype argument
To fully implement the array-like interface, `ArrayProxy.__array__` should accept the `dtype` argument. Currently neither of the following works:

```Python
mask = np.bool_(img.dataobj)
mask = np.asanyarray(img.dataobj, dtype=bool)
```

```
In [6]: np.bool_(img.dataobj)                                                                       
TypeError                                 Traceback (most recent call last)
<ipython-input-6-bad9ebe4d523> in <module>
----> 1 np.bool_(img.dataobj)

TypeError: __array__() takes 1 positional argument but 2 were given

In [7]: np.asanyarray(img.dataobj, dtype=bool)                                                      
TypeError                                 Traceback (most recent call last)
<ipython-input-7-a7890277a83c> in <module>
----> 1 np.asanyarray(img.dataobj, dtype=bool)

/System/Volumes/Data/anaconda3/lib/python3.7/site-packages/numpy/core/numeric.py in asanyarray(a, dtype, order)
    589 
    590     """
--> 591     return array(a, dtype, copy=False, order=order, subok=True)
    592 
    593 

TypeError: __array__() takes 1 positional argument but 2 were given
```

Not sure whether fixing this is more of a bug fix or a new feature. I lean toward this being a bug... What do others think?

effigies:
At least back through numpy 1.12, the `__array__` protocol has [accepted](https://docs.scipy.org/doc/numpy-1.12.0/reference/arrays.classes.html?highlight=__array__#numpy.class.__array__) a dtype parameter.

Overall this is poorly documented in numpy (the above link the *only* place I've seen reference to the `dtype` parameter), but some manual testing seems to indicate it works, and that its absence produces errors in the cases described in #843. Systematic tests are added to the proxy API. Happy to add more, if we can think of specific cases that might trip us up.

It occurs to me that this largely replaces the `get_scaled()` method added in #833, with the primary difference being that `get_scaled()` guarantees not to downcast if it can't assure there will be no overflow. `get_scaled()` will also make as few copies as possible, while `__array__` may make another if `get_scaled(dtype).dtype != np.dtype(dtype)`. If that distinction is not worth drawing, it may be worth removing `get_scaled()` before it hits a release, and make the `__array__` interface THE way to control dtypes beyond `get_fdata()`.

One interesting side effect of this is that mask or atlas data that is known to be <=255, but the scale factors are not guaranteed to be `slope = 1, inter = 0` can be retrieved with minimal up-casting via:

```Python
np.array(atlas.dataobj, np.uint8)
# or 
np.uint8(atlas.dataobj)
```

This is of course quite dangerous, and with the current implementation, non-integer data will be happily coerced into meaningless garbage. Validating coercion or using safer casting rules could be a good idea, but that would also break the following equality by causing the RHS to raise an exception:

```Python
np.uint8(np.array(atlas.dataobj)) == np.array(atlas.dataobj, np.uint8)
```

All of your thoughts are very welcome.

Fixes #843. 
Affects #842.
Follow-up to #833. 

cc @jeromedockes @kchawla-pi @adelavega @rmarkello @matthew-brett


effigies:
I'm inclined to drop `get_scaled()` in favor of a dtyped `__array__`. Any thoughts?

adelavega:
The upside of `get_scaled` is that its a bit clearer what it idoes. Using `__array__` directly seems a bit more opaque, unless well documented. But it seems this is not something you necessarily want to encourage doing anyways (versus the standard way of accessing, which is `get_fdata`), so maybe that's fine.

If so, then I don't see much need for `get_scaled` either.

effigies:
I'd be happy to add a more complete description of the `dataobj` to the docs, if that would help. Right now, it's a pretty small thing here: https://nipy.org/nibabel/images_and_memory.html#use-the-array-proxy-instead-of-get-fdata

Just to concisely summarize: `np.array(img.dataobj)` (and `np.asarray` and `np.asanyarray`) is a valid way of getting information out of the object that's guaranteed to have had scaling applied to it, but the `dtype` is ambiguous. `np.array(img.dataobj, dtype=np.int16)` is (with this) very precise, although it may involve unsafe casting.

`img.dataobj.get_scaled()` is somewhere in between, with bounded ambiguity. Which might still be useful. If I say `get_scaled(dtype='uint16')`, I know I have something to which I can safely do anything I could do to a `uint16`.

IDK if that affects your thinking here.

rmarkello:
In my mind this is similar to the recent `pandas` (v0.25.0) transition away from `DataFrame.get_values()` (now deprecated) to `np.asarray(DataFrame)` and/or `DataFrame.to_numpy()`, both of which allow unsafe casting of the sort you describe.

I'd personally be fine to see `get_scaled()` replaced entirely by `__array__`, but if you're really worried about the casting you could also add a small check (e.g., confirm whether the provided `dtype` would result in potentially unsafe casting and throw a warning so that people are aware of what they're doing). That way it would still return whatever datatype was requested so the equality you proposed in the initial post would pass, but you're not being too prescriptive.

effigies:
Okay. I've dropped `get_scaled()` for now, and will remove it from the API changes in 3.0. It can always be added back, if it's useful enough.

effigies:
Any further comments? I'm inclined to do a second RC after this, to give a bit more testing time.

PR #846, opened 2019-11-25
DOC: Discuss slicer interface

effigies:
We never documented the `slicer` interface outside the API docs, which is less prominent than the feature deserves, given its usefulness, IMO.

Comments appreciated.

Follow-up to #550.


effigies:
Doctest passing. Please open a PR if you have suggestions for improvements.

PR #840, opened 2019-11-11
[wip] converting tests from  nibabel/tests/

djarecka:
will be adding more tests

PR #847, opened 2019-11-28
TEST: Check non-integral slopes, intercepts in ArrayProxy API

effigies:
Realized the ArrayProxy API tests are only using integral slopes and intercepts. Added some 5 s.f. parameters.

Currently breaks tests. Will look in detail when I can, but more eyes are always welcome.

This should be resolved and merged into #844, to ensure that the changes to results stay within tolerance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,7 +130,6 @@ script:
           cd for_testing
           cp ../.coveragerc .
           nosetests --with-doctest --with-coverage --cover-package nibabel nibabel
-          pytest -v ../nibabel/tests/test_affines.py ../nibabel/tests/test_volumeutils.py
       else
           false
       fi
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -360,10 +360,13 @@ class ArrayProxy(object):
         scl_slope = np.asanyarray(self._slope)
         scl_inter = np.asanyarray(self._inter)
         use_dtype = scl_slope.dtype if dtype is None else dtype
-        slope = scl_slope.astype(use_dtype)
-        inter = scl_inter.astype(use_dtype)
+
+        if np.can_cast(scl_slope, use_dtype):
+            scl_slope = scl_slope.astype(use_dtype)
+        if np.can_cast(scl_inter, use_dtype):
+            scl_inter = scl_inter.astype(use_dtype)
         # Read array and upcast as necessary for big slopes, intercepts
-        scaled = apply_read_scaling(self._get_unscaled(slicer=slicer), slope, inter)
+        scaled = apply_read_scaling(self._get_unscaled(slicer=slicer), scl_slope, scl_inter)
         if dtype is not None:
             scaled = scaled.astype(np.promote_types(scaled.dtype, dtype), copy=False)
         return scaled
--- a/nibabel/tests/test_proxy_api.py
+++ b/nibabel/tests/test_proxy_api.py
@@ -221,6 +221,8 @@ class TestAnalyzeProxyAPI(_TestProxyAPI):
             offsets = (self.header_class().get_data_offset(),)
         else:
             offsets = (0, 16)
+        # For non-integral parameters, cast to float32 value can be losslessly cast
+        # later, enabling exact checks, then back to float for consistency
         slopes = (1., 2., float(np.float32(3.1416))) if self.has_slope else (1.,)
         inters = (0., 10., float(np.float32(2.7183))) if self.has_inter else (0.,)
         for shape, dtype, offset, slope, inter in product(self.shapes,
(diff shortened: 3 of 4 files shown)


nibabel/arrayproxy.py
```diff
@@ -360,10 +360,13 @@ def _get_scaled(self, dtype, slicer):
         scl_slope = np.asanyarray(self._slope)
         scl_inter = np.asanyarray(self._inter)
         use_dtype = scl_slope.dtype if dtype is None else dtype
-        slope = scl_slope.astype(use_dtype)
-        inter = scl_inter.astype(use_dtype)
+
+        if np.can_cast(scl_slope, use_dtype):
+            scl_slope = scl_slope.astype(use_dtype)
+        if np.can_cast(scl_inter, use_dtype):
+            scl_inter = scl_inter.astype(use_dtype)
```
effigies: Unconditionally casting to `int` types would truncate the new parameters.

.travis.yml
```diff
@@ -129,8 +129,16 @@ script:
...
+          PASS=true
           nosetests --with-doctest --with-coverage --cover-package nibabel nibabel
+          if [ "$?" != "0" ]; then
+              PASS=false;
+          fi
           pytest -v ../nibabel/tests/test_affines.py ../nibabel/tests/test_volumeutils.py
+          if [ "$?" != "0" ]; then
+              PASS=false;
+          fi
+          $PASS
```
effigies: @djarecka Just a note that your pytest tests are masking nosetest failures. Can you please cherry-pick 2716f8d600cb511eddd7d314854387f008e6b17a or rebase onto this branch?

nibabel/tests/test_proxy_api.py
```diff
@@ -216,11 +221,10 @@ def obj_params(self):
             offsets = (self.header_class().get_data_offset(),)
         else:
             offsets = (0, 16)
-        slopes = (1., 2.) if self.has_slope else (1.,)
-        inters = (0., 10.) if self.has_inter else (0.,)
-        dtypes = (np.uint8, np.int16, np.float32)
+        slopes = (1., 2., float(np.float32(3.1416))) if self.has_slope else (1.,)
+        inters = (0., 10., float(np.float32(2.7183))) if self.has_inter else (0.,)
```
effigies: Stupidly picked decimals that couldn't be represented exactly. Cast them to float32 to make it concrete, then back to `float` for consistency.

Can pick other numbers if this will just lead to confusion.

nibabel/tests/test_proxy_api.py
```diff
@@ -262,7 +266,7 @@ def sio_func():
                 dtype=dtype,
                 dtype_out=dtype_out,
                 arr=arr.copy(),
-                arr_out=arr * slope + inter,
+                arr_out=arr.astype(dtype_out) * slope + inter,
```
effigies: If `arr` is float32, `arr * slope + inter` is float32. This hasn't mattered with the simple factors we've been using, but the new parameters produce different rounding errors in float32 (basic numpy coercion) and float64 (nibabel array scaling).

This change ensures that we produce our expected array in the wider dtype.

.travis.yml
djarecka: sorry for this, maybe you can simply remove pytest from travis for now, and just wait for #840

effigies: Sure.

.travis.yml
```diff
@@ -130,7 +130,6 @@ script:
           cd for_testing
           cp ../.coveragerc .
           nosetests --with-doctest --with-coverage --cover-package nibabel nibabel
-          pytest -v ../nibabel/tests/test_affines.py ../nibabel/tests/test_volumeutils.py
```
effigies: Was masking nosetest failures. Removing for now.

djarecka: we also added to azure... https://github.com/nipy/nibabel/blob/master/.azure-pipelines/windows.yml#L44

effigies: Amended. Thanks.

nibabel/arrayproxy.py
adelavega: So if it can't cast, but the user asks to, should this maybe spit a warning out?

nibabel/tests/test_proxy_api.py
adelavega: Maybe add this as a comment, otherwise makes sense with explanation

nibabel/arrayproxy.py
effigies: It's a helper method intended to do a best-effort downcast, but not to the detriment of accuracy. If you ask me to return `int`s but it would change the values (beyond a floating point rounding error), then the expected behavior is to keep floats. Spitting out a warning here would be extremely noisy.

effigies:
Pretty happy with these tests (though they do add time... I probably don't need to do all the ints, uints, floats and complexfloats).

These will help with stress-testing #844, so if there are no objections I'll merge tomorrow.

PR #848, opened 2019-12-06
[REF] Deprecate kw_only_meth/func decorators

### LINKED ISSUE #841: Replace @kw_only_meth/func decorator with "*" argument
Python 3 supports the use of `*` in function signatures to denote that subsequent arguments must be supplied as keywords. Historically, `nibabel` has used `nibabel.keyword_only.kw_only_{meth,func}` to do this, instead. Since `nibabel` is no longer supporting Python 2 it might be nice to remove these and just use the '*' syntax! I'd be happy to make a PR for this, if you think it would be desirable.

rmarkello:
Closes #841.

This is a minor update, but it removes a bit of custom code (the `nibabel.keywordonly` module) forcing methods/functions to have a pre-specified number of positional arguments and instead uses the Python 3.x `*` argument to achieve the same performance.

I haven't added any additional tests to check that the `*` argument performs identically to the decorators since `test_keywordonly.py` was only testing the decorators themselves (and not the functions they wrapped)&mdash;and those functions are now removed&mdash;but let me know if you'd like to see something explicit!


rmarkello:
Also, in case it needs to be said: this is very low priority and can just hang out until after the 3.0 release! I just had a few minutes and though I'd get it off my to-do list 😅

effigies:
Since this is part of the API, I'm afraid it will need to go through a deprecation cycle.

Have a look at:

https://github.com/nipy/nibabel/blob/fcc54489d04ba218f10ba98140f1596c0ad47553/nibabel/minc.py#L1-L10

https://github.com/nipy/nibabel/blob/fcc54489d04ba218f10ba98140f1596c0ad47553/nibabel/tests/test_removalschedule.py#L4-L9

rmarkello:
Gotcha! Do you have a preference for what version to list for deprecation? I'm assuming 4.0.0 based on what's currently listed.

effigies:
I think we usually go +2 major versions, so let's say 5.0.

effigies:
Thanks!

Issue #851, opened 2019-12-11
ENH: implement `mri_convert --conform` functionality


kaczmarj:
hi nibabel devs, i would like to replicate freesurfer's `mri_convert --conform` functionality, as described in https://github.com/neuronets/kwyk/issues/5. I have the code for it below. It does the following:

- Resize to (256mm)^3 bounding box
- resample to (1mm)^3 voxels
- rescale to 0-255
- reorients to RAS

`mri_convert --conform` reorients to LIA but RAS is cooler imho.

i made this an issue instead of a pr because i don't know where this sort of thing should go in the code base. @effigies suggested `nibabel.cmdline` but i'm not sure.

```python
from pathlib import Path
import nibabel as nib
import nibabel.processing
import numpy as np

def transform_range(x, new_min, new_max):
    x = np.asarray(x)
    x_min, x_max = x.min(), x.max()
    return (((x - x_min) * (new_max - new_min)) / (x_max - x_min)) + new_min

def conform(from_img, out_shape=(256, 256, 256), 
            voxel_size=(1.0, 1.0, 1.0), order=3, cval=0.0):
    # Create fake image of the image we want to resample to.
    hdr = nib.Nifti1Header()
    hdr.set_data_shape(out_shape)
    hdr.set_zooms(voxel_size)
    dst_aff = hdr.get_best_affine()
    to_img = nib.Nifti1Image(np.empty(out_shape), affine=dst_aff, header=hdr)
    # Resample input image.
    out_img = nib.processing.resample_from_to(
        from_img=small, to_vox_map=to_img, order=order, cval=cval)
    # Cast to uint8.
    data = out_img.get_fdata()
    data = transform_range(data, new_min=0.0, new_max=255.0)
    data = data.round(out=data).astype(np.uint8)
    out_img._dataobj = data
    out_img.set_data_dtype(np.uint8)
    # Reorient to RAS.
    out_img = nib.as_closest_canonical(out_img)
    return out_img

filename = str(Path(nib.testing.data_path) / 'anatomical.nii')
small = nib.load(filename)
out_img = conform(small)
```

effigies:
It would make sense for the function to go in nibabel.processing, but a command-line interface also seems useful.

PR #853, opened 2019-12-12
NF: Conformation function and CLI tool to apply shape, orientation and zooms

kaczmarj:
closes https://github.com/nipy/nibabel/issues/851
related to https://github.com/neuronets/kwyk/issues/5

This PR adds a function to resample a volume to arbitrary shape and voxel sizes. This is meant to be a close replica of freesurfer's `mri_convert --conform` command. In freesurfer, the image is reoriented to LIA, but in this command, the image is reoriented to RAS.

So this command:
- resamples to a specified output shape
- resamples to a specified voxel size
- casts to uint8 (while also changing the range of data to [0, 255]).
- reorients to RAS

The heavy lifting is done by `nibabel.processing.resample_from_to()`.

I have not added tests yet.


kaczmarj:
ready for review!

nibabel/processing.py
```diff
@@ -310,3 +311,96 @@ def smooth_image(img,
...
+        A copy of the transformed data.
+
+    Examples
+    --------
+    >>> _transform_range([2, 4, 6], -1, 1)
+    array([-1.,  0.,  1.])
+    """
+    x = np.asarray(x)
+    x_min, x_max = x.min(), x.max()
+    return (x - x_min) * (new_max - new_min) / (x_max - x_min) + new_min
```
yarikoptic: should it become resilient to `x_max == x_min` when it would either set it to `new_min`  or blow some better than ZeroDivisonError exception?

nibabel/tests/test_processing.py
```diff
@@ -426,3 +427,33 @@ def test_against_spm_resample():
...
+    assert c.dataobj.dtype == np.dtype(np.uint8)
+    assert aff2axcodes(c.affine) == ('R', 'A', 'S')
+
+    c = conform(anat, out_shape=(100, 100, 200), voxel_size=(2, 2, 1.5))
+    assert c.shape == (100, 100, 200)
+    assert c.header.get_zooms() == (2, 2, 1.5)
+    assert c.dataobj.dtype == np.dtype(np.uint8)
+    assert aff2axcodes(c.affine) == ('R', 'A', 'S')
+
+    # Error on non-3D images.
```
yarikoptic: is that what freesurfer does?  I thought it might be nice to have this one applicable to 4D (or 5D - AFNI etc) series, to "conform" the first 3 dimensions to desired resolution/orientation etc

kaczmarj: i don't think it is what freesurfer does, but for simplicity and due to my lack of knowledge, i made it this way. how should the output shape and zooms be changed to support 4- and 5-D images?

nibabel/processing.py
kaczmarj: thank you @yarikoptic - i will account for `x_max == x_min` and i will add a test for this case.

kaczmarj:
> Is actual (command line) "command" interface coming in this PR or later?

@yarikoptic - it could come in this pr. is there a preferred entrypoint? `nib-conform`?

effigies:
So you're already not doing exactly what FreeSurfer does, so I don't think you need to stick too close to what it does.

I think you obviously want to be able to specify dimensions, zooms and orientation from the CLI. I would suggest that the dtype and rescaling might also be useful parameters. For rescaling, I would make it a boolean flag to indicate that the data should be rescaled to the full range of the dtype (True) or preserve values and adjusting scale parameters as needed (False).

Anyway, I haven't reviewed the code yet, but figured I might chime in about capabilities.

kaczmarj:
>For rescaling, I would make it a boolean flag to indicate that the data should be rescaled to the full range of the dtype (True) or preserve values and adjusting scale parameters as needed (False).

this makes sense for integer types, but i don't think it should be done for float types. 

or perhaps we could make things much easier and not cast the data.

effigies:
I agree that rescaling into the range for floats doesn't make sense.

effigies:
Hi @kaczmarj just checking in on this. Do you want to aim for a 3.1 release (soon-ish...) or 3.2?

kaczmarj:
hi @effigies - thanks for the ping. i will work on this tonight and will update. i aim for 3.1.

kaczmarj:
@effigies - i added the command-line program `nib-conform` (including tests). i also narrowed the `conform` function. now, it conforms to a requested shape and voxel size. it does not change data type or transform data. how does that sound?

nibabel/processing.py
```diff
@@ -310,3 +311,80 @@ def smooth_image(img,
...
+    to_img = Nifti1Image(np.empty(out_shape), affine=dst_aff, header=hdr)
+
+    # Resample input image.
+    out_img = resample_from_to(
+        from_img=from_img, to_vox_map=to_img, order=order, mode="constant",
+        cval=cval, out_class=out_class)
+
+    # Reorient to desired orientation.
+    ornt = axcodes2ornt(orientation, labels=list(zip('RPI', 'LAS')))
+    return out_img.as_reoriented(ornt)
```
kaczmarj, line 390: These `labels` are not the default. I included them because if I created an orientation array for `"RAS"` with the default, the image would be in `"LAS"` orientation. Any idea why this might be?

effigies: This doesn't seem to take into account your original orientation, so very likely what's happening is that you have an LAS image. `axcodes2ornt("RAS")` will get you an identity transform.

To get the transform to RAS, use [io_orientation(out_img.affine)](https://nipy.org/nibabel/reference/nibabel.orientations.html#nibabel.orientations.io_orientation). To get the target transform, use `axcodes2ornt`.

Example:

```Python
def flip_img(img, axcodes):
    start_ornt = nb.io_orientation(img.affine)
    end_ornt = nb.orientations.axcodes2ornt(axcodes)
    transform = nb.orientations.ornt_transform(to_ras, to_target)
    return img.as_reoriented(transform)
```

bin/nib-conform
```diff
@@ -0,0 +1,17 @@
...
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""
+Conform a volume to a new shape and/or voxel size.
+"""
+
+from nibabel.cmdline.conform import main
+
+if __name__ == '__main__':
+    main()
```
effigies, line 17: I don't think we install these files anywhere...

nibabel/cmdline/conform.py
```diff
@@ -0,0 +1,65 @@
...
+
+    out_img = conform(
+        from_img=from_img,
+        out_shape=kwargs["out_shape"],
+        voxel_size=kwargs["voxel_size"],
+        order=3,
+        cval=0.0,
+        orientation=kwargs["orientation"])
+
+    out_img.to_filename(kwargs["outfile"])
```
effigies, line 65: Might as well use:

```suggestion
    nb.save(out_img, kwargs["outfile"])
```

This will avoid the requirement that the output file format be the same as the input file format, and you can do things like:

```
nib-conform old_analyze.img new_nifti.nii
```

nibabel/processing.py
```diff
@@ -310,3 +311,80 @@ def smooth_image(img,
...
     return out_class(sm_data, img.affine, img.header)
+
+
+def conform(from_img,
+            out_shape=(256, 256, 256),
+            voxel_size=(1.0, 1.0, 1.0),
+            order=3,
+            cval=0.0,
+            orientation='RAS',
+            out_class=Nifti1Image):
```
effigies, line 322: From the docstring, the default behavior is to keep the input class. I think that's sensible.

```suggestion
            out_class=None):
```

nibabel/cmdline/conform.py
```diff
@@ -0,0 +1,65 @@
...
+    return p
+
+
+def main(args=None):
+    """Main program function."""
+    parser = _get_parser()
+    if args is None:
+        namespace = parser.parse_args(sys.argv[1:])
+    else:
+        namespace = parser.parse_args(args)
```
effigies, line 49: Is this distinct from:

```suggestion
    namespace = parser.parse_args(args)
```

nibabel/cmdline/conform.py
```diff
@@ -0,0 +1,65 @@
...
+
+    kwargs = vars(namespace)
```
effigies, line 51: I don't understand the value of this idiom. Instead of `namespace.infile`, you now write `kwargs["infile"]`.

Replace `namespace` with `opts`, and you're even saving characters to stick with the original object.

nibabel/tests/test_processing.py
effigies: The output shape and zooms would not be affected by rolling the first three axes.

In any event, I'm okay with getting this in and adding 4/5D image support later, but if we're going to have a test, then we should mark it as one that we intend to relax. Otherwise it may look like a desired constraint to future developers.

nibabel/processing.py
```diff
@@ -310,3 +311,80 @@ def smooth_image(img,
...
+        raise ValueError("`out_shape` must have {} values".format(required_ndim))
+    elif len(voxel_size) != required_ndim:
+        raise ValueError("`voxel_size` must have {} values".format(required_ndim))
+
+    # Create fake image of the image we want to resample to.
+    hdr = Nifti1Header()
+    hdr.set_data_shape(out_shape)
+    hdr.set_zooms(voxel_size)
+    dst_aff = hdr.get_best_affine()
+    to_img = Nifti1Image(np.empty(out_shape), affine=dst_aff, header=hdr)
```
effigies, line 381: Does `resample_from_to` care about the type of `to_vox_map`? If so, we might want to make this from `from_img.__class__` and `from_img.header_class`.

bin/nib-conform
kaczmarj: should I remove this file?

nibabel/cmdline/conform.py
kaczmarj: Thanks, I added this.

nibabel/processing.py
kaczmarj: thank you so much. i've correct the orientation code in `conform`

nibabel/processing.py
kaczmarj: I don't know if it matter to be honest, but I added this just to be safe.

nibabel/processing.py
```diff
@@ -310,3 +311,82 @@ def smooth_image(img,
...
+        raise ValueError("`out_shape` must have {} values".format(required_ndim))
+    elif len(voxel_size) != required_ndim:
+        raise ValueError("`voxel_size` must have {} values".format(required_ndim))
+
+    # Create fake image of the image we want to resample to.
+    hdr = from_img.header_class()
+    hdr.set_data_shape(out_shape)
+    hdr.set_zooms(voxel_size)
+    dst_aff = hdr.get_best_affine()
+    to_img = from_img.__class__(np.empty(out_shape), affine=dst_aff, header=hdr)
```
effigies, line 381: Consider the following image:

```Python
nb.Nifti1Image(np.zeros((256, 256, 256)), affine=np.eye(4))
```

This is already RAS, 1mm isotropic and 256^3 in dimension. However, only the first octant (bounded by `(0, 0, 0), (0, 0, 128), ... ( 128, 128, 128)`) will be shared with `to_img`. When we resample, I assume that we'll get a heavily cropped image.

Assuming I'm correct, it might be better to construct the affine such that the same point is in the center of the image.

bin/nib-conform
effigies: I think so. I'd try removing and make sure `pip install .` produces a working `nib-conform` command.

nibabel/processing.py
kaczmarj: ras_256iso
ras_128iso_conformed


I have a 256^3 image with 1^3 voxels in RAS, and I ran it through `nib-conform` to change it to 128^3 with 1^3 voxels. It is cropped, but the center of the original is the center of the output. Does this answer what you were getting at?

effigies: Nope, try the following:

```Python
import nibabel as nb
import numpy as np

img = nb.load(fname)
nb.Nifti1Image(img.dataobj, np.eye(4), img.header).to_filename("corner.nii.gz")
```

This will place the (0, 0, 0) RAS point in the corner of the image. Conform `corner.nii.gz` and you'll see what I'm talking about.

kaczmarj: OK, did that, and indeed the conformed `corner.nii.gz` is heavily cropped. How can the affine be constructed to maintain the same center?

effigies: Need to solve for:

```
ras = affine * ijk
affine' * ijk' = affine * ijk
```

Noting that the affine is

```
[rzs t
   0 1]
```

And `t` is the thing you can change.

If you've got time now, go for it, but I can work out the formula this afternoon.

kaczmarj: Let me take a shot at it this afternoon

kaczmarj: OK, this is way above level of understanding... sorry @effigies. i'll have to take linear algebra in grad school

effigies:
@kaczmarj To be clear, are you trying to de-oblique images as well? Right now if there is a rotation in the affine, this function will remove it during the resampling and have a diagonal affine. I do not believe `mri_convert -conform` does deoblique.

kaczmarj:
@effigies - i am not trying to de-oblique the images. turns out i overlooked several things when i created that empty image to which input images are conformed. the only requirements for this function are to change the voxel sizes and change the shape of the output image.

kaczmarj:
I've found the relevant bits of freesurfer's source. I will dig through it to understand how the transformation matrix is computed, and I'll update the PR.

The code that creates the template image against which the input is conformed: https://github.com/freesurfer/freesurfer/blob/4db941ef298c0ac5fb78c29fd0e95571ac363e16/utils/mri_conform.cpp#L50-L68

The `MRI` class that includes the image geometry attributes: https://github.com/freesurfer/freesurfer/blob/4db941ef298c0ac5fb78c29fd0e95571ac363e16/include/mri.h#L192-L224

The part of `mri_convert` that runs if `--conform` is specified: https://github.com/freesurfer/freesurfer/blob/ea988613231c850785b13632205467613c88d89d/mri_convert/mri_convert.cpp#L2699-L2718

nibabel/processing.py
kaczmarj: it looks like this is the freesurfer implementation of the solution we're looking for https://github.com/freesurfer/freesurfer/blob/ea988613231c850785b13632205467613c88d89d/utils/mri.cpp#L415-L523

kaczmarj:
@effigies - following https://github.com/freesurfer/freesurfer/blob/4db941ef298c0ac5fb78c29fd0e95571ac363e16/utils/mri_conform.cpp#L140-L256 , i changed the `conform` command to copy the image header and then modify the zooms and shape. If you try to conform the corner image, then the conformation is correct and looks like the output of `mri_convert --conform`.

```python
import nibabel as nb
import nibabel.processing
import numpy as np

img = nb.load(fname)
corner = nb.Nifti1Image(img.dataobj, np.eye(4), img.header)
corner.to_filename("corner.nii.gz")

nb.processing.conform(corner).to_filename("corner_conformed_nib.nii.gz")

# Similar output as
# mri_convert --conform corner.nii.gz corner_conformed_fs.nii.gz
```

However, changing the voxel dimensions from something other than 1mm^3 does not work anymore... any idea why?

effigies:
Sorry, my ability to sustain focus is compromised at the moment.

I got around to fiddling: kaczmarj#1.

kaczmarj:
Thanks @effigies. As far as I'm concerned, this PR should be good for final review. When merging, can you squash and merge? There are some commits in here are irrelevant and include incorrect code.

effigies:
Will review tomorrow or Friday. Happy to squash, if that's what you prefer.

effigies:
LGTM. @yarikoptic What do you think?

effigies:
Bump @yarikoptic, in case this notification escaped, too.

effigies:
@kaczmarj Sorry, forgot to squash. I think we'll survive a little mess.

leej3:
@kaczmarj just to flag a gotcha for the freesurfer conform code... they seem to do image histogram normalization (shift values to center white matter somewhere around 100) and I believe they truncate to uint8 unless told not to do so. We got tripped up by that in attempting an identical but all python implementation. @patrick-mcclure may have some input on details of how to resolve it (though we may have just stuck with the freesurfer output?). It might be worth taking into account if this is for neural nets.

patrick-mcclure:
@kaczmarj @leej3 For our successful projects, we, indeed, did stick to volumes pre-processed using Freesurfer's mri_convert.

leej3:
Ok, thanks, good to know. And was the diagnosis that it was primarily the histogram normalization or were there other hypotheses?

kaczmarj:
Thanks @leej3 and @patrick-mcclure. You are right that this does not perform histogram normalization or truncating to uint8. I think not truncating to uint8 should be fine, because that truncation will lose precision anyway. I'd be curious to hear if you think histogram normalization had a beneficial effect on the model performance.

leej3:
Well you certainly can't mix and match:train on mri_convert output and predict on output that was conformed without normalization. My memory was that the normalization performed by freesurfer was important for heterogeneous datasets, i.e. it was essential for any broadly usable tool. Correct me if I'm wrong @patrick-mcclure.

satra:
> they seem to do image histogram normalization

@leej3 -  is this done during `mri_convert conforming step` or later in the freesurfer process? my recollection is that this is done later because that requires getting at what white matter is. `mri_convert` is generally quite simple.

patrick-mcclure:
From what I remember, training in normalized volumes and then testing on
non-normalized volumes lead to a significant decrease in test performance
compared to testing using the same volumes with normalization.

kaczmarj:
@satra @leej3 - it looks like the histogram normalization happens during the conversion to uint8: https://github.com/freesurfer/freesurfer/blob/dev/mri_convert/mri_make_uchar.cpp

from the comments at the top of that file:

```
 * uses the Tal xform to find a ball of voxels that are mostly brain.
 * the top of the intensity histogram in this ball will then be white matter,
 * which allows us to center it at the desired value approximately (110)
```

satra:
@patrick-mcclure and @leej3 - i think there are two things getting mixed up here.

1. what the conform step in freesurfer does, which is what this should implement
2. what is used for model training and what the model can be applied to. 

let's keep 2 out of this discussion, since that relates to preprocessing for models that nibabel could not care less about.

satra:
@kaczmarj - i would still like to know how mri_convert can determine where this ball is without doing any registration. i can conform a blank mri scan if i want to.

leej3:
The mri_make_uchar tool compiled from the code @kaczmarj referenced is in fact run later, I believe [here](https://github.com/freesurfer/freesurfer/blob/88682efec81a0399643ca0363821f9cd4f6d38f3/scripts/recon-all.v6.hires#L1805) in recon-all.

It's coming back to me now. I think I had gotten confused by the output naming and assumed orig.mgz was the output of the first step of the pipeline with just a basic conform performed but it is in fact following the uint8 conversion (which does that normalization). 

So summary is that mri_convert is indeed just performing a format conversion and with the -conform option it additionally performs operations equivalent to the ones here.

Agreed, 2. is another problem

kaczmarj:
to add onto that @leej3 , `mri_make_uchar` is not used by `mri_convert --conform`. sorry for any confusion. it does look like it's used by recon-all as @leej3 noted.

`mri_convert --conform` will do the following things (i went through the code again):
1. read in volume
1. create an mri template with 1^3 mm voxels, size 256^3, and LIA orientation.
    - the header of the template indicates that the volume is uint8 (but the values have not been cast yet).
1. change data type to uint8 with [`MRIchangeType`](https://github.com/freesurfer/freesurfer/blob/666aa24824f05a25e0f6be82722c99af26718da9/utils/mri.cpp#L11516-L11614). this function does not depend on an atlas. it creates a 1000-bin histogram of the mri values and then scales the values to fit the limits of uint8 (it supports other integer types, too)

leej3:
Thanks for detailing that. 

Out of interest have you worked on a python implementation (or are you aware of one) that does the sort of white matter aware histogram normalization that mri_make_uchar  performs?

satra:
@kaczmarj - perhaps why it needs to create a histogram is relevant to implementing 3. since the basic scaling can be implemented just with nibabel's changing datatype function, which does the scaling on the range of the data.

kaczmarj:
@satra - i am still trying to figure that out. i might need to translate the c++ code to python/numpy to give a good answer.

The linear transformation is:
```
val = dest_min + scale * (val - src_min);
```

where `dest_min` is the minimum of the output data type (0 for uint8), and `val` is the value being scaled.

other values are as follows:

```
src_min, src_max = limits(mri);
src_min = (float)bin * bin_size + src_min;
src_max = (float)bin * bin_size + src_min;
scale = (dest_max - dest_min) / (src_max - src_min);
```

the `bin` values in the calculation of `src_min` and `src_max` are different and come from `for` loops. i'm having trouble understanding what's going on there.

satra:
@kaczmarj - this looks like a straight up `np.histogram`  operation. with the given keyword args. the only thing that would change would be `src_min` and `src_max` given the parameters `f_low` and `f_high`. and that would require the clipping bit.

and it seems `f_low` and `f_high` are controlling for number of elements to be discarded from the histogram in bins to determine `src_min` and `src_max`.

so should be easily reimplementable with `np.histogram`.

Issue #803, opened 2019-09-09
RFC: NEP 29 - Python and Numpy version support


effigies:
Numpy has put out https://numpy.org/neps/nep-0029-deprecation_policy.html, which proposes that scientific Python projects adopt the following approach to supporting Python and Numpy:

> * This project supports at least the minor versions of Python initially released 42 months prior to a planned project release date.
> * The project will always support at least the 2 latest minor versions of Python.
> * support minor versions of numpy initially released in the 24 months prior to a planned project release date or the oldest version that supports the minimum Python version (whichever is higher).
> * The project will always support at least the 3 latest minor versions of NumPy.
> 
> The minimum supported version of Python will be set to python_requires in setup. All supported minor versions of Python will be in the test matrix and have binary artifacts built for releases.
> The project should adjust upward the minimum Python and NumPy version support on every minor and major release, but never on a patch release.

Which works out to the following schedule:

```
On Jan 16, 2019 drop support for Numpy 1.12 (initially released on Jan 15, 2017)
On Mar 14, 2019 drop support for Python 3.5 (initially released on Sep 13, 2015)
On Jun 08, 2019 drop support for Numpy 1.13 (initially released on Jun 07, 2017)
On Jan 07, 2020 drop support for Numpy 1.14 (initially released on Jan 06, 2018)
On Jun 23, 2020 drop support for Python 3.6 (initially released on Dec 23, 2016)
On Jul 23, 2020 drop support for Numpy 1.15 (initially released on Jul 23, 2018)
On Jan 13, 2021 drop support for Numpy 1.16 (initially released on Jan 13, 2019)
On Jul 26, 2021 drop support for Numpy 1.17 (initially released on Jul 26, 2019)
On Dec 26, 2021 drop support for Python 3.7 (initially released on Jun 27, 2018)
```

This is a more rapid turnaround than we've been working with, as we've only recently dropped support for numpy < 1.12. I want to get people's thoughts on whether we should abide by this (bearing in mind that this only affects major and minor releases; patch releases can continue supporting older versions). I do think there's value in having a predictable schedule, even if it's not identical. I see a couple options (and am open to others):

1) NEP 29 + 1 year, which will more closely match Python end-of-life
2) Python end-of-life + NEP 29 schedule for numpy
3) Continue as we are, ratcheting numpy when we really need to, but not faster than NEP 29

At any rate, this is definitely something we should be aware of in terms of how the overall ecosystem will interact with nibabel if we keep a slower timeline.

[...]

PR #868, opened 2020-01-27
NEP29: Bump minimum numpy to 1.13

effigies:
As of January 16, the next minor release (3.1) is to require numpy 1.13.

TODO:

* [x] Check for any code that branches on numpy < 1.13
* [x] Titrate a minimum `packaging` while we're at it.

Ref #803.


effigies:
Looks like we can use any version of `packaging`. That's simple.

PR #871, opened 2020-02-04
TEST: Add BaseTestCase class to skip TestCases starting with ``_``

effigies:


effigies:
Cancelled the tests because they're going to :x: anyway. See https://travis-ci.com/nipy/nibabel/jobs/283443185 to verify that new failures aren't happening.

djarecka:
@effigies - looks like everything is working, the only thing worries me that when you run tests like `nibabel/tests/test_analyze.py`, there are many `SKIPPED` reports, because of the `_TestLabeledWrapStruct`. Not sure if this was the same with nose

effigies:
Yeah, I tried to get rid of them, but couldn't. One option is to `del` the class after using it, but I'm not sure it's worth it.

djarecka:
yes, probably not worth it. 
I believe we even had this discussion about `del` in DC. remember that was talking to you about my issues with this class in general

PR #872, opened 2020-02-04
TEST: Cleanups

effigies:
@djarecka @robbisg If you're both good with these updates, feel free to merge.

### DIFF
--- a/nibabel/streamlines/tests/test_array_sequence.py
+++ b/nibabel/streamlines/tests/test_array_sequence.py
@@ -9,8 +9,6 @@ import pytest
 from ...testing_pytest import assert_arrays_equal
 from numpy.testing import assert_array_equal
 
-import pytest; pytestmark = pytest.mark.skip()
-
 from ..array_sequence import ArraySequence, is_array_sequence, concatenate
 
 
@@ -37,9 +35,7 @@ def check_empty_arr_seq(seq):
     assert len(seq._lengths) == 0
     # assert_equal(seq._data.ndim, 0)
     assert seq._data.ndim == 1
-    
-    # TODO: Check assert_true
-    # assert_true(seq.common_shape == ())
+    assert seq.common_shape == ()
 
 
 def check_arr_seq(seq, arrays):
@@ -68,9 +64,9 @@ def check_arr_seq(seq, arrays):
 
 
 def check_arr_seq_view(seq_view, seq):
-    assert seq_view._is_view is True
+    assert seq_view._is_view
     assert seq_view is not seq
-    assert (np.may_share_memory(seq_view._data, seq._data)) is True
+    assert np.may_share_memory(seq_view._data, seq._data)
     assert seq_view._offsets is not seq._offsets
     assert seq_view._lengths is not seq._lengths
 
@@ -179,7 +175,6 @@ class TestArraySequence(unittest.TestCase):
         element = generate_data(nb_arrays=1,
                                 common_shape=SEQ_DATA['seq'].common_shape*2,
                                 rng=SEQ_DATA['rng'])[0]
-
         with pytest.raises(ValueError):
             seq.append(element)
 
@@ -275,7 +270,6 @@ class TestArraySequence(unittest.TestCase):
         # Test invalid indexing
         with pytest.raises(TypeError):
             SEQ_DATA['seq']['abc']
-            #SEQ_DATA['seq'].abc
 
         # Get specific columns.
         seq_view = SEQ_DATA['seq'][:, 2]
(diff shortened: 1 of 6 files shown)


effigies:
I guess also make sure these pass Travis. :-)

effigies:
@djarecka My Riot client is having trouble connecting. Assuming https://travis-ci.com/nipy/nibabel/builds/147415802 passes, then the new expectation will be that all PRs into `pytest` should pass Travis.

robbisg:
I'm ok, but I will wait @djarecka.
Then I rebase my personal branch, merge and push other modules!

effigies:
@robbisg Feel free to work off of the `pytest` branch for now, rather than wait for tests. If you avoid the files I've touched, then there shouldn't be any conflict.

djarecka:
sorry, was also having lunch and waiting for my riot to recover, but it looks it's not only me... maybe we should move somewhere once @effigies comes back

nibabel/streamlines/tests/test_array_sequence.py
```diff
@@ -9,8 +9,6 @@
 from ...testing_pytest import assert_arrays_equal
```
djarecka: this should be removed

djarecka:
two comments: 
- for some reason it was often the case that functions like `assert_array_equal` or `assert_almost_equal` were import from `..testing` (currently `..testing_pytest`) and directly from `nipype.testing`. Let's remove the imports of the numpy function from `testing_pytest` and just keep the direct imports

- do we want to keep `assert_array_equal`? This is relatively easy to convert to standard python function, e.g. `assert (arr1 == arr2).all()`

nibabel/streamlines/tests/test_array_sequence.py
effigies: This is used. Note array**s**.

djarecka: that's right

effigies:
@djarecka 

> * for some reason it was often the case that functions like `assert_array_equal` or `assert_almost_equal` were import from `..testing` (currently `..testing_pytest`) and directly from `nipype.testing`

One advantage to importing things from `nibabel.testing` instead of `numpy.testing` is that, if numpy changes their API, we can just fix it in one place. But if you feel like homogenizing, please open a PR.

> * do we want to keep `assert_array_equal`? This is relatively easy to convert to standard python function, e.g. `assert (arr1 == arr2).all()`

It's not quite. From [the docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_array_equal.html)

> In contrast to the standard usage in numpy, NaNs are compared like numbers, no assertion is raised if both objects have NaNs in the same positions.

I fixed the `optpkg` test.

I also made some style changes to `test_analyze`, since I was going to have to run the CI again anyway. Would you care to have another review?

djarecka:
@effigies - so you want to use `assert_array_equal`?  I don't like comparing NaNs, but if this is important, it's fine with me
I understood that you want to remove as much as we can and I was moving all `assert_array_equal` to `().all`, but can revert.

djarecka:
@effigies - sure, importing from `..testing` have advantages, but I  saw test files that imported the same function from both places, what is confusing

effigies:
Oh, sorry. Yeah, I meant things like numpy decorators that don't do anything numpy-specific. For testing arrays, I'm fine sticking with the `numpy.testing.assert_*`. If they're obviously not going to have `nan`s, then switching to `assert np.all(x == y)` or `assert np.array_equal(a, b)` seem fine.

Also, if they survive the translation, then clearly they were fine, so I wouldn't revert unless tests start failing.

> but I saw test files that imported the same function from both places, what is confusing

Can you show an example? I've seen what I thought were duplicates, but one had a plural "arrays".

djarecka:
> Oh, sorry. Yeah, I meant things like numpy decorators that don't do anything numpy-specific. For testing arrays, I'm fine sticking with the `numpy.testing.assert_*`. If they're obviously not going to have `nan`s, then switching to `assert np.all(x == y)` or `assert np.array_equal(a, b)` seem fine.
> 
> Also, if they survive the translation, then clearly they were fine, so I wouldn't revert unless tests start failing.

Ok, I will not revert the changes, but will also not convert `array_*` anymore.


> > but I saw test files that imported the same function from both places, what is confusing
> 
> Can you show an example? I've seen what I thought were duplicates, but one had a plural "arrays".

Can't find it now, maybe I was wrong that were in the same file, but definitely in some files functions   were imported from `nipype.testing` and `nose.tools` and in some other files, they were imported from `..testing`. Perhaps we can leave it as it is and can come back to this later.

djarecka:
@effigies - this is an example that imports some `numpy.testing` functions directly from the package and some from `..testing`:
https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_trackvis.py#L13

effigies:
Gotcha. In that case, it looks like I imported things that were available in `nibabel.testing` from there, and the other from `numpy` because it wasn't in nibabel.

But if you look at the history, it used to prefer to grab things directly from `nose` or `numpy`, so I'm okay with returning to that approach. It's probably easiest to remember.

effigies:
This one looks like it's going to pass. https://travis-ci.com/nipy/nibabel/builds/147436283

If you're good with it, I'll merge it to remove some jobs from the CI queue.

djarecka:
@effigies - sure, let's merge it. And I will rebase my branch before opening PR

PR #865, opened 2020-01-25
TEST: Drop nose for pytest

effigies:
This PR exists to make all changes related to #864 easy to view, and makes CI status more visible.

Closes #864.
Replaces #840.

PR #873, opened 2020-02-04
TEST: Make things work on Windows pipeline

anibalsolon:
- Dropping nose references on the CI runners
- Keeping nose as a dependency, so we do not get confused about what is broken in the porting


effigies:
Well, that looks like Windows settled. We might lose a little coverage by not running `nose`, but it's probably not worth the trouble of having two types of tests.

Can you restore the Travis ones, though? The goal is to make sure that we are still testing everything while we move to pytest piece-by-piece.

anibalsolon:
Got it! Will revert it.

effigies:
@anibalsolon I went ahead and cherry-picked your commits into 5695644074d5b754782350e844a15ff900db37e5. Feel free to rebase onto `upstream/pytest` at this point.

effigies:
Windows builds are working on #865, so I think we can close this one for now.

setup.cfg
```diff
@@ -59,6 +59,7 @@ nosetests =
     coverage
     nose >=0.11
     pytest
+    pytest-cov
```
effigies: We shouldn't need this for `nosetests`... We need `pytest` for imports.

anibalsolon:
> I think we need to re-add the entry to `azure-pipelines.yml` as in [bfb63fc](https://github.com/nipy/nibabel/commit/bfb63fcb37459e69bd5eb34baded1515af1d2b22).
> 
> ```yaml
>       nosetests:
>         PYTHON_VERSION: '3.6'
>         PYTHON_ARCH: 'x64'
>         CHECK_TYPE: 'nosetests'
> ```

It is not required since the default one is nosetests. But I reintroduced it for symmetry.

Issue #886, opened 2020-02-10
Old Matplotlib Travis test failing


effigies:
The `OLD_WHEELS` repository seems to be either flaking or down: https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com/

This is affecting Travis jobs requesting matplotlib 1.3.1. @matthew-brett Do you know where to get more information on this host?

In any event 1.3.1 is from 2013. The current http://wheels.scipy.org has Python 3.5 wheels for 1.5.2, which is 2016 release, which seems plenty old. If the wheel isn't coming back, then I don't see any reason not to bump to 1.5.2.

In other news, cheers to the matplotlib team for keeping their API so stable that we didn't have to change anything for 6 years.

matthew-brett:
Yes, sorry, this is a Rackspace container, and Rackspace recently cancelled
the free subscription we were using - we're just trying to work out where
to move stuff too.  Yes, that's way-old now - maybe support a few years
back?

effigies:
Okay. Let's update to 1.5. I'll submit a PR for the maintenance branches.

PR #887, opened 2020-02-11
MNT/FIX: Bump minimum matplotlib to 1.5.2 to match available wheels

effigies:
Per https://github.com/nipy/nibabel/issues/886#issuecomment-584536704, ancient wheels are no longer being hosted for free. The only affected minimum dependency we have is matplotlib, which we can bump to 1.5 without likely affecting users.

The oldest wheel for Python 3.6 is 1.5.3, so I'm going with that to avoid doing this again in a couple months.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ matrix:
     # Absolute minimum dependencies plus oldest MPL
     - python: 3.5
       env:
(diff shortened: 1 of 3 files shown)


effigies:
I'm assuming this won't be very controversial, but reviews are welcome.

PR #889, opened 2020-02-14
RF: minc - delay import of h5py until needed.

yarikoptic:
On Debian systems h5py comes with MPI enabled build.
I guess on some misconfigured systems, like my laptop, it might take awhile to import.

And since h5py here is imported upon import of `nibabel`, it makes
it long(er) to import it.  Here is timing on my laptop before the change:

    $> time python -c 'import nibabel'
    python -c 'import nibabel'  0.20s user 0.08s system 5% cpu 5.356 total

and here is after the change:

    $> time python -c 'import nibabel'
    python -c 'import nibabel'  0.13s user 0.02s system 100% cpu 0.150 total

### DIFF
--- a/nibabel/minc2.py
+++ b/nibabel/minc2.py
@@ -27,8 +27,6 @@ and compare against command line output of::
 """
 import numpy as np
 
-from ._h5py_compat import h5py
-
 from .minc1 import Minc1File, MincHeader, Minc1Image, MincError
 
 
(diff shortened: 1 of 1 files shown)


effigies:
Can't see any reason to delay it. I'll also cherry-pick into `maint/3.0.x` so that it will be in the next bug-fix release.

yarikoptic:
THANK YOU @effigies !

yarikoptic:
I am a bit confused by coverage report, ... but oh well -- it is not the first time this happens to me ;)

effigies:
Yeah, it can be a mess.

PR #891, opened 2020-02-18
TEST: Final nose purge

effigies:
We missed a couple modules:

* nibabel.streamlines.tests
* nibabel.nicom.tests.test_dwiparams

This PR also removes the `nosetests` extra and the `nosetests` checks from Travis and Azure.

With the removal of these tests, we can remove the numpy printoptions code in `nibabel.__init__.py`.

Further, the `numpy.testing.NoseTester` is nose-specific, so I rewrote `nibabel.test()` and `nibabel.bench()` to use Pytest. This can be refactored further in the future if it seems worth it. To make pytest play nicely with benchmarks, I added a config file that will be part of the package.

This is best reviewed commit-by-commit.

### DIFF
--- a/.azure-pipelines/windows.yml
+++ b/.azure-pipelines/windows.yml
@@ -40,24 +40,6 @@ jobs:
         mkdir for_testing
         cd for_testing
         cp ../.coveragerc .
-        nosetests --with-doctest --with-coverage --cover-package nibabel nibabel ^
-          -I test_data ^
-          -I test_environment ^
-          -I test_euler ^
-          -I test_giftiio ^
-          -I test_netcdf ^
-          -I test_pkg_info ^
-          -I test_quaternions ^
-          -I test_scaling ^
-          -I test_scripts ^
-          -I test_spaces ^
-          -I test_testing
-      displayName: 'Nose tests'
-      condition: and(succeeded(), eq(variables['CHECK_TYPE'], 'nosetests'))
-    - script: |
-        mkdir for_testing
-        cd for_testing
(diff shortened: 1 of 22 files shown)


anibalsolon:
For minc2.py and volumeutils.py, is there any reason for the decrease of coverage? Maybe some raised exception not catch by pytest.

For the rest, LGTM!

djarecka:
lgtm

but just a couple more things:
- [this](https://github.com/nipy/nibabel/blob/master/tools/bisect_nose.py) should not be needed 
- we are not running test for [nisext](https://github.com/nipy/nibabel/tree/master/nisext) anymore, and the test are using `nose`.  
- is [this](https://github.com/nipy/nibabel/blob/master/nibabel/cmdline/dicomfs.py#L21) still needed?

I was also wondering should we try to remove `clear_and_catch_warnings`, `error_warnings`, etc. from `testing` (we removed many usage from the tests but not all), but this is just a question.

effigies:
@anibalsolon For coverage losses, I think it's mostly that Codecov is a bit spazzy. There's very often a little +/- that happens with a given PR. We can check #865's coverage report before a final merge to see if there's anything concerning.

@djarecka 

1) I decided to keep bisect_nose for now. It can still be used, even if it's not supported. Might be good to write a similar tool for pytest.
2) For nisext, we should probably open an issue for some kind of official deprecation and removal. I'm not sure if anybody uses it.
3) Yes. It will hit for Windows or any system that doesn't have fuse installed.

PR #899, opened 2020-04-06
MNT: Drop deprecated ndarray.tostring for tobytes

effigies:
`tobytes` was added in numpy 1.9 and `tostring` is being deprecated in 1.19.


effigies:
I imagine this should be uncontroversial, but please let me know if you have concerns. Otherwise I'll merge tomorrow.

Issue #531, opened 2017-04-15
`fileslice.is_fancy` fails for slices with numpy scalar integers


pauldmccarthy:
This code:
```python
import numpy as np
import nibabel as nib

shape = (10, 20, 30, 5)
randVoxels = np.vstack([np.random.randint(0, s, 10) for s in shape[:3]]).T

for x, y, z in randVoxels:
    nib.fileslice.canonical_slicers((x, y, z, slice(None, None, None)), shape)
```

Raises this error:
```
ValueError                                Traceback (most recent call last)
<ipython-input-10-1584b7a87e3e> in <module>()
      6
      7 for x, y, z in randVoxels:
----> 8     nib.fileslice.canonical_slicers((x, y, z, slice(None, None, None)), shape)
      9

/Users/paulmc/Projects/nibabel/nibabel/fileslice.pyc in canonical_slicers(sliceobj, shape, check_inds)
     80         sliceobj = (sliceobj,)
     81     if is_fancy(sliceobj):
---> 82         raise ValueError("Cannot handle fancy indexing")
     83     can_slicers = []
     84     n_dim = len(shape)

ValueError: Cannot handle fancy indexing
```

Due to this check in [fileslice.py](https://github.com/nipy/nibabel/blob/78c9b61/nibabel/fileslice.py#L36):
```python
        if hasattr(slicer, 'dtype'):  # ndarray always fancy
            return True
```

At first, I thought this was a simple bug, that could be fixed either by making the check more specific (e.g. `if hasattr(slicer, 'dtype') and not isinstance(slicer, Integral)`, or by removing the check entirely - the code which follows should take care of any non-integral slice object. 

But, tracing back through the history, and looking at the [test cases](https://github.com/nipy/nibabel/blob/78c9b61/nibabel/tests/test_fileslice.py#L35), I'm struggling to understand why this check was deemed necessary in the first place, and am wondering if I've missed some edge case or subtle piece of complexity. Can anybody shed some light on this?

matthew-brett:
The check is because numpy fancy indexing gets triggered by any array input, and it's considerably more complicated to re-implement, so I'm just barfing when I see it.   I keep forgetting how fancy indexing works in complex cases, but my memory is, it does act almost the same in the specific case of numpy array scalars.   So, of course it would be possible to extend the possible indexing to the fancy indexing cases which are simple to implement, it's just I hadn't got around to doing that.

pauldmccarthy:
[...]

PR #903, opened 2020-04-09
ENH: allow numpy integer scalars to index fileslice

MichielCottaar:
This is done by allowing 0-dimensional numpy integer arrays to return true in the `is_fancy` function.

Tests are added to show identical behaviour as normal integers for canonical_slicers and fileslice (the two places where `is_fancy` is used).

The reason for this is that I often find myself trying to index like `dataobj[..., idx]`, where `idx` is an numpy integer scalar, and getting a `fancy indexing is not allowed` error. This can be easily fixed by casting the numpy integer to a python one like `dataobj[..., int(idx)]`. However, as the indexing rules are identical for scalar numpy integers and python built-in integers (except that the former returns a copy, while the latter returns a view), I feel this casting should not be necessary.

This is distinct from #531 as this focusses just on 0-dimensional scalars and still disallows indexing with 1- or higher dimensional arrays.

### DIFF
--- a/nibabel/fileslice.py
+++ b/nibabel/fileslice.py
@@ -48,7 +48,7 @@ def is_fancy(sliceobj):
     if not isinstance(sliceobj, tuple):
         sliceobj = (sliceobj,)
     for slicer in sliceobj:
-        if hasattr(slicer, 'dtype') and slicer.ndim > 0:  # ndarray always fancy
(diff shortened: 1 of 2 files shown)


effigies:
In the absence of objection, I'll plan to merge this tomorrow for inclusion in nibabel 3.1.0.

PR #910, opened 2020-05-15
DOC: Dicom doc. improvement

jond01:
Hi, your package and docs are great!
I found some possible improvements for the dicom documentation, as suggested in this PR:
- (-1 + N) in the affine formulas -> (N - 1) for better readability
- Further explanation about slice spacing (\Delta{s})
- Small typo

### DIFF
--- a/doc/source/dicom/dicom_orientation.rst
+++ b/doc/source/dicom/dicom_orientation.rst
@@ -242,12 +242,8 @@ Where:
 
 For later convenience we also define values useful for 3D volumes:
 
-* $s$ : slice index to the slice plane. The first slice index is zero.
-* $\Delta{s}$ - Spacing in mm between slices, given by the Spacing
-  Between Slices (0018,0088) attribute in units of mm (if exists).
-  Can also be calculated from the Image Orientation (Patient) of
-  multiple slices together with Image Orientation (Patient), as
-  explained in :ref:`dicom-affine-defs`.
+* $s$ : Slice index to the slice plane. The first slice index is zero.
+* $\Delta{s}$ - Spacing in mm between slices.
 
 .. _dicom-3d-affines:
 
@@ -334,6 +330,10 @@ and therefore:
 
 See :download:`derivations/spm_dicom_orient.py` for the derivations and
 some explanations.
+
+For a single slice $N=1$ the affine matrix is $A_{single}$. In this
+case, the slice spacing $\Delta{s}$ may be obtained by the Spacing
+Between Slices (0018,0088) attribute in units of mm, if it exists.
 
 .. _dicom-z-from-slice:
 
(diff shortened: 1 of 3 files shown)


doc/source/dicom/dicom_orientation.rst
```diff
@@ -243,7 +243,11 @@ Where:
 For later convenience we also define values useful for 3D volumes:
 
 * $s$ : slice index to the slice plane. The first slice index is zero.
-* $\Delta{s}$ - spacing in mm between slices. 
+* $\Delta{s}$ - Spacing in mm between slices, given by the Spacing
```
matthew-brett, line 246: Oh dear - I see now that it is more complicated than I thought when I wrote the original - https://stackoverflow.com/questions/14930222/how-to-calculate-space-between-dicom-slices-for-mpr  - what do you think we should do?

jond01: I think that the method with _Image Position (Patient)_ is more general, as some dicoms may miss the _Spacing Between Slices_ attribute.
If both attributes appear, then filling the third column in the affine is possible with either:
1. _Image Position (Patient)_ of multiple sorted slices
2. _Spacing Between Slices_ + the normal __n__ from _Image Orientation (Patient)_.

And they are equivalent (at least are expected to be).

Method 1 is described thoroughly in the doc.
Method 2 may be helpful in some special case: you have just one slice (N=1) but still want the third column of the 4x4 affine matrix. It can be achieved with:
- The cross product of _Image Orientation (Patient)_ = __n__
- _Spacing Between Slices_ = __Δs__

In summary, it worth mentioning but does not contradict anything.

matthew-brett: Yes - I think the question is whether to add the explicit mention of the DICOM Spacing Between Slices attribute here - because it looks like it is a pretty unreliable method of getting the thing we want, which is the distance between the center of one slice and the center of the next: https://groups.google.com/forum/#!msg/comp.protocols.dicom/gyvsgAj4y6o/KOjJywTWFlAJ

jond01: I agree that the bold part here can be moved to some footnote:
>* $\Delta{s}$ - Spacing in mm between slices, **given by the Spacing ...**

However, from my experience when this tag is there - it is correct, i.e. it agrees with the current method.
David A. Clunie is a reliable source, but it was 20 years ago. I think that today most of the manufacturers comply with the definition in the standard, which is exactly what we want:

> Spacing Between Slices | (0018,0088) | 3 | Spacing between adjacent slices, in mm. The spacing is measured from the center-to-center of each slice, and if present shall not be negative.

From:
http://dicom.nema.org/MEDICAL/dicom/2015b/output/chtml/part03/sect_C.7.6.16.2.html#table_C.7.6.16-2

matthew-brett: That's reasonable - maybe with some discussion like here so people know to be a little careful.

matthew-brett:
Thanks for persisting.  Looks good.

PR #911, opened 2020-05-16
MRG: fix Python 2, old Sympy stuff in derivations

matthew-brett:
Python 2 prints, and older uses of Sympy.

Found while checking #910

### DIFF
--- a/doc/source/dicom/derivations/dicom_mosaic.py
+++ b/doc/source/dicom/derivations/dicom_mosaic.py
@@ -1,10 +1,12 @@
 ''' Just showing the mosaic simplification '''
-import sympy
-from sympy import Matrix, Symbol, symbols, zeros, ones, eye
+
+from sympy import Matrix, Symbol, symbols, simplify
+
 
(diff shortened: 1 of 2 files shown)


jond01:
Hi, I suggest also changing `slice_thickness` and its occurrences here:
https://github.com/nipy/nibabel/blob/8471a3701332c9ec08c51e153cc88551355af1b0/doc/source/dicom/derivations/spm_dicom_orient.py#L40

To `slice_spacing`, in order to avoid confusion with the *Slice Thickness* dicom attribute.

matthew-brett:
Good point, thanks.

effigies:
The failing style check should be fixed in master. Mind hitting the PEP8 fixes requested by pep8speaks?

matthew-brett:
I did a rebase to get rid of the PEP8 errors - but now get a coverage error - it must be a false positive.

effigies:
Yeah, Codecov is overly sensitive. I'm okay to merge, if you're all set.

matthew-brett:
All set.

Issue #854, opened 2019-12-18
ExpiredDeprecationError being hit


effigies:
A bunch of `DeprecationWarning`s got promoted to `ExpiredDeprecationError` post-release, so we should update the tests.

Would be good to expect the exception in 3.0.x and remove the deprecated code in `master`.

effigies:
Finished in 3.0.1, need to follow-up in 3.1.0.

effigies:
Removing the deprecated code will make the errors less informative in case somebody missed the 3.0.x series. Let's plan to remove them in 4.0 instead.

Issue #734, opened 2019-03-01
RFC: Release schedule


effigies:
The current release schedule is ad hoc, which mostly follows a pattern that a bug gets fixed or feature gets added, and then several months later people get tired of depending on an unreleased feature and ask for a release. So I'd like to establish a predictable release schedule, while at the same time not artificially inducing churn or overburdening downstream projects with too short of deprecation timelines.

For the level of activity we see, I think quarterly minor releases would be sensible, and a yearly major release would allow us to begin to finally remove components that have been deprecated for years at this point. Major releases will have a minimum one month RC phase.

Bug-fix (micro) releases could continue on an ad hoc basis, on a monthly schedule if there's steady accumulation, shorter in the case of downstream breakage or longer for slow periods.

I'll plan on starting with a minor release (2.4.0) in March or April, and a major release (3.0.0) in September or October.

There's a related discussion to be had about Python 2 support, but perhaps that makes more sense to start another thread?

@nipy/core @nipy/team-nibabel  Let me know if you have any reservations (or unreserved support...). I'll also ping @GaelVaroquaux, @pauldmccarthy and @afni-rickr to make sure nilearn, and the Pythonic sides of FMRIB and AFNI get a chance to weigh in. Please tag in anybody else that you think might want to comment.

@yarikoptic Any concerns from a NeuroDebian perspective?

matthew-brett:
I like the plan - thanks for writing it up.

yarikoptic:
Sounds great to me, thank you!

effigies:
From the last few months, I think shooting for quarterly feature releases is probably faster than the activity level justifies, so I'm going to update the schedule to every four months. If churn picks up, we can revisit again.

The new targets are:

| Release | Month |
|-----|-----|
| 2.5.0 | July 2019
| 3.0.0-rc1 | October 2019 |
| 3.0.0 | November 2019 |
| 3.1.0 | March 2020 |
| 3.2.0 | July 2020 |
| 4.0.0-rc1 | October 2020 |
| 4.0.0 | November 2020 |

effigies:
To update the table with actual releases and project out through the end of 2020:

| Release | Month |
|-----|-----|
| 2.5.0 | July 2019 
| 3.0.0-rc1 | November 2019 |
| 3.0.0 | December 2019 |
| 3.1.0 | April 2020 |
| 3.2.0 | August 2020 |
| 4.0.0-rc1 | November 2020 |
| 4.0.0 | December 2020 |

[...]

PR #915, opened 2020-05-24
RF/MNT: Remove some rust

jond01:
- [x] `flip_axis` in orientations module -> `np.flip` + deprecation
 ==> #916
- [x] Remove v3.0 deprecations (+ related tests)
 ==> Not yet, use special Errors (out of the scope of this PR).
- [x] Refactor py2 prints, imports and modules
 ==> #917
- [x] Docstrings: use triple _double_-quoted string - change `''' ... '''` to `""" ... """`
 ==> #918
- [x] F-strings: change string formatting with `%` / `.format` to f-strings (there's still some work in this point)
 ==> #919

Might be a bit rude because there was no prior discussion... but we can discuss here :-)


effigies:
Hi, thanks a lot for these suggestions!

>     * `flip_axis` in orientations module -> `np.flip` + deprecation

Sounds good to me. Generally we go next major + 1, to be extra conservative about removing things.

>     * Remove v3.0 deprecations (+ related tests)

I'd considered doing this for 3.1 in #854, but the `ExpiredDeprecationError` is more informative than an `ImportError` will be, so I postponed removing 3.0 deprecations until 4.0.

This wasn't highly principled, but I imagined there's a reasonable likelihood somebody might have missed 3.0.0 and 3.0.1 and moved straight to 3.1.0. Also, in the run-up to 4.0 we'll have to effect the deprecations scheduled then, so it might be cleanest to make a big sweep and clear out the dead-as-of-3.0 code then, too.

Anyway, happy to discuss making a coherent policy. It might be that 3.2, which will be ~6 months after 3.0, is far enough along.

>     * Refactor py2 prints, imports and modules
>     * Docstrings: use triple _double_-quoted string - change `''' ... '''` to `""" ... """`
>     * F-strings: change string formatting with `%` / `.format` to f-strings (there's still some work in this point)

:+1:

> Might be a bit rude because there was no prior discussion... but we can discuss here :-)

Not at all! I appreciate the initiative.

One thing that might be worth considering is splitting out the efforts into separate, focused PRs. That way, a trivial change that hits many files, like using double quotes on docstrings, can be reviewed separately from a more consequential one in a single file, such as the `np.flip` one.

jond01:
Thanks for the reply!

I agree that separate PRs are better, let's do this.

Regarding the deprecations - deferring is ok, but I do not get the reason behind a special error for over-deprecated objects. It means that these deprecations will stay forever in the package. Don't we want to remove them at some point?

effigies:
> Regarding the deprecations - deferring is ok, but I do not get the reason behind a special error for over-deprecated objects.

It gives people who ignored warnings an informative error. We usually try to direct them to the replacement.

> Don't we want to remove them at some point?

Yes. I figured a major version (~1 year; see https://github.com/nipy/nibabel/issues/734) would be a reasonable interval to expect even pretty slow-moving ecosystems to run into it.

jond01:
Sounds good.
So do we want to defer v3.0 deprecations to v4.0? (in a separate PR)

effigies:
Do you mean changes like the following?

```Diff
     @property
-    @deprecate_with_version('ArrayProxy.header deprecated', '2.2', '3.0')
+    @deprecate_with_version('ArrayProxy.header deprecated', '2.2', '4.0')
     def header(self):
         return self._header
```

If so, no. That would revert the exception (which breaks things) to a warning (which can be ignored). The idea of not removing immediately is to break things, but with an informative error. In the run-up to 4.0, we can clear out all the code.

If not, could you clarify what you mean?

jond01:
Yes, that what I meant, and no problem.
Unfortunately, I have already deferred one deprecation in #916 in order to pass the CI:
https://github.com/nipy/nibabel/pull/916/commits/164d46c8e9656d206989e6689508cb239862464c

effigies:
With that one, because of the missing comma, the docstring was actually [rendered](https://nipy.org/nibabel/reference/nibabel.orientations.html#nibabel.orientations.orientation_affine):

>     orientation_affine deprecated. Please use inv_ornt_aff instead1.3
> 
>         deprecated from version: 3.0

Instead of raising a warning between 1.3 and then an error at 3.0, it only started raising a warning at 3.0.

Because we did attempt to deprecate it at 1.3, I'm a bit inclined to fast-track it and target 4.0 rather than 5.0 for raising an `ExpiredDeprecationError`.

PR #916, opened 2020-05-25
RF: Use np.flip instead of flip_axis

jond01:
The package requires numpy >= 1.13, np.flip was added in 1.12
Add deprecation (from 3.2 until 5.0) + deprecation test
Remove flip_axis test

### DIFF
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -165,7 +165,7 @@ def apply_orientation(arr, ornt):
     # apply ornt transformations
     for ax, flip in enumerate(ornt[:, 1]):
         if flip == -1:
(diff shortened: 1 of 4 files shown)


nibabel/orientations.py
```diff
@@ -230,20 +230,21 @@ def inv_ornt_aff(ornt, shape):
 
 
 @deprecate_with_version('orientation_affine deprecated. '
-                        'Please use inv_ornt_aff instead'
+                        'Please use inv_ornt_aff instead.',
                         '1.3',
```
effigies, line 234: Seems kind of unfair to change the docs to claim it was deprecated since 1.3, since it only just started saying it at 3.0.

```suggestion
                        '3.0',
```

PR #917, opened 2020-05-25
MNT: Fix configparser + python 2 deprecations

jond01:
Refactor python 2 prints and imports
Also some style

### DIFF
--- a/nibabel/tests/test_round_trip.py
+++ b/nibabel/tests/test_round_trip.py
@@ -138,7 +138,7 @@ def check_arr(test_id, V_in, in_type, out_type, scaling_type):
     rel_err = np.abs(top / arr)
     abs_err = np.abs(top)
     if slope == 1:  # integers output, offset only scaling
-        if set((in_type, out_type)) == set((np.int64, np.uint64)):
+        if {in_type, out_type} == {np.int64, np.uint64}:
             # Scaling to or from 64 bit ints can go outside range of continuous
             # integers for float64 and thus lose precision; take this into
             # account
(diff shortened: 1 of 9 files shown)


nibabel/tests/test_round_trip.py
```diff
@@ -181,14 +181,14 @@ def check_arr(test_id, V_in, in_type, out_type, scaling_type):
...
-               rel_mx_e,
-               slope, inter)
+        print(test_id,
+              np.dtype(in_type).str,
+              np.dtype(out_type).str,
+              exp_abs_mx_e,
+              abs_mx_e,
+              rel_thresh,
+              rel_mx_e,
+              slope, inter)
```
effigies, line 191: Looks like this was supposed to print a tuple but got lucky in 2-3 conversion.

```suggestion
        print((test_id,
               np.dtype(in_type).str,
               np.dtype(out_type).str,
               exp_abs_mx_e,
               abs_mx_e,
               rel_thresh,
               rel_mx_e,
               slope, inter))
```

versioneer.py
```diff
@@ -276,11 +276,7 @@
 
 """
 
-from __future__ import print_function
-try:
-    import configparser
-except ImportError:
-    import ConfigParser as configparser
+import configparser
```
effigies, line 279: While these are valid updates, I'm disinclined to change versioneer except for functional fixes, to keep any diffs focused.

nibabel/tests/test_round_trip.py
jond01: Nice, I thought it's just python 3 with a space after the print (`print ('hi')` works)

versioneer.py
jond01: Ok, what about the changes after the imports? They are deprecated in py3.6

effigies: Fair enough. Let's go ahead with these changes, then.

effigies:
LGTM. Thanks!

PR #918, opened 2020-05-26
STY: Use triple double-quoted string in docstrings (PEP-257)

jond01:
''' -> """
+ Some style in dicom/derivations


nibabel/tests/test_quaternions.py
```diff
@@ -18,7 +18,7 @@
 from .. import quaternions as nq
 from .. import eulerangles as nea
 
-# Example rotations '''
+# Example rotations """
```
effigies, line 21:
```suggestion
# Example rotations
```

PR #922, opened 2020-06-14
NEP29: Set minimum numpy to 1.14

effigies:
This allows us to drop the `numpy.set_printoptions` shim to maintain doctests for old numpy.


effigies:
Should be uncontroversial.

PR #925, opened 2020-06-25
FIX: Detect IndexedGzipFile as compressed file type

pauldmccarthy:
This PR updates `nibabel.volumeutils` so that it recognises `IndexedGzipFile` objects as compressed files, and does not attempt to memory-map them.

Background at pauldmccarthy/indexed_gzip#38.

### DIFF
--- a/nibabel/openers.py
+++ b/nibabel/openers.py
@@ -35,6 +35,9 @@ try:
     del igzip, version
 
 except ImportError:
+    # nibabel.openers.IndexedGzipFile is imported by nibabel.volumeutils
+    # to detect compressed file types, so we give a fallback value here.
+    IndexedGzipFile = gzip.GzipFile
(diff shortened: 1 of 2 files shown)


effigies:
@pauldmccarthy Thanks for this. It looks safe and bug-fixy. Would you mind rebasing on `maint/3.1.x`?

pauldmccarthy:
hey @effigies, no problem; I'm not sure if you meant to rebase against maint/3.1.x, as that would result in a PR containing all of the differences between master and maint/3.1.x. What I did instead was re-create the feature branch off maint/3.1.x, and then cherry pick the commit for this PR onto it. Let me know if that's what you wanted!

effigies:
Yup, this is exactly what I meant. Thanks!

PR #929, opened 2020-07-02
ENH: Update NIfTI header dimensions for CIFTI-2 compliance

mgxd:
Updates the `CiftiImage.update_headers()` method to configure the NIfTI header's `dim` to CIFTI-2 standards brought up in https://github.com/poldracklab/fmriprep/issues/2205

PR #930, opened 2020-07-02
FIX: Set CIFTI-2 dimensions in the last three entries of ``dim``

effigies:
Builds on and supersedes #929.


effigies:
Thanks for the review.

PR #931, opened 2020-07-06
FIX: Use seconds as time units when converting PARRECHeader to Nifti1Header

effigies:
parrec2nii produces NIfTI files with TRs in seconds but records the units as msec

#692 proposed retaining the PAR/REC msec units by removing a division by `1000.`. This PR instead proposes keeping things in seconds.

cc @Roosted7 @epongpipat

Replaces #692.

### DIFF
--- a/nibabel/parrec.py
+++ b/nibabel/parrec.py
@@ -775,7 +775,8 @@ class PARRECHeader(SpatialHeader):
                     self.general_info['exam_date'].replace(' ', ''),
                     self.general_info['protocol_name']))[:80]  # max len
         is_fmri = (self.general_info['max_dynamics'] > 1)
-        t = 'msec' if is_fmri else 'unknown'
+        # PAR/REC uses msec, but in _calc_zooms we convert to sec
+        t = 'sec' if is_fmri else 'unknown'
         xyzt_units = unit_codes['mm'] + unit_codes[t]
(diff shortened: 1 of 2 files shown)


effigies:
Would be glad if anybody wants to weigh in on which of these two approaches is best. I think this one will specifically fix conversions and not break anybody's workflows that have been depending on `PARRECImage` providing TRs in seconds. I can see an argument for keeping things close to the image, but I think standardizing on seconds makes more sense.

effigies:
Anybody care to have a look? If not I'll exercise supreme maintainer powers and hit merge tomorrow.

@Roosted7 @epongpipat I would appreciate your opinions on this one.

matthew-brett:
Hmm - it's a tough call, I agree.   In general, we try to report the file from nibabel as a native reader would have reported it, but here I guess we allowed ourselves to make the file appear more NIfTI-like.    This fix seems better for back-compatibility, but you could also make an argument for converting to seconds only when converting the image to NIfTI format (so, not in `_calc_zooms`, but in `as_analyze...`).  But - I do agree that we shouldn't make odd milliseconds NIfTI images, they would be very confusing.

effigies:
How about going with this for a bug fix, and we can move PARRECImage to report msec TR in a feature release?

Maybe it would be worth revisiting #567. I ran into some conceptual problems trying to think through the API there.

matthew-brett:
Sounds reasonable ...

PR #932, opened 2020-07-06
ENH: Add option to infer CIFTI-2 intent codes

mgxd:
Adds support for `FileBasedImage` methods `to_filename` and `instance_to_filename` to support keyword args.

Adds keyword arg `validate` to CIFTI's `to_filename` and `instance_to_filename` methods.
- If `False` (default), the intent code will not be set (except when the intent code is `"none"`)
- If `True`, the first suffix of the output filename will be compared to valid suffixes from within the CIFTI-2 specification. If there is a match, the corresponding intent code will be set, and each IndexMap within the header matrix is checked.

### DIFF
--- a/nibabel/cifti2/cifti2.py
+++ b/nibabel/cifti2/cifti2.py
@@ -93,46 +93,46 @@ CIFTI_BRAIN_STRUCTURES = ('CIFTI_STRUCTURE_ACCUMBENS_LEFT',
 # "Standard CIFTI Mapping Combinations" within CIFTI-2 spec
 # https://www.nitrc.org/forum/attachment.php?attachid=341&group_id=454&forum_id=1955
 CIFTI_CODES = Recoder((
-    ('dconn', 'NIFTI_INTENT_CONNECTIVITY_DENSE', (
+    ('.dconn.nii', 'NIFTI_INTENT_CONNECTIVITY_DENSE', (
         'CIFTI_INDEX_TYPE_BRAIN_MODELS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
-    ('dtseries', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES', (
+    ('.dtseries.nii', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES', (
         'CIFTI_INDEX_TYPE_SERIES', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
-    ('pconn', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED', (
+    ('.pconn.nii', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED', (
         'CIFTI_INDEX_TYPE_PARCELS', 'CIFTI_INDEX_TYPE_PARCELS',
     )),
-    ('ptseries', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SERIES', (
+    ('.ptseries.nii', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SERIES', (
         'CIFTI_INDEX_TYPE_SERIES', 'CIFTI_INDEX_TYPE_PARCELS',
     )),
-    ('dscalar', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SCALARS', (
+    ('.dscalar.nii', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SCALARS', (
         'CIFTI_INDEX_TYPE_SCALARS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
-    ('dlabel', 'NIFTI_INTENT_CONNECTIVITY_DENSE_LABELS', (
+    ('.dlabel.nii', 'NIFTI_INTENT_CONNECTIVITY_DENSE_LABELS', (
         'CIFTI_INDEX_TYPE_LABELS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
-    ('pscalar', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SCALAR', (
+    ('.pscalar.nii', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SCALAR', (
         'CIFTI_INDEX_TYPE_SCALARS', 'CIFTI_INDEX_TYPE_PARCELS',
     )),
-    ('pdconn', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_DENSE', (
+    ('.pdconn.nii', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_DENSE', (
         'CIFTI_INDEX_TYPE_BRAIN_MODELS', 'CIFTI_INDEX_TYPE_PARCELS',
     )),
-    ('dpconn', 'NIFTI_INTENT_CONNECTIVITY_DENSE_PARCELLATED', (
+    ('.dpconn.nii', 'NIFTI_INTENT_CONNECTIVITY_DENSE_PARCELLATED', (
         'CIFTI_INDEX_TYPE_PARCELS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
-    ('pconnseries', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SERIES', (
+    ('.pconnseries.nii', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SERIES', (
         'CIFTI_INDEX_TYPE_PARCELS', 'CIFTI_INDEX_TYPE_PARCELS', 'CIFTI_INDEX_TYPE_SERIES',
     )),
-    ('pconnscalar', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SCALAR', (
+    ('.pconnscalar.nii', 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SCALAR', (
         'CIFTI_INDEX_TYPE_PARCELS', 'CIFTI_INDEX_TYPE_PARCELS', 'CIFTI_INDEX_TYPE_SCALARS',
     )),
-    ('dfan', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES', (
+    ('.dfan.nii', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES', (
         'CIFTI_INDEX_TYPE_SCALARS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
-    ('dfibersamp', 'NIFTI_INTENT_CONNECTIVITY_UNKNOWN', (
+    ('.dfibersamp.nii', 'NIFTI_INTENT_CONNECTIVITY_UNKNOWN', (
         'CIFTI_INDEX_TYPE_SCALARS', 'CIFTI_INDEX_TYPE_SCALARS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS',
     )),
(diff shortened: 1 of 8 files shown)


nibabel/cifti2/cifti2.py
```diff
@@ -88,6 +88,23 @@ class Cifti2HeaderError(Exception):
...
+    '.ptseries': 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SERIES',
+    '.dscalar': 'NIFTI_INTENT_CONNECTIVITY_DENSE_SCALARS',
+    '.dlabel': 'NIFTI_INTENT_CONNECTIVITY_DENSE_LABELS',
+    '.pscalar': 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SCALAR',
+    '.pdconn': 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_DENSE',
+    '.dpconn': 'NIFTI_INTENT_CONNECTIVITY_DENSE_PARCELLATED',
+    '.pconnseries': 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SERIES',
+    '.pconnscalar': 'NIFTI_INTENT_CONNECTIVITY_PARCELLATED_PARCELLATED_SCALAR',
+    '.dfan': 'NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES',
+}
```
effigies, line 106: Does it make sense to add an `extension` field to `nifti1.intent_codes`? 

https://github.com/nipy/nibabel/blob/0ba55d4966cb3b382583b66f7e64edb17df7ce47/nibabel/nifti1.py#L152-L250

The `NIFTI_INTENT_CONNECTIVITY_*` are added here:

https://github.com/nipy/nibabel/blob/0ba55d4966cb3b382583b66f7e64edb17df7ce47/nibabel/cifti2/parse_cifti2.py#L50-L75

nibabel/cifti2/cifti2.py
```diff
@@ -1490,6 +1503,33 @@ def get_data_dtype(self):
...
+        # if intent code is not set, default to unknown
+        if header.get_intent()[0] == 'none':
+            header.set_intent('NIFTI_INTENT_CONNECTIVITY_UNKNOWN')
+        super().to_filename(filename)
+
+
+def _infer_intent_from_filename(filename):
+    """Parses output filename for common suffixes and fetches corresponding intent code"""
+    from pathlib import Path
+    ext = Path(filename).suffixes[0]
```
effigies, line 1530: This might be more reliable (though require an `except IndexError` case):

```suggestion
    ext = Path(filename).suffixes[-2]
```

nibabel/cifti2/cifti2.py
```diff
@@ -1490,6 +1503,33 @@ def get_data_dtype(self):
     def set_data_dtype(self, dtype):
         self._nifti_header.set_data_dtype(dtype)
 
+    def to_filename(self, filename, infer_intent=False):
```
effigies, line 1506: keyword only

```suggestion
    def to_filename(self, filename, *, infer_intent=False):
```

nibabel/cifti2/cifti2.py
```diff
@@ -1490,6 +1503,33 @@ def get_data_dtype(self):
...
+        """
+        Ensures NIfTI header intent code is set prior to saving.
+
+        Parameters
+        ----------
+        infer_intent : boolean, optional
+            If ``True``, attempt to infer and set intent code based on filename suffix.
+        """
+        header = self._nifti_header
+        if infer_intent:
```
effigies, line 1516: I think it would be reasonable to do something like:

```Python
predicted_intent = _infer_intent(...)
if infer_intent:
    set_intent()
else:
    if get_intent() != predicted_intent:
        warnings.warn(...)
```

Or we could make it `True`/`False`/`None`, where `True` overrides, `None` warns, and `False` suppresses warnings.

effigies (commented):
I had a quick look and overall looks reasonable. I'll find a little time to think through the API more thoroughly later.

nibabel/cifti2/cifti2.py
mgxd: do you mean within the parameters description tuple? or as a new element? I'm not sure how relevant that would be outside of the CIFTI scope tbh

effigies: New field. And it would be None for most. GIFTI could use it for its extensions, as well. Though on second thought, since GIFTI has at least one that is a combination of intents, we might want to create a separate structure.

I'll think more about it.

effigies: I think I agree that this probably shouldn't be shoehorned into the intent code structure.

The extensions do also come with required axis/index types. What do you think of something like the following?

```Python
CIFTI_EXTENSIONS = Recoder((
    ('dconn', 'NIFTI_INTENT_CONNECTIVITY_DENSE', ('CIFTI_INDEX_TYPE_BRAIN_MODELS', 'CIFTI_INDEX_TYPE_BRAIN_MODELS')),
    ('dtseries', 'NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES', ('CIFTI_INDEX_TYPE_SERIES', 'CIFTI_INDEX_TYPE_BRAIN_MODELS')),
    ...
    ), fields=('extension', 'niistring', 'map_types')
)
```

This would set us up to be able to validate or potentially reorder axes, as well.

mgxd:
@effigies Totally forgot about this PR - I liked your suggestion in https://github.com/nipy/nibabel/pull/932#discussion_r468995046 so I went ahead and added that. Since that now includes IndexMap information, I reworked the original `infer_intent` kwarg into `validate`, which checks for the presence of the expected IndexMaps (and also the correct order) on top of setting the intent code.

nibabel/cifti2/tests/test_cifti2.py
```diff
@@ -426,3 +426,7 @@ def make_imaker(self, arr, header=None, ni_header=None):
             )
             header.matrix.append(mim)
         return lambda: self.image_maker(arr.copy(), header, ni_header)
+
+    def validate_filenames(self, imaker, params, validate=False):
+        super().validate_filenames(imaker, params, validate=False)
```
effigies, line 431:
```suggestion
        super().validate_filenames(imaker, params, validate=validate)
```

nibabel/loadsave.py
```diff
@@ -96,7 +96,7 @@ def save(img, filename):
 
     # Save the type as expected
     try:
-        img.to_filename(filename)
+        img.to_filename(filename, **kwargs)
```
effigies, line 99: And on L147.

mgxd:
I believe the failing tests are unrelated - this is ready for review.

effigies:
Yup, opened https://github.com/pydicom/pydicom/issues/1214 for those failures.

effigies:
Overall, I'm not sure I like this UX. If I try to save a `generic_cifti.nii` without `validate=False`, this is going to give me a `KeyError`.

My inclination is that we want warnings, not errors, by default. And inferring intent codes and validating axes feel like they shouldn't be controlled by the same parameter. @MichielCottaar @satra Would be interested in your opinions (or anyone else who uses CIFTI a lot) on how we can be helpful without being burdensome.

I haven't checked this, but I suspect that the logic will fail if someone tries to save a compressed `.nii.gz`, which we've previously allowed, even though CIFTI strongly discourages it.

MichielCottaar:
Thanks for looking into these intent codes and filename extensions. I would suggest to turn around the logic to use the map types as a ground truth of what the user intends rather than the file extension. At least for the current CIFTI file types, it is possible to determine what the intent code should be for any set of map types. I think such an inference could just added in without most users noticing (ideally it would not overwrite any intent code explicitly set by the user).

I do like the idea of also checking that the filename extension matches what is being stored. However, I agree with effigies that it is probably better to keep this as a warning.

effigies:
I think using the axis types as a more authoritative indicator of intent than the extension makes a lot of sense. So something like:

```Python
try:
    expected_intent = CIFTI_CODES.niistring[axis_types]
    expected_ext = CIFTI_CODES.extension[axis_types]
except KeyError:
    expected_intent = "NIFTI_INTENT_CONNECTIVITY_UNKNOWN"
    expected_ext = ".nii"

if validate and expected_intent != found_intent:
    warn
    set intent

if validate and not fname.endswith(expected_ext):
    warn
```

This logic is simplified, since axis type tuples are not unique determinants. (`NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES` applies to `dtseries.nii` and `dfan.nii`, `(CIFTI_INDEX_TYPE_SCALARS, CIFTI_INDEX_TYPE_BRAIN_MODELS)` applies to `dscalar.nii` and `dfan.nii`, and `dfibersamp.nii` and `dfansamp.nii` share both intents and axes.)

This would also treat `.nii.gz` the same as a mismatched `.<subext>.nii`, and warn, but not fail, which feels a bit kinder to users.

With this mode, I think I'm okay with controlling both behaviors with a single parameter. `validate` seems fine to me...

satra:
@effigies - your plan looks reasonable. the one thing perhaps to think a bit more about is exception vs warning. warnings in python are mostly useless, i think, in dataflows and large scale processing, which is where such mismatches are likely to show up. the question is whether a reasonable assumption is being made by the tool when processing when ignoring a warning (since warnings are never trapped by downstream code only filtered out).

effigies:
That's a fair point. What would be a case in which you'd want to see an error? Trying to think through what the default behaviors should be. I think there should be something between "error" and "no helping", so if we do want to promote these checks to exceptions, then I'm back to wanting separate switches for inferring the intent code and performing validation. Or maybe a multi-level `validate in (True, False, "warn")`.

satra:
this is probably going to take this to a different PR/problem space, so first i will say that true, false, warn/except is a good starting way of handling this.

so the simple use cases of matrix dimension flips and others should be choices for exception (e.g., timeseries by vertices vs vertices by timeseries). although i'm not sure as i write this if cifti actually enforces this or hcp-flavored cifti uses this as convention (i think the latter).

more general approach, let validate take a callable, and one could build up an hcp-flavored validator (this can check on extension to internal mismatches, dimension mismatches, even have heuristics for certain types of information depending on intent codes).

most algorithms assume that the file given to it is the correct file instead of checking to see if it is. for example, i load a dtseries and i assume that the dimensions are vertices/voxels by time. but it may be impossible for nibabel in general to know what they intent of the algorithm/tool designer was. so let the tool designer decide what is valid for their application and nibabel, if it can validate basic things like dimensions, availability of appropriate metadata (e.g., a dtseries should have a samplingstep inside the header that is reasonable).

mgxd:
@MichielCottaar @satra thanks for the reviews - I'm fine with reducing the severity of "incorrect" CIFTIs to a warning..

> I would suggest to turn around the logic to use the map types as a ground truth of what the user intends rather than the file extension

I like this!

> warnings in python are mostly useless, i think, in dataflows and large scale processing, which is where such mismatches are likely to show up

This was my first thought when creating this, but looking back, it may be too strict of an enforcement -  I'm fine with reducing the severity of "incorrect" CIFTIs to a warning. I think warning handling is important, but that can (should?) be handled by the specific workflow (i.e. using `warnings.simplefilter` to treat them as errors).

effigies:
@mgxd Are you shooting to get this in for the release?

mgxd:
yes, I was hoping to - but this failing test reveals a problem with the approach of looking up based on map configuration.

> (CIFTI_INDEX_TYPE_SCALARS, CIFTI_INDEX_TYPE_BRAIN_MODELS) applies to dscalar.nii and dfan.nii

when trying to validate the configuration for a `*dscalar.nii` file, the information for `*dfan.nii` is fetched. How can we avoid these collisions?

effigies:
We might need to modify recorders to return multiple options if there are multiple matches.

effigies:
I think we'll need to push this off to another release. I've held up 3.2.0 for too long and almost certainly won't have time to review more this week.

effigies:
This would be good to get in soon, if possible. It will need rebasing due to mass application of blue.

effigies:
@mgxd I'm going to aim for a new feature release next week. I don't know if you'd like to take a stab at bringing this one up to date?

mgxd:
@effigies I'll try, but no promises

effigies:
No worries.

PR #934, opened 2020-07-10
TEST: Purge nose from nisext tests

MeggyCal:
When I simply launch pytest, it fails.

### DIFF
--- a/nisext/tests/test_sexts.py
+++ b/nisext/tests/test_sexts.py
@@ -6,7 +6,7 @@ import types
 
 from ..sexts import package_check
 
-from nose.tools import assert_true, assert_false, assert_equal, assert_raises
+import pytest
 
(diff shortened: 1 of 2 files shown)


nisext/tests/test_sexts.py
```diff
@@ -40,78 +42,63 @@ def test_package_check():
...
+        assert pkg_chk_sta(FAKE_NAME,
                         version='0.2',
-                        version_getter=bad_getter),
-            {'install_requires': exp_spec})
+                        version_getter=bad_getter) == {'install_requires': exp_spec}
         # Likewise for optional dependency
-        assert_equal(
-            pkg_chk_sta(FAKE_NAME,
+        assert pkg_chk_sta(FAKE_NAME,
                         version='0.2',
```
effigies, line 100: Here is an example

PR #938, opened 2020-07-21
ENH: Add bytestring serialization to CIFTI-2

effigies:
CIFTI-2 is serializable, so might as well add it in. This required making `Cifti2Header` comparable, and fixing a small bug where each attempt to save or serialize a CIFTI-2 image adds a `Cifti2Extension` without clearing out old ones.

@mgxd Would you mind reviewing this one?


effigies:
Thanks for the review.

PR #949, opened 2020-08-22
TEST: Suppress expected warnings

effigies:
Tests are currently generating a huge number of warnings, which makes outputs less informative than they could be.

Will add subdirectories soon, but want to get CI going to make sure I'm not breaking things.

### DIFF
--- a/nibabel/freesurfer/tests/test_mghformat.py
+++ b/nibabel/freesurfer/tests/test_mghformat.py
@@ -345,13 +345,13 @@ def test_deprecated_fields():
 
     # mrparams is the only deprecated field at the moment
     # Accessing hdr_data is equivalent to accessing hdr, so double all checks
-    with pytest.deprecated_call():
+    with pytest.deprecated_call(match="from version: 2.3"):
         assert_array_equal(hdr['mrparams'], 0)
     assert_array_equal(hdr_data['mrparams'], 0)
 
-    with pytest.deprecated_call():
+    with pytest.deprecated_call(match="from version: 2.3"):
         hdr['mrparams'] = [1, 2, 3, 4]
-    with pytest.deprecated_call():
+    with pytest.deprecated_call(match="from version: 2.3"):
         assert_array_almost_equal(hdr['mrparams'], [1, 2, 3, 4])
     assert hdr['tr'] == 1
     assert hdr['flip_angle'] == 2
@@ -369,7 +369,7 @@ def test_deprecated_fields():
     hdr['flip_angle'] = 6
     hdr['te'] = 7
     hdr['ti'] = 8
-    with pytest.deprecated_call():
+    with pytest.deprecated_call(match="from version: 2.3"):
         assert_array_almost_equal(hdr['mrparams'], [5, 6, 7, 8])
     assert_array_almost_equal(hdr_data['mrparams'], [5, 6, 7, 8])
 
@@ -377,7 +377,7 @@ def test_deprecated_fields():
     hdr_data['flip_angle'] = 10
     hdr_data['te'] = 11
     hdr_data['ti'] = 12
-    with pytest.deprecated_call():
+    with pytest.deprecated_call(match="from version: 2.3"):
         assert_array_almost_equal(hdr['mrparams'], [9, 10, 11, 12])
     assert_array_almost_equal(hdr_data['mrparams'], [9, 10, 11, 12])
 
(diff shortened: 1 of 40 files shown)


effigies:
@anibalsolon @djarecka @orduek @robbisg Could I trouble one of you for a review?

djarecka:
@effigies - I can take a look, but I was never working on removing the warnings, so don't have experience with it. One question I have after quickly checking pytest website - have you tried `--disable-warnings`? Or it's not what you want?

effigies:
No, I want warnings to show if they're valid, as that's one of the better ways to see upcoming deprecations in Python or numpy. But expected warnings that we produce shouldn't clutter the outputs.

robbisg:
I can help, but as @djarecka, I never worked with suppressing warnings. ;)

djarecka:
@effigies -  ok, so you're concerned about your own warnings and you want to test them and remove from the output at the sam time. got it!

nibabel/freesurfer/tests/test_mghformat.py
```diff
@@ -345,11 +345,14 @@ def test_deprecated_fields():
 
     # mrparams is the only deprecated field at the moment
     # Accessing hdr_data is equivalent to accessing hdr, so double all checks
-    assert_array_equal(hdr['mrparams'], 0)
+    with pytest.deprecated_call():
```
djarecka, line 348: fyi. if you want to check specific warning, you can also use `with pytest.deprecated_call(match="deprecated from version: 2.3")`

nibabel/streamlines/tests/test_streamlines.py
```diff
@@ -193,15 +201,13 @@ def test_save_tractogram_file(self):
...
+        with pytest.warns(ExtensionWarning) as w:
             trk_file = trk.TrkFile(tractogram)
             with self.assertRaises(ValueError):
                 nib.streamlines.save(trk_file, "dummy.tck", header={})
 
-            assert len(w) == 1
-            assert issubclass(w[0].category, ExtensionWarning)
-            assert "extension" in str(w[0].message)
+        assert len(w) == 1
+        assert "extension" in str(w[0].message)
```
djarecka, line 210: you could also use `with pytest.warns(ExtensionWarning, match="extension")`

djarecka:
@effigies -lgtm, i don't remember why we didn't move from `clear_and_catch_warnings`. I thought we had some reasons, but perhaps not very important ones. 
I recently discovered `match` in `pytest.raises` etc., and I really like it, so showed you some examples for future.

effigies:
>  i don't remember why we didn't move from `clear_and_catch_warnings`.

I don't think I realized that pytest was breaking its behavior, and it seemed like an extra task to insert after getting tests working.

Thanks for the suggestions. I think I'll go and look for all the warnings and improve them...

effigies:
Updated. Would appreciate another look through when you get a chance.

nibabel/streamlines/tests/test_tck.py
```diff
@@ -113,18 +113,14 @@ def test_load_file_with_wrong_information(self):
...
+        with pytest.warns(HeaderWarning, match="Missing 'datatype'"):
             tck = TckFile.load(BytesIO(new_tck_file))
-        assert len(w) == 1
-        assert "Missing 'datatype'" in str(w[0].message)
         assert_array_equal(tck.header['datatype'], "Float32LE")
 
         # Simulate a TCK file with no `file` field.
         new_tck_file = tck_file.replace(b"\nfile: . 67", b"")
-        with pytest.warns(HeaderWarning) as w:
+        with pytest.warns(HeaderWarning, matches="Missing 'file'") as w:
```
djarecka, line 119:
```suggestion
        with pytest.warns(HeaderWarning, match="Missing 'file'") as w:
```

effigies: Oops! Good catch.

effigies:
All green. Good to merge?

djarecka:
@effigies - I've just opened diff quickly, and wonder if there is any specific reason why you use `unittest` for skipping tests (e.g.[`unittest.skipIf`](https://github.com/nipy/nibabel/pull/949/files#diff-d945cd306bbbfd6603636e375034c729R24)) instead of `pytest.mark.skipif` ?

effigies:
Yes. When there's a standard library equivalent, I'm preferring that, on the grounds that one day we may need to migrate from pytest as we did from nose. The more things that are vanilla Python, the less work that should be.

djarecka:
ok, i'm still hoping that we don't have to move from pytest anytime soon :-)
lgtm

effigies:
Nose lasted 10 years, so if pytest does as well, we still have plenty of time.

effigies:
Thanks for reviewing!

PR #952, opened 2020-09-09
NF: volume computation function for nifti masks

JulianKlug:
Adds a function to compute the volume of a mask image in mm3.
Simple vanilla python equivalent to "fslstats /path/file.nii -V".

Reason: Useful for pure python projects where one can not rely on fsl.  

To Do:
- [ ] Add CLI


MonsieurWave:
A few questions to help me guide this PR:

1. Would this be the right location for this function? As I am still new to the source code of nibabel, I was unsure on where to add it.

2. Where should the CLI for this function be?

0rC0:
I think you've forgot a `return`, something like:

` return mask_volume_vx * voxel_volume_mm3`

or

`return mask_volume_mm3`

effigies:
If we want to implement something equivalent to `fslstats`, then I think it might be worth creating `nibabel.imagestats` (I think `nibabel.stats` might imply more analytical statistics, which are covered in nilearn) as a place to house the various functions.

The command line should be in `nibabel.cmdline.<tool>`, and expose the entrypoint `nib-<tool>`.

MonsieurWave:
> I think you've forgot a `return` :P, something like:
@0rC0 actually there is a return on line 248, it is just hidden by the codecov messages.
```python
    return mask_volume_mm3
```

effigies:
I think gradually is fine. Let's call it `nib-stats` and start with volume calculation.

Also, don't feel compelled to mimic the `fslstats` flags or output conventions exactly. Clarity is a better goal.

One thing to consider: `fslstats` produces both voxel and mm^3 results. Do we want to follow that lead, or permit users to say `--units {vox,mm3}`?

JulianKlug:
There seems to be an issue where the example part of the docstring is interpreted as code and not as a string which then lets the tests fail.
Not quite sure how to resolve this...

JulianKlug:
would be ready for #960 if the aforementioned issue is resolved

nibabel/tests/test_imagestats.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_mask_volume():
+    # Test mask volume computation
+    infile = test_data(fname="anatomical.nii")
+    img = load(infile)
+    vol_mm3 = imagestats.mask_volume(img)
+    vol_vox = imagestats.mask_volume(img, units='vox')
+
+    assert float(vol_mm3) == 2273328656.0
+    assert float(vol_vox) == 284166082.0
```
effigies, line 28:
```
$ fslstats nibabel/tests/data/anatomical.nii -V
33825 270600.000000 
```

I'm guessing (but haven't checked) the discrepancy is from values > 1.

nibabel/cmdline/stats.py
```diff
@@ -0,0 +1,39 @@
...
+
+def _get_parser():
+    """Return command-line argument parser."""
+    p = argparse.ArgumentParser(description=__doc__)
+    p.add_argument("infile",
+                   help="Neuroimaging volume to compute statistics on.")
+    p.add_argument("-V", "--Volume", action="store_true", required=False,
+                   help="Compute mask volume of a given mask image.")
+    p.add_argument("--units", default="mm3", required=False,
+                   help="Preferred output units of {mm3, vox}. Defaults to mm3")
```
effigies, line 27: I believe `argparse` lets you set choices and will add them to the help text.

```suggestion
                   choices=("mm3", "vox"), help="Preferred output units")
```

nibabel/cmdline/stats.py
```diff
@@ -0,0 +1,39 @@
...
+def main(args=None):
+    """Main program function."""
+    parser = _get_parser()
+    opts = parser.parse_args(args)
+    from_img = load(opts.infile)
+
+    if opts.Volume:
+        computed_volume = mask_volume(from_img, opts.units)
+        print(computed_volume)
+        return computed_volume
```
effigies, line 39: I'm not really sure what happens when you exit with a float error code, but I don't think we want to do that.

```suggestion
        return 0
```

bin/nib-stats.py
```diff
@@ -0,0 +1,18 @@
...
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""
+Compute image statistics
+"""
+
+from nibabel.cmdline.stats import main
+
+
+if __name__ == '__main__':
+    main()
```
effigies, line 18: These files are actually out-of-date. Just add an entry to the entry points:

https://github.com/nipy/nibabel/blob/917afabdbad80dfaf226bd1899745d4f5110cc0c/setup.cfg#L71-L81

nibabel/imagestats.py
```diff
@@ -0,0 +1,58 @@
...
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""
+Functions for computing image statistics
+"""
+
+import numpy as np
+
+
+def mask_volume(img, units='mm3'):
```
effigies, line 16: Thinking about this a bit, if `units=="vox"`, then the output should be integer; there's no good reason to have a float for counting non-zero voxels. That makes the types a bit messy. What if we did (simplified):

```Python
def count_nonzero_voxels(img):
    return np.count_nonzero(img.dataobj)

def mask_volume(img):
    nz_vox = count_nonzero_voxels(img)
    return np.prod(img.header.get_zooms()[:3]) * nz_vox
```

The conditional logic inside this function basically evaporates, and is moved into `cmdline.stats.main` to select which function to call.

effigies (commented):
Thanks for pushing on this. I've left a batch of comments. Feel free to argue and ask questions. :-)

bin/nib-stats.py
JulianKlug: Great hadn't seen that!

nibabel/cmdline/stats.py
JulianKlug: Thanks for the hint!

nibabel/cmdline/stats.py
JulianKlug: Agreed.

nibabel/tests/test_imagestats.py
JulianKlug: I thought so too, but hadn't time to check. 
The volume computation is only reliable for 0-1 masks.

nibabel/imagestats.py
```diff
@@ -0,0 +1,58 @@
...
+    mask_volume_mm3: float
+        Volume of mask expressed in mm3.
+
+    Examples
+    --------
+    >>> import nibabel as nf
+    >>> path = 'path/to/nifti/mask.nii'
+    >>> img = nf.load(path) # path is contains a path to an example nifti mask
+    >>> mask_volume(img)
+    50.3021
```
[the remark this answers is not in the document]
JulianKlug: 👍

nibabel/imagestats.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    header = img.header
+    _, vx, vy, vz, _, _, _, _ = header['pixdim']
+    voxel_volume_mm3 = vx * vy * vz
```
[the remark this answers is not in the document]
JulianKlug: Should we check the case where `spatial_axes_first(img)` is not defined but `header['pixdim']` is?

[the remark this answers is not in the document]
effigies: No, all analyze-like images are spatial axes first. I think the only currently supported exceptions are 4D MINC files.

nibabel/imagestats.py
JulianKlug: Sounds good to me. I have simply left the code a bit more verbose for better readability.

nibabel/imagestats.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+Functions for computing image statistics
+"""
+
+import numpy as np
+from nibabel.imageclasses import spatial_axes_first
+
+def count_nonzero_voxels(img):
+    """
+    Count number of non-zero voxels
```
effigies, line 18:
```suggestion
    Count number of non-zero voxels

```

nibabel/imagestats.py
```diff
@@ -0,0 +1,62 @@
...
+    Parameters
+    ----------
+    img : ``SpatialImage``
+        All voxels of the mask should be of value 1, background should have value 0.
+
+    Returns
+    -------
+    non zero voxel volume: int
```
effigies, line 26: Simple variable names are fine.

```suggestion
    count : int
```

nibabel/imagestats.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Parameters
+    ----------
+    img : ``SpatialImage``
+        All voxels of the mask should be of value 1, background should have value 0.
+
+
+    Returns
+    -------
+    mask_volume_mm3: float
```
effigies, line 44:
```suggestion
    volume : float
```

nibabel/imagestats.py
```diff
@@ -0,0 +1,62 @@
...
+    Returns
+    -------
+    non zero voxel volume: int
+        Number of non-zero voxels
+
+    """
+    return np.count_nonzero(img.dataobj)
+
+def mask_volume(img):
+    """ Compute volume of mask image.
```
effigies, line 33:
```suggestion
    """ Compute volume of mask image.

```

nibabel/tests/test_imagestats.py
```diff
@@ -0,0 +1,31 @@
...
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+""" Tests for image statistics """
+
+import numpy as np
+
+from .. import imagestats
+from .. import Nifti1Image
+
+import pytest
+
```
effigies, line 17:
```suggestion
```

effigies (commented):
Mostly style suggestions. And please use `flake8` to check for any additional issues.

nibabel/cmdline/tests/test_stats.py
```diff
@@ -0,0 +1,47 @@
...
+
+class Capturing(list):
+    def __enter__(self):
+        self._stdout = sys.stdout
+        sys.stdout = self._stringio = StringIO()
+        return self
+    def __exit__(self, *args):
+        self.extend(self._stringio.getvalue().splitlines())
+        del self._stringio    # free up some memory
+        sys.stdout = self._stdout
```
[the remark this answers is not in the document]
JulianKlug: thanks, didn't know about that one

JulianKlug:
Sorry for the style issues. Should be fixed now.

effigies:
Thanks a lot! This is a good start to a handy utility, and I look forward to seeing it grow.

PR #963, opened 2020-10-25
CI: Test on Python 3.9 with numpy nightly wheels

effigies:
Still waiting on Python 3.9 on Travis and 3.9 wheels for most dependencies, but 3.9-dev is around and there are some nightly wheels on https://anaconda.org/scipy-wheels-nightly/. Might as well find out how we do with them.

### DIFF
--- a/nibabel/nicom/ascconv.py
+++ b/nibabel/nicom/ascconv.py
@@ -89,7 +89,10 @@ def assign2atoms(assign_ast, default_class=int):
             target = target.value
             prev_target_type = OrderedDict
         elif isinstance(target, ast.Subscript):
-            index = target.slice.value.n
+            if isinstance(target.slice, ast.Constant):  # PY39
+                index = target.slice.n
+            else:  # PY38
+                index = target.slice.value.n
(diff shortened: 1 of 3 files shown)


nibabel/nicom/ascconv.py
```diff
@@ -89,7 +89,10 @@ def assign2atoms(assign_ast, default_class=int):
             target = target.value
             prev_target_type = OrderedDict
         elif isinstance(target, ast.Subscript):
-            index = target.slice.value.n
+            if isinstance(target.slice, ast.Constant):  # PY39
+                index = target.slice.n
+            else:  # PY38
+                index = target.slice.value.n
```
effigies, line 95: cc @moloney @matthew-brett Apparently there's a change in `ast` behavior in Python 3.9. IDK if we need to be checking other sections for more subtle bugs?

matthew-brett: Glad you tracked that one down.  Yes we should probably comb ...  but it will be a hard job.

PR #976, opened 2020-12-06
MNT: Test fallback using versioneer config

effigies:
This uses https://github.com/python-versioneer/python-versioneer/pull/234 and strips out all custom hacks.

The thing to check will be what happens with the build from archive.


effigies:
It worked (see [logs](https://github.com/nipy/nibabel/pull/976/checks?check_run_id=1507146137#step:8:31)). Can revisit once versioneer has a fallback mechanism in master.

PR #980, opened 2020-12-16
MNT: Remove modules, objects and attributes deprecated in 3.0

effigies:
Left these in post-3.0 to allow `ExpiredDeprecationError`s to give useful error messages. Now time to remove.


effigies:
@matthew-brett As the most conscientious defender of the API, any concerns?

effigies:
@jond01 Any interest in reviewing? I recall you had made a pass at removing these a while ago and I argued you out of it. :-)

jond01:
Thanks, I would like to. Hopefully by the end of next week, but don't wait for me if there are other plans

jond01 (approved):
Hi, I reviewed the PR and approve it, still have some related comments:
1. We remove `trackvis` module now, which is fine, but I suppose that we should also modify its item in `MODULE_SCHEDULE` (in `test_removalschedule.py`) from
   ```
    ("4.0.0", ["nibabel.trackvis"]),
   ```
   to
   ```
    ("3.0.0", ["nibabel.trackvis"]),
   ```
2. I found 4 problematic instances of `@deprecate_with_version` which lack a comma between the message and the intended 'since'. I will change it in another small PR (unrelated to 3.0), it is similar to what we had [here](https://github.com/nipy/nibabel/pull/916#discussion_r430041330)
3. Do we want to remove these docstrings about `keep_file_open='auto'` deprecation:
   ```
        .. deprecated:: 2.4.1
            ``keep_file_open='auto'`` is redundant with `False` and has
            been deprecated. It raises an error as of nibabel 3.0.
   ```
   ? It appears 9 times and anyway passing `'auto'` results in a detailed error message, so maybe it's about time.

effigies:
Yeah, I'm fine dropping the `keep_file_open` deprecation notice from the docs.

PR #987, opened 2021-01-02
BF: Fix concat_images dtype issue

### LINKED ISSUE #986: Concat_images result not consistent between load/save cycles
**Describe the bug**
After using `concat_images` to concatenate two NIfTI images, then saving the new image to disk and loading it back into memory, the values of the data arrays are no longer equal. 

In my case, the source and destination NIfTI file both have `uint16` data, so there should be no issue in floating point loss, and any conversions to `float32` or greater along the way should not result in a loss of precision.

**To Reproduce**
1) Use `concat_images` on two NIfTI images (one of which should have a integer data type, e.g., `uint16`)
2) Save the result
3) Load the result
4) Comparing the images' data arrays with `np.array_equal`

See the test written into the pull request, removing the patch in `funcs.py` to trigger the failure. I'd expect this issue to generally occur with NIfTI's using integers as their data format.

Alternatively, use the script in [this gist](https://gist.github.com/spolcyn/0b7f3f2309dbf31d3f42dacc526203dc) with your own NIfTI file.

**Expected behavior**
The post-concat in-memory representation of the data should be equal to the data after it's been saved/re-loaded from disk.

**Actual behavior**
The data was no longer equal. 

**Likely Reason**
It is almost certain that this is related to the datatype of the `dataobj` after `concat_images`. My analysis of the reason is as follows:

1. Images with `uint16` `dataobj`'s enter the function
2. `Out_data` is created with the correct shape, but with default `dtype=np.float64`
3. `Out_data` gets the data, upcasting it to a `float64`
4. `Out_data` is used to form the new image
5. The new image, when written to disk, finds a conflict between the datatype it should be on disk (`uint16`) and the current `dataobj` type (`float64`).
6. To get around this problem, a scaling slope is applied -- however, with the 32 bit float in the header, the scaling is not perfect (at least with values in my test images).
a. **Note:** This can be verified by examining the NIfTI header binary directly, and for my test I found an `scl_slope` of `0x3d4e00ce` = `0.0502937361598`, and each data element was off by exactly `19 + 5/6`. Doing the math, `1/(19 + 5/6) = 0.050420168067227`, which is close but not quite exactly the `scl_slope` value. (This may suggest choosing the `scl_slope` value in the `save()` routine could be tweaked, but without really digging into that code myself, I'm not sure -- it would appear the most accurate approximation of that number for a `float32` is `0.0504201687872409820556640625`, or `0x3d4e8561` in hex). 
7. Accordingly, when the image is re-read in, the `float64` result is slightly off (slightly less than the true value for my test images), and the `uint16` result is also incorrect (at least for my test images, as the values typically work out to be slightly less than the true value, which results in a floor to a full integer beneath the true value)

**Potential Fix**
Including dtype in `out_data = np.empty(out_shape, dtype=img0.header.get_data_dtype())` in `funcs.py` fixes the problem (alternatively, `.dataobj.dtype` could be used, but using the header seems more maintainable). 

I've put in a PR with a test modification and my proposed fix, happy to discuss it either there or here.

spolcyn:
Fixes #986

### DIFF
--- a/nibabel/funcs.py
+++ b/nibabel/funcs.py
@@ -122,7 +122,7 @@ def concat_images(images, check_affines=True, axis=None):
     if axis is None:
         # collect images in output array for efficiency
         out_shape = (n_imgs, ) + shape0
-        out_data = np.empty(out_shape)
+        out_data = np.empty(out_shape, dtype=img0.header.get_data_dtype())
     else:
         # collect images in list for use with np.concatenate
         out_data = [None] * n_imgs
--- a/nibabel/tests/test_funcs.py
+++ b/nibabel/tests/test_funcs.py
@@ -13,7 +13,7 @@ import numpy as np
 from ..funcs import concat_images, as_closest_canonical, OrientationError
 from ..analyze import AnalyzeImage
 from ..nifti1 import Nifti1Image
-from ..loadsave import save
+from ..loadsave import save, load
 
 from ..tmpdirs import InTemporaryDirectory
 
@@ -104,6 +104,14 @@ def test_concat():
                             try:
                                 all_imgs = concat_images([img0, img1],
                                                          **concat_imgs_kwargs)
+
+                                all_imgs_fname = "all_imgs.nii"
+                                save(all_imgs, all_imgs_fname)
+                                from_disk = load(all_imgs_fname)
+
+                                assert_array_equal(all_imgs.get_fdata(),
+                                                   from_disk.get_fdata())
+
                             except ValueError as ve:
                                 assert expect_error, str(ve)
                             else:


effigies:
The problem here is that `get_data_dtype()` could be `int16` while the correct interpreted dtype could be float, if the scale factors are non-trivial.

This is just to say that I have seen this and will try to look through it soon, but I haven't had time to think about it carefully.

effigies:
Apologies for the extremely slow response... 2021 kept getting away from me in a lot of ways, but hopefully I can be a little more organized this year.

Anyway, I've looked at this again, and the fix is going to be quite easy due to some other in-progress work (#1082). If you cherry-pick d4a6fedd1d1d8c934eb1795e042c17a42f023f9, you can use `get_obj_dtype(img.dataobj)` to determine the logical dtype of each image. I would probably then use [numpy.result_type](https://numpy.org/doc/stable/reference/generated/numpy.result_type.html) on the logical dtype of all images. So something like:

```
out_dtype = np.result_type(*(get_obj_dtype(img.dataobj) for img in images))
out_data = np.empty(out_shape, dtype=out_dtype)
```

The reason I'm not going purely on the single input is that you could concat a `uint8` image and a `float32` image and the result would depend on which one happened to be first.

spolcyn:
Thanks @effigies. Sounds like it'd be best to wait for that commit and #1082 to be merged before applying these suggestions. Does that sync with your thoughts?

PR #988, opened 2021-01-05
NEP29+1y: Set minimum numpy to 1.15

effigies:
According to #803, it will be time to drop numpy 1.14 support on January 7.


effigies:
January 7 has come and gone.

PR #996, opened 2021-02-04
DOC: fix grammar of headings in CoC

matthew-brett:
The document says:

```

...
```

but later the headings become imperatives:

```
```

Imperatives changed to adjectives.

[skip ci]


effigies (approved):
Guess we should also update https://nipy.org/conduct.html.

arokem:
+1 to this change and to changes upstream. In addition to the grammatical mismatch, upon re-reading this text, the imperative is possibly a tad too assertive. Anyway, +1 all around from me.

PR #999, opened 2021-02-25
ENH: Add support for GIFTI ExternalFileBinary

pauldmccarthy:
This PR adds support for loading GIFTI data arrays from external files, using the `ExternalFileName` attribute of the `<DataArray>` element.

I had a query from a user who is working with a [SPM toolbox](http://www.neuro.uni-jena.de/cat/) which generates GIFTI files where the data arrays are stored in external `.dat` files, and the `<DataArray>` element has  `encoding="ExternalFileBinary"`. This led me to discover that `nibabel` does not support GIFTI files which use this feature, presumably because the GIFTI spec is slightly under-specified regarding the expected format of the external file - from section 7.0 of the [GIFTI spec](https://www.nitrc.org/projects/gifti/):

 > For external data storage, the Encoding attribute of the DataArray element will have a value of “ExternalFileBinary” indicating that the data is located in an external file and in binary format.

However, the CAT Toolbox simply stores its data as a raw uncompressed binary file, and I think that this is a sensible and reasonable interpretation of the above definition. If this assumption is made, then all of the necessary information (file offset, data type, row-/colum-major ordering, shape) will be available via the other attributes of the `<DataArray>` element, so the required changes to `nibabel` are quite trivial.

Happy to add in some unit tests and error checking if you think this is a useful contribution.

### DIFF
--- a/nibabel/gifti/parse_gifti_fast.py
+++ b/nibabel/gifti/parse_gifti_fast.py
@@ -30,22 +30,34 @@ class GiftiParseError(ExpatError):
     """ Gifti-specific parsing error """
 
 
-def read_data_block(encoding, endian, ordering, datatype, shape, data):
+def read_data_block(encoding,
+                    endian,
+                    ordering,
+                    datatype,
+                    shape,
+                    data,
+                    darray):
     """ Tries to unzip, decode, parse the funny string data """
     enclabel = gifti_encoding_codes.label[encoding]
     dtype = data_type_codes.type[datatype]
+
     if enclabel == 'ASCII':
         # GIFTI_ENCODING_ASCII
         c = StringIO(data)
         da = np.loadtxt(c, dtype=dtype)
         return da  # independent of the endianness
-
-    elif enclabel == 'External':
-        # GIFTI_ENCODING_EXTBIN
-        raise NotImplementedError("In what format are the external files?")
-
-    elif enclabel not in ('B64BIN', 'B64GZ'):
+    elif enclabel not in ('B64BIN', 'B64GZ', 'External'):
         return 0
+
+    # GIFTI_ENCODING_EXTBIN
+    # We assume that the external data file is raw uncompressed binary, with
+    # the data type/endianness/ordering specified by the other DataArray
+    # attributes
+    if enclabel == 'External':
+        with open(darray.ext_fname, 'rb') as f:
+            f.seek(darray.ext_offset)
+            nbytes = np.prod(shape) * dtype().itemsize
+            buff = f.read(nbytes)
 
     # Numpy arrays created from bytes objects are read-only.
     # Neither b64decode nor decompress will return bytearrays, and there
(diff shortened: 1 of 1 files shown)


effigies:
Yes, sounds good to me. Especially if there are files in the wild, we should support them.

I notice that the file is being opened relative to the cwd, not the gifti, which seems the more likely interpretation.

pauldmccarthy:
Ahh, I hadn't thought of that - the CAT toolbox file which I have specifies the external file as relative to the GIFTI:

```
   <DataArray  ArrayIndexingOrder="ColumnMajorOrder"
               DataType="NIFTI_TYPE_FLOAT32"
               Dim0="64984"
               Dimensionality="1"
               Encoding="ExternalFileBinary"
               Endian="LittleEndian"
               ExternalFileName="s12.mesh.thickness.resampled_32k.TESTCAT.dat"
               ExternalFileOffset="0"
               Intent="NIFTI_INTENT_NONE">
      <MetaData>
      </MetaData>
      <Data></Data>
   </DataArray>
```

effigies:
That makes sense. Just need to update the code to reflect that. Right now it's an open can with no path adjustment.

pauldmccarthy:
No problems - I'll follow up with that, and some tests as well - might find some more time this evening..

pauldmccarthy:
Will add a bit more testing soon ..

nibabel/gifti/parse_gifti_fast.py
```diff
@@ -30,45 +31,76 @@ class GiftiParseError(ExpatError):
...
+        if fname is None:
+            raise GiftiParseError('ExternalFileBinary is not supported '
+                                  'when loading from in-memory XML')
+        ext_fname = op.join(op.dirname(fname), darray.ext_fname)
+        if not op.exists(ext_fname):
+            raise GiftiParseError('Cannot locate external file ' + ext_fname)
+        with open(ext_fname, 'rb') as f:
+            f.seek(darray.ext_offset)
+            nbytes = np.prod(darray.dims) * dtype().itemsize
+            buff = f.read(nbytes)
```
effigies, line 78: This would certainly work, but if somebody's gone through the trouble of giving us a binary file for more efficient access, should we consider return a `np.memmap` or even an `ArrayProxy` so that the data is only loaded on-demand?

pauldmccarthy: Yes, good point - I had thought of that, but then just went with what is done for other `<DataArray>` types. I suppose this should be user-selectable as well - I'll add a `mmap=True` option to `GiftiImage.from_filename` and related ...

pauldmccarthy:
Ok, I've added mem-map support, and added a few extra test cases - ready for a review!

effigies (approved):
This LGTM, thanks Paul. I'll give it a couple days in case anybody else wants to weigh in.

Issue #943, opened 2020-07-31
Very slow and memory hungry tractogram loader (TRK)


emanuele:
At least for tractograms in TRK (Trackvis) format, in the case of a large number of streamlines (>1 million), loading streamlines with dipy.io.streamline.load_tractogram() is extremely slow and requires an unusual amount of memory. Example: for a tractogram of 10 million streamlines, loading takes:
- 15 minutes and approximately 10Gb of RAM with Nibabel
- 1 minute and approximately 5Gb of RAM with our custom, numpy-based, TRK loader (https://github.com/emanuele/load_trk.git)
- 6 seconds and 3.5Gb of RAM if we previously save the tractogram as a numpy array (numpy.save()) and then load it with numpy.load(). But in this case, the streamlines must be resampled to a fixed number of points.

grlee77:
> if we previously save the tractogram as a numpy array (numpy.save()) and then load it with numpy.load(). But in this case, the streamlines must be resampled to a fixed number of points.

The following blog post has a suggestion for how to store ragged arrays (aka jagged arrays), that should fit the case where not all streamlines have the same number of points. It basically would involve concatenating all streamlines along the dimension that varies in length and recording the offsets to each streamline in a separate index list. These two arrays (concatenated and offsets) can then be stored in a single `.npz` file via `np.savez`. There is a concrete example in the following blog post:
https://tonysyu.github.io/ragged-arrays.html#.XyRsJBF7kr4

If the overall memory use becomes a concern, it would be possible to use [xarray + Dask](https://examples.dask.org/xarray.html) to allow indexing into the concatenated array without having the load the whole thing in memory at once.

effigies:
We currently have the `ArrayProxy` construct, which just records the offset, dtype and shape, and will pull data when requested. By a similar principle we could have an `ArrayProxySequence` which just pins multiple `ArrayProxy`s to different offsets in the file, or an `ArraySequenceProxy`, which indexes the entire file and acts like an `ArraySequence`.

soichih:
I am also seeing a very slow streamline loading performance for tck as well. It looks like this part of the code is severely CPU bounded 
> https://github.com/nipy/nibabel/blob/master/nibabel/streamlines/array_sequence.py#L32

emanuele:
[...]

PR #1000, opened 2021-02-25
ENH: Speed up tractogram.apply_affine

MarcCote:
This PR adds the `inplace` option to `affines.apply_affine(...)` which `Tractogram.apply_affine` can leverage for some speed and memory gains.

The speedup is actually two-fold:
1. Allowing `affines.apply_affine` to perform the transformation in-place reduces memory peak usage (esp. when dealing with 10 million streamlines).
2. If the `ArraySequence` representing the tractogram's streamlines is not a sliced view, the affine transformation can be applied directly on the underlying numpy array that contains the data. Doing that avoids having to transform each streamline one by one as it is currently done.

Also, this PR partially addresses issue #943 (see my comment over there for more details).


nibabel/affines.py
```diff
@@ -80,8 +83,18 @@ def apply_affine(aff, pts):
...
+
+    if inplace:
+        try:
+            np.dot(pts, rzs.T, out=pts)
+            pts += trans[None, :]
+        except ValueError:
+            # Fallback on the not inplace version.
+            pts = np.dot(pts, rzs.T) + trans[None, :]
+    else:
+        pts = np.dot(pts, rzs.T) + trans[None, :]
```
MarcCote, line 95: If anyone has a cleaner code for this, I'm a taker.

arokem: Under what circumstances do you end up with the `ValueError` (that wouldn't raise for the `out=None` case)?

MarcCote: According to Numpy's documentation:
> out ndarray, optional
Output argument. This must have the exact kind that would be returned if it was not used. In particular, it must have the right type, must be C-contiguous, and its dtype must be the dtype that would be returned for dot(a,b). This is a performance feature. Therefore, if these conditions are not met, an exception is raised, instead of attempting to be flexible.

https://numpy.org/doc/stable/reference/generated/numpy.dot.html

MarcCote: It does happen in the tests. That's why I added it. If we don't want it in the `affines.apply_affine` I can check it in the Tractogram class.

effigies: A couple things:

1) If `np.dot(pts, rzs.T, out=pts)` succeeds but `pts += trans[None, :]` fails, then the result will be `pts = pts @ rzs.T @ rzs.T + trans[None, :]`. This seems unlikely but we should be a bit defensive.
2) I would probably not duplicate `pts = np.dot(pts, rzs.T) + trans[None, :]` and instead encode the fallback explicitly in the branching structure. This is debatable, aesthetically, so feel free to reject.

Combining the two (and using `@` instead of `np.dot`):

```suggestion
    if inplace:
        try:
            np.dot(pts, rzs.T, out=pts)
        except ValueError:
            inplace = False
        else:
            pts += trans[None, :]
    if not inplace:
        pts = pts @ rzs.T + trans[None, :]
```

MarcCote: I like it. Thanks.

effigies (approved):
Sorry for the extremely slow review. This LGTM.

MarcCote:
@effigies no worries. Happy it's in before 4.0.0 :).

PR #1013, opened 2021-05-05
ENH: Provide clear error message when files with zip extensions don't match file contents

### LINKED ISSUE #966: More helpful error messages when uncompressed file is passed with compressed extension
We have a pretty one-size-fits-all error when we can't determine the file type:

```Shell
$ head -c 1000 /dev/zero > test.nii.gz
$ python -c "import nibabel as nb; nb.load('test.nii.gz')"
  File ".../nibabel/nibabel/loadsave.py", line 55, in load
    raise ImageFileError(f'Cannot work out file type of "{filename}"')
nibabel.filebasedimages.ImageFileError: Cannot work out file type of "test.nii.gz"
```

A relatively common situation is an uncompressed NIfTI that got renamed `x.nii.gz`. It's fairly straightforward to identify data that isn't gzip or bzip2 by magic number, so we could help people out with a message like `File "test.nii.gz" is not a gzip file.`

jeromedockes:
fixes #966 
I added this check next to where `load` checks the file existence and size. please let me know if you would rather have it done some other way!

maybe another problem that could deserve a dedicated error message would be when the user doesn't have permissions to read the file?

### DIFF
--- a/nibabel/loadsave.py
+++ b/nibabel/loadsave.py
@@ -53,7 +53,7 @@ def _signature_matches_extension(filename):
     try:
         with open(filename, "rb") as fh:
             found_signature = fh.read(len(expected_signature))
-    except Exception:
+    except OSError:
         return False, f"Could not read file: {filename}"
     if found_signature == expected_signature:
         return True, ""
--- a/nibabel/loadsave.py
+++ b/nibabel/loadsave.py
@@ -60,7 +60,7 @@ def _signature_matches_extension(filename, sniff):
                 sniff = fh.read(len(expected_signature))
         except OSError:
             return False, f"Could not read file: {filename}"
-    if not sniff.startswith(expected_signature):
+    if sniff.startswith(expected_signature):
         return True, ""
     format_name = signatures[ext]["format_name"]
     return False, f"File {filename} is not a {format_name} file"
--- a/nibabel/loadsave.py
+++ b/nibabel/loadsave.py
@@ -46,7 +46,7 @@ def _signature_matches_extension(filename, sniff):
     """
     signatures = {
         ".gz": {"signature": b"\x1f\x8b", "format_name": "gzip"},
-        ".bz2": {"signature": b"\x42\x5a\x68", "format_name": "bzip2"}
+        ".bz2": {"signature": b"BZh", "format_name": "bzip2"}
     }
     filename = _stringify_path(filename)
     *_, ext = splitext_addext(filename)
@@ -54,15 +54,13 @@ def _signature_matches_extension(filename, sniff):
     if ext not in signatures:
         return True, ""
     expected_signature = signatures[ext]["signature"]
-    if sniff is not None and len(sniff) >= len(expected_signature):
-        found_signature = sniff[:len(expected_signature)]
-    else:
+    if sniff is None or len(sniff) < len(expected_signature):
         try:
             with open(filename, "rb") as fh:
-                found_signature = fh.read(len(expected_signature))
+                sniff = fh.read(len(expected_signature))
         except OSError:
             return False, f"Could not read file: {filename}"
-    if found_signature == expected_signature:
+    if not sniff.startswith(expected_signature):
         return True, ""
     format_name = signatures[ext]["format_name"]
     return False, f"File {filename} is not a {format_name} file"
--- a/nibabel/loadsave.py
+++ b/nibabel/loadsave.py
@@ -20,6 +20,47 @@ from .arrayproxy import is_proxy
 from .deprecated import deprecate_with_version
 
 
+def _signature_matches_extension(filename):
+    """Check if signature aka magic number matches filename extension.
+
+    Parameters
+    ----------
+    filename : str or os.PathLike
+        Path to the file to check
+
+    Returns
+    -------
+    matches : bool
+       - `True` if the filename extension is not recognized (not .gz nor .bz2)
+       - `True` if the magic number was successfully read and corresponds to
+         the format indicated by the extension.
+       - `False` otherwise.
+    error_message : str
+       An error message if opening the file failed or a mismatch is detected;
+       the empty string otherwise.
+
+    """
+    signatures = {
+        ".gz": {"signature": b"\x1f\x8b", "format_name": "gzip"},
+        ".bz2": {"signature": b"\x42\x5a\x68", "format_name": "bzip2"}
+    }
+    filename = _stringify_path(filename)
+    *_, ext = splitext_addext(filename)
+    ext = ext.lower()
+    if ext not in signatures:
+        return True, ""
+    expected_signature = signatures[ext]["signature"]
+    try:
+        with open(filename, "rb") as fh:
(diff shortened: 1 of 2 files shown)


nibabel/loadsave.py
```diff
@@ -44,6 +85,9 @@ def load(filename, **kwargs):
         raise FileNotFoundError(f"No such file or no access: '{filename}'")
     if stat_result.st_size <= 0:
         raise ImageFileError(f"Empty file: '{filename}'")
+    matches, msg = _signature_matches_extension(filename)
+    if not matches:
+        raise ImageFileError(msg)
```
effigies, line 90: My concern here is that this adds a penalty of a double read on *good* files, which can be add up when reading many files over a network.

But I think we can use the existing machinery to avoid extra reads. Note that L94 (was L50) has a `sniff` block, which is the first N bytes of an opened file.

https://github.com/nipy/nibabel/blob/5c4f39cef1f04a9afd241242c7f0148d7e5935c9/nibabel/loadsave.py#L50

Because we don't really care about the I/O penalty of a failure, I'm okay letting the loop fail a few times before we check if it's the obvious problem.

```Diff
     sniff = None
     for image_klass in all_image_classes:
         is_valid, sniff = image_klass.path_maybe_image(filename, sniff)
         if is_valid:
             img = image_klass.from_filename(filename, **kwargs)
             return img
 
+    matches, msg = _signature_matches_extension(filename, sniff)
+    if not matches:
+        raise ImageFileError(msg)
 
     raise ImageFileError(f'Cannot work out file type of "{filename}"')
```

This would require us to add a `sniff=None` parameter to  `_signature_matches_extension` so the `open().read()` call could be skipped if `not None`.

nibabel/loadsave.py
```diff
@@ -20,6 +20,47 @@
...
+    filename = _stringify_path(filename)
+    *_, ext = splitext_addext(filename)
+    ext = ext.lower()
+    if ext not in signatures:
+        return True, ""
+    expected_signature = signatures[ext]["signature"]
+    try:
+        with open(filename, "rb") as fh:
+            found_signature = fh.read(len(expected_signature))
+    except Exception:
```
effigies, line 56:
```suggestion
    except OSError:
```

nibabel/tests/test_loadsave.py
```diff
@@ -76,6 +76,21 @@ def test_load_empty_image():
     assert str(err.value).startswith('Empty file: ')
 
 
+@pytest.mark.parametrize("extension", [".gz", ".bz2"])
+def test_load_bad_compressed_extension(tmp_path, extension):
+    file_path = tmp_path.joinpath(f"img.nii{extension}")
```
effigies, line 81: Just being picky...

```suggestion
    file_path = tmp_path / f"img.nii{extension}"
```

nibabel/tests/test_loadsave.py
```diff
@@ -76,6 +76,21 @@ def test_load_empty_image():
...
+    file_path.write_bytes(b"bad")
+    with pytest.raises(ImageFileError, match=".*is not a .* file"):
+        load(file_path)
+
+
+def test_load_file_that_cannot_be_read(tmp_path):
+    subdir = tmp_path.joinpath("img.nii.gz")
```
effigies, line 88:
```suggestion
    subdir = tmp_path / "img.nii.gz"
```

effigies (commented):
Thanks! Wrote most of my comments in one spot, so see below.

nibabel/loadsave.py
```diff
@@ -20,6 +20,54 @@
...
+    expected_signature = signatures[ext]["signature"]
+    if sniff is not None and len(sniff) >= len(expected_signature):
+        found_signature = sniff[:len(expected_signature)]
+    else:
+        try:
+            with open(filename, "rb") as fh:
+                found_signature = fh.read(len(expected_signature))
+        except OSError:
+            return False, f"Could not read file: {filename}"
+    if found_signature == expected_signature:
```
effigies, line 65: Maybe odd styling preference, but I like to do:

```Python
if data not ready:
    prep_data
proceed as usual
```

An `else` block is a thing to keep in mind, so I tend to look for ways to avoid them. Here I'd say:

```suggestion
    if sniff is None or len(sniff) < len(expected_signature):
        try:
            with open(filename, "rb") as fh:
                sniff = fh.read(len(expected_signature))
        except OSError:
            return False, f"Could not read file: {filename}"
    if not sniff.startswith(expected_signature):
```

nibabel/loadsave.py
```diff
@@ -20,6 +20,54 @@
...
+         the format indicated by the extension.
+       - `False` otherwise.
+    error_message : str
+       An error message if opening the file failed or a mismatch is detected;
+       the empty string otherwise.
+
+    """
+    signatures = {
+        ".gz": {"signature": b"\x1f\x8b", "format_name": "gzip"},
+        ".bz2": {"signature": b"\x42\x5a\x68", "format_name": "bzip2"}
```
effigies, line 49: Any reason to use hex codes instead of ASCII values here?

```suggestion
        ".bz2": {"signature": b"BZh", "format_name": "bzip2"}
```

jeromedockes: no reason!

nibabel/loadsave.py
jeromedockes: sounds good!

effigies (approved):

nibabel/loadsave.py
```diff
@@ -20,6 +20,52 @@
...
+    if ext not in signatures:
+        return True, ""
+    expected_signature = signatures[ext]["signature"]
+    if sniff is None or len(sniff) < len(expected_signature):
+        try:
+            with open(filename, "rb") as fh:
+                sniff = fh.read(len(expected_signature))
+        except OSError:
+            return False, f"Could not read file: {filename}"
+    if not sniff.startswith(expected_signature):
```
effigies, line 63: Think I got this backwards...

```suggestion
    if sniff.startswith(expected_signature):
```

HippocampusGirl:
This is amazing! I was wondering that if the new code detects the file signature/magic numbers, then why raise an error when there is a mismatch? Why not just proceed to read the image?

effigies:
The biggest barrier to doing that is that we handle images by passing around the filename and only opening them when we need to read them, rather than potentially holding a bunch of open filehandles. This is done with the `ImageOpener` class, which determines whether it should use `__builtins__.open`, `gzip.open` or `BZ2file` to open files based on the extension. The `ImageOpener` itself is created statelessly each time the file needs to be opened, so either it would need to be stateful, or we'd need to keep a registry of filenames with mismatched compression states, or something else that would achieve this. A stateful `ImageOpener` would also need to be careful to close its filehandles; right now we are sure things get closed because we throw away the objects.

I'd review a PR along those lines, but tbh when the fix is for the user to rename a file to be more correct, I wouldn't put in the effort myself.

HippocampusGirl:
Thank you for the explanation. I agree that it seems to be quite a lot of effort without a lot to gain :-)

PR #1014, opened 2021-05-06
NF Operator class

### LINKED ISSUE #939: Implement arithmetic and logical operator support?
This has been suggested before (by me, and probably others), but it would be quite helpful to have nibabel classes (at least in the nifti hierarchy) support arithmetic and logical operations as syntactic sugar for common operations—at least in cases where the semantics can be defined with little or no ambiguity.

E.g., `img1 & img2` would give the logical conjunction of two images; `img1 - img2` the difference; etc. It doesn't seem like users are likely to be confused about what these operators do, and obviously they would only work for operands with identical dimensions/affines.

htwangtw:
Closes #939.

This is a draft for feedback. Any suggestion is welcome! 
I am not very familiar with python class so please let me know if there's things in python I haven't taken advantage of already. I would like to know:
1. Are there header/affine check functions that already exist in `nibabel` that I can reuse?
2. More efficient way to add more operators? I don't quite understand the example in the original issue:
`    __and__ = partial(_op, op=operator.__and__)`

**To-do**
- [x] check any other operator we want to cover
- [x] better/flexible evaluation of data shape  - I would use `fslmaths` as a reference dealing with masking 4D array with 3D image
- [ ] more test cases
- [ ] better error handling
- [x] docstring
- [ ] warning related to caching


effigies:
Hi @htwangtw, I'm interpreting the `DRAFT` heading to mean you want to push on this more pre-review, but please just ping me whenever you want feedback, or if you have design questions you want to raise.

htwangtw:
I found the draft function not particular meaningful to me so I just make it a normal PR...
@effigies I have a few questions:
1. Memory management: A normal image loaded by `nibabel`, `dataobj` is some reference to the data on disc. When we do an operation, the `dataobj` will be read and stored in the memory, and the same for the new nifti image generate. Am I missing anything?

2. `ZeroDivisionError`: To my surprise numpy returns `inf`, not  `ZeroDivisionError` when dividing a value by zero. Is this a thing better be flagged as error before the operation or just let it be?

3. Behaviour when dealing with 3D vs 4D image: I don't have a good answer on this one. We can assume people are familiar with `fslmaths` and get similar image outputs. We can just disallow this kind of case, but I think it would be nice to have `fslmaths` like behaviour. Thoughts???

4. Operators that only require one input: I can see `abs(img)`, `-img`, `+img`  really useful. I just don't have an idea how they can fit into the current `_op` method. Thoughts?

effigies:
> I found the draft function not particular meaningful to me so I just make it a normal PR...
> @effigies I have a few questions:
> 
> 1. Memory management: A normal image loaded by `nibabel`, `dataobj` is some reference to the data on disc. When we do an operation, the `dataobj` will be read and stored in the memory, and the same for the new nifti image generate. Am I missing anything?

Nope. No avoiding it. The only other way to do it would be by allowing `ArrayProxy`s to build up a sequence of partial applications, but that is a much bigger task than is called for here.

> 2. `ZeroDivisionError`: To my surprise numpy returns `inf`, not  `ZeroDivisionError` when dividing a value by zero. Is this a thing better be flagged as error before the operation or just let it be?

It's a valid IEEE754 value. You pays your money, you takes your chances.

> 3. Behaviour when dealing with 3D vs 4D image: I don't have a good answer on this one. We can assume people are familiar with `fslmaths` and get similar image outputs. We can just disallow this kind of case, but I think it would be nice to have `fslmaths` like behaviour. Thoughts???

I'm not sure what the `fslmaths` behavior is here, but I would expect if the volumetric dimensions match, to broadcast along the time dimension. I suspect this is the numpy default?

> 4. Operators that only require one input: I can see `abs(img)`, `-img`, `+img`  really useful. I just don't have an idea how they can fit into the current `_op` method. Thoughts?

You could do something like:

```Python
class OperableImage:
    def _binop(self, val, *, op):
        ...
    def _unop(self, *, op):
        ...
```

htwangtw:
Hey @effigies  - 

I tried to add some `Nifti1Image.dataobj` type test and realise the `Nifti1Image` class itself has pretty comprehensive checks upon image creation. Can you think of a chase that invalid `Nifti1Image.dataobj` value can be loaded?

PR #1017, opened 2021-05-12
MRG: add BIAP from Wiki

matthew-brett:


matthew-brett:
I think this one is OK to merge - it fixes a few links, but it will also allow us to point to the main docs for the BIAPs.  When merged, I'll redirect from the current wiki pages.

effigies:
Not at a laptop, so can't push a new build. Might want to set up CI to build the docs one of these days.

PR #1024, opened 2021-06-04
ENH: Create gzip header deterministically by default

### LINKED ISSUE #1023: Setting gzip mtime to zero on image save
Right now `ImageOpener()` does not allow us to pass an `mtime`, which means that saving identical images as `.nii.gz` 1s apart will produce two images with two different checksums.

Is there any value in saving a non-zero mtime? I see a few options, in my order of preference, most preferred first:

1) Always save as 0.
2) Default to 0, allow user to override.
3) Default to a timestamp (current), allow user to override (new).

Note that the way we use `GzipFile`, it also sets the filename, which means saving the `a.nii.gz` and `b.nii.gz` with the same timestamp would still not have identical checksums. I would also be inclined to open GzipFiles in such a way as to not set filenames.

effigies:
As discussed in #1023, we currently allow `gzip.GzipFile` to set the gzip header as it will, including filename and mtime fields. This means that files with identical unzipped contents and compression levels have different gzipped representations, which is not ideal for verifying sources of non-determinism in a complex workflow.

This PR attempts to balance sensible default behaviors with user control. It changes the default `mtime` from `time.time()` to `0`, while allowing users to set the `mtime` as an argument to `Opener()`.

It *removes* the behavior of setting the original filename in the gzip header. In a medical imaging context, the original filenames may include PHI (e.g., subject name, scan date, etc) that will follow the file through renames unless the file is rewritten entirely. A flag to preserve the old behavior does not seem worth the effort.

Fixes #1023.

### DIFF
--- a/nibabel/tests/test_openers.py
+++ b/nibabel/tests/test_openers.py
@@ -16,7 +16,7 @@ import hashlib
 import time
 
 from numpy.compat.py3k import asstr, asbytes
-from ..openers import Opener, ImageOpener, HAVE_INDEXED_GZIP, BZ2File
+from ..openers import Opener, ImageOpener, HAVE_INDEXED_GZIP, BZ2File, DeterministicGzipFile
 from ..tmpdirs import InTemporaryDirectory
 from ..volumeutils import BinOpener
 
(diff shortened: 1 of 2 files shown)


effigies:
@yarikoptic @ltetrel Would either of you care to review?

nibabel/tests/test_openers.py
```diff
@@ -367,3 +370,126 @@ def test_iter():
...
+def test_DeterministicGzipFile():
+    with InTemporaryDirectory():
+        msg = b"Hello, I'd like to have an argument."
+
+        # No filename, no mtime
+        with open("ref.gz", "wb") as fobj:
+            with GzipFile(filename="", mode="wb", fileobj=fobj, mtime=0) as gzobj:
+                gzobj.write(msg)
+        anon_chksum = md5sum("ref.gz")
+
```
yarikoptic, line 389: minor: since file is small, and time is stored up to a second, shouldn't we sleep here for a second to 100% ensure that time is different between two invocations?

effigies: This is checking changing the filename. We mock mtime on L412+

yarikoptic (approved):
Looks great, thank you @effigies !!!

nibabel/tests/test_openers.py
yarikoptic: great, thanks!

effigies:
Thanks for the review.

kousu:
This is great! Thank you for being proactive about squashing the indeterminism/data leakage.

PR #1030, opened 2021-07-09
WIP: Add memory efficient meta data summary

moloney:
This is some work-in-progress for adding data structures for creating a memory efficient summary of a sequence of meta data dictionaries (assuming a large number of keys/values repeat) and then using this to determine how to sort the associated images into an nD array.

This approach was inspired by [this dcmstack issue](https://github.com/moloney/dcmstack/issues/37).


nibabel/metasum.py
```diff
@@ -0,0 +1,583 @@
...
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+'''Memory efficient tracking of meta data dicts with repeating elements
+'''
+from dataclasses import dataclass
```
effigies, line 11: This is going to fail on Python 3.6. We may need to create a stub `dataclass` that causes `DimIndex` to fail when initialized, but avoid crashing when just importing this file.

moloney: There is a backport available for 3.6: https://pypi.org/project/dataclasses/

Otherwise I could replace this with a more verbose regular class.

effigies: I guess let's add a conditional dependency for 3.6.

setup.cfg
```diff
@@ -31,6 +31,7 @@ python_requires = >=3.6
 install_requires =
     numpy >=1.13
     packaging >=14.3
+    bitarray
```
effigies, line 34:
```suggestion
    bitarray
    dataclasses ; python_version < "3.7"
```

effigies:
Merged master in to resolve conflicts and get the tests going. Let me know if you'd prefer I didn't do that.

ZviBaratz:
@matthew-brett  / @moloney / @effigies
If DICOM related functionality is moved to dicom_parser, do you still think the `MetaSummary` implementation will be required?
I feel like we could simply cache a dictionary of lazily evaluated header values within each `Series` instance. The higher level `Dataset` class (to be implemented) can simply query those.

moloney:
@ZviBaratz Can you explain in more detail what you have in mind?  I don't see how a cache helps to solve the problem of determining what meta data is varying when someone hands us a list of Dicom files we have never seen before (that could come from multiple Dicom series).

ZviBaratz:
The idea is that there will be a `Dataset` class which will receive a root directory and iterate its files to create the representations for the contained series. When a user tries to query based on any particular header field, the dataset queries all the created `Series` instances headers to retrieve the value (at which point it could be saved to a cache dictionary in order to avoid repeating computations). Of course some evaluation time is to be expected, but I don't think it should be anything too bad up to a few dozen series. If you're working with more than that, it might be best to export the metadata to some external table anyway.

moloney:
We really don't want to require all the files live in a single directory.  The assumption is you are passed a list of files that could be massive even for a single series (e.g. 36K) that you have never seen before and you want to efficiently convert them into an xarray on the fly.  My original implementation in dcmstack wasn't totally naive, meta data values that were constant were only stored once, and yet it required orders of magnitude more memory (18GB vs ~800MB) compared to this approach.

ZviBaratz:
I see. 
I'll be working on the issues that are already piling up in dicom_parser for the next couple of weeks, after that I'll start thinking on how this would best be integrated into dicom_parser. We could discuss it in more detail in our next meeting.

moloney:
If we want to support using multiprocessing to speed up the parsing of very large series, this would also provide a nice compact representation to pass around.

effigies:
Sorry, I lost track of this one. What's the status? Are we still trying to get this into nibabel?

PR #1043, opened 2021-08-09
FIX: LGTM alerts

DimitriPapadopoulos:
This fixes some alerts where the solution is obvious.

Other alerts can be found here:
https://lgtm.com/projects/g/nipy/nibabel/


effigies (approved):
All looks good...

PR #1044, opened 2021-08-09
DOC: Typos found by codespell

DimitriPapadopoulos:


nibabel/_version.py
```diff
@@ -284,7 +284,7 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, run_command=run_command):
         # TAG-NUM-gHEX
         mo = re.search(r'^(.+)-(\d+)-g([0-9a-f]+)$', git_describe)
         if not mo:
-            # unparseable. Maybe git-describe is misbehaving?
+            # unparsable. Maybe git-describe is misbehaving?
```
effigies, line 287: We should fix issues in Versioneer in https://github.com/python-versioneer/python-versioneer, rather than making edits here.

DimitriPapadopoulos: Done.

effigies (approved):

nibabel/_version.py
DimitriPapadopoulos: Note that this has already been fixed in Versioneer, just 14 daus ago:
https://github.com/python-versioneer/python-versioneer/pull/260

PR #1048, opened 2021-08-22
FIX: Return error codes in dicomfs, rather than raise

DimitriPapadopoulos:
> Illegal raise
> 
> Illegal class '`int`' raised; will result in a `TypeError` being raised instead.

I believe the intent was to `return`, not to `raise`. @chrispycheng Can you confirm?


chrispycheng:
@DimitriPapadopoulos you got it! nice catch, thanks.

effigies:
Thanks! Test failures are unrelated.

PR #1049, opened 2021-08-22
FIX: LGTM error

DimitriPapadopoulos:
> Non-callable called
> 
> Call to a `non-callable` of class `GiftiDataArray`.

This error is a false positive (https://github.com/github/codeql/issues/6527), and this commit merely attempts to work around it.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -380,20 +380,20 @@ class GiftiDataArray(xml.XmlSerializable):
         return len(self.dims)
 
     # Setter for backwards compatibility with pymvpa
-    @num_dim.setter
     @deprecate_with_version(
         "num_dim will be read-only in future versions of nibabel",
         '2.1', '4.0')
+    @num_dim.setter
     def num_dim(self, value):
         if value != len(self.dims):
             raise ValueError(f'num_dim value {value} != number of '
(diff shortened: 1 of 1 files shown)


effigies:
Unfortunately, I don't think this transposition works. Any other thoughts?

DimitriPapadopoulos:
Not really, apart from fixing the LGTM issue github/codeql#6527! Why doesn't it work? Does `@deprecate_with_version` need to be the last decorator?

effigies:
> Does `@deprecate_with_version` need to be the last decorator?

It shouldn't. It's a pretty generic decorator:

https://github.com/nipy/nibabel/blob/85c7b37886589c9e8a18bdc0749ae76ee6c81021/nibabel/deprecator.py#L176-L187

But `classmethod` and `property.setter` change the function to a different kind of thing, which the deprecator doesn't know isn't just a normal function. So it doesn't seem too surprising that those things need to be last (compositionally, so first in order).

DimitriPapadopoulos:
I see. I suppose I have to close this pull request then.

effigies:
Yeah I think there's not much to do here for now. If you do figure out an alternative strategy for decorating that would allow us to reorder, or that would pass-through the relevant features for LGTM, then let's reopen.

PR #1050, opened 2021-09-06
MNT: Set minimum pydicom to 1.0.0

effigies:
Pydicom<1 used the setuptools' ``use_2to3`` option to achieve Python 3 compatiblity. Setuptools has removed lib2to3 in pypa/setuptools#2731, which causes our minimum requirements tests to fail.

Will refactor `nicom` to not depend on `pydicom_compat` so we can deprecate it, as well.

### DIFF
--- a/nibabel/dft.py
+++ b/nibabel/dft.py
@@ -24,8 +24,9 @@ import numpy
 from io import BytesIO
 
 from .nifti1 import Nifti1Header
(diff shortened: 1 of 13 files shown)


effigies:
Should be a straightforward review. Plan to merge in 24 hours if nobody objects.

PR #1051, opened 2021-09-07
FIX: Work around numpy SystemError to maintain expected error types

effigies:
This should fix the pre-release tests. I expect pydicom to still cause problems in the minimum requirements tests.


effigies:
Any objections?

PR #1052, opened 2021-09-08
FIX: LGTM error: Clear-text logging of sensitive information

DimitriPapadopoulos:
Disable false positives that cannot be fixed by LGTM, because their heuristic specifically flags string "`id`" which is common in DICOM, for example in "`UID`".


effigies:
Thanks!

PR #1055, opened 2021-09-16
FIX: Suggest `nibabel.save()` on calls to deprecated `giftiio.write()`

anibalsolon:
Small adjustment to refer to correct function in deprecation notice.


effigies (approved):
Oh sure, if you insist on *useful* error messages.

PR #1056, opened 2021-09-17
DOC: First pass at SurfaceImage BIAP

effigies:
Haven't fully covered what's in the [HackMD](https://hackmd.io/ZXcVpr1wQvmQIq9Sl1Vidg), but did expand a bit where we made some decisions that didn't get written down. Also tried to pin down some terminology, as I think it's easy to speak imprecisely here. Some of this may be conventions that we need to enforce by transforming data on input (as we do with volumetric images come with affines that project into something besides RAS+).

Everything is still negotiable, so please feel free to make drastic suggestions.

Will start on implementations next week unless I feel inspired sooner.

cc @larsoner @htwangtw @NicolasGensollen @matthew-brett

### DIFF
--- a/doc/source/devel/biaps/biap_0009.rst
+++ b/doc/source/devel/biaps/biap_0009.rst
@@ -1,0 +1,159 @@
+.. _biap9:
+
+#############################
+BIAP9 - The Surface Image API
+#############################
+
+:Author: Chris Markiewicz
+:Status: Draft
+:Type: Standards
+:Created: 2021-09-16
+
+**********
+Background
+**********
+
+Surface data is generally kept separate from geometric metadata
+===============================================================
+
+In contrast to volumetric data, whose geometry can be fully encoded in the
+shape of a data array and a 4x4 affine matrix, data sampled to a surface
+requires the location of each sample to be explicitly represented by a
+coordinate. In practice, the most common approach is to have a geometry file
+and a data file.
+
+A geometry file consists of a vertex coordinate array and a triangle array
+describing the adjacency of vertices, while a data file is an n-dimensional
+array with one axis corresponding to vertex.
+
+Keeping these files separate is a pragmatic optimization to avoid costly
+reproductions of geometric data, but presents an administrative burden to
+direct consumers of the data.
+ 
+Terminology
+===========
+
+For the purposes of this BIAP, the following terms are used:
+
+* Coordinate - a triplet of floating point values in RAS+ space
+* Vertex - an index into a table of coordinates
+* Triangle (or face) - a triplet of adjacent vertices (A-B-C);
+  the normal vector for the face is ($\overline{AB}\times\overline{AC}$)
+* Topology - vertex adjacency data, independent of vertex coordinates,
+  typically in the form of a list of triangles
+* Geometry - topology + a specific set of coordinates for a surface
+* Patch - a connected subset of vertices (can be the full topology)
+* Data array - an n-dimensional array with one axis corresponding to the the
+  vertices (typical) OR faces (more rare) in a patch
+
+
+Currently supported surface formats
+===================================
+
+* FreeSurfer
+   * Geometry (e.g. ``lh.pial``):
+     :py:func:`~nibabel.freesurfer.io.read_geometry` / 
+     :py:func:`~nibabel.freesurfer.io.write_geometry`
+   * Data
+      * Morphometry:
+        :py:func:`~nibabel.freesurfer.io.read_morph_data` /
+        :py:func:`~nibabel.freesurfer.io.write_morph_data`
+      * Labels: :py:func:`~nibabel.freesurfer.io.read_label`
+      * MGH: :py:class:`~nibabel.freesurfer.mghformat.MGHImage`
+* GIFTI: :py:class:`~nibabel.gifti.gifti.GiftiImage`
+   * Every image contains a collection of data arrays, which may be
+     coordinates, topology, or data (further subdivided by type and intent)
+* CIFTI-2: :py:class:`~nibabel.cifti2.cifti2.Cifti2Image`
+   * Pure data array, with image header containing flexible axes
+   * Geometry referred to by an associated ``wb.spec`` file
+     (no current implementation in NiBabel)
+
+
+*********************************
+Desiderata for a SurfaceImage API
+*********************************
+
+The following are provisional guiding principles 
+
+1. A surface image (data array) should carry a reference to geometric metadata
+   that is easily transferred to a new image.
+2. Partial images (data only or geometry only) should be possible. Absence of
+   components should have a well-defined signature, such as a property that is
+   ``None`` or a specific ``Exception`` is raised.
+3. All arrays (coordinates, triangles, data arrays) should be proxied to
+   avoid excess memory consumption
+4. Selecting among coordinates (e.g., gray/white boundary, inflated surface)
+   for a single topology should be possible.
+5. Combining multiple brain structures (canonically, left and right hemispheres)
+   in memory should be easy; serializing to file may be format-specific.
+6. Splitting a data array into independent patches that can be separately
+   operated on and serialized should be possible.
+
+
+Prominent use cases
+===================
+
+* Arithmetic/modeling - per-vertex mathematical operations
+* Smoothing - topology/geometry-respecting smoothing
+* Plotting - paint the data array as a texture on a surface
+* Decimation - subsampling a topology (possibly a subset, possibly with
+  interpolated vertex locations)
+* Resampling to a geometrically-aligned surface
+  * Downsampling by decimating, smoothing, resampling
+  * Inter-subject resampling by using ``?h.sphere.reg``
+* Interpolation of per-vertex and per-face data arrays
+
+These are not necessarily operations that NiBabel must provide, but the
+components needed for each should be readily retrieved.
+
+********
+Proposal
+********
+
+.. code-block:: python
+
+    class SurfaceHeader:
+        @property
+        def ncoords(self):
+            """ Number of coordinates """
+
+        @property
+        def nfaces(self):
+            """ Number of faces """
+
+        def get_coords(self, name=None):
+            """ Nx3 array of coordinates in RAS+ space """
+
+        def get_faces(self, name=None):
+            """ Mx3 array of indices into coordinate table """
+
+        def get_mesh(self, name=None):
+            return self.get_coords(name=name), self.get_faces(name=name)
+
+        def decimate(self, *, ncoords=None, ratio=None):
+            """ Return a SurfaceHeader with a smaller number of vertices that
+            preserves the geometry of the original """
+            # To be overridden when a format provides optimization opportunities
+
+        def load_vertex_data(self):
+            """ Return a SurfaceImage with data corresponding to each vertex """
+
+        def load_face_data(self):
+            """ Return a SurfaceImage with data corresponding to each face """
+
+
+    class SurfaceImage:
+        @property
+        def header(self):
+            """ A SurfaceHeader or None """
+
+        @property
+        def dataobj(self):
+            """ An ndarray or ArrayProxy with one of the following properties:
+            
+            1) self.dataobj.shape[0] == self.header.ncoords
+            2) self.dataobj.shape[0] == self.header.nfaces
+            """
+
+        def load_header(self, pathlike):
+            """ Specify a header to a data-only image """
--- a/doc/source/devel/biaps/biap_0009.rst
+++ b/doc/source/devel/biaps/biap_0009.rst
@@ -116,6 +116,8 @@ make sense for NiBabel to provide methods.
 Proposal
 ********
 
+The basic API is as follows:
+
 .. code-block:: python
 
     class SurfaceHeader:
@@ -168,3 +170,57 @@ Proposal
 
         def load_header(self, pathlike):
             """ Specify a header to a data-only image """
+
+
+Here we present common use cases:
+
+
+Modeling
+--------
+
+.. code-block:: python
+
+    from nilearn.glm.first_level import make_first_level_design_matrix, run_glm
+
+    bold = SurfaceImage.from_filename("/data/func/hemi-L_bold.func.gii")
+    dm = make_first_level_design_matrix(...)
+    labels, results = run_glm(bold.get_data(), dm)
+    betas = bold.__class__(results["betas"], bold.header)
+    betas.to_filename("/data/stats/hemi-L_betas.mgz")
+
+Data images have their own metadata apart from geometry that needs handling in a header:
+
+* Axis labels (time series, labels, tensors)
+* Data dtype
+
+
+Plotting
+--------
+
+Nilearn currently provides a
+`plot_surf <https://nilearn.github.io/modules/generated/nilearn.plotting.plot_surf.html>`_ function.
+With the proposed API, we could interface as follows:
+
+.. code-block:: python
+
+    def plot_surf_img(img, surface="inflated"):
+        from nilearn.plotting import plot_surf
+        coords, triangles = img.header.get_mesh(name=surface)
+
+        data = tstats.get_data()
+
+        return plot_surf((triangles, coords), data)
+
+    tstats = SurfaceImage.from_filename("/data/stats/hemi-L_contrast-taskVsBase_tstat.mgz")
+    tstats.load_header("/data/subjects/fsaverage5")
+    plot_surf_img(tstats)
+
+This assumes that ``load_header()`` will be able to identify a FreeSurfer subject directory.
+
+******************
+* Open questions *
+******************
+
+1) Can/should image and header objects be promiscuous? If I load a GIFTI header,
+   should ``header.load_vertex_data()`` accept a FreeSurfer curvature file, or
+   should there be a bit more ecosystem-local logic?
--- a/doc/source/devel/biaps/biap_0009.rst
+++ b/doc/source/devel/biaps/biap_0009.rst
@@ -43,7 +43,7 @@ For the purposes of this BIAP, the following terms are used:
   typically in the form of a list of triangles
 * Geometry - topology + a specific set of coordinates for a surface
 * Patch - a connected subset of vertices (can be the full topology)
-* Data array - an n-dimensional array with one axis corresponding to the the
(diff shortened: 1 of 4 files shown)


doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,159 @@
...
+   * Data
+      * Morphometry:
+        :py:func:`~nibabel.freesurfer.io.read_morph_data` /
+        :py:func:`~nibabel.freesurfer.io.write_morph_data`
+      * Labels: :py:func:`~nibabel.freesurfer.io.read_label`
+      * MGH: :py:class:`~nibabel.freesurfer.mghformat.MGHImage`
+* GIFTI: :py:class:`~nibabel.gifti.gifti.GiftiImage`
+   * Every image contains a collection of data arrays, which may be
+     coordinates, topology, or data (further subdivided by type and intent)
+* CIFTI-2: :py:class:`~nibabel.cifti2.cifti2.Cifti2Image`
```
satra, line 66: i wouldn't completely call cifti-2 a surface format. there is no geometry information stored in the file itself (unless someone hacked it. it's a point cloud that happens to have been extracted from some combination of a surface geometry and a volume geometry. 

also many of the cifti-2 types are parcel based.

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,159 @@
...
+
+* Coordinate - a triplet of floating point values in RAS+ space
+* Vertex - an index into a table of coordinates
+* Triangle (or face) - a triplet of adjacent vertices (A-B-C);
+  the normal vector for the face is ($\overline{AB}\times\overline{AC}$)
+* Topology - vertex adjacency data, independent of vertex coordinates,
+  typically in the form of a list of triangles
+* Geometry - topology + a specific set of coordinates for a surface
+* Patch - a connected subset of vertices (can be the full topology)
+* Data array - an n-dimensional array with one axis corresponding to the the
```
NicolasGensollen, line 46:
```suggestion
* Data array - an n-dimensional array with one axis corresponding to the
```

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,159 @@
...
+    class SurfaceHeader:
+        @property
+        def ncoords(self):
+            """ Number of coordinates """
+
+        @property
+        def nfaces(self):
+            """ Number of faces """
+
+        def get_coords(self, name=None):
```
NicolasGensollen, line 124: If I remember our discussion, `name` would be used to select the hemisphere, inflated, pial, and so on... right?
Would the `SurfaceHeader` provide then a `get_names` method or something similar to get which name values can be used here?

NicolasGensollen (commented):
Thanks @effigies !
I just made a quick pass but it looks good to me so far.

doc/source/devel/biaps/biap_0009.rst
htwangtw: Indeed CIFTI supports multimodal data not purely for surface. It kind of deserve its own category. 
IMO with the popularity of HCP data and fmriprep supporting output in fsLR template, surface based data support for CIFTI is too common to ignore. A major road block for user is data I/O. Working on how CIFTI-2 image relates to geometry template is a starting point none the less.

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,159 @@
...
+direct consumers of the data.
+ 
+Terminology
+===========
+
+For the purposes of this BIAP, the following terms are used:
+
+* Coordinate - a triplet of floating point values in RAS+ space
+* Vertex - an index into a table of coordinates
+* Triangle (or face) - a triplet of adjacent vertices (A-B-C);
```
htwangtw, line 40: Should we change it to `Face (or triangle)` as in the API draft below, we are using face primarily?

bthirion: `triangle` is more precise I think.

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,159 @@
...
+
+********
+Proposal
+********
+
+.. code-block:: python
+
+    class SurfaceHeader:
+        @property
+        def ncoords(self):
```
bthirion, line 117: maybe n_coords ?

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,159 @@
...
+            """ Number of coordinates """
+
+        @property
+        def nfaces(self):
```
bthirion, line 121: and n_face or n_triangles bzw ?

bthirion (commented):
Thx. Just left tiny comments.

doc/source/devel/biaps/biap_0009.rst
htwangtw: I agree. So change the API below then.

doc/source/devel/biaps/biap_0009.rst
effigies: That's what I meant by "Pure data array". Fair point that it accepts data that has no geometric basis (parcels). Made a note of that case.

danjgale:
@effigies thanks for the tag. TBH I found working with surfaces really easy when putting together surfplot, and I'm trying to recall any issues I had. When I eventually get back  to polishing it up (soon🤞) and come across any issues, I can let you know. 

The biggest challenges I've had primarily concern manipulating CIFTIs, such as changing the label table, creating a new CIFTI file from a modified data array, etc. [Here's an example manipulation](https://github.com/danjgale/nixtract/blob/5f0519b1b96633ee5f59136f2434cb272426e37d/tests/setup_mock_data.py#L259-L319) I've done in a different package, though I'm not sure how applicable these challenges are to the current topic.

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,165 @@
...
+
+
+Currently supported surface formats
+===================================
+
+* FreeSurfer
+   * Geometry (e.g. ``lh.pial``):
+     :py:func:`~nibabel.freesurfer.io.read_geometry` / 
+     :py:func:`~nibabel.freesurfer.io.write_geometry`
+   * Data
```
larsoner, line 57: Might be worth detailing the number of dimensions supported in each of these, unless they all support arbitrary dimensions. Some or all (or none, I'm not sure!) of these could be a single scalar per vertex, so you can't for example have the `n-dimensional`ity mentioned in the `Data array` definition above, you're stuck with n=0 (per vertex)

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,165 @@
...
+5. Combining multiple brain structures (canonically, left and right hemispheres)
+   in memory should be easy; serializing to file may be format-specific.
+6. Splitting a data array into independent patches that can be separately
+   operated on and serialized should be possible.
+
+
+Prominent use cases
+===================
+
+* Arithmetic/modeling - per-vertex mathematical operations
```
larsoner, line 97: I think as long as there is a way to expose per-vertex data `ndarray`, we can just say "use NumPy-compatible tools" for this

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,165 @@
...
+* Smoothing - topology/geometry-respecting smoothing
+* Plotting - paint the data array as a texture on a surface
```
larsoner, line 99: Having worked with VTK / Mayavi / PyVista / VisPy / mplot3d each a bit, I think this is going to be more difficult than it seems. This to me seems better tackled by a separate package, otherwise maintenance will be difficult.

I see the text below about NiBabel not necessarily providing each operation, so maybe adding to the `Proposal` below explicitly what functionality is out of scope would be good?

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,165 @@
...
+* Decimation - subsampling a topology (possibly a subset, possibly with
+  interpolated vertex locations)
+* Resampling to a geometrically-aligned surface
+  * Downsampling by decimating, smoothing, resampling
+  * Inter-subject resampling by using ``?h.sphere.reg``
+* Interpolation of per-vertex and per-face data arrays
+
+These are not necessarily operations that NiBabel must provide, but the
+components needed for each should be readily retrieved.
```
larsoner, line 108: I would put this at the top of this section -- when reading the list first, I assumed that the goal was to support/implement all prominent use cases.

larsoner (commented):
Just some preliminary small stuff, looks good so far!

doc/source/devel/biaps/biap_0009.rst
effigies: I considered that `n=0`, since one axis is the vertex.

larsoner:
Sorry about missing the meeting today. I didn't get to the FreeSurfer stuff in the last two weeks but it looks like you started (yay!). Let me know if you want me to try it or add to it or anything else

effigies:
Hey @larsoner, I've spent the morning fiddling with a collection API, if you want to have a look. I did the FreeSurfer stuff just to ground myself before diving into CIFTI and CaretSpecFile (now workbench specs), so probably won't be spending much more (self-directed) time on it, so you should feel free to play with it and see what it still needs to be useful.

neurolabusc:
@effigies you may want to consider handling the [mz3](https://github.com/neurolabusc/surf-ice/tree/master/mz3) format, the page includes Python, Blender, JavaScript code as well as a full specification and examples datasets. It attempts to live at the Pareto frontier of file size and speed. It is both faster and smaller than GIfTI. The layout design maps directly onto modern GPU architectures (and hence WebGL, Metal and Vulkan APIs). I developed this primarily for the sample files distributed with Surfice, to reduce download times, disk usage and the time to run the demos / regression testing. It is intentionally simple, and does not have the flexibility of other formats. It does not attempt to do everything, but it can fill a niche.

effigies:
@neurolabusc Thanks! Opened an issue to avoid this getting lost...

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,309 @@
...
+Smoothing
+=========
+
+.. code-block:: python
+
+    bold = SurfaceImage.from_filename("/data/func/hemi-L_bold.func.gii")
+    bold.load_geometry("/data/anat/hemi-L_midthickness.surf.gii")
+    # Not implementing networkx weighted graph here, so assume we have a method
+    graph = bold.geometry.get_graph()
+    distances = distance_matrix(graph)  # n_coords x n_coords matrix
```
larsoner, line 232: This is maybe going to be harder than we think:
```
>>> rr, tris = nib.freesurfer.read_geometry('/mnt/bakraid/data/structurals/fsaverage/surf/lh.white')
>>> from mne.surface import mesh_dist
>>> adjacency = mesh_dist(tris, rr)
>>> t0 = time.time(); dist_matrix = dijkstra(adjacency, directed=False, limit=10.); print(time.time() - t0)
...
numpy.core._exceptions.MemoryError: Unable to allocate 200. GiB for an array with shape (163842, 163842) and data type float64
```
So we'll need to improve SciPy's dijkstra to support `sparse` outputs start...

If we don't use the distance-based method and instead use the "smudging" operation by pushing data to neighboring vertices using a number of steps, this process is avoided. I guess this is a bit of an actual implementation detail, but we should think about these two methods as options with different tradeoffs.

nibabel/tests/test_surfaceimages.py
```diff
@@ -0,0 +1,221 @@
...
+        return self.meshes[self._default].vnum
+
+    @property
+    def n_triangles(self):
+        return self.meshes[self._default].fnum
+
+
+class FreeSurferSubject(GeometryCollection):
+    @classmethod
+    def from_subject(klass, subject_id, subjects_dir=None):
```
larsoner, line 206: So far:
```
>>> from nibabel.tests.test_surfaceimages import FreeSurferSubject
>>> subj = FreeSurferSubject.from_subject('fsaverage', '/mnt/bakraid/data/structurals')
  File "<stdin>", line 1, in <module>
  File "/home/larsoner/python/nibabel/nibabel/tests/test_surfaceimages.py", line 210, in from_subject
    return klass.from_directory(Path(subjects_dir) / subject_id)
AttributeError: type object 'FreeSurferSubject' has no attribute 'from_directory'
```
So some unit tests are clearly needed :)

And if I change the call to use `from_spec`, then there is no `self` referred to in that class method

nibabel/tests/test_surfaceimages.py
```diff
@@ -0,0 +1,221 @@
...
+
+    @property
+    def triangles(self):
+        offset = self.offset + 12 * self.vnum
+        ap = ArrayProxy(self._file_like, ((self.fnum, 3), ">i4", offset))
+        ap.order = 'C'
+        return ap
+
+
+class FreeSurferHemisphere(SurfaceGeometry):
```
larsoner, line 166: This seems to work well as a subclass in local testing

larsoner (commented):
Just some preliminary testing from me, but looks like a good start. I can see the direction from the code and it makes sense.

doc/source/devel/biaps/biap_0009.rst
effigies: Yes, full pairwise distances is painfully naive; not sure if iterated smudging is equivalent to distance-based (up to accumulated floating point error) but seems reasonable.

Mostly as a bookmark to myself, it might be worth looking into https://github.com/gilShamai/Fast-Pairwise-Geodesic-Distance-Computation to see if they have any good ideas.

jbteves:
@larsoner @effigies where should I put some toy code as an example of converting an STC to a GIFTI? Should that go here, or should that be something we put in MNE first, and then add to this API later?

effigies:
Would a gist be sensible in the short term?

jbteves:
Sure, here you go:

https://gist.github.com/jbteves/384bfdd3777190f59620c3d988a314fd

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,335 @@
...
+.. code-block:: python
+
+    img = nb.load("sub-01_task-rest_bold.dtseries.nii")  # Assume CIFTI CoordinateImage
+    parcel = nb.load("sub-fsLR_hemi-L_label-DLPFC_mask.label.gii") # GiftiImage
+    structure = parcel.meta.metadata['AnatomicalStructurePrimary'] # "CortexLeft"
+    vtx_idcs = np.where(parcel.agg_data())[0]
+    dlpfc_idcs = img.coordaxis.get_indices(parcel=structure, indices=vtx_idcs)
+
+    # Subsampled coordinate axes will override any duplicate information from header
+    dlpfc_img = CoordinateImage(img.dataobj[dlpfc_idcs], img.coordaxis[dlpfc_idcs], img.header)
```
larsoner, line 330: Or in syntactic sugar mode via `__getitem__`:
```
dlpfc_img = img[dlpfc_idcs]
```
and under the hood you just do the operation above?

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,335 @@
...
+
+    # Now load geometry so we can plot
+    wbspec = CaretSpec("fsLR.wb.spec")
+    dlpfc_img.coordaxis.load_structures(wbspec)
+    ...
```
larsoner, line 335: Add an example of trivial HDF5-based round-trip?
```
Native HDF5 format support
--------------------------
To support saving without loss of information, we also support serialization
to HDF5:
...
```

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,335 @@
...
+========
+
+.. code-block:: python
+
+    from nilearn.glm.first_level import make_first_level_design_matrix, run_glm
+
+    bold = CoordinateImage.from_filename("/data/func/hemi-L_bold.func.gii")
+    dm = make_first_level_design_matrix(...)
+    labels, results = run_glm(bold.get_fdata(), dm)
+    betas = CoordinateImage(results["betas"], bold.coordaxis, bold.header)
```
larsoner, line 266: One thing to note here is that `betas` will have collapsed over the time axis of the data, so the `bold.header` might be incorrect now, or otherwise need adjustment.

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,335 @@
...
+
+.. code-block:: python
+
+    bold = CoordinateImage.from_filename("/data/func/hemi-L_bold.func.gii")
+    bold.coordaxis.load_structures({"lh": "/data/anat/hemi-L_midthickness.surf.gii"})
+    # Not implementing networkx weighted graph here, so assume we have a function
+    # that retrieves a graph for each structure
+    graphs = get_graphs(bold.coordaxis)
+    distances = distance_matrix(graphs['lh'])  # n_coords x n_coords matrix
+    weights = normalize(gaussian(distances, sigma))
```
larsoner, line 289:
```suggestion
    weights = normalize(gaussian(distances, sigma))  # sparse matrix
```

doc/source/devel/biaps/biap_0009.rst
```diff
@@ -0,0 +1,335 @@
...
+    # Wildly inefficient smoothing algorithm
+    smoothed = CoordinateImage(weights @ bold.get_fdata(), bold.coordaxis, bold.header)
```
larsoner, line 291: As you mentioned, this will kill most people's machines. Maybe you need to do this for GIfTI, but you shouldn't need to for the native HDF5 format, or for most operations (including plotting) -- it can just be done on the fly.

WDYT about an `upsampler` property or way to specify how to go from the subsampled / parcel space to the full-resolution space? By default / None could mean "map directly to the specified parcel, leave all else zero", but you in principle should be able to have this also be a `scipy.sparse` CSR or CSC matrix of shape `(n_full_space, n_parcel)` that you can `@` with your data to get to the full-res space i.e., high-res surface for surfaces, or T1 / 1mm voxel spacing (usually) for volumetric.

larsoner (commented):
Feel free to merge and address my comments later @effigies , already looks like a great start!

PR #1059, opened 2021-09-30
FIX: Reshape CIFTI-2 affines to 4x4 when encoded as row-major sequence

vanandrew:
Found this issue while working with some dconn files in our lab. If the file was written out by [`ft_write_cifti`](https://www.fieldtriptoolbox.org/reference/ft_write_cifti/), nibabel (3.2.1) fails to load the file with the following exception:

```python
ValueError                                Traceback (most recent call last)
<ipython-input-3-fc915084f155> in <module>
----> 1 img = nib.load("/home/vanandrew/Data/SIC02_MarguliesSimilarity_xdist30.dconn.nii")

~/Projects/nibabel/nibabel/loadsave.py in load(filename, **kwargs)
     98         is_valid, sniff = image_klass.path_maybe_image(filename, sniff)
     99         if is_valid:
--> 100             img = image_klass.from_filename(filename, **kwargs)
    101             return img
    102 

~/Projects/nibabel/nibabel/dataobj_images.py in from_filename(klass, filename, mmap, keep_file_open)
    469             raise ValueError("mmap should be one of {True, False, 'c', 'r'}")
    470         file_map = klass.filespec_to_file_map(filename)
--> 471         return klass.from_file_map(file_map, mmap=mmap,
    472                                    keep_file_open=keep_file_open)
    473 

~/Projects/nibabel/nibabel/cifti2/cifti2.py in from_file_map(klass, file_map, mmap, keep_file_open)
   1418         # Use array proxy object where possible
   1419         dataobj = nifti_img.dataobj
-> 1420         return Cifti2Image(reshape_dataobj(dataobj, dataobj.shape[4:]),
   1421                            header=cifti_header,
   1422                            nifti_header=nifti_img.header,

~/Projects/nibabel/nibabel/cifti2/cifti2.py in __init__(self, dataobj, header, nifti_header, extra, file_map)
   1381         self.update_headers()
   1382 
-> 1383         if self._dataobj.shape != self.header.matrix.get_data_shape():
   1384             warn(f"Dataobj shape {self._dataobj.shape} does not match shape "
   1385                  f"expected from CIFTI-2 header {self.header.matrix.get_data_shape()}")

~/Projects/nibabel/nibabel/cifti2/cifti2.py in get_data_shape(self)
   1231         base_shape = [None] * (max(self.mapped_indices) + 1)
   1232         for mim in self:
-> 1233             size = len(cifti2_axes.from_index_mapping(mim))
   1234             for idx in mim.applies_to_matrix_dimension:
   1235                 base_shape[idx] = size

~/Projects/nibabel/nibabel/cifti2/cifti2_axes.py in from_index_mapping(mim)
    142                    'CIFTI_INDEX_TYPE_BRAIN_MODELS': BrainModelAxis,
    143                    'CIFTI_INDEX_TYPE_PARCELS': ParcelsAxis}
--> 144     return return_type[mim.indices_map_to_data_type].from_index_mapping(mim)
    145 
    146 

~/Projects/nibabel/nibabel/cifti2/cifti2_axes.py in from_index_mapping(cls, mim)
    415                     shape = mim.volume.volume_dimensions
    416                     affine = mim.volume.transformation_matrix_voxel_indices_ijk_to_xyz.matrix
--> 417         return cls(name, voxel, vertex, affine, shape, nvertices)
    418 
    419     def to_mapping(self, dim):

~/Projects/nibabel/nibabel/cifti2/cifti2_axes.py in __init__(self, name, voxel, vertex, affine, volume_shape, nvertices)
    310                 raise ValueError("Affine and volume shape should be defined "
    311                                  "for BrainModelAxis containing voxels")
--> 312             self.affine = np.asanyarray(affine)
    313             self.volume_shape = volume_shape
    314 

~/Projects/nibabel/nibabel/cifti2/cifti2_axes.py in affine(self, value)
    570             value = np.asanyarray(value)
    571             if value.shape != (4, 4):
--> 572                 raise ValueError('Affine transformation should be a 4x4 array')
    573         self._affine = value
    574 

ValueError: Affine transformation should be a 4x4 array
> /home/vanandrew/Projects/nibabel/nibabel/cifti2/cifti2_axes.py(572)affine()
    570             value = np.asanyarray(value)
    571             if value.shape != (4, 4):
--> 572                 raise ValueError('Affine transformation should be a 4x4 array')
    573         self._affine = value
    574 
```

The issue seems to be that [`ft_write_cifti`](https://www.fieldtriptoolbox.org/reference/ft_write_cifti/) writes the affine as a single dimension vector, rather than a 4x4 affine:

```python
ipdb> value.shape
(16,)
```

My fix adjusts the affine setter, so that it reshapes the array to 4x4 if it is a single dimension:

```python
@affine.setter
def affine(self, value):
    if value is not None:
        value = np.asanyarray(value)
        if value.shape == (16,):  # <-- If a single dim
            value = value.reshape(4, 4)  # <-- reshape to (4, 4)
        if value.shape != (4, 4):
            raise ValueError('Affine transformation should be a 4x4 array')
    self._affine = value
```
which seems to solve the issue.


effigies:
Reading through the CIFTI spec:

> **TransformationMatrixVoxelIndicesIJKtoXYZ Element**
> * *Description* – Contains a matrix that translates Voxel IJK Indices to spatial XYZ coordinates (+X=>right, +Y=>anterior, +Z=> superior). The resulting coordinate is the center of the voxel.
> * *Attributes*
>   * **MeterExponent** - Integer, specifies that the coordinate result from the transformation matrix should be multiplied by 10 to this power to get the spatial coordinates in meters (e.g., if this is “-3”, then the transformation matrix is in millimeters).
> * *Child Elements: [NA]*
> * *Text Content* - Sixteen floating-point values, in row-major order, that form a 4x4 homogeneous transformation matrix.
> * *Parent Element* – **Volume**
> 
> The transformation matrix below is encoded into the XML as “m1 m2 m3 m4 m5 m6 m7 m8 m9 m10 m11 m12 m13 m14 m15 m16”. The last row (elements m13, m14, m15, and m16) will always be [0, 0, 0, 1].
> ```
>  m1  m2  m3  m4
>  m5  m6  m7  m8
>  m9 m10 m11 m12
> m13 m14 m15 m16
> ```

Weirdly, it looks like we're expecting it to have newlines, while the standard seems to clearly indicate that it should be space-separated, so we should definitely accept those. (Looking at `wb_command`-generated dtseries, I see they contain newlines, so I guess even they aren't following the standard exactly.)

I don't think this is the correct place to fix this, as this is an API that expects affine matrices. I think the better place is in the parser:

https://github.com/nipy/nibabel/blob/b8a1039f1cb0b344153dc66e6b67052720a8c5d8/nibabel/cifti2/parse_cifti2.py#L537-L542

With something like:

```Diff
         elif self.write_to == 'TransformMatrix':
             # conversion to numpy array
             c = BytesIO(data.strip().encode('utf-8'))
             transform = self.struct_state[-1]
-            transform.matrix = np.loadtxt(c, dtype=np.float64)
+            matrix = np.loadtxt(c, dtype=np.float64)
+            transform.matrix = matrix.reshape(4, 4)
             c.close()
```

vanandrew:
Just tested your fix, and it seems to work on the file I'm having trouble with. I'll update my request in a bit.

effigies:
Thanks! Would you be able to add a small test? Either could bundle a small (<4KB) file in `nibabel/tests/data`, a larger file in `nibabel-data/nitest-cifti2`, or do something like:

```Python
def test_space_separated_affine():
    img = # Some Cifti2Image with the second axis being a brain model
    bstr = img.to_bytes()
    # Replace newlines in affine matrix with spaces
    affine_bstr = img.header.get_index_map(1).volume.to_xml()
    spacesep_bstr = bstr.replace(affine_bstr, affine_bstr.replace(b"\n", b" "))
    # Verify still parseable
    new_img = nb.Cifti2Image.from_bytes(spacesep_bstr)
```

vanandrew:
Wasn't sure where to put the test, so I added it under `nibabel/cifti2/tests/test_cifti2.py`

effigies:
It looks like `nibabel/cifti2/tests/test_cifti2io_header.py` does some other test loads, so that might be a better spot.

vanandrew:
Alright. Moved it over to `nibabel/cifti2/tests/test_cifti2io_header.py`

effigies (approved):
Perfect. Thanks!

PR #1066, opened 2021-11-19
ENH: Add MZ3 format

effigies:
Currently implements reading the header and "scalar" array. Haven't tested if we need to switch ordering to handle multiple scalar arrays. Does not yet implement reading vertices, faces or rgba.

Two significant wrinkles to this format:

1) Gzipped and uncompressed versions have the same extension.
2) The data array size must be calculated from the file size. For the compressed case, I first try to use the `member-size mod 2**32` field in the gzip footer to detect the size. If we hit 20GB we bet that it's a weird file and just seek to the end of the gzip stream.

Loading face/header data will be a good test for BIAP 0009 (#1056), but don't want to clog up that PR at the moment.

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -86,7 +86,8 @@ class ArrayProxy(object):
     # Assume Fortran array memory layout
     order = 'F'
 
-    def __init__(self, file_like, spec, *, mmap=True, keep_file_open=None):
+    def __init__(self, file_like, spec, *,
+                 mmap=True, keep_file_open=None, compression=None):
         """Initialize array proxy instance
 
         Parameters
@@ -125,6 +126,7 @@ class ArrayProxy(object):
             If ``file_like`` is an open file handle, this setting has no
             effect. The default value (``None``) will result in the value of
             ``KEEP_FILE_OPEN_DEFAULT`` being used.
+        compression : { None, "gz", "bz2", "zst" }, optional, keyworld only
         """
         if mmap not in (True, False, 'c', 'r'):
             raise ValueError("mmap should be one of {True, False, 'c', 'r'}")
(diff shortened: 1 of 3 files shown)


neurolabusc:
@effigies thanks for supporting this. The [mz3 specification is here](https://github.com/neurolabusc/surf-ice/tree/master/mz3). 
 - It is true that a single extension is used for both compressed and uncompressed data. You can discriminate these by looking at the first two bytes of the file. Specifically, all GZip files begin with hex = 0x1F8B, all uncompressed MZ3 files begin with hex = 0x4D5A
 - You do not need to know the file size to compute the array size. The header `NFACE` and `NVERT` explicitly describe the number of indexed triangles (faces) and spatial points (vertices).

Happy to work directly with you to tidy this up.

nibabel/openers.py
```diff
@@ -153,13 +154,13 @@ class Opener:
 
     fobj: io.IOBase
 
-    def __init__(self, fileish: str | io.IOBase, *args, **kwargs):
+    def __init__(self, fileish: str | io.IOBase, *args, compression: str | None, **kwargs):
```
effigies, line 157:
```suggestion
    def __init__(self, fileish: str | io.IOBase, *args, compression: str | None = None, **kwargs):
```

nibabel/arrayproxy.py
```diff
@@ -119,10 +119,10 @@ class ArrayProxy(ArrayLike):
     See :mod:`nibabel.minc1`, :mod:`nibabel.ecat` and :mod:`nibabel.parrec` for
     examples.
     """
-
     _default_order = 'F'
 
-    def __init__(self, file_like, spec, *, mmap=True, order=None, keep_file_open=None):
+    def __init__(self, file_like, spec, *,
+                 mmap=True, keep_file_open=None, compression=None):
```
effigies, line 125:
```suggestion
                 mmap=True, order=None, keep_file_open=None, compression=None):
```

Issue #1071, opened 2022-01-19
Gzipped gifti files I/O


alexisthual:
Loading `.gii.gz` files will crash with `nibabel.filebasedimages.ImageFileError: Cannot work out file type of "/path/to/file"`.

Here is a snippet to reproduce:

```python
import nibabel as nib
from nilearn import datasets

fs5 = datasets.fetch_surf_fsaverage()
nib.load(fs5.pial_left)
```

matthew-brett:
For someone passing by, the key line causing the error boils down to:

```python
nib.load('/some/path/pial_left.gii.gz')
```

This is because we don't recognize `.gii.gz` files as Gifti files.   See PR #1072

PR #1072, opened 2022-01-19
MRG: Allow compressed Gifti images

matthew-brett:
Allow .gz and .bz2 compressed Gifti images.

Closes gh-1071


effigies:
I think we left these off because they weren't really seen in practice, and the GIFTI approach to compression is to compress data arrays, not the entire file.

If they exist in the wild, we should support them.

nibabel/gifti/tests/test_parse_gifti_fast.py
```diff
@@ -462,3 +462,11 @@ def test_external_file_failure_cases():
     parser = GiftiImageParser()
     with pytest.raises(GiftiParseError):
         img = parser.parse(xmldata)
+
+
+def test_load_compressed():
+    for ext in ('', '.gz', '.bz2'):
+        fn = pjoin(IO_DATA_PATH, 'external.gii' + ext)
+        img7 = load(fn)
```
effigies, line 470: I think you need to create `external.gii.gz` and `external.gii.bz2` in https://github.com/nipy/nibabel/tree/master/nibabel/gifti/tests/data.

neurolabusc:
The [GIfTI specification](https://www.nitrc.org/projects/gifti/) does not describe whole file compression (e.g. `.gii.gz`). Rather, only the  `<Data>` elements are compressed, while the XML header remains uncompressed.

As a concrete example, a compressed GIfTI might describe compressed triangle indices like this:

```
<DataArray  ArrayIndexingOrder="RowMajorOrder"
   DataType="NIFTI_TYPE_INT32"
   Dim0="34438"
   Dim1="3"
   Dimensionality="2"
   Encoding="GZipBase64Binary"
   Endian="LittleEndian"
   ExternalFileName=""
   ExternalFileOffset=""
   Intent="NIFTI_INTENT_TRIANGLE">
<MetaData>
```

In contrast, the uncompressed data would change the encoding tag to read `Encoding="Base64Binary"`.

I can still see the benefit in this PR: the GIfTI requirement that the compressed payload is stored in Base64 makes the resulting stream [33% larger on disk](https://developer.mozilla.org/en-US/docs/Glossary/Base64). Gzip will be able to compress base64 effectively.

As an aside, any discussion of GIfTI compression should include the caveat that GIfTI implementations (including those provided with the specification) that include data compressed with `GZipBase64Binary` must only write/read a DEFLATE-compressed payload, without the [Gzip header](https://en.wikipedia.org/wiki/Gzip).

effigies:
@matthew-brett Rebase?

matthew-brett:
Hmm - changed the test - to what seemed to be the equivalent, but it fails, because, contrrary to what the test appeared to be asserting, there is a `ResourceWarning`, but it's this:

```
'{message : ResourceWarning("unclosed file <_io.BufferedReader name=\'/Volumes/zorg/mb312/dev_trees/nibabel/nibabel/gifti/tests/data/ascii.gii\'>"), category : \'ResourceWarning\', filename : \'/Volumes/zorg/mb312/dev_trees/nibabel/nibabel/gifti/gifti.py\', lineno : 913, line : None}'
```

Indeed, I do get this warning running a simple reproducer script from the command line:

```python
# to_gii.py
import nibabel as nib
img = nib.load('nibabel/gifti/tests/data/ascii.gii')
img.to_filename('test.gii')
```

```
$ python -W error to_gii.py 
Exception ignored in: <_io.FileIO name='nibabel/gifti/tests/data/ascii.gii' mode='rb' closefd=True>
  File "/Volumes/zorg/mb312/dev_trees/nibabel/nibabel/gifti/gifti.py", line 913, in from_file_map
    parser.parse(fptr=file_map['image'].get_prepare_fileobj('rb'))
ResourceWarning: unclosed file <_io.BufferedReader name='nibabel/gifti/tests/data/ascii.gii'>
```

I'm not sure why my new code doesn't give the result that the previous code seems to expect.  Any ideas?

effigies:
The test was testing writing, but picked up an unclosed handle during reading. I think we need this patch:

```diff
diff --git a/nibabel/gifti/gifti.py b/nibabel/gifti/gifti.py
index 4e205eb2..b704bfc5 100644
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -909,7 +909,8 @@ class GiftiImage(xml.XmlSerializable, SerializableImage):
         img : GiftiImage
         """
         parser = klass.parser(buffer_size=buffer_size, mmap=mmap)
-        parser.parse(fptr=file_map['image'].get_prepare_fileobj('rb'))
+        with file_map['image'].get_prepare_fileobj('rb') as fptr:
+            parser.parse(fptr=fptr)
         return parser.img
 
     @classmethod
```

matthew-brett:
Ah - thank you - applied.

effigies:
@neurolabusc 

> The [GIfTI specification](https://www.nitrc.org/projects/gifti/) does not describe whole file compression (e.g. `.gii.gz`). Rather, only the `<Data>` elements are compressed, while the XML header remains uncompressed.

Agreed that this is counter-to-spec, but if they exist in the real world, I think we need to support reading them at least.

> As an aside, any discussion of GIfTI compression should include the caveat that GIfTI implementations (including those provided with the specification) that include data compressed with `GZipBase64Binary` must only write/read a DEFLATE-compressed payload, without the [Gzip header](https://en.wikipedia.org/wiki/Gzip).

It looks like we use `zlib.compress` and `zlib.decompress`, which adds a 2 byte header (`x\x9c`) and 4 byte CRC, but does not add a gzip header. I believe we are conformant here, and would error if we found a data array with a gzip header.

Just to check, are there issues with our implementation that need to be addressed? And do you have concerns with merging this as-is?

effigies:
Will go ahead and merge. Any further issues can be addressed separately.

neurolabusc:
@effigies sorry for not replying. My issue is with the specification, which does not describe the sample datasets nor the reference implementations. Years ago, I asked that the specification be updated to reflect the observed usage. My sense is that GIfTI and CIfTI specifications did not benefit from the community engagement and refinement we saw with the NIfTI specification.

PR #1073, opened 2022-01-20
MNT: Drop uses of deprecated distutils

effigies:
CI is failing. None of these cases seemed to hinge on the exact semantics of `LooseVersion`. In most cases, `packaging.version.Version` works as well. For CIFTI-2, the spec says the string must be "2", so let's not parse that unless tests show we've been explicitly trying to be flexible there.

### DIFF
--- a/nibabel/cifti2/parse_cifti2.py
+++ b/nibabel/cifti2/parse_cifti2.py
@@ -7,10 +7,10 @@
 #
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
-from distutils.version import LooseVersion
-
 import numpy as np
 from io import BytesIO
+
+from packaging.version import Version, parse
 
 from .cifti2 import (Cifti2MetaData, Cifti2Header, Cifti2Label,
                      Cifti2LabelTable, Cifti2VertexIndices,
@@ -151,9 +151,9 @@ class Cifti2Parser(xml.XmlParser):
         if name == 'CIFTI':
             # create cifti2 image
             self.header = Cifti2Header()
-            self.header.version = attrs['Version']
-            if LooseVersion(self.header.version) < LooseVersion('2'):
-                raise ValueError('Only CIFTI-2 files are supported')
+            self.header.version = ver = attrs['Version']
+            if parse(ver) < Version("2"):
+                raise ValueError(f'Only CIFTI-2 files are supported; found version {ver}')
             self.fsm_state.append('CIFTI')
             self.struct_state.append(self.header)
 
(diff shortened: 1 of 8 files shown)


nibabel/cifti2/parse_cifti2.py
```diff
@@ -152,7 +150,7 @@ def StartElementHandler(self, name, attrs):
             # create cifti2 image
             self.header = Cifti2Header()
             self.header.version = attrs['Version']
-            if LooseVersion(self.header.version) < LooseVersion('2'):
+            if self.header.version not in ("2", "2.0"):
```
matthew-brett, line 153: Is that right - that these are the only possible spellings for the version >= 2?   What about versions > 2?  Why not use `Version('2')` here instead?

matthew-brett (commented):
Harmonize Cifti with other Version calls?

nibabel/cifti2/parse_cifti2.py
effigies: Can restore rough equivalence, just unclear why to expect future versions to be parseable by the same parser.

matthew-brett: Do we have some reason to think that Cifti versions won't be in standard version format?   Or, put another way, what's the advantage of restricting ourselves to the strings here?

effigies:
> Do we have some reason to think that Cifti versions won't be in standard version format? Or, put another way, what's the advantage of restricting ourselves to the strings here?

The spec requires it to be "2":

> ### CIFTI Element
> * Description – The root of the CIFTI XML.
> * *Attributes*
>    * Version – Indicates version of the CIFTI XML. For CIFTI-2, this must have the value “2”. If the value is “1” or “1.0”, this document does not apply, and instead the file should follow the CIFTI-1 specification: http://www.nitrc.org/plugins/mwiki/index.php/cifti:MainPage. It should be noted that CIFTI-2 compliant software is not required to be able to read CIFTI-1.

Our tests explicitly check a file with "2.0" in it, and it seems reasonable to continue with that. It's unclear whether it's a good idea to support undefined version declarations like "2.1" or "3.0", but at least it's not a regression to permit them. I'm a little inclined to emit a warning like "Unknown CIFTI version {ver}; will attempt to parse as CIFTI-2".

I switched to `packaging.version.parse`. If that fails to produce a `Version()` object, it will produce a `LegacyVersion()` that always compares less than any `Version()`, so this check should remain flexible.

nibabel/cifti2/parse_cifti2.py
matthew-brett: If you think it's likely that Cifti will go to some other version convention, we can just stick with the string comparison and run an error?

Issue #1046, opened 2021-08-19
Helper function for saving compact integer datatypes


neurolabusc:
I aided a nibabel user who wrote a function to segment a brain into 4 tissue types (0=non brain, 1= gray matter, 2= white matter, 3= CSF). They were surprised at the huge disk space resulting from the NIfTI images, which were saved as [DT_INT64](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h) datatype. Beyond being large, this is a very unusual datatype for NIfTI (it did not exist in the ancestor Analyze type).  

The native NumPy datatype for integers is int64. So nibabel is preserving this datatype. I do wonder if the [Nifti1Image](https://nipy.org/nibabel/reference/nibabel.nifti1.html#nibabel.nifti1.Nifti1Image) command could include an optional parameter that chooses the most compact data type if the input data is of an integer type (e.g. in this example, the voxel intensities range from 0..47, so it could be saved losslessly as DT_UINT8, requiring 1/8th the disk space.

When I look at the example [nibabel examples](https://nipy.org/nibabel/gettingstarted.html) they all show sensible use setting the dtype
```
data = np.ones((32, 32, 15, 100), dtype=np.int16)
```
However, the rational for these is never described. Perhaps the documentation could not that choice of datatype impacts file size and some tools only support a limited range of these datatypes (the classic Analyze data types are UINT8, INT16, FLOAT32 and some tools can act in unexpected ways with other datatypes, e.g. AFNI promotes UINT16 to FLOAT32).

```
import nibabel as nib
import numpy as np

data = np.arange(4*4*3).reshape(4,4,3)
#data.dtype = 'int64'
#data values are integers 0..47
img = nib.Nifti1Image(data, affine=np.eye(4))
nib.save(img, 'test.nii')

data8 = np.arange(4*4*3, dtype=np.uint8).reshape(4,4,3)
img8 = nib.Nifti1Image(data8, affine=np.eye(4))
nib.save(img8, 'test8.nii')

data32 = np.arange(4*4*3, dtype=np.float32).reshape(4,4,3)
img32 = nib.Nifti1Image(data32, affine=np.eye(4))
nib.save(img32, 'test32.nii')
```

effigies:
Thinking a little about this, I think we can do something like:

```Python
def get_min_dtype(img):
    arr = np.asanyarray(img.dataobj)
    min_dt = arr.dtype
    if np.issubdtype(min_dt, np.integer):
[...]

PR #1082, opened 2022-02-11
NF: Pass dtype to Analyze-like images at initialization/serialization, warn on creation of NIfTI images with 64-bit ints (API change)

effigies:
No test yet, but a quick prototype based on the discussion in #1046.

I'm not sure what people's intuition about the behavior of a `dtype` parameter should be. What I've done for now is just to `set_data_dtype()` so that the array is untouched but will be coerced at write time. However, I could see an argument that the expectation is that the array would be immediately cast so that the result of `np.asanyarray(img.dataobj)` will have the dtype passed to the image.

Additionally, this currently assumes that passing a header is as good as passing a dtype, but this is going to have the problem that passing the header of a 64-bit image will succeed silently.

I have not made any attempt to implement a `from nibabel.__future__ import int64_error` at this point. Figured I'd get some code out and argue out the semantics before refining.

### DIFF
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -992,7 +992,7 @@ class AnalyzeImage(SpatialImage):
         """
         return file_map['header'], file_map['image']
 
-    def to_file_map(self, file_map=None):
+    def to_file_map(self, file_map=None, dtype=None):
         """ Write image to `file_map` or contained ``self.file_map``
 
         Parameters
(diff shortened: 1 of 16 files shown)


matthew-brett:
I think the logic is right - that the `dtype` parameter overrides whatever is in the header, if present.  I am personally OK with not warning about int64 in the header - if the header got to here, the user must be OK with int64 images.  After this change, they will be difficult to make by accident.

matthew-brett:
Thinking further about the future import - I think we'll have to make that a module global in fact, as in:

```
# Using it
nib.__future__.int64_error = True
```

effigies:
Dropped the `__future__` config, since we're just going to do a rapid warn/error cycle.

effigies:
This is ready for review. This PR aims to establish the API and semantics for explicit dtypes. I have not complicated this PR by trying to add in keywords like "compat" or "mask"; I think that will be cleaner as a separate PR that builds on what we end up with here.

@matthew-brett @jbteves @jeromedockes

jeromedockes:
I don't know nibabel well enough to do an actual review but this seems to be in line with what was decided during the nilearn office hours.

effigies:
@jeromedockes Sorry, yes, in pinging you, I mostly meant "Does this API work for nilearn's purposes?"

jeromedockes:
No problem, on the contrary thanks for pinging me! yes I do think that API will work for nilearn. We haven't really made progress because the nilearn developers team is at a somewhat reduced capacity at the moment but we should get to it soon. if you do open the follow-up PR adding some keywords please ping us there too as we might take advantage of these as well.

effigies:
@matthew-brett Could I bug you for a review?

effigies:
@matthew-brett Any chance of a review? I think this one should be relatively straightforward while #1082 will take more careful thought.

effigies:
Going to merge. Can fix up later if something is found. Realizing we didn't get CIFTI fully supported on the creation side, but that's turning out to be complicated enough to be worth its own PR.

PR #1083, opened 2022-02-11
ENH: Allow save-time passing of on-disk dtype to save functions

effigies:
A common idiom is to use a higher-precision internal data type before saving to a lower-precision on-disk data type. The `set_data_dtype()` method is a little clunky. This permits the following:

```Python
nii.to_filename(fname, dtype="int16")
nii.to_bytes(dtype="int16")
nb.save(nii, fname, dtype="int16")
```

Only implemented for Analyze-like images here. Related to discussion in #1046, but does not (currently) warn for int64 data being saved without a dtype argument.

### DIFF
--- a/nibabel/analyze.py
+++ b/nibabel/analyze.py
@@ -989,7 +989,7 @@ class AnalyzeImage(SpatialImage):
         """
         return file_map['header'], file_map['image']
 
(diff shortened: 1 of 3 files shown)


effigies:
Combining into #1082.

PR #1091, opened 2022-02-23
RF: Unify Caret-XML-style metadata structure as dict-like

effigies:
We have two implementations of the same structure in GIFTI and CIFTI-2. The Cifti one is much nicer to deal with, but also a little short of a straight drop-in replacement for `dict()`.

Spurred by writing a parser for `CaretSpecFile`s in #1090. Mostly used the Cifti implementation, but dropped the `difference_update()` (not sure why this exists, so don't want to copy it without a use case). Used a private `_data` object to allow the `GiftiMetaData.data` to be a deprecated list view.

Scheduled dropping most of the custom bits of each subclass for 6.0 to give a good long warning period.

### DIFF
--- a/nibabel/caret.py
+++ b/nibabel/caret.py
@@ -1,0 +1,117 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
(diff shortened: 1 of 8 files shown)


jbteves (approved):
LGTM with the caveat that I'm not sure what the approval checklist for nibabel looks like 😉

htwangtw (approved):
LGTM! A more parsable gifti metadata makes me really happy.

effigies:
Reviewing guidelines: https://nipy.org/nibabel/devel/core_developer.html#reviewing

PR #1093, opened 2022-02-23
FIX: Resize ArraySequence.data without helper function to avoid reference increment

### LINKED ISSUE #995: Improving performance of TckFile when reading huge MRtrix tck files
Hi,

  First off, thank you all for the great toolkit!  Wanted to point out a performance issue and post a possible solution for others to leverage.

  Using the default `.tck` reader (`TckFile`), 500,000 streamlines takes approximately 22 minutes to read.  Profiling the code reveals most of the reading time is spent on [these lines of array_sequence.py](https://github.com/nipy/nibabel/blob/95e1fbece92b92fc5a5f5c107bd0222438bdb0cc/nibabel/streamlines/array_sequence.py#L237-L238).  Not quite sure what's going on, but reading one streamline at a time thrashes with huge numbers of calls to `numpy.copy`.  Using the code below, the same file is loaded in around 15 seconds.  

  This version made reading faster by reading 512Mb of the 3.5Gb file at a time, then loading all the streamlines into memory at once.  `ArraySequence` performed very well due to [this code](https://github.com/nipy/nibabel/blob/95e1fbece92b92fc5a5f5c107bd0222438bdb0cc/nibabel/streamlines/array_sequence.py#L302-L313).

```python

def load_tck(filename, buffer_size=512):
    """Load a TCK file with a _much_ larger buffer, 512Mb"""
    header = nib.streamlines.tck.TckFile._read_header(filename)
    reader = nib.streamlines.tck.TckFile._read(filename, header, buffer_size=buffer_size)

    start_time = time.time()
    e = [x for x in reader]
    logging.debug(f"read streamline file {filename} in {time.time() - start_time} seconds")
    streamlines = ArraySequence(e)
    tractogram = Tractogram(streamlines)

    # By definition.
    tractogram.affine_to_rasmm = np.eye(4)
    header[Field.VOXEL_TO_RASMM] = np.eye(4)

    return TckFile(tractogram, header=header)
```

MarcCote:
Fixes #995 

Passing `self._data` to `_safe_resize` makes an unnecessary reference to that ndarray, thus causing a ValueError when trying to resize it in-place.

### DIFF
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -281,7 +281,7 @@ class ArraySequence(object):
                 self._data.resize(new_shape)
             except ValueError:
                 self._data = self._data.copy()
-                self._data.resize(new_shape)
+                self._data.resize(new_shape, refcheck=False)
 
     def shrink_data(self):
         self._data.resize((self._get_next_offset(),) + self.common_shape,
--- a/nibabel/streamlines/array_sequence.py
+++ b/nibabel/streamlines/array_sequence.py
@@ -24,16 +24,6 @@ def is_ndarray_of_int_or_bool(obj):
             np.issubdtype(obj.dtype, np.bool_)))
 
 
-def _safe_resize(a, shape):
-    """ Resize an ndarray safely, using minimal memory """
-    try:
-        a.resize(shape)
-    except ValueError:
-        a = a.copy()
-        a.resize(shape, refcheck=False)
-    return a
-
-
 class _BuildCache(object):
     def __init__(self, arr_seq, common_shape, dtype):
         self.offsets = list(arr_seq._offsets)
@@ -287,7 +277,11 @@ class ArraySequence(object):
         if self._data.size == 0:
             self._data = np.empty(new_shape, dtype=build_cache.dtype)
         else:
-            self._data = _safe_resize(self._data, new_shape)
+            try:
+                self._data.resize(new_shape)
+            except ValueError:
+                self._data = self._data.copy()
+                self._data.resize(new_shape)
 
     def shrink_data(self):
         self._data.resize((self._get_next_offset(),) + self.common_shape,


nibabel/streamlines/array_sequence.py
```diff
@@ -287,7 +277,11 @@ def _resize_data_to(self, n_rows, build_cache):
         if self._data.size == 0:
             self._data = np.empty(new_shape, dtype=build_cache.dtype)
         else:
-            self._data = _safe_resize(self._data, new_shape)
+            try:
+                self._data.resize(new_shape)
+            except ValueError:
+                self._data = self._data.copy()
+                self._data.resize(new_shape)
```
effigies, line 284:
```suggestion
                self._data.resize(new_shape, refcheck=False)
```

MarcCote: Actually, do we need `refcheck=False` ? In theory, we just did a copy on the line before. Maybe we can replace those two lines with `self._data = np.resize(self._data, new_shape)`?.

MarcCote: Oh. Just saw the errors on Travis (but some of them passed?!).

effigies: I believe the reason we were doing this is because of the zero-fill behavior of `ndarray.resize`:

```
numpy.resize = resize(a, new_shape)
    Return a new array with the specified shape.
    
    If the new array is larger than the original array, then the new
    array is filled with repeated copies of `a`.  Note that this behavior
    is different from a.resize(new_shape) which fills with zeros instead
    of repeated copies of `a`.
```

MarcCote: Ah right. Ok. I'll add the refcheck=False, seems for some systems it is needed. I believe it is safe to do so, since we are doing a copy just before (so in theory, there should be no references).

effigies:
Passing. Given that our tests are fairly frequently hitting the `except` case, I'm not sure if this is going to universally speed things up, but anything's better than nothing.

Okay to merge?

MarcCote:
That works for me. I do observe a great speedup on my laptop (Python 3.9, numpy 1.22). There are some tests that explicitly trigger that except, so that's expected.

effigies:
Thanks!

PR #1094, opened 2022-02-25
ENH: Add rounding to ``voxel_sizes()``

oesteban:
Adds a convenience argument to round the voxel sizes calculated from affines.

### DIFF
--- a/nibabel/affines.py
+++ b/nibabel/affines.py
@@ -269,7 +269,7 @@ def dot_reduce(*args):
     return reduce(lambda x, y: np.dot(y, x), args[::-1])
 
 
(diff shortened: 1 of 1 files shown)


matthew-brett:
This is a minor back-compat break - the values will change at the 6th decimal from this code, compared to the previous code.   I suppose this could change a few affines in later code and cause some confusion where resliced images are not exactly the same.

And - why put the argument in the function call?  Can't the user just do it themselves, more transparently?

```
v = img.voxel_sizes().round(6)
```

oesteban:
I see the point. thanks

PR #1097, opened 2022-03-19
MNT: Update submodule URLs to use https

effigies:
Should fix failing CI.


effigies:
Seems to have done the trick. Every thing is getting to the test stage.

PR #1103, opened 2022-05-07
DOC: Update documentation for InTemporaryDirectory to mark it as not thread-safe

### LINKED ISSUE #1102: Thread safety concerns with reliance on nibabel.tmpdirs.InTemporaryDirectory
I opened a similar ticket against [`dipy`](https://github.com/dipy/dipy/issues/2587) but figured it would be useful to open here as well as nibabel is the root cause...

Reading through the nibabel source code for [`InTemporaryDirectory`](https://github.com/nipy/nibabel/blob/a8382384ca24564719ca9bc03488226733c3f2ec/nibabel/tmpdirs.py) its doing some quite sketchy things like:

```python
self._pwd = os.getcwd()
os.chdir(self.name)
```

To support programs that are mutlithreaded -- as a lot of API server implementations are in the Python universe -- I suggest the implementation be revisited so as to not accidentally introduce race conditions into the code of consuming applications.

I can work on a reproduction but logically since Python threads share a single process and `os.chdir` operates at the process level, conceptually I think its obvious nibabel is making some bad assumptions

jacobdr:
Closes #1102

### DIFF
--- a/nibabel/tmpdirs.py
+++ b/nibabel/tmpdirs.py
@@ -51,8 +51,11 @@ class TemporaryDirectory(object):
 class InTemporaryDirectory(TemporaryDirectory):
     """ Create, return, and change directory to a temporary directory
 
-    NOTE: This implementation is not thread-safe and should be used with
-    caution by external consumers of this library.
+    Notes
+    ------
+    As its name suggests, the class temporarily changes the working
+    directory of the Python process, and this is not thread-safe.  We suggest
+    using it only for tests.
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


nibabel/tmpdirs.py
```diff
@@ -51,6 +51,9 @@ def __exit__(self, exc, value, tb):
 class InTemporaryDirectory(TemporaryDirectory):
     """ Create, return, and change directory to a temporary directory
 
+    NOTE: This implementation is not thread-safe and should be used with
+    caution by external consumers of this library.
+
```
matthew-brett, line 56: How about:

```suggestion
    Notes
    ------
    As its name suggests, the class temporarily changes the working
    directory of the Python process, and this is not thread-safe.  We suggest
    using it only for tests.

```

matthew-brett:
Thanks - looks fine.

PR #1105, opened 2022-05-07
RF: Cleanup Makefile to bring parity with CI, remove tox + nose references

jacobdr:
* Align on venv as virtual environment convention
* Update makefile test target to re-use same script used in CI to orchestrate testing
* Remove refererences to tox and nose
* Fix dev-requirements.txt and doc-requirements.txt to pin to versions that are non-breaking


Makefile
```diff
@@ -252,11 +252,13 @@ sdist-venv: clean
 	rm -rf dist venv
```
jacobdr, line 252: I wasn't sure if this target is still being used anywhere -- so I kind of hacked it to install the dev dependencies to get it to pass on my local

doc-requirements.txt
```diff
@@ -1,6 +1,8 @@
 # Requirements for building docs
 -r requirements.txt
 sphinx<3
-numpydoc
+jinja2<3
```
jacobdr, line 4: I needed to pin these to get version compatibility to work. In the future might be better to use a tool like poetry so that we can generate a lockfile and not have to deal with this sort of not-fun dependency pinning

setup.cfg
```diff
@@ -96,10 +96,17 @@ nibabel =
 max-line-length = 100
 ignore = D100,D101,D102,D103,D104,D105,D200,D201,D202,D204,D205,D208,D209,D210,D300,D301,D400,D401,D403,E24,E121,E123,E126,E226,E266,E402,E704,E731,F821,I100,I101,I201,N802,N803,N804,N806,W503,W504,W605
 exclude =
+    venv
+    build
     *test*
     *sphinx*
     nibabel/externals/*
     */__init__.py
+    # temporary -- should be removed in another PR
```
jacobdr, line 105: Need a follow-up PR to clean up the remaining issues. I added these as ignores so that local flake8 passes. These were being covered up by the configuration of pep8speaks [only looking at the diff](https://github.com/nipy/nibabel/blob/master/.pep8speaks.yml#L2)

tools/ci/activate.sh
```diff
@@ -1,10 +1,10 @@
-if [ -e virtenv/bin/activate ]; then
-    source virtenv/bin/activate
-elif [ -e virtenv/Scripts/activate ]; then
-    source virtenv/Scripts/activate
+if [ -e venv/bin/activate ]; then
```
jacobdr, line 1: This change was made so that we are consistent in local dev and remote dev with the virtualenv

doc-requirements.txt
jacobdr: I needed this to get my local to pass but it seems like CI might not like it.... I'll try reverting to see if if this is what is breaking the pre-release CI jobs

Makefile
```diff
@@ -83,22 +82,23 @@ $(WWW_DIR):
 # Tests
 #
 
-test: unittest testmanual
+test: test-style unittest testmanual
 
-
-ut-%: build
```
jacobdr, line 89: I removed this as I did not see it being used anywhere. I can revert this if a maintainer feels it should be brought back

jacobdr:
@effigies all the pre-release tests failed for the same reason -- a string representation in a doctest for `netcdf_file`. I don't immediately see how my changes would have introduced this. Looking at this now, but any thoughts as to root cause?

jacobdr:
Trying to narrow differences... since `mmap` is coming from the standard lib I suspected a change in Python versions and how `repr(mmap_obj)` or `str(mmap_obj)` is displayed... Its possible, though still smells to me so needs a sanity check. 

As evidence, though, comparing against a prior (chose #1096 as the most recent unmerged green build I saw....)

**Python version comparison**

  * [3.10.4 this MR](https://github.com/nipy/nibabel/runs/6336648709?check_suite_focus=true#step:3:12) vs [3.10.2 prior success](https://github.com/nipy/nibabel/runs/5739141260?check_suite_focus=true#step:3:12)
  * [3.9.12 this MR](https://github.com/nipy/nibabel/runs/6336648414?check_suite_focus=true#step:3:12) vs [3.9.10 prior success](https://github.com/nipy/nibabel/runs/5739141130?check_suite_focus=true#step:3:12)
  * the [scheduled cron job against master](https://github.com/nipy/nibabel/actions/runs/2288182520) is failing for the same reason.

Makefile
```diff
@@ -83,22 +82,23 @@ $(WWW_DIR):
...
-	@cd doc/source && PYTHONPATH=../..:$(PYTHONPATH) $(NOSETESTS) --with-doctest --doctest-extension=.rst . dicom
+	export CHECK_TYPE=doc; ./tools/ci/check.sh
 
+coverage: unittest
 
-coverage: build
-	@PYTHONPATH=.:$(PYTHONPATH) $(NOSETESTS) --with-coverage --cover-package=nibabel
+.PHONY: test-style
+test-style:
+	export CHECK_TYPE=style; ./tools/ci/check.sh
```
effigies, line 97: I would rather put `make style` in `check.sh` than call `check.sh` from here. And likewise with other tests.

```suggestion
.PHONY: style
style:
	flake8
```

Issue #1108, opened 2022-05-08
Pre-release job is failing against master


jacobdr:
[most recent failing job as of this writing May 7](https://github.com/nipy/nibabel/actions/runs/2288182520)

Trying to narrow differences... since `mmap` is coming from the standard lib I suspected a change in Python versions and how `repr(mmap_obj)` or `str(mmap_obj)` is displayed... Its possible, though still smells to me so needs a sanity check. 

As evidence, though, comparing against a prior (chose #1096 as the most recent unmerged green build I saw....)

**Python version comparison**

  * [3.10.4 this MR](https://github.com/nipy/nibabel/runs/6336648709?check_suite_focus=true#step:3:12) vs [3.10.2 prior success](https://github.com/nipy/nibabel/runs/5739141260?check_suite_focus=true#step:3:12)
  * [3.9.12 this MR](https://github.com/nipy/nibabel/runs/6336648414?check_suite_focus=true#step:3:12) vs [3.9.10 prior success](https://github.com/nipy/nibabel/runs/5739141130?check_suite_focus=true#step:3:12)
  * the [scheduled pre-release cron job against master](https://github.com/nipy/nibabel/actions/runs/2288182520) is failing for the same reason.

Final note: the only pre-release job that passes is the one that does not have the PRE_PIP_FLAGS environment variable set....

_Originally posted by @jacobdr in https://github.com/nipy/nibabel/pull/1105#issuecomment-1120350450_

effigies:
Manually bisecting the pre-release wheels of numpy (https://pypi.anaconda.org/scipy-wheels-nightly/simple/numpy/) I can say that this arose between numpy-1.23.0.dev0+1065.gf996a2b62 and numpy-1.23.0.dev0+1087.g0eaa40db3.

Looking at https://github.com/numpy/numpy/compare/f996a2b62...0eaa40db3, I see they merged https://github.com/numpy/numpy/pull/21324 in that interval, which seems likely to be the cause. It does not look like it actually performs a copy, so that's probably okay. It looks like the reason the test was written this way:

    NetCDF files, when opened read-only, return arrays that refer
    directly to memory-mapped data on disk:

        >>> data = time[:]
        >>> data.base.base  # doctest: +ELLIPSIS
        <mmap.mmap ...>

is because the object is not actually a `numpy.memmap` but a custom `mmap.mmap`-backed ndarray.

[...]

PR #1110, opened 2022-05-23
MNT: Update netcdf to latest scipy

effigies:
Seems numpy's release has forced our hand. No longer clear that tests will ensure we're getting `mmap`s. I'll open an issue with scipy.

Fixes https://github.com/nipy/nibabel/issues/1108.


effigies:
Issue opened FWIW: https://github.com/scipy/scipy/issues/16285

PR #1111, opened 2022-05-26
ENH: Add dtype argument to Cifti2Image

effigies:
To bring into line with other Analyze-like formats.

Given that the subject that brought us here was inappropriate use of image dtypes, I am also constraining the dtypes to those listed in the standard:

> The datatype field in the NIfTI header can be any of float, double, and signed or unsigned int8,
int16, int32, or int64, with bitpix set correspondingly, per the NIfTI standard. Composite
datatypes, such as complex or RGB, and bitwise storage are not supported in CIFTI. Multiple
scalar maps, or separate files, can be used instead of composite datatypes.

I also dropped float128, which is not specifically restricted, but nor is it specifically permitted. Seems unlikely to be a good idea.


effigies:
Tests forthcoming.

Issue #1089, opened 2022-02-18
Intent to deprecate: 64-bit integer NIfTI images


effigies:
64 bit integer images are easy to generate in nibabel due to numpy semantics, but almost universally incompatible with non-Python tools. We must balance breaking existing Python scripts, the Python library ecosystem, and the wider neuroimaging ecosystem that uses NIfTI as an interchange format.

Recognizing that we are currently contributing to breakage in the wider ecosystem, we have decided to deprecate 64-bit integer image creation except via explicit override. The following path is proposed in order to bring major Python libraries into alignment over a 6-month window through escalating warnings and errors.

NiBabel will take the following steps:

* In NiBabel 4.0.0 (March 2022):
  * Provide `dtype=` parameter to `Nifti1Image()` (and similar) to pass to `set_data_dtype()` after initialization (#1082)
  * Warn on `Nifti1Image(data, affine)` if data.dtype is int64 or uint64 unless dtype or header is provided (#1082)
  * Provide `nib-cast` (name may change) command-line tool to recast existing files to new data type (#1113)
  * Provide `dtype=` parameter to `Nifti1Image.to_filename()` (and similar) to `set_data_dtype()` before serializing 
    * If `img.get_data_dtype()` is (u)int64 and `dtype` is not passed, raise a warning, to be promoted to error in 5.0
    * Provide "mask" (uint8), "label" (int16 or int32 depending on values) aliases (#1096)
  * For both `__init__` and `to_filename()`:
    * Raise `TypeError` if `dtype` is `int` or `"int"` (`np.dtype(int) == np.dtype("int") == np.dtype("int64"))` so users need to explicitly request `np.int64`.
  * Add documentation page to link in warnings/errors
* In NiBabel 5.0.0 (September 2022):
  * `img.to_filename(dtype=None)` to raise exception if `img.get_data_dtype()` is (u)int64

cc @matthew-brett @jbteves @neurolabusc @mrneont @jeromedockes @NicolasGensollen

Please add anything I've missed or let me know if anything needs adjustment.

jbteves:
Thanks for writing this up! I would add to the dtype parameter that you should explicitly test to make sure that if somebody passes the `int` alias, it errors out as invalid so that we avoid the int32/int64 ambiguity when discussing "nifti" int versus the "numpy" int. I'm not sure exactly how the type resolution might work so I wanted to make sure that was down.

neurolabusc:
@effigies thanks for the concise summary.

[...]

PR #1113, opened 2022-06-07
NF: nib-convert CLI tool

effigies:
In #1089 we proposed:

> Provide nib-cast (name may change) command-line tool to recast existing files to new data type

Also found an error in NIfTI writing if `get_data_dtype(finalize=True)` wasn't called prior to `to_file_map()`.

### DIFF
--- a/nibabel/cmdline/tests/test_convert.py
+++ b/nibabel/cmdline/tests/test_convert.py
@@ -32,11 +32,21 @@ def test_convert_noop(tmp_path):
     assert converted.shape == orig.shape
     assert converted.get_data_dtype() == orig.get_data_dtype()
 
+    infile = test_data(fname='resampled_anat_moved.nii')
+
     with pytest.raises(FileExistsError):
         convert.main([str(infile), str(outfile)])
 
     convert.main([str(infile), str(outfile), '--force'])
     assert outfile.is_file()
+
+    # Verify that we did overwrite
+    converted2 = nib.load(outfile)
+    assert not (
+        converted2.shape == converted.shape
+        and np.allclose(converted2.affine, converted.affine)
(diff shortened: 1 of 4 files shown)


effigies:
Can I interest anyone in a review? Want to get this in for 4.0 as it includes a fix.

nibabel/cmdline/tests/test_convert.py
```diff
@@ -0,0 +1,152 @@
...
+    converted = nib.load(outfile)
+    assert np.allclose(converted.affine, orig.affine)
+    assert converted.shape == orig.shape
+    assert converted.get_data_dtype() == orig.get_data_dtype()
+
+    with pytest.raises(FileExistsError):
+        convert.main([str(infile), str(outfile)])
+
+    convert.main([str(infile), str(outfile), '--force'])
+    assert outfile.is_file()
```
arokem, line 39: Not too important, but is there a way to make sure that this is the file that was created in the forced run, and not the one that existed there before? You might worry that the `force` flag would somehow make the program do nothing.

arokem (commented):
Overall, looks good. I had a couple of small comments.

effigies:
Thanks @arokem. I think I addressed both of your concerns. Please have another look when you get a minute.

arokem:
Yep. This looks good to me and I am +1 to merge. Does nibabel have a protocol about needing additional reviewers, or should I hit "merge"?

effigies:
Nope, one reviewer is enough, unless several people have been active in the discussion. (We have some docs here now: https://nipy.org/nibabel/devel/core_developer.html#reviewing)

arokem:
Got it! Thanks. I'll wait on that one remaining CI run and then hit the merge button.

Issue #1065, opened 2021-11-16
Add MZ3 surface file format


effigies:
Via @neurolabusc:

> you may want to consider handling the [mz3](https://github.com/neurolabusc/surf-ice/tree/master/mz3) format, the page includes Python, Blender, JavaScript code as well as a full specification and examples datasets. It attempts to live at the Pareto frontier of file size and speed. It is both faster and smaller than GIfTI. The layout design maps directly onto modern GPU architectures (and hence WebGL, Metal and Vulkan APIs). I developed this primarily for the sample files distributed with Surfice, to reduce download times, disk usage and the time to run the demos / regression testing. It is intentionally simple, and does not have the flexibility of other formats. It does not attempt to do everything, but it can fill a niche.

neurolabusc:
I forgot to mention Matlab code is provided. @gllmflndn contributed to the code and specification, and thanks to him it is supported by [spm12](https://github.com/spm/spm12/search?q=mz3) and [PALM](https://github.com/andersonwinkler/PALM).

effigies:
Thanks, Chris. That sounds reasonable to me.

PR #1114, opened 2022-06-09
ENH: Reading and writing JSON/binary JSON based surface data (.jmsh and .bmsh)

fangq:
hi @effigies, I am following up on the discussions from #1065, and would like to use this thread to propose a patch to support .jmsh/.bmsh based surface mesh data formats.

To recap on my previous presentation during the surface mesh format meeting:
- `.jmsh` is basically a plain JSON file using [JMesh annotations](https://github.com/NeuroJSON/jmesh/blob/master/JMesh_specification.md) to store surface mesh data structures (nodes as `MeshVertex3`, triangles as `MeshTri3`)
- `.bmsh` is the same as .jmsh except that the underlying serialization uses a binary JSON format - [BJData](https://github.com/NeuroJSON/bjdata)
- via the [JData specification](https://github.com/NeuroJSON/jdata/blob/master/JData_specification.md) (data-structure level annotations), both formats support [data-record-level compression](https://github.com/NeuroJSON/jdata/blob/master/JData_specification.md#compressed-array-storage-format) (zlib, gzip, 'lzma' and potentially others)

thanks to @neurolabusc's mesh format benchmarks, the loading speed and file sizes of these files are summarized/compared in [this post](https://nipy.discourse.group/t/data-formats-review-surface-and-surface-sampled-data/123/22?u=fangq).

Here, I would like to show you a lightweight patch (draft) to allow nibabel to read/write .jmsh and .bmsh files. A screenshot showing the interface can be found below. 

This patch requires two external lightweight packages as dependencies, both can be installed via pip:
- [jdata (14 kB) - handling JData annotation encoding/decoding and file IO](https://pypi.org/project/jdata/) and 
- [bjdata (63kB) - handling bjdata reading/writing](https://pypi.org/project/bjdata/)

because I am not entirely familiar with nibabel's IO classes, currently, I based the JMesh class over `FileBasedImage`. I understand that `CoordinateImage` is currently being developed in #1090, and perhaps it is the better root class for adding new mesh formats.

Let me know if you have any comments about this draft. I would be happy to add some tests and open to suggestions to continue polishing this patch.

jmesh_nibabel


effigies:
Thanks @fangq! This is great to see. If I'm reading this right, `jmsh`/`bmsh` are specifically surface meshes, not data sampled to surface meshes? If so, then it corresponds to the `TriangularMesh` API from [BIAP 9](https://nipy.org/nibabel/devel/biaps/biap_0009.html#proposal), not really the `CoordinateImage` itself.

One question I have is if it would be possible to load the metadata from files and then construct something like an [array proxy](https://nipy.org/nibabel/reference/nibabel.arrayproxy.html#nibabel.arrayproxy.ArrayProxy) so that the data arrays do not need to be loaded until called for? If I recall, your files are not actually gzipped, only the data arrays, so we can't use our existing file handlers and proxies as drop-in loaders, but I suspect we could figure something out.

fangq:
> Thanks @fangq! This is great to see. If I'm reading this right, jmsh/bmsh are specifically surface meshes, not data sampled to surface meshes? If so, then it corresponds to the TriangularMesh API from [BIAP 9](https://nipy.org/nibabel/devel/biaps/biap_0009.html#proposal), not really the CoordinateImage itself.

Thanks for the feedback. The JMesh spec actually permits one to attach [node or triangle-associated data](https://github.com/NeuroJSON/jmesh/blob/master/JMesh_specification.md#common-geometry-properties) as [Properties](https://github.com/NeuroJSON/jmesh/blob/master/JMesh_specification.md#structure-form). The currently supported property keywords include `Normal`, `Color`, `Value`, `Tag`, and `Size`. However, for simplicity, in this PR, I only exposed the `Tag` properties as JMesh class's `nodelabel` and `facelabel` properties - others can also be easily exposed if see a need.

In this case, should I derive JMesh from TriangularMesh class?

> One question I have is if it would be possible to load the metadata from files and then construct something like an [array proxy](https://nipy.org/nibabel/reference/nibabel.arrayproxy.html#nibabel.arrayproxy.ArrayProxy) so that the data arrays do not need to be loaded until called for?

Loading array buffers from the file is handled in the parser level (i.e. the external `bjdata` and `jdata` moduels), and it currently always loads the data to memory when parsing a file. When the data buffer is compressed, my understanding is that `numpy.memmap` or `arrayproxy` may not likely to work. But for raw/uncompressed bjdata (raw.bmsh) files, I could potentially add some code to allow the parser to return `numpy.memmap` objects if user prefers. For now, let's assume this memmap based access is not yet supported.

By the way, do you have a code formatter (something like [this](https://github.com/fangq/mcx/blob/master/src/Makefile#L268-L291) for C++) so that I can get rid of CI warnings? 

also, how to run the tests in nibabel? I just want to follow the examples of other modules and start building some tests.

fangq:
@effigies, sorry for dropping the ball. I would like to spend some time to get this PR polished.

I see you had added a formatter (blue) in #1124, I would like to know what is the command I should run to apply the style to format my code.

Also, per your earlier comment, jmesh can associate values to vertices/triangles, please let me know if the `FileBasedImage` class that I currently derived the `JMesh` class from is still appropriate.

thanks

effigies:
Thanks @fangq. I'll try to have a look at this in more detail in the coming week.

PR #1118, opened 2022-06-18
TEST: Suppress new numpy warning on nan-to-int cast

effigies:
https://github.com/numpy/numpy/pull/21437 does not require us to change our behavior, but does require us to change our expectations of what warnings are raised. Since this is quite an edge case, it seems to make more sense to suppress the warning and eventually expect it when numpy 1.24 becomes our minimum.

@matthew-brett What do you think?

### DIFF
--- a/nibabel/tests/test_volumeutils.py
+++ b/nibabel/tests/test_volumeutils.py
@@ -413,7 +413,7 @@ def test_a2f_nan2zero():
     # How weird?  Look at arr.astype(np.int64)
     with np.errstate(invalid='ignore'):
         data_back = write_return(arr, str_io, np.int64, nan2zero=False)
-    assert_array_equal(data_back, arr.astype(np.int64))
+        assert_array_equal(data_back, arr.astype(np.int64))
 
 
 def test_a2f_nan2zero_scaling():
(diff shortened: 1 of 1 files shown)


nibabel/tests/test_volumeutils.py
```diff
@@ -692,9 +692,15 @@ def test_a2f_nan2zero_range():
...
         with pytest.warns(warn_type) if warn_type else error_warnings():
-            nan_cast = np.array(np.nan, dtype=dt).astype(np.int8)
+            # XXX NP1.24
+            # Casting nan to int will produce a RuntimeWarning in numpy 1.24
+            # Change to expecting this warning when this becomes our minimum
+            with np.errstate(invalid='ignore'):
+                nan_cast = np.array(np.nan, dtype=dt).astype(np.int8)
         with pytest.warns(warn_type) if warn_type else error_warnings():
-            back_arr = write_return(arr, fobj, np.int8, intercept=129, nan2zero=False)
+            # XXX NP1.24 - expect RuntimeWarning
```
matthew-brett, line 701: What do you think about a Numpy version check here and checking for RunTime error, so we don't forget?

effigies: Was trying to avoid it at 3am, but I guess it does make sense to go ahead and put the anticipated final test in now. Using Version('1.24.dev0+239') to match https://github.com/numpy/numpy/commit/abc26a351e14c890bdd67785ed4f3c60c637b0fa.

matthew-brett: Sorry to make you do that!

matthew-brett:
Sorry - I didn't check - but our pre-release checks will hit the relevant branch?

effigies:
> Sorry - I didn't check - but our pre-release checks will hit the relevant branch?

Relevant branch of numpy? Yes, that commit is in nightly, which is why pre-release tests are failing right now.

Not sure if I'm interpreting correctly, though, so please feel free to clarify your question.

matthew-brett:
Yes, sorry - that's what I meant.

PR #1124, opened 2022-07-23
STY: Normalize to blue and isort

effigies:
Looking at #1114, I think we're past the days of asking people to follow style guides. Code formatters have gotten a lot better in the last decade, and it makes more sense to just give people the configuration to auto-format.

Here I'm proposing we use [blue](https://blue.readthedocs.io/), which is like black but doesn't insist on double-quotes everywhere.

I'm also suggesting [isort](https://pycqa.github.io/isort/) just because I've seen a number of PRs where people's editors auto-sort imports for them, so let's just tell their editors what to do.

The style config becomes:

```ini
[flake8]
max-line-length = 99
extend-ignore = E203,E266,E402,E731
exclude =
    *test*
    *sphinx*
    nibabel/externals/*
    */__init__.py

[blue]
line_length = 99
extend-exclude = _version.py

[isort]
extend_skip = _version.py
profile = black
line_length = 99
```

There are some things I want to revert and protect from re-formatting, most notably the NIfTI intent recoder. But on the whole I find very little objectionable and some things look noticeably better. Any objections / other things to revert and protect?


ZviBaratz:
How do you feel about specifying:

```ini
[flake8]
...
per-file-ignores =
    */__init__.py: F401
```
for *init* files? Or are there other codes we should be ignoring? 

Otherwise, I think it would be great to move forward with this. We can always revert changes in particular files or sections if we find it less readable, and hopefully at some point we will be able to update `blue` with your PR having been merged. LMK if there's anything I can do to help with this.

effigies:
Let's break all the PRs...

PR #1127, opened 2022-07-26
FIX: Make GiftiMetaData.data a list proxy, deprecate

effigies:
This is a follow-up to #1091 with things that should have been done there, including proxying actions on the `GiftiMetaData.data` internal list and deprecating the property altogether.

I added a comprehensive test that checks that attempting to use `GiftiMetaData.data` as one might have pre-4.0 works once again.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -26,6 +26,42 @@ from .util import (array_index_order_codes, gifti_encoding_codes,
 from ..deprecated import deprecate_with_version
 
 
+class _GiftiMDList(list):
+    """List view of GiftiMetaData object that will translate most operations"""
+    def __init__(self, metadata):
+        self._md = metadata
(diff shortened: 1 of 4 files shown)


effigies:
@htwangtw Could I bug you for a review here?

effigies:
I have, somewhat tediously, looked through every instance of `GiftiNVPairs` I could find on GitHub, and there are many uses cases where people `darray.meta.data.insert(0, GiftiNVPairs(...))`, but fortunately nobody seems to be slicing or anything like that.

One thing that isn't addressed here is that we no longer offer anything but insertion order, so people who previously put `insert(0, ...)` will not get the same result out. It doesn't seem like a major crime, though.

htwangtw (approved):
The test looks comprehensive and I cannot think of anything else. LGTM

effigies:
Thanks!

Issue #1128, opened 2022-08-05
Load image with non-conforming filename?


quasarj:
Is there a way to load a file if it does not end in .nii or .nii.gz?

I have looked through the code and read over the issues, but don't see any info on how to do this. I'm a little afraid that the code looks like nibabel assumes the filenames are accurate, which is... highly dangerous, to say the least. I'm also confused by why it would even do this, because it looks like it also reads some bytes from the file to "sniff" the type?

matthew-brett:
Could you use something like:

```python
with open('test.dat', 'rb') as fobj:
    img = nib.Nifti1Image.from_bytes(fobj.read())
```

?

effigies:
If you're looking for the reasons the code is the way it is, there are several problems we need to solve:

1) Handle images with multiple paired files (`.hdr` + `.img`) and allow users to load the pair by passing either filename
2) Handle image types with multiple possible representations (`.hdr` + `.img` OR `.nii`)
3) Handle multiple image types that share header formats (e.g., Analyze, SPM-style Analyze, NIfTI-1)
4) Handle compression transparently for all of the above
5) Delay loading of data array until called

(2) is the big pain that makes even using `nib.Nifti1Image.from_filename('test.dat')` fail, as trying to determine how to read a NIfTI file automatically requires knowing what variant it is. To override this detection, you can create a `file_map`:

```Python
file_map = nib.Nifti1Image.make_file_map({'image': 'test.dat'})
img = nib.Nifti1Image.from_file_map(file_map)
```

This is somewhat irreducibly complex (in the current architecture) because we have to consider paired images and delayed data loading. It's possible that automatically detecting compression from the magic number instead of from the extension could be done without a lot of rearchitecting.

quasarj:
Thank you for the reply! I had forgotten about the .hdr + .img situation, and I can definitely see how that may add complexity.

And thank you for the idea for overriding the detection. That does work if my .nii file is not compressed, but if it's gzipped I get an error.  Is there some way to tell it to expect it compressed? I only see examples passing the keys "image" and "header" to make_file_map.

effigies:
No. Right now our image openers that store filenames so that you can close and reopen files as needed (I forgot to mention that we try to avoid keeping filehandles open longer than needed, as HPC clusters often have strict quotas) rely on the extension of a gzip-compressed file to be `.gz`. This limitation is what I was trying to get at here:

[...]

PR #1129, opened 2022-08-15
ENH: Add to/from_stream methods and from_url classmethod to SerializableImage

effigies:
Extends #639 to allow `from_url()` to load an image. Does not currently handle gzipped (see #1128 for some more discussion) streams, though this should be doable.

Abstracts to using `io.IOBase` to permit easy extension to other streams, but does not expose it generically at this point.

Tested to work on https://openneuro.org/crn/datasets/ds000117/snapshots/1.0.5/files/sub-01:ses-mri:fmap:sub-01_ses-mri_magnitude1.nii and a `file:///...` option.

### DIFF
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -540,27 +540,27 @@ class SerializableImage(FileBasedImage):
     """
 
     @classmethod
-    def _filemap_from_iobase(klass, ioobject: io.IOBase):
+    def _filemap_from_iobase(klass, io_obj: io.IOBase):
         """For single-file image types, make a file map with the correct key"""
         if len(klass.files_types) > 1:
             raise NotImplementedError(
                 "(de)serialization is undefined for multi-file images"
             )
-        return klass.make_file_map({klass.files_types[0][0]: ioobject})
+        return klass.make_file_map({klass.files_types[0][0]: io_obj})
 
     @classmethod
-    def _from_iobase(klass, ioobject: io.IOBase):
+    def _from_iobase(klass, io_obj: io.IOBase):
         """Load image from readable IO stream
 
         Convert to BytesIO to enable seeking, if input stream is not seekable
         """
-        if not ioobject.seekable():
-            ioobject = io.BytesIO(ioobject.read())
-        return klass.from_file_map(klass._filemap_from_iobase(ioobject))
+        if not io_obj.seekable():
+            io_obj = io.BytesIO(io_obj.read())
+        return klass.from_file_map(klass._filemap_from_iobase(io_obj))
 
-    def _to_iobase(self, ioobject: io.IOBase, **kwargs):
+    def _to_iobase(self, io_obj: io.IOBase, **kwargs):
         """Save image from writable IO stream"""
-        self.to_file_map(self._filemap_from_iobase(ioobject), **kwargs)
+        self.to_file_map(self._filemap_from_iobase(io_obj), **kwargs)
 
     @classmethod
     def from_bytes(klass, bytestring: bytes):
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -549,17 +549,30 @@ class SerializableImage(FileBasedImage):
         return klass.make_file_map({klass.files_types[0][0]: io_obj})
 
     @classmethod
-    def _from_iobase(klass, io_obj: io.IOBase):
+    def from_stream(klass, io_obj: io.IOBase):
         """Load image from readable IO stream
 
         Convert to BytesIO to enable seeking, if input stream is not seekable
+
+        Parameters
+        ----------
+        io_obj : IOBase object
+            Readable stream
         """
         if not io_obj.seekable():
             io_obj = io.BytesIO(io_obj.read())
         return klass.from_file_map(klass._filemap_from_iobase(io_obj))
 
-    def _to_iobase(self, io_obj: io.IOBase, **kwargs):
-        """Save image from writable IO stream"""
+    def to_stream(self, io_obj: io.IOBase, **kwargs):
+        """Save image to writable IO stream
+
+        Parameters
+        ----------
+        io_obj : IOBase object
+            Writable stream
+        \*\*kwargs : keyword arguments
+            Keyword arguments that may be passed to ``img.to_file_map()``
+        """
         self.to_file_map(self._filemap_from_iobase(io_obj), **kwargs)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


nibabel/tests/test_image_api.py
```diff
@@ -572,6 +572,45 @@ def validate_to_from_bytes(self, imaker, params):
...
+
+    def validate_from_file_url(self, imaker, params):
+        tmp_path = self.tmp_path
+
+        img = imaker()
+        import uuid
+        fname = tmp_path / f'img-{uuid.uuid4()}{self.standard_extension}'
+        img.to_filename(fname)
+
+        rt_img = img.__class__.from_url(f"file://{fname}")
```
effigies, line 606: Was assuming paths started with `/`. An extra `/` won't hurt posix, but is needed for Windows.

```suggestion
        rt_img = img.__class__.from_url(f"file:///{fname}")
```

effigies:
Anybody interested in reviewing?

nibabel/filebasedimages.py
```diff
@@ -538,7 +540,30 @@ class SerializableImage(FileBasedImage):
     """
 
     @classmethod
-    def from_bytes(klass, bytestring):
+    def _filemap_from_iobase(klass, ioobject: io.IOBase):
```
matthew-brett, line 543: Consider `io_obj` instead of `ioobject`?   Just because it's a bit easier to read, and might be easier for non-English-speakers to parse.

nibabel/filebasedimages.py
```diff
@@ -538,7 +540,30 @@ class SerializableImage(FileBasedImage):
...
+            )
+        return klass.make_file_map({klass.files_types[0][0]: ioobject})
+
+    @classmethod
+    def _from_iobase(klass, ioobject: io.IOBase):
+        """Load image from readable IO stream
+
+        Convert to BytesIO to enable seeking, if input stream is not seekable
+        """
+        if not ioobject.seekable():
```
matthew-brett, line 557: Do we need a specific error if the stream is not readable?

nibabel/filebasedimages.py
```diff
@@ -538,7 +540,30 @@ class SerializableImage(FileBasedImage):
     """
 
     @classmethod
-    def from_bytes(klass, bytestring):
+    def _filemap_from_iobase(klass, ioobject: io.IOBase):
+        """For single-file image types, make a file map with the correct key"""
+        if len(klass.files_types) > 1:
+            raise NotImplementedError(
```
matthew-brett, line 546: Worth a test on e.g. an Analyze image?

matthew-brett (commented):
Can we automatically recognize the URL and plumb this into `nib.load` etc?

nibabel/filebasedimages.py
```diff
@@ -538,7 +540,30 @@ class SerializableImage(FileBasedImage):
...
+                "(de)serialization is undefined for multi-file images"
+            )
+        return klass.make_file_map({klass.files_types[0][0]: ioobject})
+
+    @classmethod
+    def _from_iobase(klass, ioobject: io.IOBase):
```
matthew-brett, line 552: Any reason not to expose this as something `from_stream`?  (and `to_stream`)?

matthew-brett (commented):
Oops - sorry - forgot to add this comment.

nibabel/filebasedimages.py
effigies: This is what we'll currently get:

```Python
In [1]: f = open('deleteme', 'w')

In [2]: f.read()
UnsupportedOperation                      Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 f.read()

UnsupportedOperation: not readable
```

Any thoughts on a better exception than `io.UnsupportedOperation`?

effigies:
> Can we automatically recognize the URL and plumb this into `nib.load` etc?

Probably. Feels like a bit of a can of worms that I didn't want to open in this PR. Okay to push off to another one?

nibabel/filebasedimages.py
effigies: Not that I can think of.

nibabel/filebasedimages.py
matthew-brett: But - I guess we'll get that in due course, either here, if the object is not `seekable`, or later when we try and read from the image?   Do you think there's any advantage here for checking if the stream is readable and raising here, as opposed to later?

effigies: Well, let's just look at what we get:

```Python
In [7]: nb.Nifti1Image.from_stream(f)
UnsupportedOperation                      Traceback (most recent call last)
<ipython-input-7-174c60215a44> in <module>
----> 1 nb.Nifti1Image.from_stream(f)

~/Projects/nipy/nibabel-mnt/nibabel/filebasedimages.py in from_stream(klass, io_obj)
    557         if not io_obj.seekable():
    558             io_obj = io.BytesIO(io_obj.read())
--> 559         return klass.from_file_map(klass._filemap_from_iobase(io_obj))
    560 
    561     def to_stream(self, io_obj: io.IOBase, **kwargs):

~/Projects/nipy/nibabel-mnt/nibabel/analyze.py in from_file_map(klass, file_map, mmap, keep_file_open)
    967         hdr_fh, img_fh = klass._get_fileholders(file_map)
    968         with hdr_fh.get_prepare_fileobj(mode='rb') as hdrf:
--> 969             header = klass.header_class.from_fileobj(hdrf)
    970         hdr_copy = header.copy()
    971         imgf = img_fh.fileobj

~/Projects/nipy/nibabel-mnt/nibabel/nifti1.py in from_fileobj(klass, fileobj, endianness, check)
    684     @classmethod
    685     def from_fileobj(klass, fileobj, endianness=None, check=True):
--> 686         raw_str = fileobj.read(klass.template_dtype.itemsize)
    687         hdr = klass(raw_str, endianness, check)
    688         # Read next 4 bytes to see if we have extensions.  The nifti standard

~/Projects/nipy/nibabel-mnt/nibabel/openers.py in read(self, *args, **kwargs)
    191 
    192     def read(self, *args, **kwargs):
--> 193         return self.fobj.read(*args, **kwargs)
    194 
    195     def readinto(self, *args, **kwargs):

UnsupportedOperation: not readable
```

It's three layers deeper in the call stack, but the traceback is pretty readable. My preference is to only preempt errors if they're going to be difficult to interpret when the downstream user gets them, so I'm inclined to keep this one as it is. Unless you'd rather hide the `ImageOpener()` machinery?

matthew-brett: No - that's fine - I'm happy if you're happy.

matthew-brett:
> Probably. Feels like a bit of a can of worms that I didn't want to open in this PR. Okay to push off to another one?

Yes, that's fine.

effigies:
Thanks!

PR #1131, opened 2022-08-20
ENH: Make layout order an initialization parameter of ArrayProxy

effigies:
For #1090, it's been useful to set the proxy order dynamically, rather than subclassing `ArrayProxy`. This makes it an initialization argument. I've left it as a class variable on the off-chance that somebody actually uses something like:

```Python
class CArrayProxy(ArrayProxy):
    order = 'C'
```

Minor in-passing cleanups, such as removing the explicit `(object)` superclass and moving the test loops to parametrizations.

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -119,7 +119,8 @@ class ArrayProxy:
         order : {None, 'F', 'C'}, optional, keyword only
             `order` controls the order of the data array layout. Fortran-style,
             column-major order may be indicated with 'F', and C-style, row-major
-            order may be indicated with 'C'. The default order is 'F'.
+            order may be indicated with 'C'. None gives the default order, that
+            comes from the `_default_order` class variable.
         keep_file_open : { None, True, False }, optional, keyword only
             `keep_file_open` controls whether a new file handle is created
             every time the image is accessed, or a single file handle is
@@ -133,7 +134,7 @@ class ArrayProxy:
         if mmap not in (True, False, 'c', 'r'):
             raise ValueError("mmap should be one of {True, False, 'c', 'r'}")
         if order not in (None, 'C', 'F'):
-            raise ValueError("order should be one of {'C', 'F'}")
+            raise ValueError("order should be one of {None, 'C', 'F'}")
         self.file_like = file_like
         if hasattr(spec, 'get_data_shape'):
             slope, inter = spec.get_slope_inter()
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -115,7 +115,7 @@ class ArrayProxy:
             True gives the same behavior as ``mmap='c'``.  If `file_like`
             cannot be memory-mapped, ignore `mmap` value and read array from
             file.
-        order : {'F', 'C'}, optional, keyword only
+        order : {None, 'F', 'C'}, optional, keyword only
             `order` controls the order of the data array layout. Fortran-style,
             column-major order may be indicated with 'F', and C-style, row-major
             order may be indicated with 'C'. The default order is 'F'.
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -53,7 +53,7 @@ raised.
 KEEP_FILE_OPEN_DEFAULT = False
 
 
-class ArrayProxy(object):
+class ArrayProxy:
     """ Class to act as proxy for the array that can be read from a file
 
     The array proxy allows us to freeze the passed fileobj and header such that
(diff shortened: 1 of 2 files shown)


effigies:
Anybody up for a review?

matthew-brett:
Will do, tomorrow ...

nibabel/arrayproxy.py
```diff
@@ -116,6 +115,10 @@ def __init__(self, file_like, spec, *, mmap=True, keep_file_open=None):
             True gives the same behavior as ``mmap='c'``.  If `file_like`
             cannot be memory-mapped, ignore `mmap` value and read array from
             file.
+        order : {'F', 'C'}, optional, keyword only
```
matthew-brett, line 118: Shouldn't this be:

```suggestion
        order : {None, 'F', 'C'}, optional, keyword only
```

?

nibabel/tests/test_arrayproxy.py
```diff
@@ -154,33 +154,55 @@ def test_nifti1_init():
...
     fobj = BytesIO()
-    fobj.write(b'\0' * offset)
+    fobj.write(bytes(offset))
     fobj.write(arr.tobytes(order='F'))
     prox = ArrayProxy(fobj, hdr)
     sliceobj = (None, slice(None), 1, -1)
     assert_array_equal(arr[sliceobj] * 2.0 + 1.0, prox[sliceobj])
 
 
+@pytest.mark.parametrize("order", ("C", "F"))
```
matthew-brett, line 193: Also test `order=None`?

matthew-brett (commented):
Hmm - I see what you mean, but the interface is now a bit clunky, because there are two ways to override the order.   I guess the most direct interface would be to make the class variable be `default_order`.     Is there a good way of going there without breaking people's code?

nibabel/tests/test_arrayproxy.py
effigies: Does not work for `fobj.write(arr.tobytes(order=order))` (currently L198). This is testing override specifically, so I'm not sure "use default" is really relevant.

effigies:
> Hmm - I see what you mean, but the interface is now a bit clunky, because there are two ways to override the order. I guess the most direct interface would be to make the class variable be `default_order`. Is there a good way of going there without breaking people's code?

Could change `order` to `default_order`, and then look for an `order` class variable (`hasattr(self.__class__, "order")`) in `__init__` and raise a `DeprecationWarning`.

matthew-brett:
> > 
> 
> Could change `order` to `default_order`, and then look for an `order` class variable (`hasattr(self.__class__, "order")`) in `__init__` and raise a `DeprecationWarning`.

Nice idea.

effigies:
Matthew, any further comments?

nibabel/tests/test_arrayproxy.py
```diff
@@ -57,6 +60,10 @@ def copy(self):
 
 class CArrayProxy(ArrayProxy):
     # C array memory layout
+    _default_order = 'C'
+
+
+class DeprecatedCArrayProxy(ArrayProxy):
```
matthew-brett, line 66: Timed out depraction again (as above)?

matthew-brett (commented):
Oops - sorry - some of these I forgot to add in another review.

nibabel/arrayproxy.py
```diff
@@ -142,11 +148,25 @@ def __init__(self, file_like, spec, *, mmap=True, keep_file_open=None):
...
             raise TypeError('spec must be tuple of length 2-5 or header object')
 
+        # Warn downstream users that the class variable order is going away
+        if hasattr(self.__class__, 'order'):
+            warnings.warn(f'Class {self.__class__} has an `order` class variable. '
+                          'ArrayProxy subclasses should rename this variable to `_default_order` '
+                          'to avoid conflict with instance variables.\n'
+                          '* deprecated in version: 5.0\n'
+                          '* will raise error in version: 7.0\n',
+                          DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
effigies: It's a decorator that works for deprecating an entire function/method/class. It does not work for cases like this. This is why I added the check below:

nibabel/tests/test_arrayproxy.py
```diff
@@ -154,33 +166,87 @@ def test_nifti1_init():
...
+    # We don't really care about the original order, just that the behavior
+    # of the deprecated mode matches the new behavior
+    fprox = ArrayProxy(fobj, (shape, arr.dtype), order='F')
+    cprox = ArrayProxy(fobj, (shape, arr.dtype), order='C')
+
+    # Start raising errors when we crank the dev version
+    if Version(__version__) >= Version('7.0.0.dev0'):
+        cm = pytest.raises(ExpiredDeprecationError)
+    else:
+        cm = pytest.deprecated_call()
```
effigies, line 234: ... here.

nibabel/tests/test_arrayproxy.py
effigies: This is specifically for testing. The warning/expiration will show up in `test_deprecated_order_classvar`. When we remove that function, we can remove this class. I'll add a comment...

effigies, line 66:
```suggestion
class DeprecatedCArrayProxy(ArrayProxy):
    # Used in test_deprecated_order_classvar. Remove when that test is removed (8.0)
```

matthew-brett (commented):
LGTM ...

effigies:
Thanks!

Issue #859, opened 2020-01-08
pip editable install for user broken since adding pyproject.toml


matthew-brett:
Before commit c0321382, that added `pyproject.toml`, I could do this:

```
pip install --user -e .
```

After that commit, I get:

```
    usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
       or: setup.py --help [cmd1 cmd2 ...]
       or: setup.py --help-commands
       or: setup.py cmd --help
    
    error: option --user not recognized
```

I'm not quite sure why.  Any ideas?

effigies:
Ugh. Looks like this is a known issue: https://github.com/pypa/pip/issues/6375

The solution seems to be `--no-use-pep517`, added in https://github.com/pypa/pip/pull/6370.

I'm looking into whether `pyproject.toml` can be fixed to allow pip to handle this case without a special flag.

effigies:
Might just drop `pyproject.toml`. I was hoping that it would avoid issues with an old `setuptools`, but it seems like a rat's nest.

effigies:
To dig into it a bit more, I think I (sort of) understand what's going on.

| | ~~`--editable`~~ | `--editable` |
|--|--|--|
| ~~`--user`~~ | build wheel, install in environment | Call `setup.py develop` |
| `--user` | build wheel, install in user directory | XXXX |

* `pyproject.toml` enables build isolation, which we were using to ensure that a recent setuptools got used without depending on users to have a very recent `setuptools` and `pip`.
* `pip` will handle `pip install --user .` by building the wheel in isolation, and then installing the wheel according to `--user`.
* Pip knows enough about `-e` to know that it's setuptools' problem, so passes everything that way. Setuptools, of course, doesn't know anything about `--user`.

More broadly, PEP-517 simply doesn't address the notion of editable mode, which is a setuptools-specific concept. No other build systems have an equivalent.

The only resolution that currently exists to do both `--user` and `--editable` is to disable PEP-517, which can currently be done in `pip` with `--no-use-pep517` or by deleting `pyproject.toml`.

effigies:
@matthew-brett Do you feel we should remove `pyproject.toml`, or are you content to use `--no-use-pep517` in this context?

I'm honestly not happy with any options at this point, so can be persuaded to go either way.

matthew-brett:
I'm fine either way - but - what is the advantage of `pyproject.toml`?

effigies:
[...]

PR #1133, opened 2022-09-04
MNT: Re-add pyproject.toml

effigies:
Time to try using more modern build metadata again.

* Use `pyproject.toml` to mark build-time dependencies ~~, which also allows us to move to the non-vendored flavor of versioneer~~
* Move most package metadata out of `setup.cfg`. I don't really want to rewrite the extras interpolation, so I'm marking that dynamic for now.

@matthew-brett I've verified that `pip install --user -e .` works with recent setuptools, which was the issue in #859 that made us drop this before.


matthew-brett (approved):
Looks good to me.    Are you still iterating?

effigies:
Nope, thanks for the review!

PR #1137, opened 2022-09-13
Fixes a bug when a .gz / .bz2 without a nii is opened

Sikerdebaard:
When a .gz / .bz2 / .zst is opened with nibabel.load and no valid nii file is present after decompression the packages throws an exception on [loadsave.py:65](https://github.com/nipy/nibabel/blob/master/nibabel/loadsave.py#L65). This is because the variable `sniff` is a tuple as returned from the various image classes iterated in [loadsave.py:98](https://github.com/nipy/nibabel/blob/master/nibabel/loadsave.py#L98). The code expects a byte-array, or at least something with a `.startswith` function. Instead an `AttributeError  'tuple' object has no attribute 'startswith'` is raised as a tuple does not have this function.

This code fixes the problem by resetting the variable `sniff` to `None`. In my opinion that is the most sensible way to deal with the issue as it checks the filetype of the file and not the filetype of the decompressed file.

I've added a pytest to `nibabel/tests/test_image_types.py` to test this code. Relevant datafiles have been added to `nibabel/tests/data` as well.

Note: for the unit test - if you wish to enable testing of `.zst` files make sure you add the signature of these files to [loadsave.py:49](https://github.com/nipy/nibabel/blob/master/nibabel/loadsave.py#L49). The header of these files seem to be `\xfd\x2f\xb5\x28`.


nibabel/tests/test_image_types.py
```diff
@@ -121,3 +123,29 @@ def test_sniff_and_guessed_image_type_randomized():
...
+            assert not is_valid
+            if is_valid:
+                img = image_klass.from_filename(filename, **kwargs)
+                return img
+
+        # reset sniff to None as we are not interested in the header of the decompressed file
+        sniff = None
+        matches, msg = _signature_matches_extension(filename, sniff)
+
+        assert matches == True
```
effigies, line 151: Here we're re-implementing (corrected) `nb.load()` logic. However, this passes because it doesn't trigger the error condition being fixed. Instead, we should make sure that `nb.load()` behaves in the correct way (gives the right output or raises the right error) when given various types of inputs.

We already have:

https://github.com/nipy/nibabel/blob/2838c06389165c9563551d35988df793fb15263d/nibabel/tests/test_loadsave.py#L77-L82

This is getting us the right answer, albeit for the wrong reason because `sniff` is a tuple and not a `bytes`. To detect the actual bug, we need a new test with inputs that are actually zipped but don't contain images. You provide some, but we can just as easily create them on the fly.

```Python
from nibabel.openers import Opener

@pytest.mark.parametrize("extension", [".gz", ".bz2"])
def test_load_good_extension_with_bad_data(tmp_path, extension):
    file_path = tmp_path / f"img.nii{extension}"
    with Opener(file_path, "wb") as fobj:
        fobj.write(b"bad")
    with pytest.raises(ImageFileError, match="Cannot work out file type of .*"):
        load(file_path)
```

WDYT?

effigies (commented):
Thanks for catching this! I think we should go in a different direction for both the test and the fix, see comments below.

Also, I agree with adding `.zst` to the `signatures` to check for.

Sikerdebaard:
@effigies looks good to me. I'm a bit time-constrained right now so I can't verify this code but I'm assuming it works.

PR #1138, opened 2022-09-13
FIX: Testing for extension/signature mismatches shouldn't use sniff 

effigies:
Bug introduced in #1013 and reported in #1137. Added a test that fails with an unexpected error, and then updated `_signature_matches_extension` to not accept any `sniff`. (If the file was compressed, the sniff would be read from the uncompressed stream. If it wasn't compressed, the sniff would be empty.)

Some updates to make the same machinery work transparently for `.zst`.

Closes #1137.

### DIFF
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -430,7 +430,7 @@ class FileBasedImage(object):
         try:
             with ImageOpener(meta_fname, 'rb') as fobj:
                 binaryblock = fobj.read(sniff_nbytes)
-        except IOError:
+        except (IOError, EOFError):
             return None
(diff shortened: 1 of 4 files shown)


effigies:
@Sikerdebaard Would you care to try this branch out?

Sikerdebaard:
This code works in my particular environment. The new unit test case also succeeds. Looks good to me. :)

Any idea in what version this fix will be merged? I've monkeypatched it for current versions, but I'd like to add a version check so that it doesn't monkeypatch starting at the fixed version.

effigies:
The next release will be 5.0.0.

PR #1140, opened 2022-09-28
FIX: Handles invalid TCK files, prevents infinite loop

### LINKED ISSUE #1139: TCK streamlines gets stuck when file is not a TCK
Currently, the `nibabel.streamlines.tck.TckFile` (haven't looked at the other formats) does not verify that a file is a .tck (i.e. has the "magic number" `mrtrix tracks`).

Since it does not verifies it, the code goes straight to try to read the key-value headers and get [stuck in a loop](https://github.com/nipy/nibabel/blob/feb20638ceca8e852b78ba1be09c3fa2d73279c2/nibabel/streamlines/tck.py#L316), waiting for an "END". `readline` does not check for EOF.

If [`_read_header`](https://github.com/nipy/nibabel/blob/feb20638ceca8e852b78ba1be09c3fa2d73279c2/nibabel/streamlines/tck.py#L291) was a class method, it would be possible to do:

```python
if magic_number.strip() != cls.MAGIC_NUMBER:
    msg = ("TCK magic number not found.")
    raise HeaderError(msg)
```

however, being a static method, one could only reference the `TckFile.MAGIC_NUMBER` (or any other way really).
Also, a better way to read the lines would be better, and detect EOF.

This is my new `while true` (and a test case for this issue):

`echo "hello\nthere" > test.tck; python -c "import nibabel as nb; nb.streamlines.load('test.tck')"`

anibalsolon:
Proper check for the magic number and the header's end, stopping to look for it at EOF.

Resolves #1139


effigies:
Thanks @anibalsolon! This looks perfectly sensible. Two thoughts:

1) The `Opener` class is intended to allow us to ignore whether we were passed a filename or an open filehandle and act like a file handle. We should never need to access the `fobj` instance variable and instead just use the opener directly, such as `opener.read()` or `for line in opener:`. If there are methods that you need on it, we should add them, but I think everything you use should be available.

2) The use of `asstr()` seems like a code smell, as it indicates we're unsure whether we're getting text or binary data. It does seem worth figuring out how/why this is ever passed a file object in text mode and see if we can make sure that it's always `'rb'`.

The first would be good to address here. The second can wait for another PR unless you're feeling really eager to explore this.

effigies:
Looks good. Thanks!

anibalsolon:
@effigies Hey! Sorry, I started answering here and got distracted. Thank you for the review, good one about the `.fobj`

For 2,  the TCK file should be in binary mode, given that the data will be binary. I am working on a fix for that.

Regarding this text/binary mode guarantee, when providing a file object for Opener, would it be beneficial to make sure the file object is in the same mode as the requested for Opener? e.g.
```
fd = open('text.txt', 'r')

with Opener(fd):  # raise exception, default is 'rb'
    ...

with Opener(fd, 'w'):  # raise exception
    ...

with Opener(fd, 'r'):  # all good
    ...
```

Given the file object is provided directly to the Opener in some interfaces/formats (TCK and TRK), we do not know what the mode the fd is in. Or would it be better to do the checking in another layer of the code? I usually prefer to do these checkings instead of letting a random exception explode in the user's face, but not sure what's the philosophy for nibabel.

effigies:
Ah, sorry for jumping the gun.

I agree it would make sense for the opener to raise an exception for a mode mismatch, but I'm not sure if we should start enforcing it when there is no explicit mode. In any event, this would be an API change. So in the short term:

```Python
fname = 'file'
fd = open('file', 'r')

assert Opener(fname).mode == 'rb'
assert Opener(fname, mode='r').mode == 'r'
assert Opener(fd).mode == 'r'
assert Opener(fd, mode='r').mode == 'r'

with assert_raises(IOError):
    Opener(fd, 'rb')

with assert_raises(IOError):
    Opener(fd, 'w')
```

If we do want to make `Opener(open(fname, 'r'))` raise the same error as `Opener(open(fname, 'r'), mode='rb')`, we'll need to do a `DeprecationWarning` cycle. Which would be the next release +2 major versions (7.0).

PR #1153, opened 2022-12-12
MNT: Apply refurb suggestions

DimitriPapadopoulos:
* [FURB108]: Use `x in (y, z)` instead of `x == y or x == z`
* [FURB110]: Use `x or y` instead of `x if x else y`
* [FURB113]: Use `x.extend(...)` instead of repeatedly calling `x.append()`
* [FURB115]: Replace `len(x) > 0` with `x`
* [FURB118]: Replace lambda with `operator.eq`
* [FURB123]: Replace `int(x)` with `x`
* [FURB125]: Return is redundant here


nibabel/tests/data/gen_standard.py
```diff
@@ -35,19 +35,17 @@ def _gen_straight_streamline(start, end, steps=3):
...
+    ]
 
     # Get the coordinates of voxels 'on' in the mask.
     coords = np.array(zip(*np.where(mask)))
 
-    streamlines = []
-    for c in coords:
-        for line in X:
-            streamlines.append((line + c) * voxel_size)
+    streamlines = [(line + c) * voxel_size for line in X for c in coords]
```
effigies, line 48:
```suggestion
    streamlines = [(line + c) * voxel_size for c in coords for line in X]
```

nibabel/tests/test_volumeutils.py
```diff
@@ -947,9 +947,7 @@ def test_seek_tell():
                 # zeros in their wake.  BZ2Files can't seek when writing,
                 # unless we enable the write0 flag to seek_tell
                 # ZstdFiles also does not support seek forward on write
-                if (not write0 and
-                    (in_file == 'test.bz2' or
-                     in_file == 'test.zst')):  # Can't seek write in bz2, zst
+                if (not write0 and in_file in ('test.bz2', 'test.zst')):
```
effigies, line 950:
```suggestion
                if not write0 and in_file in ('test.bz2', 'test.zst'):
```

effigies (commented):
Overall looks reasonable. Thanks! Two suggestions.

effigies (approved):

PR #1155, opened 2022-12-13
TEST: Use class variables and close open file handles in ECAT tests

DimitriPapadopoulos:
`ResourceWarning: unclosed file <_io.BufferedReader name='/home/runner/work/nibabel/nibabel/virtenv/lib/python3.9/site-packages/nibabel/tests/data/tinypet.v'>`

### DIFF
--- a/nibabel/tests/test_ecat.py
+++ b/nibabel/tests/test_ecat.py
@@ -59,7 +59,7 @@ class TestEcatHeader(tws._TestWrapStructBase):
             hdr.get_data_dtype()
 
     def test_header_codes(self):
-        fid = open(ecat_file, 'rb')
+        fid = open(self.example_file, 'rb')
         hdr = self.header_class()
         newhdr = hdr.from_fileobj(fid)
         fid.close()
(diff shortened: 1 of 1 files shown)


effigies:
Tests should be fixed if you rebase on master...

DimitriPapadopoulos:
I am still getting errors, they seem related to CI instability.

effigies:
Hmm. Looks like there might be a between-test dependency that can get broken by `pytest-xdist`. I ran several times and thought I would have caught a non-determinism...

effigies:
Okay, let's try again...

effigies:
This doesn't seem to have eliminated the warning.

DimitriPapadopoulos:
Removing all warnings requires a rewrite of the testing classes.

In any case, this does the right thing by changing `ecat_file` → `self.example_file`.

effigies:
Yeah, I think we need to move to a different fixture approach to ensure better handling of resources. Are you good to merge this as-is, or are there things you still want to try to do here?

DimitriPapadopoulos:
I'm good, I can rewrite the commit message if needed.

PR #1156, opened 2022-12-13
BF: Fix error message

DimitriPapadopoulos:


nibabel/ecat.py
```diff
@@ -611,8 +611,8 @@ def _get_oriented_data(self, raw_data, orientation=None):
         elif orientation == 'radiological':
             orientation = patient_orient_radiological[0]
         else:
-            raise ValueError('orientation should be None,\
-                neurological or radiological')
+            raise ValueError('orientation should be None, '
```
effigies, line 614: Might be cleaner just to put it on one line. Does that pass `flake8`?

DimitriPapadopoulos: Indeed, let's try.

effigies:
Looks like it didn't raise any alarms in https://github.com/nipy/nibabel/actions/runs/3707291213/jobs/6283525379. Thanks!

PR #1157, opened 2022-12-16
CI: Run tests with available parallelism

effigies:
Uses pytest-xdist to hopefully reduce the amount of time we're spending on tests.

One check needs to be made more tolerant of this case, as it assumed that two tests could edit variables in a shared scope.


effigies:
We get an ~2x speedup, so will merge tomorrow unless there's some reason not to.

PR #1158, opened 2022-12-18
FIX: Ensure exclusive access to dft database

effigies:
Using `pytest-xdist`, it turns out we have some racing tests in `test_dft.py`. This updates the database connection to allow sqlite's own locking mechanisms to ensure consistency. While seeing what was necessary, I refactored the `dft.DB` object.

The essential change is:

```diff
-     DB = sqlite3.connect(DB_FNAME, check_same_thread=False)
+     DB = sqlite3.connect(DB_FNAME, isolation_level='EXCLUSIVE')
```

But I find the current approach more readable.


effigies:
@chaselgrove @matthew-brett Any chance of a review? Making databases multiprocessing safe is not something I've done a lot of.

chaselgrove:
> @chaselgrove @matthew-brett Any chance of a review? Making databases multiprocessing safe is not something I've done a lot of.

@effigies I can't say I ever have!  So if pytest-xdist is happy, then that's a win.  The changes as a whole look fine to me.  There is the side effect that the database file is created when `dft` is imported (rather than when the database is accessed, as before), but I won't lose sleep over it.

effigies:
The database was always initialized at import, but I wouldn't be opposed to shifting it to first access.

effigies:
I'm going to merge to get tests passing again. Happy to make some post-merge adjustments if y'all see any.

PR #1161, opened 2022-12-28
Fixed documentation build error

### LINKED ISSUE #1160: Docs build failure
Sphinx's conf.py tries to read a missing `[metadata]` attribute from the *setup.cfg* file, which causes a build error.

Possible solution: use the `toml` library to read the required metadata from *pyproject.toml*.

ZviBaratz:
The *conf.py* file included an extraction of a `[metadata]` attribute from *setup.cfg* using `ConfigParser`. Since the metadata was moved to *pyproject.toml*, this raised an exception. To solve this, I added the `toml` library to the doc requirements and replaced `ConfigParser.read()` with `toml.load()`. Sphinx also required a version update to build successfully. This PR resolved #1160.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,8 +10,9 @@ dev =
 doc =
     matplotlib >= 1.5.3
     numpydoc
-    sphinx >=0.3,<3
+    sphinx ~= 5.3
     texext
+    toml
 minc2 =
     h5py
 spm =
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ Thumbs.db
 # Things specific to this project #
 ###################################
 doc/source/reference
+doc/source/generated
 venv/
 .buildbot.patch
 .vscode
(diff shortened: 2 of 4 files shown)


matthew-brett:
MRG: update doc build

Fixes for configuration change.

setup.cfg
```diff
@@ -10,8 +10,9 @@ dev =
 doc =
     matplotlib >= 1.5.3
     numpydoc
-    sphinx >=0.3,<3
+    sphinx ~= 5.3
     texext
+    toml
```
effigies, line 15: Just a note that we should be using `tomli; python_version < "3.11"` and importing as:

```Python
try:
    import tomllib
except ImportError:
    import tomli as tomllib
```

The toml package is unmaintained, and `tomli` has become `tomllib` in the stdlib.

ZviBaratz: My bad, resolved in 4b693b84fd267f4a53a65070487c2d90d8f64ad6.

PR #1163, opened 2022-12-28
Fixed documentation build error and warnings

### LINKED ISSUE #1162: RST warnings on documentation build
There are some warnings raised during the build error re RST and docstring formatting.

ZviBaratz:
This resolves #1162, however, there are still some documents not included in any toctree, which raises warnings.

```
.../nibabel/doc/source/devel/bv_formats.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/devel/data_pkg_discuss.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/devel/data_pkg_uses.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/devel/image_design.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/devel/scaling.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/installing_data.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/notebooks/index.rst: WARNING: document isn't included in any toctree
.../nibabel/doc/source/old/ioimplementation.rst: WARNING: document isn't included in any toctree
```

Should any of these be removed? Or included in any existing / new section?


doc/source/devel/biaps/biap_0003.rst
```diff
@@ -199,8 +199,10 @@ The base level header will usually also have image metadata fields giving
 information about the whole image. A field is an "image metadata field" if it
 is defined at the top level of the header.  For example::
 
->>> hdr = dict(nipy_header_version='1.0',
-...            Manufacturer="SIEMENS")
+.. code-block:: python
+
+  hdr = dict(nipy_header_version='1.0',
+             Manufacturer="SIEMENS")
```
effigies, line 205: Will indentation not do? This is inconsistent with the rest of the doc.

```suggestion
      >>> hdr = dict(nipy_header_version='1.0',
      ...            Manufacturer="SIEMENS")
```

ZviBaratz:
@effigies there are also suddenly doctest failures that weren't there before. Any idea what heppened?

effigies:
I think there's something that happened with upgrading sphinx that changed the behavior of autosummary. I believe that the `:toctree: reference` had something to do with it, but I haven't been able to figure it out yet.

ZviBaratz:
Are these really meant to be doctests? If so, shouldn't we either include the imports or add `doctest_global_setup` to *conf.py*?

effigies:
Yes, the idea is that all of the example code should run, so we know the documentation is valid. Could be that there was an implicit context in earlier sphinx that now needs to be managed directly.

ZviBaratz:
No problem. Do you think it would be better to import all the relevant objects in the global context or per test? E.g., in *quaternions.py*:

```python
>>> import numpy as np
>>> from nibabel.quaternions import quat2mat  # Added
>>> q = [0, 1, 0, 0] # 180 degree rotation around axis 0
>>> M = quat2mat(q) # from this module
>>> vec = np.array([1, 2, 3]).reshape((3,1)) # column vector
>>> tvec = np.dot(M, vec)
```

or add all the required imports in `doctest_global_setup`?

effigies:
Wait, so doctests in the current file don't have access to the variables in that file? That seems like a regression in sphinx, or possibly some option that was switched from defaulting true to false.

effigies:
Seems this is a known issue: https://github.com/sphinx-doc/sphinx/issues/6590

ZviBaratz:
From what I gather, it doesn't seem like there's an easy workaround, nor that this will be resolved any time soon. Any suggestions?

effigies:
Let's just import the current module.

ZviBaratz:
There's a `ModuleNotFoundError` for `pytest`, I tried adding it to the workflow's dependencies installation step, but strangely it's still there. Will have more time to try to figure it out during the weekend. Meanwhile, any ideas would be very welcome.

doc/source/conf.py
```diff
@@ -69,9 +73,10 @@
               'matplotlib.sphinxext.plot_directive',
               ]
 
-# the following doesn't work with sphinx < 1.0, but will make a separate
-# sphinx-autogen run obsolete in the future
-#autosummary_generate = True
+# Autosummary always wants to use a `generated/` directory.
+# We generate with `make api-stamp`
+# This could change in the future
+autosummary_generate = False
```
effigies, line 79: @matthew-brett Is the `make api-stamp` replaceable with some invocation of `autosummary`? I don't fully understand everything that's going on here.

effigies:
I'm going to merge this. We can do additional fixups if you spot anything...

PR #1167, opened 2023-01-01
DOC: Removed long description from info.py

### LINKED ISSUE #1166: Remove README duplication from info.py
The current *README.rst* file is generated as part of the [release checklist](https://nipy.org/nibabel/devel/make_release.html#release-checklist) using the `make refresh-readme` command. This is no longer required for RST files to be displayed correctly on PyPI and may be removed.

ZviBaratz:
RST no longer requires special handling for correct rendering on PyPI. Closes #1166.


ZviBaratz:
I apologize for the mess – 
This bit is really complicating things:
https://github.com/nipy/nibabel/blob/d0e63635869608b51ab763c276963b095ab045f4/nibabel/__init__.py#L12-L38
I'm assuming the reason the quickstart is included in the source directory is so that it is tested.
Can we think of a better alternative? Using the `include` directive isn't supported on PyPI, and therefore `twine check` fails when I try to include the README contents in the source directory. I would like to avoid duplication as much as possible, but I think we might have to simply move the quickstart section to the README and have a copy maintained within the top-level init.

effigies:
I think the idea was that there are three contexts in which we want to render the same text:

1) Package description on PyPI. It's just `long_description`.
2) README on GitHub. This will have CI/coverage shields.
3) Docs landing page. This will need additional content.

I'm not really sure that the need for some sync mechanism among these has changed. I'm inclined to leave this as it is, as it is still single-sourced and we have a documented process for keeping them in sync.

ZviBaratz:
Got it. There's something about the duplication that just feels wrong, but I see how it might be simpler to rely on the current setup. Thank you for clarifying.

PR #1170, opened 2023-01-01
MNT: Officially deprecate nisext

effigies:
@matthew-brett Does this seem like enough, or do we need to be doing anything else to warn the (possibly nonexistent) user base?


matthew-brett:
Looks fine to me.  We should should probably get Dipy to remove their use of Nisext in their `Makefile`.

effigies:
@skoudoro Do you all still use these rules? https://github.com/dipy/dipy/blob/089a02ad0520895bd30370591f965c685293e480/Makefile#L80-L97

PR #1171, opened 2023-01-01
MNT: Replace versioneer with setuptools_scm

effigies:
setuptools_scm now has git-archive support, is better maintained than versioneer (I say as the de facto maintainer of versioneer), can be used with other build backends than setuptools, and generally a smaller footprint in the package.

I'm interested to see if this passes all of our tests.


effigies:
This is ready for review.

nibabel/pkg_info.py
```diff
@@ -85,20 +96,25 @@ def pkg_commit_hash(pkg_path=None):
...
-        hash_from = 'archive substitution'
-    elif versions['version'] == '0+unknown':
-        hash_from, hash_str = '(none found)', '<not found>'
-    else:
-        hash_from = 'repository'
-    return hash_from, hash_str
-
-
-def get_pkg_info(pkg_path):
+    from subprocess import run
```
ZviBaratz, line 99: Is there any benefit to importing this within the function rather than at the top of the module? AFAIK, as long as there are no dependency issues or critical performance considerations, the latter is generally encouraged.

ZviBaratz (approved):
Awesome!

nibabel/pkg_info.py
effigies: The idea was that we don't import subprocess outside this very rarely-used function, but we do import numpy which imports subprocess, so I guess it makes no difference.

effigies:
Thanks for the review!

PR #1172, opened 2023-01-03
RF: Rely on stdlib implementations for tmpdirs tools

effigies:
I noticed that we have almost an identical implementation of `TemporaryDirectory` as the stdlib (https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), so I scheduled it for deprecation and removal.

With the coming [chdir](https://docs.python.org/3/library/contextlib.html#contextlib.chdir) context manager, the remaining tools can be made convenience wrappers on stdlib 1-or-2 liners, so I went ahead and did it. IMO it's a bit cleaner to read a single function with the setup-yield-teardown structure than a two-or-three method class.

### DIFF
--- a/nibabel/tests/test_removalschedule.py
+++ b/nibabel/tests/test_removalschedule.py
@@ -15,6 +15,12 @@ MODULE_SCHEDULE = [
 
 OBJECT_SCHEDULE = [
     (
+        '8.0.0',
+        [
+            ('nibabel.tmpdirs', 'TemporaryDirectory'),
+        ],
+    ),
+    (
         '7.0.0',
         [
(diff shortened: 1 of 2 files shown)


ZviBaratz (approved):
LGTM. 
If you'd like, I think using `pathlib` instead of `os` for everything besides `chdir` would be preferable.

effigies:
From an API perspective, it's good to accept either `str` or `Path,` but maintaining existing return types is important. These will accept either input, but have been emitting strings. We would need to raise a `FutureWarning` to switch to returning `Path`s, which hardly seems worth it when people can already wrap with `Path()` if they want it.

Without that change, using the `pathlib` API internally seems like it would make this more complicated, not less.

PR #1173, opened 2023-01-03
ENH: Schedule int64 warning to convert to error at 5.0

effigies:
Follow-up to #1082, which will need to transition from a warning to an error on the release of 5.0 (which I'm aiming for next week).

Created a new function in `nibabel.deprecated` to turn what we need to do here into a template that can be used elsewhere.


effigies:
@matthew-brett Any opinions?

PR #1175, opened 2023-01-03
ENH: Support multiline header fields in TCK

effigies:
@mattcieslak's initial PR was lost in the refactor by @anibalsolon in #1140. This PR takes the data file shared by @valeriejill, with just the first streamline (zeroed out) to confirm that a header that was failing to be read now can be. I've also added an extra `command_history` line later in the header to ensure that we aggregate duplicate keys instead of overwrite, in response to https://github.com/nipy/nibabel/pull/957#issuecomment-737673131 by @thijsdhollander.

@valeriejill can you confirm that we have permission to publish this file as part of our test data? If not, I think I've learned enough about working with the format to make my own.

@anibalsolon @MarcCote I would appreciate code reviews from you.

Closes #957.


effigies:
Any chance of a review? I'm comfortable with the test coverage and will plan to merge tomorrow unless someone asks me to wait.

mattcieslak:
This looks good to me, thanks for revisiting it @effigies!

effigies:
Thanks for the review! And thanks for your patience with the... sporadic activity here.

MarcCote:
Thanks @effigies for maintaining this code :).

anibalsolon:
Getting back-
Nice! Hopefully there's no colon in new lines 🤷🏻‍♂️ it would be great if new lines were escaped (\\n) when mrtrix writes them

PR #1176, opened 2023-01-06
MNT: Switch to hatchling build backend

### LINKED ISSUE #1087: Reorganize/compress test files to reduce package footprint
It would be great if the file nibabel/nibabel/nicom/tests/data/4d_multiframe_test.dcm could be compressed, similar to several other test files in this folder. The command `pigz -11 -n 4d_multiframe_test.dcm` will reduce the file from 58.8mb to 68kb. This reduces the disk requirements for nibabel.

effigies:
I've become increasingly frustrated with setuptools across various projects, in large part due to the weird blend of backward compatibility and churn related to trying to support modern standards.

#1133 converted most of our build metadata to backend-agnostic `pyproject.toml` metadata.
#1171 converted the rest and replaced the setuptools-specific versioneer with the (surprisingly) backend-agnostic setuptools_scm.

This PR finishes the job by selecting a new backend, and one that actually works better with `setuptools_scm` by separating version calculation from version file writing and both of these from determining which files belong in the sdist or wheel. Even better, you can separately decide what goes in an sdist and what goes in a wheel.

See [Hatch](https://hatch.pypa.io/) and [hatch-vcs](https://github.com/ofek/hatch-vcs) for further details.

This will also let us resolve #1087 in the short term by excluding it from the wheel/installation, while keeping it in the sdist for testing purposes.

Fixes #1087.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# See also nibabel/.gitignore for test data rules that
+# are meant to make you think before you `git add --force`
+
 # Editor temporary/working/backup files #
 #########################################
 .#*
@@ -35,23 +38,6 @@
 *.o
 *.py[oc]
 *.so
-
-# Packages #
-############
-# it's better to unpack these files and commit the raw source
-# git has its own built in compression methods
-*.7z
-*.bz2
-*.bzip2
-*.dmg
-*.gz
-*.iso
-*.jar
-*.rar
-*.tar
-*.tbz2
-*.tgz
-*.zip
 
 # Python files #
 ################
--- a/.gitignore
+++ b/.gitignore
@@ -51,12 +51,14 @@ dist/
 .coverage
 .ropeproject/
 htmlcov/
+.*_cache/
 
 # Logs and databases #
 ######################
 *.log
 *.sql
 *.sqlite
+*.sqlite3
 
 # OS generated files #
 ######################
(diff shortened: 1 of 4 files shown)


effigies:
Seems it likes to exclude .gz data files by default. Not sure if that's a bug or configuration choice.

ofek:
> Seems it likes to exclude .gz data files by default. Not sure if that's a bug or configuration choice.

- https://hatch.pypa.io/latest/config/build/#vcs
- https://github.com/nipy/nibabel/blob/02c7a3429cad8bca7272da80a2eda0334b016d15/.gitignore#L47

effigies:
Thanks, @ofek! I was thinking it might be something along those lines but ran out of energy to fiddle last night.

effigies:
@ZviBaratz Would you mind reviewing if you have a couple minutes?

ZviBaratz (approved):
Looks good :+1: 
Thank you for taking this on. Hatch looks really great, and I haven't had the chance to use it yet. I think we could probably leverage it further down the road to simplify some of our testing and CI workflows, etc.

effigies:
> I think we could probably leverage it further down the road to simplify some of our testing and CI workflows, etc.

Sure. Feel free to look into this. My next project along the lines of these PRs will be to purge setuptools' pkg_resources for pkgutil and importlib-resources.

PR #1178, opened 2023-01-06
ENH: Add type annotations for image API (FileBasedImage and DataobjImage)

effigies:
Since this much is done, I'm putting it up here. If anybody is up for a review before the release, this can go in 5.0. If not, it can wait for 5.1.

SpatialImage is next on my list, but I didn't want to hold this up if it could get in.

### DIFF
--- a/nibabel/arrayproxy.py
+++ b/nibabel/arrayproxy.py
@@ -25,11 +25,13 @@ The proxy API is - at minimum:
 
 See :mod:`nibabel.tests.test_proxy_api` for proxy API conformance checks.
 """
+import typing as ty
 import warnings
 from contextlib import contextmanager
 from threading import RLock
 
 import numpy as np
+import numpy.typing as npt
 
 from . import openers
 from .fileslice import canonical_slicers, fileslice
@@ -53,7 +55,25 @@ raised.
 KEEP_FILE_OPEN_DEFAULT = False
 
 
-class ArrayProxy:
+@ty.runtime_checkable
+class ArrayLike(ty.Protocol):
+    """Protocol for numpy ndarray-like objects
+
+    This is more stringent than :class:`numpy.typing.ArrayLike`, but guarantees
+    access to shape, ndim and slicing.
+    """
+
+    shape: tuple[int, ...]
+    ndim: int
+
+    def __array__(self, dtype: npt.DTypeLike | None = None, /) -> npt.NDArray:
+        ...
+
+    def __getitem__(self, key, /) -> npt.NDArray:
+        ...
+
+
+class ArrayProxy(ArrayLike):
     """Class to act as proxy for the array that can be read from a file
 
     The array proxy allows us to freeze the passed fileobj and header such that
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -10,6 +10,8 @@
 from __future__ import annotations
 
 import io
+import os
+import typing as ty
 from copy import deepcopy
 from typing import Type
 from urllib import request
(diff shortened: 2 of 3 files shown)


nibabel/arrayproxy.py
```diff
@@ -53,7 +57,24 @@
...
+    """Protocol for numpy ndarray-like objects
+
+    This is more stringent than :class:`numpy.typing.ArrayLike`, but guarantees
+    access to shape, ndim and slicing.
+    """
+
+    shape: tuple[int, ...]
+    ndim: int
+
+    def __array__(self, dtype: npt.DTypeLike | None = None, /) -> npt.NDArray:
```
ZviBaratz, line 73: I'm sorry if I'm missing something, but shouldn't this be:
```suggestion
    def __array__(self, *, dtype: npt.DTypeLike | None = None) -> npt.NDArray:
```
I think `/` is only used for positional arguments, whereas keyword arguments require `*`. Sorry if I'm missing something.

nibabel/filebasedimages.py
```diff
@@ -18,6 +20,10 @@
 from .filename_parser import TypesFilenamesError, splitext_addext, types_filenames
 from .openers import ImageOpener
 
+FileSpec = ty.Union[str, os.PathLike]
```
ZviBaratz, line 23: Since we're already going for 3.10 style annotation, we might as well:
```suggestion
FileSpec = str | os.PathLike
```
Shouldn't we?

ZviBaratz (commented):
Assuming I'm just missing something in the first comment re `/`, I think it looks great. 

Also, I think using:
```python
CACHING_OPTIONS = ty.Literal['fill', 'unchanged']
MMAP_OPTIONS = ty.Literal['c', 'r']
```
and maybe even:
```python
HeaderType = FileBasedHeader | ty.Mapping | None
```
could be nice, in terms of more concise function signatures and code readability. But it might also be too much. Your call.

nibabel/arrayproxy.py
effigies:
```
numpy.ndarray.__array__ = __array__(...)
    a.__array__([dtype], /) -> reference if type unchanged, copy otherwise.
    
    Returns either a new reference to self if dtype is not given or a new array
    of provided data type if dtype is different from the current dtype of the
    array.
```

nibabel/filebasedimages.py
effigies: This line gets run by the interpreter, so fails before Python 3.10. I can try hiding it behind `if ty.TYPE_CHECKING:`. but then I can't import `FileSpec` into other modules without guarding that behind `ty.TYPE_CHECKING` either.

nibabel/arrayproxy.py
ZviBaratz: Huh. Yeah. I am missing something. If you understand why/how and wouldn't mind explaining sometime, I'd be happy to learn.

nibabel/filebasedimages.py
ZviBaratz: I see. Thank you for clarifying. Definitely sounds like keeping the `Union` makes more sense in this case.

nibabel/arrayproxy.py
ZviBaratz: I thought that once a default value is assigned (including `None`, as it is here), the parameter is immediately considered a keyword argument, and that `/` is used only for positional arguments. 
Looking at [the relevant PEP](https://peps.python.org/pep-0570/#syntax-and-semantics), I learned that: 

> The logic around optional values for positional-only parameters remains the same as for positional-or-keyword parameters.

PR #1179, opened 2023-01-07
ENH: Annotate SpatialImage API, improve  superclass annotations

effigies:
The most interesting thing about annotating the `SpatialImage` was annotating the slicer, which needs to know what kind of thing it returns when sliced. This required learning about [generic types](https://docs.python.org/3/library/typing.html#typing.Generic) and [type variables](https://docs.python.org/3/library/typing.html#typing.TypeVar).

To get a sense of how they work, here's an excerpt:

```Python
SpatialImgT = ty.TypeVar('SpatialImgT', bound='SpatialImage')

class SpatialFirstSlicer(ty.Generic[SpatialImgT]):
    def __init__(self, img: SpatialImgT): ...
    def __getitem__(self, slicer: object) -> SpatialImgT: ...

class SpatialImage:
    @property
    def slicer(self: SpatialImgT) -> SpatialFirstSlicer[SpatialImgT]: ...
```

`SpatialImgT` will match any type that is a subclass of `SpatialImage`. So when we call `Nifti1Image().slicer` we get back a `SpatialFirstSlicer[Nifti1Image]`. So `Nifti1Image().slicer[:, :, :, 0]` will have type `Nifti1Image`.

The type variables also helped with figuring out the right types for the `@classmethod`s, which I then went back and added to filebasedimages and dataobj_images. This led to some mypy complaints deeper in the hierarchy. I don't think any were actually bugs, but we don't lose functionality by "fixing" them.

### DIFF
--- a/nibabel/filebasedimages.py
+++ b/nibabel/filebasedimages.py
@@ -24,6 +24,11 @@ FileSpec = ty.Union[str, os.PathLike]
 FileMap = ty.Mapping[str, FileHolder]
 FileSniff = ty.Tuple[bytes, str]
 
+ImgT = ty.TypeVar('ImgT', bound='FileBasedImage')
+HdrT = ty.TypeVar('HdrT', bound='FileBasedHeader')
+
+StreamImgT = ty.TypeVar('StreamImgT', bound='SerializableImage')
+
 
 class ImageFileError(Exception):
     pass
@@ -33,7 +38,7 @@ class FileBasedHeader:
     """Template class to implement header protocol"""
 
     @classmethod
-    def from_header(klass, header=None):
+    def from_header(klass: type[HdrT], header: FileBasedHeader | ty.Mapping | None = None) -> HdrT:
         if header is None:
             return klass()
         # I can't do isinstance here because it is not necessarily true
@@ -47,19 +52,19 @@ class FileBasedHeader:
         )
 
     @classmethod
-    def from_fileobj(klass, fileobj: io.IOBase):
-        raise NotImplementedError
+    def from_fileobj(klass: type[HdrT], fileobj: io.IOBase) -> HdrT:
+        raise NotImplementedError  # pragma: no cover
 
-    def write_to(self, fileobj: io.IOBase):
-        raise NotImplementedError
+    def write_to(self, fileobj: io.IOBase) -> None:
+        raise NotImplementedError  # pragma: no cover
 
-    def __eq__(self, other):
-        raise NotImplementedError
+    def __eq__(self, other: object) -> bool:
+        raise NotImplementedError  # pragma: no cover
 
-    def __ne__(self, other):
+    def __ne__(self, other: object) -> bool:
         return not self == other
 
-    def copy(self) -> FileBasedHeader:
+    def copy(self: HdrT) -> HdrT:
         """Copy object to independent representation
 
         The copy should not be affected by any changes to the original
@@ -153,6 +158,7 @@ class FileBasedImage:
     """
 
     header_class: Type[FileBasedHeader] = FileBasedHeader
+    _header: FileBasedHeader
     _meta_sniff_len: int = 0
     files_types: tuple[tuple[str, str | None], ...] = (('image', None),)
     valid_exts: tuple[str, ...] = ()
@@ -186,7 +192,7 @@ class FileBasedImage:
         self._header = self.header_class.from_header(header)
         if extra is None:
             extra = {}
-        self.extra = extra
+        self.extra = dict(extra)
 
         if file_map is None:
             file_map = self.__class__.make_file_map()
@@ -196,7 +202,7 @@ class FileBasedImage:
     def header(self) -> FileBasedHeader:
         return self._header
 
-    def __getitem__(self, key):
+    def __getitem__(self, key) -> None:
         """No slicing or dictionary interface for images"""
         raise TypeError('Cannot slice image objects.')
 
@@ -221,7 +227,7 @@ class FileBasedImage:
         characteristic_type = self.files_types[0][0]
         return self.file_map[characteristic_type].filename
 
-    def set_filename(self, filename: str):
+    def set_filename(self, filename: str) -> None:
         """Sets the files in the object from a given filename
 
         The different image formats may check whether the filename has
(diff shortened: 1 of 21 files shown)


nibabel/filebasedimages.py
```diff
@@ -239,16 +245,16 @@ def set_filename(self, filename: str):
...
 
     @classmethod
-    def from_file_map(klass, file_map: FileMap):
-        raise NotImplementedError
+    def from_file_map(klass: type[ImgT], file_map: FileMap) -> ImgT:
+        raise NotImplementedError  # pragma: no cover
 
     @classmethod
-    def filespec_to_file_map(klass, filespec: FileSpec):
+    def filespec_to_file_map(klass, filespec: FileSpec) -> FileMap:
```
ZviBaratz, line 257:
```suggestion
    def filespec_to_file_map(klass: type[ImgT], filespec: FileSpec) -> FileMap:
```
There are a few `classmethod`s here with this annotation and a few without. Seems to me like it would be better to keep it consistent and annotate all of them, if it doesn't break anything.

nibabel/spatialimages.py
```diff
@@ -152,21 +185,33 @@ class HeaderTypeError(Exception):
...
+    _dtype: np.dtype
+    _shape: tuple[int, ...]
+    _zooms: tuple[float, ...]
 
-    def __init__(self, data_dtype=np.float32, shape=(0,), zooms=None):
+    def __init__(
+        self,
+        data_dtype: npt.DTypeLike = np.float32,
+        shape: Sequence[int] = (0,),
+        zooms: Sequence[float] | None = None,
```
ZviBaratz, line 202: These should be replaced with `collections.abc.Sequence` at some point, when support for 3.8 is dropped. I realize there's time, but perhaps we can keep track of it somewhere? Or should we just wait for the errors that will be raised in some future Python version?

nibabel/spatialimages.py
```diff
@@ -349,21 +416,21 @@ def check_slicing(self, slicer, return_spatial=False):
...
                 raise IndexError('New axis not permitted in spatial dimensions')
             elif isinstance(subslicer, int):
                 raise IndexError(
                     'Scalar indices disallowed in spatial dimensions; Use `[x]` or `x:x+1`.'
                 )
-        return spatial_slices if return_spatial else slicer
+        return spatial_slices if return_spatial else canonical
 
-    def slice_affine(self, slicer):
+    def slice_affine(self, slicer: object) -> np.ndarray:
```
ZviBaratz, line 433: Again, I'm probably missing something, but just making sure `object` is the best choice here.

ZviBaratz (approved):

nibabel/filebasedimages.py
```diff
@@ -47,19 +52,19 @@ def from_header(klass, header=None):
...
-        raise NotImplementedError
+    def __eq__(self, other: object) -> bool:
+        raise NotImplementedError  # pragma: no cover
 
-    def __ne__(self, other):
+    def __ne__(self, other: object) -> bool:
         return not self == other
 
-    def copy(self) -> FileBasedHeader:
+    def copy(self: HdrT) -> HdrT:
```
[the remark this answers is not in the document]
effigies: The idea here is that you have a `.copy()` being called on a `FileBasedHeader` instance, which will almost always be some subclass, and return an instance of the same type. To assert that the result is not just a `FileBasedHeader` but the same subclass of `FileBasedHeader`, we use a type variable bound to `FileBasedHeader`.

As it was, `Nifti1Header().copy().get_sform()` would not type check. Now it will.

nibabel/filebasedimages.py
```diff
@@ -47,19 +52,19 @@ def from_header(klass, header=None):
...
+        raise NotImplementedError  # pragma: no cover
 
-    def write_to(self, fileobj: io.IOBase):
-        raise NotImplementedError
+    def write_to(self, fileobj: io.IOBase) -> None:
+        raise NotImplementedError  # pragma: no cover
 
-    def __eq__(self, other):
-        raise NotImplementedError
+    def __eq__(self, other: object) -> bool:
```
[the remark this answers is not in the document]
effigies: `ty.Any` would mean that the internals do not get type checked and you might raise an `AttributeError` or something else that will break rather than return True/False. There's a good StackOverflow thread: https://stackoverflow.com/questions/57905233/how-to-annotate-eq-when-it-uses-an-attribute

nibabel/filebasedimages.py
```diff
@@ -47,19 +52,19 @@ def from_header(klass, header=None):
         )
 
     @classmethod
-    def from_fileobj(klass, fileobj: io.IOBase):
-        raise NotImplementedError
+    def from_fileobj(klass: type[HdrT], fileobj: io.IOBase) -> HdrT:
+        raise NotImplementedError  # pragma: no cover
```
ZviBaratz, line 56: From the SO link you sent re the `object` vs. `Any` comment, I also learned that it is better to:
```suggestion
        return NotImplementedError  # pragma: no cover
```
[ref](https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror)

nibabel/filebasedimages.py
effigies: Those `classmethods` return an image of a given type. Here the return type does not depend on the `klass` type, so we don't need to use a type variable. Using one type variable would fail in any case. If we want to be thorough, we could do:

```suggestion
    def filespec_to_file_map(klass: FileBasedImage, filespec: FileSpec) -> FileMap:
```

But this is just the default annotation, like saying `x: int = 1`. I'm -0 on it as clutter, but if you feel strongly I'm okay with adding it.

nibabel/spatialimages.py
```diff
@@ -131,18 +131,51 @@
 """
 from __future__ import annotations
 
-from typing import Type
+import io
+import typing as ty
+from typing import Literal, Sequence
```
effigies, line 136:
```suggestion
from collections.abc import Sequence
from typing import Literal
```

nibabel/spatialimages.py
effigies: Made a suggestion above.

nibabel/spatialimages.py
effigies: Here the point is that you really can pass anything to this. We narrow the type by using `check_slicing`, which narrows it by using `canonical_slicers`. If it would be helpful, I have some updates to `flieslice` (where `canonical_slicers` is defined). I didn't want to muddy this PR, but I can pull those in if you like.

ZviBaratz: No need, thank you. Just wanted to double-check with you (and learn).

effigies:
Awesome, thanks for the review! I hope you don't mind me continuing to bug you with these. Feel free to say you don't have time or don't want to...

PR #1182, opened 2023-01-13
ENH: Set symmetric threshold for identifying unit quaternions in qform calculation

### LINKED ISSUE #1181: Output of `Nifti1Header.get_qform()` returns different affine values than `fslhd`
Hi there! I'm trying to track down discrepancies between the qform outputs of both `nibabel` and [`fslhd`](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/Fslutils), a FSL command for printing Nifti header metadata. 

For [this sample image](https://github.com/nipy/nibabel/files/10394137/sample_image.nii.gz), the relevant output of `fslhd` is:

```
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	-0.625000 0.000000 -0.000000 80.000000 
qto_xyz:2	0.000000 0.570521 1.469992 -102.004517 
qto_xyz:3	-0.000000 -0.255207 3.286208 -137.370850 
qto_xyz:4	0.000000 0.000000 0.000000 1.000000 
sform_name	Scanner Anat
sform_code	1
sto_xyz:1	-0.625000 0.000000 -0.000000 80.000000 
sto_xyz:2	0.000000 0.570521 1.469992 -102.004517 
sto_xyz:3	0.000000 -0.255207 3.286208 -137.370850 
sto_xyz:4	0.000000 0.000000 0.000000 1.000000 
```

However, when I load the same image using `nibabel`, then call `header.get_sform()` and `header.get_qform()`, I see the following:

```python
# Avoid exp notation for easier readability
>>> import numpy as np
>>> np.set_printoptions(suppress=True)

# Load image
>>> import nibabel as nib
>>> im_sample = nib.load("sample_image.nii.gz")

# Print sform and qform
>>> sform = im_sample.header.get_sform()
>>> print(sform)
[[  -0.625         0.           -0.           80.        ]
 [   0.            0.57052135    1.46999204 -102.0045166 ]
 [   0.           -0.25520656    3.28620815 -137.37084961]
 [   0.            0.            0.            1.        ]]
>>> qform = im_sample.header.get_qform()
>>> print(qform)
[[  -0.62499996    0.00004513   -0.00121767   80.        ]
 [  -0.00004513    0.57052135    1.46999205 -102.0045166 ]
 [  -0.0002114    -0.25520655    3.28620791 -137.37084961]
 [   0.            0.            0.            1.        ]]
```

The `sform` matches expectations, but the `qform` does not.

* Do you know how these discrepancies might have arisen? (Perhaps, differing methods of [computing the `qform`](https://nipy.org/nibabel/nifti_images.html#the-qform-affine)?)
* And, do you know what the "correct" interpretation of the qform is?

Thanks much!

effigies:
The `quaternions.fillpositive` function is intended to augment a partial quaternion to make a unit 4-vector by finding `w`, given `(x, y, z)`, such that `|(w, x, y, z)| == 1`. Floating point precision error means that a `(w, x, y, z)` selected to be a unit vector can still sometimes result in `|(w, x, y, z)| != 1` by a small amount. Currently if `1 - |(x, y, z)| > -3*eps`, we call that close enough to zero that `w = 0` will still produce a unit quaternion.

This is asymmetric, as if `1 - 3*eps < |(x, y, z)| < 1`, we will get `w = sqrt(1 - |(x, y, z)|) > 0`, which can result in a surprising failure to preserve an affine that encodes a rotation in only one plane (see #1181 for an example). This is not exactly wrong, but the asymmetry is unsatisfying and inconsistent with `fslhd` at the least.

I'm starting this PR with a (failing) test to assert that a passing `fillpositive()` unit 3-vector, up to the precision allowed by its dtype, will result in a 0 `w`.

Following failing tests, I will use the change proposed in https://github.com/nipy/nibabel/issues/1181#issuecomment-1379211496 to satisfy the constraint imposed by the test.

Closes #1181.

### DIFF
--- a/nibabel/tests/test_quaternions.py
+++ b/nibabel/tests/test_quaternions.py
@@ -108,7 +108,7 @@ def test_fillpositive_simulated_error(dtype):
     # as xyz with small error, and we want to recover the w of 0
 
     # Permit 1 epsilon per value (default, but make explicit here)
-    w2_thresh = 3 * -np.finfo(dtype).eps
+    w2_thresh = 3 * np.finfo(dtype).eps
 
     pos_error = neg_error = False
     for _ in range(50):
--- a/nibabel/tests/test_quaternions.py
+++ b/nibabel/tests/test_quaternions.py
@@ -91,6 +91,15 @@ def test_fillpositive_plus_minus_epsilon(dtype):
     assert nq.fillpositive(plus)[0] == 0.0
     assert nq.fillpositive(minus)[0] == 0.0
 
+    # |(x, y, z)| > 1, no real solutions
+    plus = baseline * nptype(1 + 2 * np.finfo(dtype).eps)
+    with pytest.raises(ValueError):
+        nq.fillpositive(plus)
+
+    # |(x, y, z)| < 1, two real solutions, we choose positive
+    minus = baseline * nptype(1 - 2 * np.finfo(dtype).eps)
+    assert nq.fillpositive(minus)[0] > 0.0
+
 
 @pytest.mark.parametrize('dtype', ('f4', 'f8'))
 def test_fillpositive_simulated_error(dtype):
@@ -107,18 +116,7 @@ def test_fillpositive_simulated_error(dtype):
     for _ in range(50):
         xyz = norm(gen_vec(dtype))
 
-        wxyz = nq.fillpositive(xyz, w2_thresh)
-        assert wxyz[0] == 0.0
-
-        # Verify that we exercise the threshold
-        magnitude = xyz @ xyz
-        if magnitude < 1:
-            pos_error = True
-        elif magnitude > 1:
-            neg_error = True
-
-    assert pos_error, 'Did not encounter a case where 1 - |xyz| > 0'
-    assert neg_error, 'Did not encounter a case where 1 - |xyz| < 0'
+        assert nq.fillpositive(xyz, w2_thresh)[0] == 0.0
 
 
 def test_conjugate():
--- a/nibabel/tests/test_quaternions.py
+++ b/nibabel/tests/test_quaternions.py
@@ -16,6 +16,18 @@ from numpy.testing import assert_array_almost_equal, assert_array_equal
 from .. import eulerangles as nea
 from .. import quaternions as nq
 
+
+def norm(vec):
+    # Return unit vector with same orientation as input vector
+    return vec / np.sqrt(vec @ vec)
+
+
+def gen_vec(dtype):
+    # Generate random 3-vector in [-1, 1]^3
+    rand = np.random.default_rng()
+    return rand.uniform(low=-1.0, high=1.0, size=(3,)).astype(dtype)
+
+
 # Example rotations
 eg_rots = []
 params = (-pi, pi, pi / 2)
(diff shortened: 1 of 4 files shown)


nibabel/tests/test_quaternions.py
```diff
@@ -69,6 +69,20 @@ def test_fillpos():
     assert wxyz[0] == 0.0
 
 
+@pytest.mark.parametrize('dtype', ('f4', 'f8'))
+def test_fillpos_precision(dtype):
```
matthew-brett, line 73: Is the idea that one of these 50 random vectors will generate a small negative difference from zero?  Worth testing that explicitly?

matthew-brett: Would this be easier to read with some example of a positive and negative error?  Here we are leaving it to chance whether the test fails, are we not?

effigies:
Suddenly `os.getcwd()` is producing `FileNotFoundError`s that I can't reproduce locally...

nibabel/tests/test_quaternions.py
effigies: Just to be clear, you're suggesting two vectors that are exemplars of positive and negative error? Would random generation work, as long as we verify that we get them?

```Python
def make_unit_vector(shape, dtype, constraint=lambda x: True, tries=50):
    for _ in range(tries):
        vec = rand.uniform(low=-1, high=1, size=shape).astype(dtype)
        vec /= np.sqrt(vec @ vec)
        if constraint(vec):
            return vec

short = make_unit_vector((3,), dtype, lambda x: x @ x < 1)
long = make_unit_vector((3,), dtype, lambda x: x @ x > 1)

assert nq.fillpositive(short, w2_thresh)[0] == 0
assert nq.fillpositive(long, w2_thresh)[0] == 0
```

matthew-brett: It seems to me cleaner to find some examplar vectors that we know give positive and negative < threshold differences, and a > threshold difference, and then test some random vectors - rather than hoping (even with reasonable expectation) that we'll get suitable vectors.

nibabel/tests/test_quaternions.py
```diff
@@ -69,6 +74,53 @@ def test_fillpos():
...
+
+    # Obviously, |(x, y, z)| == 1
+    baseline = np.array([0, 0, 1], dtype=dtype)
+
+    # Obviously, |(x, y, z)| ~ 1
+    plus = baseline * nptype(1 + np.finfo(dtype).eps)
+    minus = baseline * nptype(1 - np.finfo(dtype).eps)
+
+    assert nq.fillpositive(plus)[0] == 0.0
+    assert nq.fillpositive(minus)[0] == 0.0
```
effigies, line 92: @matthew-brett How's this for a deterministic test?

nibabel/tests/test_quaternions.py
```diff
@@ -69,6 +74,53 @@ def test_fillpos():
...
+
+
+@pytest.mark.parametrize('dtype', ('f4', 'f8'))
+def test_fillpositive_simulated_error(dtype):
+    # Nondeterministic test for fillpositive threshold
+    # Create random vectors, normalize to unit length, and count on floating point
+    # error to result in magnitudes larger/smaller than one
+    # This is to simulate cases where a unit quaternion with w == 0 would be encoded
+    # as xyz with small error, and we want to recover the w of 0
```
effigies, line 101: I still like this as a bit of a stress test that will hit more realistic rounding errors, and added a comment to explain more. Can drop it if it seems excessive.

effigies:
@matthew-brett Gentle bump on this, if you get a few minutes.

nibabel/tests/test_quaternions.py
matthew-brett: Nice - but - consider pusing to threshold to show where it fails?

nibabel/tests/test_quaternions.py
```diff
@@ -69,6 +74,53 @@ def test_fillpos():
...
+        assert wxyz[0] == 0.0
+
+        # Verify that we exercise the threshold
+        magnitude = xyz @ xyz
+        if magnitude < 1:
+            pos_error = True
+        elif magnitude > 1:
+            neg_error = True
+
+    assert pos_error, 'Did not encounter a case where 1 - |xyz| > 0'
```
matthew-brett, line 120: But do you still want to fail the tests here, on a random criterion?

matthew-brett (commented):
Thanks - nice cleanups.  A couple of comments.

nibabel/tests/test_quaternions.py
effigies: Sure. It actually fails at 2*eps, since the error compounds. Updated.

nibabel/tests/test_quaternions.py
effigies: Good call. Just went ahead and checked the frequencies:

```Python
In [26]: vec = np.array([np.sqrt(np.sum(norm(gen_vec('f4'))**2)) for _ in range(10000)])

In [27]: np.unique(np.sign(1 - vec), return_counts=True)
Out[27]: (array([-1.,  0.,  1.], dtype=float32), array([  81, 6946, 2973]))
```

Seems negative errors are much less likely and we might have failed out for no good reason. I had assumed that the majority would be non-zero error and roughly split in either direction.

PR #1188, opened 2023-01-29
TYP: Annotate deprecation, one-time property, and optional package tooling

effigies:
These are mostly self-contained pieces that allow us to transparently handle edge cases. So annotating them is relatively free of rabbit holes and should help avoid unnoticed `Any`s as we annotate portions of the code that they do wrap.

### DIFF
--- a/nibabel/deprecated.py
+++ b/nibabel/deprecated.py
@@ -2,11 +2,14 @@
 """
 from __future__ import annotations
 
+import typing as ty
 import warnings
-from typing import Type
 
 from .deprecator import Deprecator
 from .pkg_info import cmp_pkg_version
+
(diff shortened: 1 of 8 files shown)


nibabel/deprecated.py
```diff
@@ -30,14 +33,14 @@ class ModuleProxy:
     module.
     """
 
-    def __init__(self, module_name):
+    def __init__(self, module_name: str):
```
ZviBaratz, line 36:
```suggestion
    def __init__(self, module_name: str) -> None:
```
I realize having the type annotation for *module_name* makes this addition functionally redundant, but if you don't disagree, I think it's better to be explicit.

ZviBaratz (approved):

PR #1189, opened 2023-01-29
TYP: Annotate volumeutils

effigies:
This PR has two significant issues that required changes:

1) `Recoder`s set instance variables based on their input parameters. These variables are always dict-like, but we can't know their names. Adding and annotating `__getattr__` resolves this.
2) The `DtypeMapper` used to allow for `Recoder`s with numpy dtypes was dict-ish, but not quite enough to make annotating it straightforward. Subclassing from `dict` was sufficient.

Otherwise, it's a fiddly file, so it needs a careful check that I haven't changed the logic or missed renaming a variable.

### DIFF
--- a/nibabel/volumeutils.py
+++ b/nibabel/volumeutils.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import gzip
 import sys
+import typing as ty
 import warnings
 from collections import OrderedDict
 from functools import reduce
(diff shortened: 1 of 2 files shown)


ZviBaratz:
Sorry, busy couple of days ahead. Will review beginning of next week if it's nothing urgent.

effigies:
Please do not consider any review requests urgent. I'm treating it as a "this is ready" notification.

ZviBaratz (approved):

effigies:
Thanks, Zvi!

PR #1190, opened 2023-01-30
MNT: Finish removing distutils and setuptools dependencies

effigies:
`distutils` is removed in Python 3.12. Looking at https://github.com/python/cpython/blob/3467991/Lib/distutils/ccompiler.py#L919-L956, it turns out the single place we were still importing it was redundant with checking `os.name`.

Setuptools has recommended against using `pkg_resources` for some time, and it seems that `importlib.resources` has settled on an API that has been stable since Python 3.9. We can use the `importlib_resources` backport for Python 3.8.

### DIFF
--- a/nibabel/tests/test_testing.py
+++ b/nibabel/tests/test_testing.py
@@ -15,8 +15,8 @@ from ..testing import (
     data_path,
     error_warnings,
     get_fresh_mod,
+    get_test_data,
     suppress_warnings,
-    test_data,
 )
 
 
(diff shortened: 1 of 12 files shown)


effigies:
This is pretty straightforward, IMO, if anybody wants to do a review. Drive-bys welcome.

Will merge at the end of the week if no reviews.

nibabel/tests/test_testing.py
```diff
@@ -171,12 +171,14 @@ def test_assert_re_in(regex, entries):
...
-    assert test_data() == data_path
-    assert test_data() == os.path.abspath(
+    assert str(test_data()) == str(data_path)
+    assert str(test_data()) == os.path.abspath(
         os.path.join(os.path.dirname(__file__), '..', 'tests', 'data')
     )
     for subdir in ('nicom', 'gifti', 'externals'):
-        assert test_data(subdir) == os.path.join(data_path[:-10], subdir, 'tests', 'data')
+        assert str(test_data(subdir)) == os.path.join(
+            data_path.parent.parent, subdir, 'tests', 'data'
```
ZviBaratz, line 180: Is `data_path` a `Path` instance? If so, wouldn't it be more natural to use the `/` operator instead of `os.path.join`? Also, maybe the assertions could use the `samefile` method?

ZviBaratz (approved):
👍🏼 
Left one comment with something I wasn't sure about.

nibabel/tests/test_testing.py
effigies: I added fa5f9207fbc1bee9e39bac865c80afb6987e13e1 to address this. LMK what you think.

effigies:
@ZviBaratz Thanks for the review. In addition to fa5f9207fbc1bee9e39bac865c80afb6987e13e1, I also added ad439f5e9f9c3c65d16969683b08cb15b37d7ee4 to rename `test_data` to `get_test_data` while we're here. This resolves a complaint from pytest whenever we import the function at the module level of a test file:

```
PytestReturnNotNoneWarning: Expected None, but nibabel/cmdline/tests/test_conform.py::test_data returned '/home/chris/Projects/nipy/nibabel/nibabel/tests/data', which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
```

Figured I'd throw it in while I was here.

nibabel/tests/test_testing.py
ZviBaratz: LGTM. 
I might, at some point in the future, if you don't think it's a bad idea, make a refactoring PR with a bunch of more explicit variable naming suggestions, e.g., here my instinct would be to do something like:
```python
expected = data_path.parent.parent / subdir / 'tests' / 'data'
assert str(test_data(subdir)) == expected
```

effigies: You're more than welcome to.

PR #1194, opened 2023-02-02
BF: Support ragged voxel arrays in ParcelsAxis

MichielCottaar:
In the past we used `np.asanyarray(voxels)`, which would produce an array with dtype="object" if provided with a ragged array. 
This no longer works in numpy 1.24, where the following error is raised:
```
>       as_array = np.asanyarray(voxels)
E       ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (4,) + inhomogeneous part.

../nibabel/nibabel/cifti2/cifti2_axes.py:778: ValueError
```

I was notified of this bug by @pauldmccarthy .

I've adjusted the logic a bit to ensure that `self.voxels` is a 1D array. Dach element is supposed to be the set of voxels in that parcel (i.e., a (N, 3) array for N voxels in that parcel). This avoids calling `np.asanyarray` directly on the supplied voxels, which might produce a multi-dimensional rather than a 1D array if all parcels contain the same number of voxels.

### DIFF
--- a/nibabel/cifti2/cifti2_axes.py
+++ b/nibabel/cifti2/cifti2_axes.py
@@ -776,8 +776,8 @@ class ParcelsAxis(Axis):
         """
         self.name = np.asanyarray(name, dtype='U')
         self.voxels = np.empty(len(voxels), dtype='object')
-        for idx in range(len(voxels)):
-            self.voxels[idx] = voxels[idx]
+        for idx, vox in enumerate(voxels):
+            self.voxels[idx] = vox
         self.vertices = np.asanyarray(vertices, dtype='object')
         self.affine = np.asanyarray(affine) if affine is not None else None
         self.volume_shape = volume_shape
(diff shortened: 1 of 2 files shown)


nibabel/cifti2/cifti2_axes.py
```diff
@@ -775,14 +775,9 @@ def __init__(self, name, voxels, vertices, affine=None, volume_shape=None, nvert
...
-        if as_array.ndim == 1:
-            voxels = as_array.astype('object')
-        else:
-            voxels = np.empty(len(voxels), dtype='object')
-            for idx in range(len(voxels)):
-                voxels[idx] = as_array[idx]
-        self.voxels = np.asanyarray(voxels, dtype='object')
+        self.voxels = np.empty(len(voxels), dtype='object')
+        for idx in range(len(voxels)):
+            self.voxels[idx] = voxels[idx]
```
effigies, line 780:
```suggestion
        for idx, vox in enumerate(voxels):
            self.voxels[idx] = vox
```

effigies (approved):
LGTM. One small suggestion.

effigies:
Thanks Michiel!

PR #1197, opened 2023-02-14
TYPE: Annotate file-handling modules

effigies:
This PR annotates the following modules:

* `filename_parser`, which covers coercion of `os.PathLike`s to `str`, and detecting related files based on extensions
* `fileholders`, which allows us to carry around filenames or open streams somewhat agnostically
* `openers`, which allows us to read/write from filenames or open streams, and transparently enables compression

### DIFF
--- a/nibabel/fileholders.py
+++ b/nibabel/fileholders.py
@@ -7,6 +7,10 @@
 #
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Fileholder class"""
+from __future__ import annotations
+
+import io
+import typing as ty
 from copy import copy
 
 from .openers import ImageOpener
@@ -19,7 +23,12 @@ class FileHolderError(Exception):
 class FileHolder:
     """class to contain filename, fileobj and file position"""
 
-    def __init__(self, filename=None, fileobj=None, pos=0):
+    def __init__(
+        self,
+        filename: str | None = None,
+        fileobj: io.IOBase | None = None,
+        pos: int = 0,
+    ):
         """Initialize FileHolder instance
 
         Parameters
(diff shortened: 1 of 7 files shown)


nibabel/fileholders.py
```diff
@@ -87,12 +95,15 @@ def same_file_as(self, other):
...
+    def file_like(self) -> str | io.IOBase | None:
         """Return ``self.fileobj`` if not None, otherwise ``self.filename``"""
         return self.fileobj if self.fileobj is not None else self.filename
 
 
-def copy_file_map(file_map):
+FileMap = ty.Mapping[str, FileHolder]
+
+
+def copy_file_map(file_map: FileMap) -> FileMap:
```
ZviBaratz, line 106: Definitely nothing critical, but this might be a little more accurate as:

```suggestion
FileMap = ty.Mapping[str, FileHolder]
FileDict = dict[str, FileHolder]


def copy_file_map(file_map: FileMap) -> FileDict:
```

ZviBaratz (approved):

nibabel/fileholders.py
effigies: I doubt anybody else will bother constructing this type, and if they do it would be wild to use something less convenient than dict.

ZviBaratz: So why not:
```suggestion
FileMap = dict[str, FileHolder]


def copy_file_map(file_map: FileMap) -> FileMap:
```
?

effigies: Ah right, we can't subscript `dict` until a later version than 3.8 (it's either 3.9 or 3.10) unless we hide it behind a `ty.TYPE_CHECKING` guard.

PR #1199, opened 2023-02-17
FIX: Disable direct creation of non-conformant GiftiDataArrays

### LINKED ISSUE #1198: Nibabel creates invalid GIfTI files
The GIfTI format specification is [available from here](https://www.nitrc.org/projects/gifti/). Section `2.3.4.2 DataType` notes `This required attribute describes the numeric type of the data contained in a Data Array and are limited to the types displayed in the table.`

| Value                                 | Description                                              |
| ------------------------ | -------------------------------------- | 
| NIFTI_TYPE_UINT8         | Unsigned, 8-bit bytes.                            |
| NIFTI_TYPE_INT32          | Signed, 32-bit integers.                         |
| NIFTI_TYPE_FLOAT32     | 32-bit single precision floating point.  |

However, the following code creates an invalid GIfTI file which popular tools can't read, using `DataType="NIFTI_TYPE_FLOAT64" `

```python
# %% make_surface.py
import pathlib

import nibabel as nib

from nilearn import datasets, surface

fsaverage = datasets.fetch_surf_fsaverage("fsaverage7")

motor_images = datasets.fetch_neurovault_motor_task()
stat_img = motor_images.images[0]
surface_map = surface.vol_to_surf(stat_img, fsaverage.pial_left)

surface_map_path = "./surface_map.gii"

img = nib.gifti.gifti.GiftiImage()
img.add_gifti_data_array(
    nib.gifti.gifti.GiftiDataArray(
        surface_map,
        intent="NIFTI_INTENT_ZSCORE",
    )
)
nib.save(img, surface_map_path)
```

For example, [Connectome Workbench](https://github.com/Washington-University/workbench) is unable to view this image:

connectome

This is similar to previous issues where nibabel used excessive integer precision for [GIfTI](https://github.com/nipy/nibabel/issues/792) and [NIfTI](https://github.com/nipy/nibabel/issues/1046) images. However, it is easier to establish the optimal data type for discrete integers rather than floating point data. For science, it is common to use float64 for internal calculations (15-17 significant decimal places) to avoid rounding errors, but for storage float32 is typically more than sufficient (about 7 decimal places). If there is a strong rationale for increasing the precision for GIfTI, the specification should be updated and implementation support provided to the smaller teams that create popular tools that support this format.

effigies:
GIFTI only defines three valid datatypes. Nibabel has generally allowed users to use any defined `NIFTI_TYPE`. In limiting this scope, there's a balance to strike between:

1) Making it easy/default to generate spec-conforming GIFTI files
2) Permitting loading of non-conforming GIFTI files (including some previously generated with nibabel)
3) Respecting the user's intent when made explicit, even when it violates the spec

(1) is the primary goal here. For (2) making ourselves intentionally unable to read a file seems like a very bad idea.

(3) seems like an open question. We do not uniformly allow users to violate standards, such as allowing fixed-length strings as the voxel data in NIfTI, even though numpy would permit it. So it could be fine to simply refuse to write files with invalid `DataArray` types. On the other hand, you could imagine a pipeline of functions that take in a `GiftiImage` and return a `GiftiImage` to ensure metadata follows data, and it's only at write where you would really want to enforce a precision reduction. (It would in any event be very difficult to entirely prevent the construction in memory of non-conformant images, due to Python's object model.)

---

My initial thought was that we could auto-convert to the nearest dtype, but that could silently introduce unexpected results for users, so @matthew-brett suggested raising errors. This is a fairly simple API change.

Because the objects are constructed piecemeal by the parser, this does not prevent loading files. We might still want to find a way to warn on load.

A couple options for making it harder to write a non-conforming GIFTI:

1) Raise error during `GiftiImage.to_xml()` if any array dtype is not conforming. This could be overridden with a keyword argument `force`.
2) Raise warning during `GiftiImage.__init__()` if any data arrays have non-conforming dtypes, or if they are added through API mechanisms (`GiftiImage.add_gifti_data_array()`).

Closes #1198.

### DIFF
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -88,7 +88,8 @@ class GiftiMetaData(CaretMetaData):
         <GiftiMetaData {'key': 'val'}>
         >>> GiftiMetaData({"key": "val"})
         <GiftiMetaData {'key': 'val'}>
-        >>> nvpairs = GiftiNVPairs(name='key', value='val')
+        >>> with pytest.deprecated_call():
+        ...     nvpairs = GiftiNVPairs(name='key', value='val')
         >>> with pytest.warns(FutureWarning):
         ...     GiftiMetaData(nvpairs)
         <GiftiMetaData {'key': 'val'}>
--- a/nibabel/gifti/gifti.py
+++ b/nibabel/gifti/gifti.py
@@ -16,7 +16,8 @@ from __future__ import annotations
 import base64
 import sys
 import warnings
-from typing import Type
+from copy import copy
+from typing import Type, cast
 
 import numpy as np
 
@@ -26,6 +27,12 @@ from ..deprecated import deprecate_with_version
 from ..filebasedimages import SerializableImage
 from ..nifti1 import data_type_codes, intent_codes, xform_codes
 from .util import KIND2FMT, array_index_order_codes, gifti_encoding_codes, gifti_endian_codes
+
+GIFTI_DTYPES = (
+    data_type_codes['NIFTI_TYPE_UINT8'],
+    data_type_codes['NIFTI_TYPE_INT32'],
+    data_type_codes['NIFTI_TYPE_FLOAT32'],
+)
 
 
 class _GiftiMDList(list):
@@ -462,11 +469,7 @@ class GiftiDataArray(xml.XmlSerializable):
         if datatype is None:
             if self.data is None:
                 datatype = 'none'
-            elif self.data.dtype in (
-                np.dtype('uint8'),
-                np.dtype('int32'),
-                np.dtype('float32'),
-            ):
+            elif data_type_codes[self.data.dtype] in GIFTI_DTYPES:
                 datatype = self.data.dtype
             else:
                 raise ValueError(
@@ -848,20 +851,45 @@ class GiftiImage(xml.XmlSerializable, SerializableImage):
             GIFTI.append(dar._to_xml_element())
         return GIFTI
 
-    def to_xml(self, enc='utf-8') -> bytes:
+    def to_xml(self, enc='utf-8', *, mode='strict') -> bytes:
         """Return XML corresponding to image content"""
+        if mode == 'strict':
+            if any(arr.datatype not in GIFTI_DTYPES for arr in self.darrays):
+                raise ValueError(
+                    'GiftiImage contains data arrays with invalid data types; '
+                    'use mode="compat" to automatically cast to conforming types'
+                )
+        elif mode == 'compat':
+            darrays = []
+            for arr in self.darrays:
+                if arr.datatype not in GIFTI_DTYPES:
+                    arr = copy(arr)
+                    # TODO: Better typing for recoders
+                    dtype = cast(np.dtype, data_type_codes.dtype[arr.datatype])
+                    if np.issubdtype(dtype, np.floating):
+                        arr.datatype = data_type_codes['float32']
+                    elif np.issubdtype(dtype, np.integer):
+                        arr.datatype = data_type_codes['int32']
+                    else:
+                        raise ValueError(f'Cannot convert {dtype} to float32/int32')
+                darrays.append(arr)
+            gii = copy(self)
+            gii.darrays = darrays
+            return gii.to_xml(enc=enc, mode='strict')
+        elif mode != 'force':
+            raise TypeError(f'Unknown mode {mode}')
         header = b"""<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE GIFTI SYSTEM "http://www.nitrc.org/frs/download.php/115/gifti.dtd">
 """
         return header + super().to_xml(enc)
 
     # Avoid the indirection of going through to_file_map
-    def to_bytes(self, enc='utf-8'):
-        return self.to_xml(enc=enc)
+    def to_bytes(self, enc='utf-8', *, mode='strict'):
+        return self.to_xml(enc=enc, mode=mode)
 
     to_bytes.__doc__ = SerializableImage.to_bytes.__doc__
 
-    def to_file_map(self, file_map=None, enc='utf-8'):
+    def to_file_map(self, file_map=None, enc='utf-8', *, mode='strict'):
         """Save the current image to the specified file_map
 
         Parameters
@@ -877,7 +905,7 @@ class GiftiImage(xml.XmlSerializable, SerializableImage):
         if file_map is None:
             file_map = self.file_map
         with file_map['image'].get_prepare_fileobj('wb') as f:
-            f.write(self.to_xml(enc=enc))
+            f.write(self.to_xml(enc=enc, mode=mode))
 
     @classmethod
     def from_file_map(klass, file_map, buffer_size=35000000, mmap=True):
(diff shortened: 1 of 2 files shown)


neurolabusc:
The primary strength of GIfTI is that it is the standard interchange format for our community, and it was designed from the outset to support domain-specific needs such as describing the anatomical structures, labeling, and spatial transforms. However, GIfTI has inherently [poor speed and file size](https://github.com/neurolabusc/MeshFormatsJS). 

My sense is that datatypes of GIfTI were carefully chosen to match the needs of our community (e.g. SNR of our data is low) and the capabilities of our hardware (APIs and GPUs limit of 2 billion vertices and vertex precision limited to float32).  While a user might intentionally request extreme datatypes believing that `more is better`, I suspect that they will experience a lot of unintended consequences. 

I appreciate @satra's ability to see the big picture, while I tend to focus on implementation details. Perhaps he can share his thoughts.

effigies:
Yes, GIFTI is an interchange format, but the nibabel objects are not very heavy. So for internal use, it's not unreasonable for someone to create a `float64` `GiftiDataArray` or even `GiftiImage` and pass it from function to function, and only want to cast down to `float32` at serialization time.

I can also see a use for wanting to inspect the object without conversion, possibly through a schema-agnostic XML viewer.

So I would suggest something like the following on `GiftiImage`:

```Python
class GiftiImage:
    def to_xml(self, enc='utf-8', *, mode='strict'):  # Called by to_filename()
        if mode == 'strict':
            if any(arr.datatype not in ('uint8', 'int32', 'float32')):
                raise ValueError(
                    'GiftiImage contains data arrays with invalid data types; '
                    'use mode="compat" to automatically cast to conforming types'
                )
        elif mode == 'compat':
            gii = copy(self)
            # Convert unsupported float/int types to float32 or int32 if possible
            return gii.to_xml(enc=enc, mode='strict')
        elif mode != 'force':
            raise TypeError(f'Unknown mode {mode}')
        ...
```

Combined with explicit datatypes, building out a non-conformant image by hand would require two specific overrides:

```Python
import numpy as np
import nibabel as nb

darr = nb.gifti.GiftiDataArray(np.zeros((5,), dtype=np.float64), datatype='NIFTI_TYPE_FLOAT64')
gii = nb.GiftiImage([darr])
gii.to_filename('64bit.gii', mode='force')
```

This would allow someone to create `GiftiDataArray`s internally and pass them around freely without premature casts. If they attempt to write them to disk without casting, they will get an error unless compatible, so we avoid silent changes in behavior (@matthew-brett's concern). They can then choose whether to coerce to be conformant with the specification or to force writing to disk.

If the force write use case is still uncompelling and two overrides are insufficient friction, I would be okay with dropping it. For loading 64-bit GIFTIs in another nibabel instance, pickling will be much faster: 

```Python
In [3]: img = nb.load('/data/openneuro/ds002790/.git/annex/objects/2j/3z/MD5E-s3485910--1942b389e2cc
   ...: f637934bf1d62c6f20c2.surf.gii/MD5E-s3485910--1942b389e2ccf637934bf1d62c6f20c2.surf.gii')

In [4]: from pickle import dumps, loads

In [5]: %timeit loads(dumps(img))
2.27 ms ± 28.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit img.from_bytes(img.to_bytes())
308 ms ± 507 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

And of course, you can still inspect a data array by hand without putting it in a `<GIFTI Version="1.0"></GIFTI>` wrapper.

```Python
In [7]: _ = img.darrays[0].to_xml()
```

effigies:
Went ahead and pushed a proposal since I wanted to test it. Not intended to preempt discussion; happy to consider alternatives.

effigies:
Thanks for the comments, @alexisthual. Happy to see further discussion or a code review. In the absence of both, I'll try to look at this with fresh eyes some time this week and plan to merge next Monday.

nibabel/gifti/gifti.py
```diff
@@ -834,20 +852,45 @@ def _to_xml_element(self):
...
+                if arr.datatype not in GIFTI_DTYPES:
+                    arr = copy(arr)
+                    # TODO: Better typing for recoders
+                    dtype = cast(np.dtype, data_type_codes.dtype[arr.datatype])
+                    if np.issubdtype(dtype, np.floating):
+                        arr.datatype = data_type_codes['float32']
+                    elif np.issubdtype(dtype, np.integer):
+                        arr.datatype = data_type_codes['int32']
+                    else:
+                        raise ValueError(f'Cannot convert {dtype} to float32/int32')
```
alexisthual, line 875: I get how this part corrects the darray's `datatype` attribute, but I don't understand where the actual data is casted to the new datatype

nibabel/gifti/tests/test_gifti.py
```diff
@@ -126,6 +128,44 @@ def assign_metadata(val):
...
 
 
+@pytest.mark.parametrize('label', data_type_codes.value_set('label'))
+def test_image_typing(label):
+    dtype = data_type_codes.dtype[label]
+    if dtype == np.void:
+        return
+    arr = 127 * rng.random(
+        20,
+    )
```
alexisthual, line 138: This could be one line I guess :slightly_smiling_face:

alexisthual (commented):
I'm not familiar with the codebase so I might be missing a lot of important points.
What is missing thus far? Some warnings for non-conformant legacy images?

nibabel/gifti/gifti.py
```diff
@@ -834,20 +852,45 @@ def _to_xml_element(self):
             GIFTI.append(dar._to_xml_element())
         return GIFTI
 
-    def to_xml(self, enc='utf-8') -> bytes:
+    def to_xml(self, enc='utf-8', *, mode='strict') -> bytes:
```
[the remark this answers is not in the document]
effigies: The `*` means you have to call `mode=` as a keyword argument. So `to_xml('utf-8', 'force')` will fail, `to_xml('utf-8', mode='force')` will pass. I would be inclined to make `enc` keyword-only as well, I just didn't want to make it part of this PR...

nibabel/gifti/gifti.py
effigies: The data is only cast at write time. Serialization to XML will call `DataArray._to_xml_element()`:

https://github.com/nipy/nibabel/blob/3a4cc5e327c785774b58bf2dee0510bbff9f044f/nibabel/gifti/gifti.py#L487-L522

Which calls `_data_tag_element()`:

https://github.com/nipy/nibabel/blob/3a4cc5e327c785774b58bf2dee0510bbff9f044f/nibabel/gifti/gifti.py#L371-L391

L380 is the one that finally does it: `out = np.asanyarray(dataarray, dtype).tobytes(order)`

nibabel/gifti/tests/test_gifti.py
effigies:
```suggestion
    arr = 127 * rng.random(20)
```

nibabel/gifti/gifti.py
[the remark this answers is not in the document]
alexisthual: That's interesting, thanks!

effigies:
Thanks for the review, @alexisthual!

PR #1200, opened 2023-02-18
DOC: Standardize module-level docstrings format

ZviBaratz:
The majority of nibabel's docstrings do not contain a period in the summary. The exceptions stick out a little in [the API docs](https://nipy.org/nibabel/api.html#api). This PR fixes these exceptions and homogenizes spacing within and after module-level docstrings.


effigies:
Thanks!

Issue #1201, opened 2023-02-18
Documentation site build error


ZviBaratz:
The execution of [`build_modref_templates.py`](https://github.com/nipy/nibabel/blob/master/doc/tools/build_modref_templates.py) in [`doc/Makefile`](https://github.com/nipy/nibabel/blob/cbd7690143d9c203e1a8ebdf67bb5832f719db06/doc/Makefile#L46) raises multiple `AttributeError`s from `nibabel._version`.

effigies:
You need to run `hatch build --hooks-only` to populate that file.

ZviBaratz:
@effigies Which file? Also, I am still getting the same exceptions.

effigies:
Hmm. Maybe try `pip install -e .`. If you have a non-editable install, then you might be getting a version mismatch.

effigies:
And what's your command? I always just do `make -C doc html` and that works for me if I've previously run `pip install -e ".[doc]"`.

ZviBaratz:
I am running an editable install, and I was using `make html` from within `doc/`. Tried both re-running the editable installation, and using `make -C doc html` from the package root, but I am still getting the same messages 😕

effigies:
Want to share the errors?

ZviBaratz:
Of course:
```
make: Entering directory '/home/zvi/Projects/nibabel/doc'
mkdir -p ../build/html ../build/doctrees
sphinx-build -b html -d ../build/doctrees   source ../build/html
Running Sphinx v5.3.0
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 2 changed, 0 removed
reading sources... [100%] reference/nibabel._version                                                                                                                                          
/home/zvi/Projects/nibabel/doc/source/reference/nibabel._version.rst:8: WARNING: autosummary: failed to import NotThisMethod.
Possible hints:
* ModuleNotFoundError: No module named 'nibabel._version.NotThisMethod'; 'nibabel._version' is not a package
* AttributeError: module 'nibabel._version' has no attribute 'NotThisMethod'
* ValueError: not enough values to unpack (expected 2, got 1)
* ModuleNotFoundError: No module named 'NotThisMethod'
* KeyError: 'NotThisMethod'
/home/zvi/Projects/nibabel/doc/source/reference/nibabel._version.rst:8: WARNING: autosummary: failed to import VersioneerConfig.
Possible hints:
[...]

PR #1202, opened 2023-02-18
RF: doc.tools partial refactoring

ZviBaratz:
There's still a lot of work to do here, but I hope this counts as an improvement.

### DIFF
--- a/doc/tools/apigen.py
+++ b/doc/tools/apigen.py
@@ -16,15 +16,21 @@ This is a modified version of a script originally shipped with the PyMVPA
 project, then adapted for use first in NIPY and then in skimage. PyMVPA
 is an MIT-licensed project.
 """
+from __future__ import annotations
 
-# Stdlib imports
+import contextlib
 import os
 import re
 from inspect import getmodule
-from types import BuiltinFunctionType, FunctionType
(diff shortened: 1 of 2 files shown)


effigies:
I have resisted the urge to update these modules more than absolutely necessary because I think they are older than current autodoc capabilities and we may be able to full-on remove them and do something more standard with Sphinx. I just haven't figured out how, as my patience with Sphinx is a finite resource.

I'm okay reviewing this, but it might be worth checking in first for future improvements, as deeper changes might be a better use of your time.

ZviBaratz:
Yeah, that makes sense. I started refactoring it after I ran into #1201 and thought it must be something in there. I figured it would be a good way to get acquainted with that module and hopefully find out the source of the problem as I go – by the time I was done, I also couldn't help but think that this definitely needs to ultimately be replaced with `autodoc`.

No need to waste your time on this then; I'll see if I can come up with something sometime soon.

effigies:
It would be good to get @matthew-brett's take, or possibly @Garyfallidis, as dipy is mentioned in the source. I'm curious if there's a known migration path from these tools or they still serve a particular purpose.

PR #1205, opened 2023-02-21
DOC: Merged duplicate README and index sections

### LINKED ISSUE #1204: Duplicate installation section in the documentation site
The README update in #1195 included an [installation section](https://github.com/nipy/nibabel#installation) which wasn't there before and creates [a redundancy](https://nipy.org/nibabel/index.html#download-and-installation) in the documentation site. 
Looking at the [download and installation instructions](https://nipy.org/nibabel/installation.html#installation) page, my impression is that it might be possible to remove that page altogether. There mighe be a couple of bits which could be ported to other locations, but nothing crucial. WDTY?

ZviBaratz:
Resolves #1204.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -118,26 +118,40 @@ When working on NiBabel itself, it may be useful to install in "editable" mode::
 For more information on previous releases, see the `release archive`_ or
 `development changelog`_.
 
+Please find detailed `download and installation instructions
+<installation>`_ in the manual.
+
 .. _current release: https://pypi.python.org/pypi/NiBabel
-.. _release archive: https://github.com/nipy/NiBabel/releases
 .. _development changelog: https://nipy.org/nibabel/changelog.html
+.. _installation: https://nipy.org/nibabel/installation.html#installation
+.. _release archive: https://github.com/nipy/NiBabel/releases
 
-Mailing List
-============
+Support
+=======
 
-Please send any questions or suggestions to the `neuroimaging mailing list
-<https://mail.python.org/mailman/listinfo/neuroimaging>`_.
+If you have problems installing the software or questions about usage,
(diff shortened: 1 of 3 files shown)


effigies:
Okay, so it looks like the reason things ended up like this is that there's no way to unify reference links like ``:ref:`heading` `` and `` `text <heading>`_`` in Sphinx. We could do something like manually dereference so `links_names.rst` contains something like:

```RST
.. _appendix of the manual: ./legal.html#license
.. _installation: ./installation.html#installation
```

And README.rst could have:

```RST
.. _appendix of the manual: https://nipy.org/nibabel/legal.html#license
.. _installation: https://nipy.org/nibabel/installation.html#installation
```

It's not great, but we don't really reorganize much.

ZviBaratz:
Sorry if I'm missing something, but couldn't we just use the second option anywhere there's a `:ref:`? 
E.g., we could replace:
```rst
* :ref:`User Documentation <manual>` (manual)
```
with:
```rst
* `User Documentation <manual>`_ (manual)
```
and include:
```rst
.. _manual: https://nipy.org/nibabel/manual.html#manual
```
in `links_names.txt`? 

Also, seems to me like the content in parentheses under the [Documentation](https://nipy.org/nibabel/index.html#id1) heading could be removed for a cleaner look.

effigies:
The reason not to use an absolute URL in `links_names.rst` is so that the link stays within site for testing. Right now if I test with `python -m http.server -d build`, I get `https://nipy.org/nibabel/legal.html` instead of `http://0.0.0.0:8000/legal.html`. If for some reason we moved things around, we wouldn't see the breakage until we uploaded the docs to `gh-pages`.

ZviBaratz:
@effigies please see the proposed changes. Hope I understood correctly.

Issue #1207, opened 2023-02-27
cant usse nibabel imagestats


Factral:
I installed the latest version from the repository and it doesn't let you use this module

![image](https://user-images.githubusercontent.com/74687828/221685171-d7fd5fb9-76e8-413e-acfb-c62fb5db68fc.png)

effigies:
It is not imported into `__init__.py`. For now, `import nibabel.imagestats`. Would you mind opening a PR to import `imagestats` into `__init__.py`?

PR #1208, opened 2023-02-27
ENH: Make imagestats module available on ``import nibabel``

Factral:
this add the import of imagestats to __init__.py

context of imagestats: #952 

Solve #1207

### DIFF
--- a/nibabel/__init__.py
+++ b/nibabel/__init__.py
@@ -44,6 +44,7 @@ from . import nifti1 as ni1
 from . import spm2analyze as spm2
 from . import spm99analyze as spm99
 from . import streamlines, viewers
+from . import imagestats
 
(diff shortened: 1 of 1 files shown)


Factral:
I already run pre-commit but I'm getting this error, any idea why? I did not touch any other file or configuration

image

effigies:
Ah, for the mypy one, you need to run `hatch build --hooks-only`. Or simpler `pip install -e .`.

Factral:
> Ah, for the mypy one, you need to run `hatch build --hooks-only`. Or simpler `pip install -e .`.

thanks!, the error was because i was trying to commit from windows 😂

effigies:
Thanks! The failures are #1209, so unrelated to this.

PR #1212, opened 2023-03-10
RF: Pull compression detection logic into a central private module

### LINKED ISSUE #1209: Failing zstd tests
The latest pyzstd release has changed the error type of:

```Python
>>> import io
>>> import pyzstd
>>> f = pyzstd.ZstdFile(io.BytesIO(b'bad'), mode='rb')
>>> f.read()
```

And tests are now failing.

@vanandrew Do you have time to look into this? I suspect what we want to do is coerce exceptions in `Opener.read()` to match expected exceptions from `IOBase.read()`, and possibly other methods.

effigies:
This change is motivated by a change in `pyzstd`, which exposes failures to read bad streams as custom `ZstdError`s instead of `OSError`s. We had previously been depending on a conflation of these, as `BZ2File` returns a straightforward `OSError` and `GzipFile` returns `BadGzipFile`, which subclasses from `OSError`.

Now there's no clear way of catching the various ways that reading a stream might fail without creating a collection of exceptions and adding to it when an interface that emits different errors is found.

Therefore, I've created a private `nibabel._compression` module. This houses `openers.COMPRESSED_FILE_LIKES` and aliases `IndexedGzipFile` to `GzipFile` when it is unavailable, as well as provides a new `COMPRESSION_ERRORS` exception type tuple that can be used in `except` statements.

Fixes #1209.

### DIFF
--- a/nibabel/_compression.py
+++ b/nibabel/_compression.py
@@ -1,0 +1,49 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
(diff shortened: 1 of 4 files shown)


effigies:
No change to the API, and it fixes the failing CI. Happy to discuss alterations after the merge.

PR #1228, opened 2023-04-22
RF: refactor find_private_element

matthew-brett:
Neater and more readable version of find_private_element, extend tests.

### DIFF
--- a/nibabel/nicom/utils.py
+++ b/nibabel/nicom/utils.py
@@ -33,7 +33,7 @@ def find_private_section(dcm_data, group_no, creator):
         match_func = creator.search
     else:  # assume string / bytes
         creator = asstr(creator)
-        match_func = lambda x : x == creator
+        match_func = asstr(creator).__eq__
     # Group elements assumed ordered by tag (groupno, elno)
     for element in dcm_data.group_dataset(group_no):
         elno = element.tag.elem
--- a/nibabel/nicom/utils.py
+++ b/nibabel/nicom/utils.py
@@ -32,7 +32,6 @@ def find_private_section(dcm_data, group_no, creator):
     if hasattr(creator, 'search'):
         match_func = creator.search
     else:  # assume string / bytes
-        creator = asstr(creator)
         match_func = asstr(creator).__eq__
     # Group elements assumed ordered by tag (groupno, elno)
     for element in dcm_data.group_dataset(group_no):
--- a/nibabel/nicom/utils.py
+++ b/nibabel/nicom/utils.py
@@ -27,26 +27,20 @@ def find_private_section(dcm_data, group_no, creator):
     Returns
     -------
     element_start : int
-        Element number at which named section starts
+        Element number at which named section starts.
     """
-    is_regex = hasattr(creator, 'search')
-    if not is_regex:  # assume string / bytes
+    if hasattr(creator, 'search'):
+        match_func = lambda x : creator.search(x)
+    else:  # assume string / bytes
         creator = asstr(creator)
-    for element in dcm_data:  # Assumed ordered by tag (groupno, elno)
-        grpno, elno = element.tag.group, element.tag.elem
-        if grpno > group_no:
-            break
-        if grpno != group_no:
-            continue
+        match_func = lambda x : x == creator
+    # Group elements assumed ordered by tag (groupno, elno)
+    for element in dcm_data.group_dataset(group_no):
+        elno = element.tag.elem
         if elno > 0xFF:
             break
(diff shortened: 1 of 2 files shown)


nibabel/nicom/utils.py
```diff
@@ -27,26 +27,20 @@ def find_private_section(dcm_data, group_no, creator):
...
     element_start : int
-        Element number at which named section starts
+        Element number at which named section starts.
     """
-    is_regex = hasattr(creator, 'search')
-    if not is_regex:  # assume string / bytes
+    if hasattr(creator, 'search'):
+        match_func = lambda x : creator.search(x)
+    else:  # assume string / bytes
         creator = asstr(creator)
```
effigies, line 35: On the whole, I think we should be aiming to remove `py3k` functions. Is it safe to say that `creator` has the same type as `element.value`, or can there be mixed-and-matched `str`/`bytes`? Or is `element.value` at least always bytes or always str?

nibabel/nicom/utils.py
```diff
@@ -27,26 +27,20 @@ def find_private_section(dcm_data, group_no, creator):
     Returns
     -------
     element_start : int
-        Element number at which named section starts
+        Element number at which named section starts.
     """
-    is_regex = hasattr(creator, 'search')
-    if not is_regex:  # assume string / bytes
+    if hasattr(creator, 'search'):
+        match_func = lambda x : creator.search(x)
```
effigies, line 33: Unneeded indirection.

```suggestion
        match_func = creator.search
```

nibabel/nicom/utils.py
```diff
@@ -27,26 +27,20 @@ def find_private_section(dcm_data, group_no, creator):
...
+    else:  # assume string / bytes
         creator = asstr(creator)
-    for element in dcm_data:  # Assumed ordered by tag (groupno, elno)
-        grpno, elno = element.tag.group, element.tag.elem
-        if grpno > group_no:
-            break
-        if grpno != group_no:
-            continue
+        match_func = lambda x : x == creator
```
effigies, line 36: Likewise, we could use `creator.__eq__`, but I wouldn't insist on that if you prefer the readability of the `lambda`. Either way, I think it would be cleaner to include the `asstr` in this line. Either:

```suggestion
        match_func = lambda x: x == asstr(creator)
```

Or:

```suggestion
        match_func = asstr(creator).__eq__
```

effigies (approved):
All makes sense. Minor comments.

nibabel/nicom/utils.py
effigies: To be clear, this does not need to be addressed in this PR, but when we actually know the types, we should work with them directly. `asstr()` says to me "here be dragons".

matthew-brett: I don't think `creator` need have the same type as `element.value` because we are testing all `element` values, and only some of them may be specifying offsets (though maybe not-offset elements would be an error).

nibabel/nicom/utils.py
matthew-brett: It has the annoying feature that we needlessly call `asstr` on `creator` for each element we test.

effigies: Fair enough.

nibabel/nicom/utils.py
matthew-brett: I agree the `asstr` is worrying, and in this case, it's because I do not know why `.value` will sometimes be `bytes` and sometimes `str`.   I should investigate, but it would take a little while to determine why differences would arise, and be sure that these won't arise in the wild.

effigies: Okay. Let's leave it alone for now.

nibabel/nicom/utils.py
matthew-brett: Sorry - no - I'm wrong for the second - that's fine - the conversion is done once.

nibabel/nicom/utils.py
```diff
@@ -27,26 +27,20 @@ def find_private_section(dcm_data, group_no, creator):
...
     element_start : int
-        Element number at which named section starts
+        Element number at which named section starts.
     """
-    is_regex = hasattr(creator, 'search')
-    if not is_regex:  # assume string / bytes
+    if hasattr(creator, 'search'):
+        match_func = creator.search
+    else:  # assume string / bytes
         creator = asstr(creator)
```
effigies, line 35: Last little cleanup.

```suggestion
```

effigies:
Good to merge?

PR #1234, opened 2023-06-26
TEST: Only use stable argsorts in PARREC tests

effigies:
numpy/numpy#23707 introduced an optimized argsort that will be used on some platforms, including GitHub actions' runners, making this a nasty issue to figure out. I'm very glad that it wasn't on some obscure architecture that one user hit, as I have no idea how I would have debugged that.

----

~~Between the last good and first bad runs, numpy released 1.25 and GitHub Actions released new Python images. First starting by verifying that the previous versions of both resolve the issue. Then will relax numpy.~~

### DIFF
--- a/nibabel/tests/test_parrec.py
+++ b/nibabel/tests/test_parrec.py
@@ -173,8 +173,6 @@ DTI_PAR_BVECS = np.array(
 
 # DTI.PAR values for bvecs
 DTI_PAR_BVALS = [1000] * 6 + [0, 1000]
-# Numpy's argsort can be unstable so write indices manually
-DTI_PAR_BVALS_SORT_IDCS = [6, 0, 1, 2, 3, 4, 5, 7]
 
 EXAMPLE_IMAGES = [
     # Parameters come from load of Philips' conversion to NIfTI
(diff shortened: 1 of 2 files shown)


effigies:
Okay. So it is Numpy 1.25. Still unclear why it's 100% of the time here and I've been unable to reproduce even stochastically on my local machine.

effigies:
Looking through the commit logs between the last succeeding and first failing pre-releases: https://github.com/numpy/numpy/compare/126b46c...v1.25.0

I think https://github.com/numpy/numpy/pull/23707 is a likely candidate. Will try finding alternatives to argsort for the failing tests.

effigies:
@grlee77 Would you mind having a look? Does this seem like a thing worth reporting upstream, or is instability in argsort considered to be acceptable?

effigies:
Bug report here: https://github.com/numpy/numpy/issues/24064

We'll see if we need to include this patch or if we should just tolerate failures until numpy releases a fix.

effigies:
Nevermind. This was our fault for not explicitly requesting a stable argsort. Merging.

PR #1241, opened 2023-07-17
MNT: Move nisext imports from distutils to setuptools

penguinpee:
- TEST: Only use stable argsorts in PARREC tests
- ENH: Warn on invalid MINC2 spacing declarations, treat as missing
- Fix tests for Python 3.12

### DIFF
--- a/nisext/sexts.py
+++ b/nisext/sexts.py
@@ -2,10 +2,10 @@
 
 import os
 from configparser import ConfigParser
-from distutils import log
-from distutils.command.build_py import build_py
-from distutils.command.install_scripts import install_scripts
-from distutils.version import LooseVersion
+import logging as log
+from setuptools.command.build_py import build_py
+from setuptools.command.install_scripts import install_scripts
+from looseversion import LooseVersion
 from os.path import join as pjoin
 from os.path import split as psplit
 from os.path import splitext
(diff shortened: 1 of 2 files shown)


effigies:
I'm okay with this, assuming you wouldn't prefer just to remove `nisext` from your package altogether.

nisext/sexts.py
```diff
@@ -2,10 +2,10 @@
...
 import os
 from configparser import ConfigParser
-from distutils import log
-from distutils.command.build_py import build_py
-from distutils.command.install_scripts import install_scripts
-from distutils.version import LooseVersion
+import logging as log
+from setuptools.command.build_py import build_py
+from setuptools.command.install_scripts import install_scripts
+from looseversion import LooseVersion
```
effigies, line 8: On the off-chance that anybody's using `nisext`, it would be a pain to cause them to fail this import because `looseversion` is not a dependency.
```suggestion
from setuptools._distutils.version import LooseVersion
```

penguinpee: I'm okay with using `setuptools._distutils.version` instead of `looseversion`. But I also put `looseversion` as a dependency in `pyproject.toml`. Unless, I'm missing something, that should pull in `loosversion` when `nibabel` is installed.

For me, at least, both approaches work.

penguinpee: > For me, at least, both approaches work.

I was wrong. The dependency is not resolved automatically in the build environment. It was still cached from an earlier build. I dropped `loosversion` in the latest commit.

PS: I meant to comment and resolve at the same time. But that didn't work.

PR #1247, opened 2023-08-15
CI: Begin testing on Python 3.12

effigies:
Numpy 1.26b1 is out, allowing tests on Python 3.12.

Not yet ready to test on numpy 2.0, so leaving the nightly build URLs as-is for now.


effigies:
Well, looks like there are numeric issues, or possibly `isinstance()` or similar checks are now producing different results. Debugging locally seems tricky, as Python 3.12rc1 won't build with pyenv on Ubuntu 22.04. Will try a Docker image when I get a chance.

effigies:
Not a numeric issue, but gzip: https://github.com/python/cpython/issues/108111

PR #1250, opened 2023-08-28
Replace np.sctypes for numpy 2.0 compat

mscheltienne:
With nightly numpy release, the following attribute error is raised:

```
AttributeError: `np.sctypes` was removed in the NumPy 2.0 release. . Did you mean: 'dtypes'?
```

corresponding to this bullet point in the release notes:

> np.issctype, np.maximum_sctype, np.obj2sctype, np.sctype2char, np.sctypeDict, np.sctypes, np.issubsctype were all removed from the main namespace without replacement, as they where niche members.

This PR replaces:

- [x] `np.sctypes` calls with `np.core.sctypes`
- [x] `arr.newbyteorder` with `arr.view(arr.dtype.newbyteorder())`
- [x] some implicit casts when multiplying by python `int` to explicit casts then multiply
- [x] `np.can_cast` usage


effigies:
Actually, would you mind running `pipx run blue nibabel` to sort out the styling?

mscheltienne:
I don't read this as a deprecation of `sctypes` entirely, rather as a removal of the "shortcut" `np.sctypes`, since they explicitly mention `namespace`.

mscheltienne:
Perfect, at least that will be a clear answer :)

larsoner:
> I don't read this as a deprecation of sctypes entirely, rather as a removal of the "shortcut" np.sctypes, since they explicitly mention namespace.

Looks like `np.core` should not be used

https://github.com/numpy/numpy/pull/24376#issuecomment-1695909212

mscheltienne:
Well that was quick. I don't have time today, but I can have a look tomorrow to replace `sctypes` with something else.

nibabel/casting.py
```diff
@@ -23,6 +23,15 @@ class CastingError(Exception):
...
 
+# np.sctypes is deprecated in numpy 2.0 and np.core.sctypes should not be used instead.
+sctypes = {
+    "int": [np.int8, np.int16, np.int32, np.int64],
+    "uint": [np.uint8, np.uint16, np.uint32, np.uint64],
+    "float": [np.float16, np.float32, np.float64, np.float128],
+    "complex": [np.complex64, np.complex128, np.complex256],
+    "others": [bool, object, bytes, str, np.void],
+}
+
```
mscheltienne, line 34: I'm not entirely convince this is enough to keep a working status on all platforms. It looks to me like `np.sctypes` is defined per platform, and if this is correct, we might need to add some logic to define the mapping between strings and scalar types per platform as well.

effigies:
Hi, thanks for this, and I think this will work. I need to find a little time to think clearly about this problem. Just to let you know that I haven't forgotten about this, just a little slow to get back to it.

larsoner:
Looks like this branch is getting out of date, trying to use it in MNE I get:
```
/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/nibabel/quaternions.py:32: in <module>
    MAX_FLOAT = np.maximum_sctype(float)
/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/numpy/__init__.py:383: in __getattr__
    raise AttributeError(
E   AttributeError: `np.maximum_sctype` was removed in the NumPy 2.0 release.
```

mscheltienne:
So soon.. looking forward to that 2.0 release, it's going to be interesting 😄 
That should fix the `maximum_sctype` calls, but I there are also 3 remaining `sctypeDict` calls which needs to be addressed.

mscheltienne:
That should do it, still not 100% convince by the `getattr / hasattr` approach, but I built the `sctypes_named` variable from numpy's documentation (https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types) and got a hit with `set(np.sctypeDict.values())`.

larsoner:
@mscheltienne looks like there are some style issues

effigies:
@larsoner Can you review and merge this when ready?

larsoner:
FYI @effigies I tweaked the repo properties to "Always suggest updating pull requests" because I wanted this button:

Screenshot from 2023-10-18 12-49-31

Feel free to revert it in Settings if you dislike it

larsoner:
@mscheltienne to fix the doctest failures you could adapt this approach:

https://github.com/mne-tools/mne-python/pull/11836

It seems easier than casting everything (what we ended up doing in MNE).

larsoner:
Okay there is a bit more urgency now that the failure in MNE is:
```
mne/surface.py:884: in read_surface
    ret = read_geometry(fname, read_metadata=read_metadata)
/opt/hostedtoolcache/Python/3.11.6/x64/lib/python3.11/site-packages/nibabel/freesurfer/io.py:181: in read_geometry
    raise ValueError('File does not appear to be a Freesurfer surface')
E   ValueError: File does not appear to be a Freesurfer surface
```
https://github.com/mne-tools/mne-python/actions/runs/6642971086/job/18049032091?pr=12135#step:17:4752

@mscheltienne I'll try to push some commits to fix this

nibabel/streamlines/trk.py
```diff
@@ -1,4 +1,4 @@
-# Definition of trackvis header structure:
++# Definition of trackvis header structure:
```
mscheltienne, line 1:
```suggestion
# Definition of trackvis header structure:
```

mscheltienne:
How does it keep getting worst.. thanks for continuing the fix, I lacked the time to come back to this one.

larsoner:
> How does it keep getting worst

😆 NumPy 2.0 continues to add breaking changes. At least we're tackling them bit by bit rather than all at once later!

nibabel/nifti1.py
```diff
@@ -2443,9 +2443,13 @@ def _get_analyze_compat_dtype(arr):
         return np.dtype('int16' if arr.max() <= np.iinfo(np.int16).max else 'int32')
 
     mn, mx = arr.min(), arr.max()
-    if np.can_cast(mn, np.int32) and np.can_cast(mx, np.int32):
+    if (isinstance(mn, int) and isinstance(mx, int)) or (
```
larsoner, line 2446: The problem with this is that on most platforms Python `int` is 64-bit. So you could have a value of `2 ** 33` represented properly using it. But saving in `np.int32` is not safe. So I think this check needs to change to something involving `np.iinfo('int32')` max/min :disappointed:

larsoner:
Green! (Other than some coverage misses that I think probably exist in `master` as well?)

@effigies feel free to review/merge if you're happy, only ended up at `+170 −97`

nibabel/tests/test_volumeutils.py
```diff
@@ -733,9 +739,10 @@ def test_apply_scaling():
     f32_arr = np.zeros((1,), dtype=f32)
     i16_arr = np.zeros((1,), dtype=np.int16)
     # Check float upcast (not the normal numpy scalar rule)
-    # This is the normal rule - no upcast from scalar
-    assert (f32_arr * f64(1)).dtype == np.float32
-    assert (f32_arr + f64(1)).dtype == np.float32
+    # This is the normal rule - no upcast from Python scalar
+    # (on NumPy 2.0 it *will* upcast from a np.float64 scalar!)
+    assert (f32_arr * 1.).dtype == np.float32
+    assert (f32_arr + 1.).dtype == np.float32
```
effigies, line 745: The goal here seems to be contrasting to numpy casting rules, not numpy + Python scalar casting rules. I would just branch on numpy version about the expected dtype for using `f64(...)`.

nibabel/casting.py
```diff
@@ -456,6 +495,7 @@ def int_to_float(val, flt_type):
     # -1
     if not isinstance(val, Integral):
         val = int(str(val))
+    val = int(val)
```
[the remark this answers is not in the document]
larsoner: NumPy 2.0 handles Python int differently from `np.int64`. Can't remember offhand why it was needed here but I like your suggestion anyway!

nibabel/tests/test_floating.py
```diff
@@ -148,9 +153,15 @@ def test_as_int():
...
+    # TODO: Should this actually still overflow? Does it matter?
+    if FP_OVERFLOW_WARN:
+        ctx = pytest.raises(OverflowError)
+    else:
+        ctx = nullcontext()
+    with ctx:
         as_int(val)
-    with pytest.raises(OverflowError):
+    with ctx:
         as_int(-val)
```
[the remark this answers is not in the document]
larsoner: It does not overflow. Agreed it's weird that it does not

effigies:
Okay, I'm starting to understand what's going on here. Numpy previously had broken behavior around longdoubles and ints. If you install Python 3.6 and numpy 1.11, you get things like:

```python
In [1]: import numpy as np

In [2]: val = np.longdouble(2**1024) * 2
OverflowError                             Traceback (most recent call last)
<ipython-input-2-4bc60941a286> in <module>
----> 1 val = np.longdouble(2**1024) * 2

OverflowError: int too large to convert to float

In [6]: val = np.longdouble(2**1023)

In [7]: val
Out[7]: 8.9884656743115795386e+307

In [8]: val * 2
Out[8]: 1.7976931348623159077e+308

In [9]: val * 4
Out[9]: 3.5953862697246318155e+308

In [10]: int(val * 4)
OverflowError                             Traceback (most recent call last)
<ipython-input-10-af59c627c981> in <module>
----> 1 int(val * 4)

OverflowError: cannot convert float infinity to integer
```

However, if you use Python 3.8 and numpy 1.20 (nibabel current minimum requirements):

```python
In [1]: import numpy as np

In [2]: nexp64 = 1024

In [3]: val = np.longdouble(2**1024) * 2

In [4]: val
Out[4]: 3.5953862697246318155e+308

In [5]: int(val)
Out[5]: 359538626972463181545861038157804946723595395788461314546860162315465351611001926265416954644815072042240227759742786715317579537628833244985694861278948248755535786849730970552604439202492188238906165904170011537676301364684925762947826221081654474326701021369172596479894491876959432609670712659248448274432
```

I think at this point, probably the correct thing is just to get rid of `int_to_float` and `as_int`. But that's not needed for this PR.

PR #1253, opened 2023-09-08
TYP: Annotate SpatialImage accepting affine=None

blakedewey:
This corrects the typing for the `SpatialImage` `__init__` function by allowing `None` for `affine`.

### DIFF
--- a/nibabel/spatialimages.py
+++ b/nibabel/spatialimages.py
@@ -481,7 +481,7 @@ class SpatialImage(DataobjImage):
     def __init__(
         self,
         dataobj: ArrayLike,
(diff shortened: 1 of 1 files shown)


effigies:
Reading the `mypy` failure, I think we need to add an `assert self._affine is not None` before line 569.

effigies:
Looking good. If you `pip install` this version into your pycharm environment, does it resolve your `Nifti1Image(data, None, header)` issue?

blakedewey:
Yes! It does.

effigies:
Great! Thanks for reporting and fixing the issue!

PR #1255, opened 2023-09-19
ENH: Add copy() method to ArrayProxy

effigies:
Small addition to ArrayProxy class to allow copying. Used in #1090, but not central to it.


effigies:
@pauldmccarthy This is bound to interact with indexed gzip. If you have time for a quick read-through, I would appreciate your thoughts.

effigies:
Thanks for the review and test! If you think of a more strenuous test or something we can do upstream, happy to discuss more.

effigies:
Test failure unrelated.

PR #1256, opened 2023-09-19
ENH: Expand CIFTI2 constants to use a synonym recoder

effigies:
For increased flexibility in the CIFTI2 axes API, this creates a set of synonyms for different CIFTI structures, mapping them onto the `CIFTI_STRUCTURE_*`.

Used for resolving synonyms in caret spec files in #1090, but mergeable on its own.

### DIFF
--- a/nibabel/cifti2/cifti2.py
+++ b/nibabel/cifti2/cifti2.py
@@ -30,7 +30,7 @@ from ..dataobj_images import DataobjImage
 from ..filebasedimages import FileBasedHeader, SerializableImage
 from ..nifti1 import Nifti1Extensions
 from ..nifti2 import Nifti2Header, Nifti2Image
-from ..volumeutils import make_dt_codes
+from ..volumeutils import Recoder, make_dt_codes
 
 
 def _float_01(val):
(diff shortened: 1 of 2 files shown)


effigies:
@mgxd Would you mind having a look when you get a chance?

nibabel/cifti2/cifti2_axes.py
```diff
@@ -520,7 +520,7 @@ def to_cifti_brain_structure_name(name):
         ValueError: raised if the input name does not match a known anatomical structure in CIFTI-2
         """
         if name in cifti2.CIFTI_BRAIN_STRUCTURES:
```
mgxd, line 522: seems like this should be
```suggestion
        if name in cifti2.CIFTI_BRAIN_STRUCTURES.ciftiname:
```

nibabel/cifti2/cifti2.py
```diff
@@ -80,39 +80,64 @@ class Cifti2HeaderError(Exception):
...
-    'CIFTI_STRUCTURE_OTHER_GREY_MATTER',
-    'CIFTI_STRUCTURE_OTHER_WHITE_MATTER',
-    'CIFTI_STRUCTURE_PALLIDUM_LEFT',
-    'CIFTI_STRUCTURE_PALLIDUM_RIGHT',
-    'CIFTI_STRUCTURE_PUTAMEN_LEFT',
-    'CIFTI_STRUCTURE_PUTAMEN_RIGHT',
-    'CIFTI_STRUCTURE_THALAMUS_LEFT',
-    'CIFTI_STRUCTURE_THALAMUS_RIGHT',
+
+def _full_structure(struct):
```
mgxd, line 84: might as well
```suggestion
def _full_structure(struct: str):
```

mgxd (approved):
Few things, otherwise looks good

nibabel/cifti2/cifti2_axes.py
effigies:
```python
>>> from nibabel.cifti2 import CIFTI_BRAIN_STRUCTURES
>>> 'CortexLeft' in CIFTI_BRAIN_STRUCTURES
True
>>> 'CortexLeft' in CIFTI_BRAIN_STRUCTURES.ciftiname
True
>>> CIFTI_BRAIN_STRUCTURES['CortexLeft']
'CORTEX_LEFT'
>>> CIFTI_BRAIN_STRUCTURES.name['CortexLeft']
'CORTEX_LEFT'
>>> CIFTI_BRAIN_STRUCTURES.ciftiname['CortexLeft']
'CIFTI_STRUCTURE_CORTEX_LEFT'
>>> CIFTI_BRAIN_STRUCTURES.guiname['CortexLeft']
'CortexLeft'
```

The idea here is that we want to accept any of these names, but then we'll map them out to the `ciftiname` form.

PR #1257, opened 2023-09-19
ENH: Add TriangularMesh data structure [BIAP9]

effigies:
#1251 Added `Pointset` and `Grid`. #1090 also proposes `TriangularMesh`, but with the refactoring seen in #1251, that seemed like a bit of extra work for minimal benefit to the transformations use-case. This PR re-adds triangular meshes.

Overall, the approach of multiple names feels clunky and pushes the concerns of surface families into pointsets. #1251 removed that, and this PR extends this by keeping `TriangularMesh` focused on the base data structure and moves `TriangularMeshFamily` logic into a `CoordinateFamilyMixin` class that permits swapping out coordinates with a `.with_name()` method (open to bikeshedding). Hence, we would change:

```Python
left_surfaces.get_coords(name='pial')
```

to:

```Python
left_surfaces.with_name('pial').get_coords()
```

It also follows the lead of `Grid` and gets rid of the omni-`__init__` in favor of targeted factory classmethods.

----

Notes from previous discussion:

* @oesteban wrote: "As a note, if some time we support other meshes with more than three edges, then n_triangles should be better named n_faces."
   * This was part of early discussions of BIAP9. IIRC the universality of triangular meshes makes them worth special-casing for now. FreeSurfer has old formats that use non-triangular meshes, and both FreeSurfer and we convert them to triangular on load: https://github.com/nipy/nibabel/blob/590b226d7d29c8086c45cc82ece76181e598c474/nibabel/freesurfer/io.py#L145-L168

### DIFF
--- a/nibabel/pointset.py
+++ b/nibabel/pointset.py
@@ -48,6 +48,11 @@ class CoordinateArray(ty.Protocol):
         ...  # pragma: no cover
 
 
+class HasMeshAttrs(ty.Protocol):
+    coordinates: CoordinateArray
+    triangles: CoordinateArray
+
+
 @dataclass
 class Pointset:
     """A collection of points described by coordinates.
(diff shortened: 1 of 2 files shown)


nibabel/pointset.py
```diff
@@ -144,6 +149,77 @@ def get_coords(self, *, as_homogeneous: bool = False):
...
+        """List of surface names that can be passed to :meth:`with_name`"""
+        return list(self._coords)
+
+    def with_name(self, name: str) -> Self:
+        new = replace(self, coordinates=self._coords[name])
+        new._coords = self._coords
+        return new
+
+    def add_coordinates(self, name, coordinates):
+        self._coords[name] = coordinates
```
effigies, line 220: This is simple but perhaps a little clunky. We could also do:

```suggestion
    def add_coordinates(self, *args, **kwargs):
        self._coords.update(dict(*args, **kwargs))
```

This would allow it to take any inputs that would generate a `dict`, which might be a bit more ergonomic. I'm not 100% on this, so I haven't made the change.

PR #1260, opened 2023-09-29
ENH: Allow ~/ in paths

ReinderVosDeWael:
This PR modifies `_stringify_path` to allow it to expand home directories (`~/`). Its output should remain the same on all other valid inputs. 

Tests are included to validate a variety of cases. As no prior tests existed for this function I cannot confirm that its behavior remains identical in all other cases.

### DIFF
--- a/nibabel/filename_parser.py
+++ b/nibabel/filename_parser.py
@@ -10,6 +10,7 @@
 from __future__ import annotations
 
 import os
+import pathlib
 import typing as ty
 
 if ty.TYPE_CHECKING:  # pragma: no cover
@@ -37,9 +38,7 @@ def _stringify_path(filepath_or_buffer: FileSpec) -> str:
     Adapted from:
     https://github.com/pandas-dev/pandas/blob/325dd68/pandas/io/common.py#L131-L160
     """
-    if isinstance(filepath_or_buffer, os.PathLike):
-        return filepath_or_buffer.__fspath__()
-    return filepath_or_buffer
+    return str(pathlib.Path(filepath_or_buffer).expanduser().resolve())
 
 
 def types_filenames(
--- a/nibabel/filename_parser.py
+++ b/nibabel/filename_parser.py
@@ -38,7 +38,8 @@ def _stringify_path(filepath_or_buffer: FileSpec) -> str:
     Adapted from:
     https://github.com/pandas-dev/pandas/blob/325dd68/pandas/io/common.py#L131-L160
     """
-    return str(pathlib.Path(filepath_or_buffer).expanduser())
+    full_path = pathlib.Path(filepath_or_buffer).expanduser()
+    return str(full_path)
 
 
 def types_filenames(
--- a/nibabel/filename_parser.py
+++ b/nibabel/filename_parser.py
@@ -38,7 +38,7 @@ def _stringify_path(filepath_or_buffer: FileSpec) -> str:
     Adapted from:
     https://github.com/pandas-dev/pandas/blob/325dd68/pandas/io/common.py#L131-L160
     """
-    return str(pathlib.Path(filepath_or_buffer).expanduser().resolve())
+    return str(pathlib.Path(filepath_or_buffer).expanduser())
 
 
 def types_filenames(
--- a/nibabel/tests/test_filename_parser.py
+++ b/nibabel/tests/test_filename_parser.py
@@ -7,10 +7,11 @@
 #
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Tests for filename container"""
+import pathlib
 
 import pytest
 
-from ..filename_parser import TypesFilenamesError, parse_filename, splitext_addext, types_filenames
+from ..filename_parser import TypesFilenamesError, parse_filename, splitext_addext, types_filenames, _stringify_path
 
 
 def test_filenames():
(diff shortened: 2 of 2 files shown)


nibabel/filename_parser.py
```diff
@@ -37,9 +38,7 @@ def _stringify_path(filepath_or_buffer: FileSpec) -> str:
     Adapted from:
     https://github.com/pandas-dev/pandas/blob/325dd68/pandas/io/common.py#L131-L160
     """
-    if isinstance(filepath_or_buffer, os.PathLike):
-        return filepath_or_buffer.__fspath__()
-    return filepath_or_buffer
+    return pathlib.Path(filepath_or_buffer).expanduser().as_posix()
```
effigies, line 41:
```suggestion
    return str(pathlib.Path(filepath_or_buffer).expanduser())
```

nibabel/tests/test_filename_parser.py
```diff
@@ -123,3 +124,24 @@ def test_splitext_addext():
...
+def test__stringify_path():
+    current_directory = pathlib.Path.cwd()
+    res = _stringify_path('fname.ext.gz')
+    assert res == 'fname.ext.gz'
+    res = _stringify_path(pathlib.Path('fname.ext.gz'))
+    assert res == 'fname.ext.gz'
+
+    home = pathlib.Path.home()
+    res = _stringify_path(pathlib.Path('~/fname.ext.gz'))
+    assert res == str(home) + '/fname.ext.gz'
```
effigies, line 138:
```suggestion
    assert res == os.path.join(home, '/fname.ext.gz')
```

nibabel/tests/test_filename_parser.py
```diff
@@ -123,3 +124,24 @@ def test_splitext_addext():
...
+
+    res = _stringify_path(pathlib.Path('./fname.ext.gz'))
+    assert res == 'fname.ext.gz'
+    res = _stringify_path(pathlib.Path('../fname.ext.gz'))
+    assert res == '../fname.ext.gz'
```
effigies, line 143:
```suggestion
    assert res == os.path.join('..', 'fname.ext.gz')
```

nibabel/tests/test_filename_parser.py
```diff
@@ -123,3 +124,24 @@ def test_splitext_addext():
...
+    
+
+
+
```
effigies, line 147:
```suggestion
```

effigies (commented):
I don't think there's any advantage to forcing to POSIX apart from to please the tests. I would just us `os.path.join` to ensure the tests are consistent with the OS.

It also won't work as-is, since `Path.home()` will contain backslashes in Windows.

ReinderVosDeWael:
Some of the failures were in other, existing, tests, would you want these modified accordingly?

ReinderVosDeWael:
See test_parse_filename in the same test file as an example.

effigies:
Oh, I see. Well, if we're munging plain strings instead of passing them through untouched, then I suppose it doesn't matter much whether they become POSIX-style or match `os.sep`.

I'm trying to think if there is any other action at a distance that this will cause. I think `img.get_filename()` will now become normalized to whatever we choose, which should only affect someone who is actually comparing that to the paths they loaded. That seems pretty unlikely...

ReinderVosDeWael:
Also apologies for the repeated test pushes; I don't have access to a windows machine so I can't test locally

effigies:
@ReinderVosDeWael I assume you saw that a new thing has broken:

```
___________________________ TestEcatImage.test_file ___________________________
[gw1] win32 -- Python 3.9.13 D:\a\nibabel\nibabel\virtenv\Scripts\python.exe

self = <nibabel.tests.test_ecat.TestEcatImage testMethod=test_file>

    def test_file(self):
>       assert self.img.file_map['header'].filename == self.example_file
E       AssertionError: assert 'D:/a/nibabel...ata/tinypet.v' == 'D:\\a\\nibab...ta\\tinypet.v'
E         - D:\a\nibabel\nibabel\virtenv\lib\site-packages\nibabel\tests\data\tinypet.v
E         ?   ^ ^       ^       ^       ^   ^             ^       ^     ^    ^
E         + D:/a/nibabel/nibabel/virtenv/lib/site-packages/nibabel/tests/data/tinypet.v
E         ?   ^ ^       ^       ^       ^   ^             ^       ^     ^    ^

..\virtenv\lib\site-packages\nibabel\tests\test_ecat.py:186: AssertionError
____________________________ test_mgh_load_fileobj ____________________________
[gw0] win32 -- Python 3.9.13 D:\a\nibabel\nibabel\virtenv\Scripts\python.exe

    def test_mgh_load_fileobj():
        # Checks the filename gets passed to array proxy
        #
        # This is a bit of an implementation detail, but the test is to make sure
        # that we aren't passing ImageOpener objects to the array proxy, as these
        # were confusing mmap on Python 3.  If there's some sensible reason not to
        # pass the filename to the array proxy, please feel free to change this
        # test.
        img = MGHImage.load(MGZ_FNAME)
>       assert img.dataobj.file_like == MGZ_FNAME
E       AssertionError: assert 'D:/a/nibabel...data/test.mgz' == 'D:\\a\\nibab...ata\\test.mgz'
E         - D:\a\nibabel\nibabel\virtenv\lib\site-packages\nibabel\tests\data\test.mgz
E         ?   ^ ^       ^       ^       ^   ^             ^       ^     ^    ^
E         + D:/a/nibabel/nibabel/virtenv/lib/site-packages/nibabel/tests/data/test.mgz
E         ?   ^ ^       ^       ^       ^   ^             ^       ^     ^    ^
```

Possibly the better check for these things is to assert that `Path(x) == Path(y)`?

ReinderVosDeWael:
Thanks for the ping - I had not seen this. I'll see if I can find a couple minutes later today (EST timezone) to resolve this.

ReinderVosDeWael:
That should do it; I've changed the failing tests to check for pathlib.Path equality. Again I can't test it on windows, so I'll await the CI.

nibabel/tests/test_filename_parser.py
```diff
@@ -123,3 +124,20 @@ def test_splitext_addext():
     assert res == ('..', '', '')
     res = splitext_addext('...')
     assert res == ('...', '', '')
+
+
+def test__stringify_path():
+    current_directory = pathlib.Path.cwd().as_posix()
```
effigies, line 130:
```suggestion
```

nibabel/tests/test_filename_parser.py
```diff
@@ -123,3 +124,20 @@ def test_splitext_addext():
...
+    res = _stringify_path('fname.ext.gz')
+    assert res == 'fname.ext.gz'
+    res = _stringify_path(pathlib.Path('fname.ext.gz'))
+    assert res == 'fname.ext.gz'
+
+    home = pathlib.Path.home().as_posix()
+    res = _stringify_path(pathlib.Path('~/fname.ext.gz'))
+    assert res == str(home) + '/fname.ext.gz'
```
effigies, line 138: Might as well be modern about this.

```suggestion
    assert res == f'{home}/fname.ext.gz'
```

effigies (approved):
Two small cleanups to the tests, but this is good to go.

PR #1262, opened 2023-10-06
MNT: Resurrect tox config to cover dev workflows and CI checks 

effigies:
This PR aims to centralize a lot of details for of our project in a single place: `tox.ini`. We've had this file for a while, but it hasn't been updated in ages. 

This fully replaces our custom `tools/ci` scripts with a maintained environment manager that works equally well locally and on CI. It thus also allows us to reduce our CI configuration significantly, as most jobs are now just running a corresponding tox environment.

This approach allows us to keep track of our minimum supported dependencies, both optional and required, and track the environments they can be reliably installed on. For example, we can document and test that we support a minimum dependency of scipy 1.6, but this is unavailable for 32-bit builds of Python 3.10+:

```ini
  # Scipy stopped producing win32 wheels at py310
  py3{8,9}-full-x86,x64: scipy >=1.6
```

This PR has gone on long enough. In separate pull requests, it would be good to:

1) Figure out why `pydicom@master` is failing, so we can restore that in pre-release tests.
2) Consider switching to `ruff`, a much faster and more fully-featured alternative to `flake8`.
3) Build and publish docs from GitHub actions. We should also fix Sphinx warnings and start running with warnings as errors.


effigies:
@matthew-brett @oesteban What do you think?

PR #1267, opened 2023-10-18
MNT: Add py312-dev-x64 environment

effigies:
cc @larsoner

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,6 +119,10 @@ jobs:
           - os: ubuntu-latest
             python-version: 3.8
             dependencies: 'min'
+          # NumPy 2.0
+          - os: ubuntu-latest
+            python-version: '3.12'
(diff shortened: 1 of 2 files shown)


larsoner:
Also need to add an entry to https://github.com/nipy/nibabel/blob/e356408049ff9ecf42a843824e9b2728f7f0b80e/.github/workflows/test.yml#L113-L128, no?

I think a single run of NP2.0 is enough

larsoner:
Seems to be failing as intended!

https://github.com/nipy/nibabel/actions/runs/6562773154/job/17825525527?pr=1267#step:7:62

```
E   AttributeError: `np.sctypes` was removed in the NumPy 2.0 release. . Did you mean: 'dtypes'?
```

larsoner:
@effigies feel free to merge and then we can see if https://github.com/nipy/nibabel/pull/1250 fixes things

PR #1270, opened 2023-10-26
MNT: Assorted refurb suggestions

DimitriPapadopoulos:


nibabel/externals/netcdf.py
effigies, line 1: We should only touch externals to fix bugs or update to a newer upstream version.

DimitriPapadopoulos: I have fixed this. Thank you for catching this.

effigies:
Thanks!

PR #1272, opened 2023-11-02
RF: Deprecate longdouble<->int hacks

effigies:
We had a pair of functions `as_int()` and `int_to_float()` that handled bugs in numpy that existed at least as recently as 1.11, but which no longer exist in 1.20, our current minimum.

This PR replaces `as_int(...)` with `int(...)` and `int_to_float(..., dtype)` with `dtype(...)` and deprecates these functions. Regression tests to ensure that these native Python/numpy functions work as expected are retained.

### DIFF
--- a/nibabel/arraywriters.py
+++ b/nibabel/arraywriters.py
@@ -30,7 +30,7 @@ larger ints and smaller.
 """
 import numpy as np
 
-from .casting import best_float, floor_exact, int_abs, int_to_float, shared_range, type_info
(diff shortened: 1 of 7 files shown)


effigies:
@matthew-brett You added these back in #70, but I think I'm right that they can go away at this point. Any objections? Any other regression tests worth putting in?

@larsoner or @mscheltienne  Would you mind having a review, since you were just dealing with these?

effigies:
Thanks for the review!

PR #1273, opened 2023-11-07
MNT: Update test dependencies for 3.12 and dev environments

effigies:
`indexed_gzip` has 3.12 wheels now, and we should go ahead and check for interactions with numpy 2-dev for all optional dependencies.


larsoner (approved):
:crossed_fingers: that it passes!

effigies:
Something's taking a surprising amount of time on 3.12-full (all OSs) but not 3.12-pre.

PR #1275, opened 2023-11-08
MNT: do not refer to the optional data packages

### LINKED ISSUE #1274: `file` constructor in `doc/source/devel/register_me.py`
Noticed this code, where `file` is neither a Python 3 built-in, nor imported:
https://github.com/nipy/nibabel/blob/1ee4983a55fae5da47fbf197fb97d17157a22534/doc/source/devel/register_me.py#L41

This looks like a Python 2 left-over:
https://python.readthedocs.io/en/v2.7.2/library/functions.html#file
https://portingguide.readthedocs.io/en/latest/builtins.html#removed-file

Has [`doc/source/devel/register_me.py`](https://github.com/nipy/nibabel/blame/master/doc/source/devel/register_me.py) ever been run recently? Is it still useful?

DimitriPapadopoulos:
The rationale is that script `doc/source/devel/register_me.py` is a Python 2 script that is not compatible with Python 3. Looks like the whole machinery has not been used for ages.

Fixes #1274.


doc/source/devel/devdiscuss.rst
```diff
@@ -21,7 +21,6 @@ progress.
 
     spm_use
     modified_images
-    data_pkg_design
     data_pkg_discuss
     data_pkg_uses
```
effigies, line 25: Should we go ahead and drop these two as well?

matthew-brett: Yes, they can all go.

DimitriPapadopoulos:
Removed file `data_pkg_uses.rst` and any reference to `data_pkg_discuss` and `data_pkg_uses`.

PR #1280, opened 2023-12-03
MNT: Deprecate nibabel.pydicom_compat

effigies:
This was added for the <1.0 to 1.0 transition. We no longer use it internally, and it's out-of-date with more recent deprecations in pydicom 2 that cause test failures in pydicom 3.0.0.dev0.

This PR switches the deprecated `read_file` to `dcmread` and deprecates our own compatibility shim. I suspect nobody is using it, so we'll give one release cycle and remove it in 7.0 (roughly 1yr).

### DIFF
--- a/nibabel/dft.py
+++ b/nibabel/dft.py
@@ -238,7 +238,7 @@ class _StorageInstance:
         return val
 
     def dicom(self):
-        return pydicom.read_file(self.files[0])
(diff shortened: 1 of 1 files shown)


effigies:
@ZviBaratz @moloney Any chance you could have a very quick look?

PR #1284, opened 2023-12-12
MNT: Add tool for generating GitHub-friendly release notes from RST changelog

effigies:
GitHub releases don't automatically convert RST to Markdown, so a direct copy ends up with warts, such as too many line breaks in the rendered output and unlinked references.

This PR fixes the former, mostly, and implements an extremely basic dereferencer. I also replace `pr/...` with `#...` to link to PRs.

This will probably need minor updates with each release, but it's a good starting point and should save time overall.


effigies:
See it in action: https://github.com/nipy/nibabel/releases/tag/5.2.0

Issue #1285, opened 2023-12-12
1 test fails: test_pkg_info


yurivict:
```
========================================================================================= FAILURES ==========================================================================================
_______________________________________________________________________________________ test_pkg_info _______________________________________________________________________________________

    def test_pkg_info():
        """Smoke test nibabel.get_info()
    
        Hits:
            - nibabel.get_info
            - nibabel.pkg_info.get_pkg_info
            - nibabel.pkg_info.pkg_commit_hash
        """
>       info = nib.get_info()

nibabel/tests/test_pkg_info.py:18: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
nibabel/__init__.py:81: in get_info
    return _get_pkg_info(os.path.dirname(__file__))
nibabel/pkg_info.py:129: in get_pkg_info
    src, hsh = pkg_commit_hash(pkg_path)
nibabel/pkg_info.py:106: in pkg_commit_hash
    proc = run(
/usr/local/lib/python3.9/subprocess.py:505: in run
    with Popen(*popenargs, **kwargs) as process:
/usr/local/lib/python3.9/subprocess.py:951: in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Popen: returncode: 255 args: ('git', 'rev-parse', '--short', 'HEAD')>, args = ['git', 'rev-parse', '--short', 'HEAD'], executable = b'git', preexec_fn = None, close_fds = True
[...]

PR #1286, opened 2023-12-13
FIX: Tolerate missing git

effigies:
Closes gh-1285.

The alternative would be to catch the exception in the test, but I can't think of a good reason for us to crash the interpreter over this.

Any objections?


effigies:
I'll consider this reviewed by @yurivict. Thanks!

PR #1288, opened 2023-12-25
MNT: Apply Repo-Review suggestions

DimitriPapadopoulos:
Apply some [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=nipy%2Fnibabel&branch=master) suggestions.

I went for the low-hanging fruits, other changes are more invasive and deserve a PR of their own, such as using a formatter like ruff of black.

Also update minimal Python version for blue. By the way, isn't blue able to understand `requires-python`?

### Fixed issues

* [MY103](https://learn.scientific-python.org/development/guides/style#MY103): MyPy warn unreachable
  Must have `warn_unreachable = true` to pass this check. There are occasionally false positives (often due to platform or Python version static checks), so it's okay to ignore this check. But try it first - it can catch real bugs too.
  ```toml
  [tool.mypy]
  warn_unreachable = true
  ```
* [MY104](https://learn.scientific-python.org/development/guides/style#MY104): MyPy enables ignore-without-code
  Must have `"ignore-without-code"` in `enable_error_code = [...]`. This will force all skips in your project to include the error code, which makes them more readable, and avoids skipping something unintended.
  ```toml
  [tool.mypy]
  enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
  ```
* [MY105](https://learn.scientific-python.org/development/guides/style#MY105): MyPy enables redundant-expr
  Must have `"redundant-expr"` in `enable_error_code = [...]`. This helps catch useless lines of code, like checking the same condition twice.
  ```toml
  [tool.mypy]
  enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
  ```
* [MY106](https://learn.scientific-python.org/development/guides/style#MY106): MyPy enables truthy-bool
 Must have `"truthy-bool"` in `enable_error_code = []`. This catches mistakes in using a value as truthy if it cannot be falsey.
  ```toml
  [tool.mypy]
  enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
  ```

### Issues ignored

Left for later because it requires extensive fixes:
* [MY101](https://learn.scientific-python.org/development/guides/style#MY101): MyPy strict mode
  Must have `strict` in the mypy config. MyPy is best with strict or nearly strict configuration. If you are happy with the strictness of your settings already, ignore this check or set `strict = false` explicitly.
  ```toml
  [tool.mypy]
  strict = true
  ```

### DIFF
--- a/nibabel/xmlutils.py
+++ b/nibabel/xmlutils.py
@@ -32,7 +32,7 @@ class XmlSerializable:
             Additional keyword arguments to :func:`xml.etree.ElementTree.tostring`.
         """
         ele = self._to_xml_element()
-        return b'' if ele is None else tostring(ele, enc, **kwargs)
+        return tostring(ele, enc, **kwargs)
 
 
 class XmlBasedHeader(FileBasedHeader, XmlSerializable):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,7 +111,7 @@ __version_tuple__ = version_tuple = {version_tuple!r}
 
 [tool.blue]
 line_length = 99
-target-version = ["py37"]
+target-version = ["py38"]
 force-exclude = """
 (
   _version.py
@@ -130,6 +130,8 @@ python_version = "3.11"
 exclude = [
   "/tests",
 ]
+warn_unreachable = true
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 
 [tool.codespell]
 skip = "*/data/*,./nibabel-data"
--- a/nibabel/_compression.py
+++ b/nibabel/_compression.py
@@ -17,7 +17,7 @@ import typing as ty
 from .optpkg import optional_package
 
 if ty.TYPE_CHECKING:  # pragma: no cover
-    import indexed_gzip  # type: ignore
+    import indexed_gzip  # type: ignore[import-not-found]
     import pyzstd
 
     HAVE_INDEXED_GZIP = True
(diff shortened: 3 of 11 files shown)


pyproject.toml
```diff
@@ -111,7 +111,7 @@ __version_tuple__ = version_tuple = {version_tuple!r}
 
 [tool.blue]
 line_length = 99
-target-version = ["py37"]
+target-version = ["py38"]
```
DimitriPapadopoulos, line 114: Doesn't blue understand `requires-python = ">=3.8"`?

nibabel/xmlutils.py
```diff
@@ -32,7 +32,7 @@ def to_xml(self, enc='utf-8', **kwargs) -> bytes:
             Additional keyword arguments to :func:`xml.etree.ElementTree.tostring`.
         """
         ele = self._to_xml_element()
-        return b'' if ele is None else tostring(ele, enc, **kwargs)
+        return tostring(ele, enc, **kwargs)
```
DimitriPapadopoulos, line 35: As far as I can see, `_to_xml_element` does nor return None, it raises exceptions. Besides it is declared to return `-> Element`, not `None`.

pyproject.toml
effigies: It does not. We'll probably switch to ruff, as blue's development seems done.

nibabel/xmlutils.py
effigies: Yeah, with type-checks this runtime check seems less necessary.

effigies (approved):

pyproject.toml
DimitriPapadopoulos: Then I will provide a PR to switch to ruff with [`quote-style`](https://docs.astral.sh/ruff/settings/#format-quote-style) set to `single`.

effigies: That would be great!

Issue #1211, opened 2023-03-03
Unable to get attribute from Philips DICOM: Calculated shape does not match number of frames.


yarikoptic:
Originally brought to try on these DICOMs in the context of https://github.com/nipy/heudiconv/issues/575#issuecomment-1443996960  and trying on Philips DICOMs within http://github.com/neurolabusc/dcm_qa_enh by @neurolabusc for enhanced Philips dicoms for  sw 5.3.0:

<details>
<summary>where attempt to run heudiconv crashed</summary> 

```shell
❯ heudiconv -f convertall -c dcm2niix --bids -o out-heudiconv -s unknown --files Philips
INFO: Running heudiconv version 0.11.6 latest 0.12.0
INFO: Analyzing 3 dicoms
Traceback (most recent call last):
  File "/usr/bin/heudiconv", line 33, in <module>
    sys.exit(load_entry_point('heudiconv==0.11.6', 'console_scripts', 'heudiconv')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/heudiconv/cli/run.py", line 24, in main
    workflow(**kwargs)
  File "/usr/lib/python3/dist-packages/heudiconv/main.py", line 327, in workflow
    study_sessions = get_study_sessions(dicom_dir_template, files,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/heudiconv/parser.py", line 178, in get_study_sessions
    seqinfo_dict = group_dicoms_into_seqinfos(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/heudiconv/dicoms.py", line 202, in group_dicoms_into_seqinfos
    mwinfo = validate_dicom(filename, dcmfilter)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/heudiconv/dicoms.py", line 106, in validate_dicom
    del mw.series_signature[sig]
        ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/nibabel/onetime.py", line 142, in __get__
    val = self.getter(obj)
[...]

PR #1302, opened 2024-02-27
ENH: Improve text for dicomwrapper errors in shape calculation

yarikoptic:
To make https://github.com/nipy/nibabel/issues/1211 more informative.

In the particular case at hand of [http://github.com/neurolabusc/dcm_qa_fmap](https://github.com/neurolabusc/dcm_qa_fmap)  IM_0027_fMAP.dcm we get 

```
nibabel.nicom.dicomwrappers.WrapperError: Calculated # of frames (128=4*32) of shape (64, 64, 32, 2, 2) does not match NumberOfFrames 64.
```

and my wild uneducated guess is that we need to filter out few more of those entries from frame_indices where we have 

```
> p dim_seq
[(0020, 9057), (2005, 106e), (2005, 1011)]
```

where only the first one "known" to pydicom's dictionary, and the other two something are likely something to filter out...
@neurolabusc -- what do you remember about that  `IM_0027_fMAP.dcm` or may be these particular DICOM tags?

### DIFF
--- a/nibabel/nicom/dicomwrappers.py
+++ b/nibabel/nicom/dicomwrappers.py
@@ -565,8 +565,11 @@ class MultiframeWrapper(Wrapper):
         ns_unique = [len(np.unique(row)) for row in self._frame_indices.T]
         shape = (rows, cols) + tuple(ns_unique)
         n_vols = np.prod(shape[3:])
-        if n_frames != n_vols * shape[2]:
-            raise WrapperError('Calculated shape does not match number of frames.')
+        n_frames_calc = n_vols * shape[2]
+        if n_frames != n_frames_calc:
+            raise WrapperError(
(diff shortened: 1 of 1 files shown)


effigies:
I would guess that the 2x2 volumes are echo times x phase/magnitude.

I never use nicom, so I don't understand how people are using this. My understanding was that heudiconv wrapped dcm2niix, so what are you actually using nicom for?

effigies (approved):
This seems fine as an error text improvement. Do you want to do more here to handle the erroring case?

yarikoptic:
not sure if I would have time any time soon -- let's may be just merge as is for now.

we use nibabel.nicom.dicomwrappers in heudiconv to group dicoms based on their metadata, and extracting some metadata to give to heuristics to decide what to name them, but may be worth reapproaching this...

PR #1314, opened 2024-03-29
WIP: generalize class for opening image

matthew-brett:
Allow FileHolder subclasses to change the ImageOpener class.

I wanted to do this to work with the `fsspec` file objects.

But - I couldn't see immediately what the right way to do the type checking is
- I was considering using OpenDef from `openers.py`, but that is only defined
if `ty.TYPE_CHECKING` is True.  I'd love some advice!

### DIFF
--- a/nibabel/fileholders.py
+++ b/nibabel/fileholders.py
@@ -23,6 +23,9 @@ class FileHolderError(Exception):
 class FileHolder:
     """class to contain filename, fileobj and file position"""
 
+    # Default image opener class / callable
+    make_image_opener = ImageOpener
+
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


effigies:
> * I was considering using OpenDef from `openers.py`, but that is only defined
>       if `ty.TYPE_CHECKING` is True.  I'd love some advice!

We could pull that out of the `ty.TYPE_CHECKING` block. The main point of putting it in there is that it was only used internally and it didn't seem usable at runtime. If you're also only using it as a type annotation, then it will be available to import into other modules like:

```python
if ty.TYPE_CHECKING:
    from nibabel.openers import OpenerDef
```

PR #1319, opened 2024-04-25
BUG: wrong origin location in orthoview, update OrthoSlicer3D._set_position in viewers.py

gjpcbecq:
correct a bug with image.orthoview() from class nibabel.viewers.OrthoSlicer3D

There is a wrong selection of indices from original data using the inverse affine transform, leading to weird selection of voxels

This bug is also related to strange behavior with special acquisition, for example with small animal settings such as rodents where PA and IS axis are transposed (LSA system), leading to wrong location of origin (0,0,0) with image.orthoview()

For example, this small example will generate wrong localization of the referential with the main code and will result in good representation with the proposed correction: 

```python 
data1 = np.random.rand(20, 20, 40)
data1[10, 10, :] = 0
data1[10, :, 30] = 0
data1[:, 10, 30] = 0
aff1 = np.array([[1, 0, 0, -10], [0, 0, 1, -30], [0, 1, 0, -10], [0, 0, 0, 1]])
nib.viewers.OrthoSlicer3D(data1, aff1)
```

### DIFF
--- a/nibabel/tests/test_viewers.py
+++ b/nibabel/tests/test_viewers.py
@@ -102,3 +102,35 @@ def test_viewer():
     v2.link_to(v1)  # shouldn't do anything
     v1.close()
     v2.close()
+
+
+@needs_mpl
+def test_viewer_nonRAS():
+    data1 = np.random.rand(10, 20, 40)
+    data1[5, 10, :] = 0
+    data1[5, :, 30] = 0
+    data1[:, 10, 30] = 0
+    # RSA affine
+    aff1 = np.array([[1, 0, 0, -5], [0, 0, 1, -30], [0, 1, 0, -10], [0, 0, 0, 1]])
+    o1 = OrthoSlicer3D(data1, aff1)
+    sag = o1._ims[0].get_array()
+    cor = o1._ims[1].get_array()
+    axi = o1._ims[2].get_array()
+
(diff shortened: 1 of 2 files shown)


effigies:
Thanks for this! Would it be possible to get a test that clearly demonstrates the error before and is fixed after?

effigies:
Hi @gjpcbecq! I'm going to try to turn your MRE into a test.

Apologies for taking so long to get back to this. I'm commenting here to try to pressure myself to spend some time on it. :-)

effigies:
Okay, so I've had a go at reproducing this, and your fix resolves the initial placement at the origin, but if you attempt to click around to other locations with the mouse, you no longer go the place you click. Removing your patch makes the mouse work.

So I think there's something more fundamentally wrong here. It's possible that we just need to do the inverse of this in the `_on_mouse()` function... I'm going to spend a bit more time on this, but if you have a chance to look at it, feel free to take a stab.

gjpcbecq:
Thank you Chris for working on this. 
I was working how to check for the data in the axes, but your solution is very nice. 
There is yet the problem with the _on_mouse function. 
I am working on it and I will propose you next week a solution in a new pull request with a test on the test_viewers.py to check it. 
Best regards, 
Guillaume

PR #1328, opened 2024-06-18
BF: Fix for 'split' (concatenated?) multiframe DICOM

moloney:
Can't just use number of frame indices to determine shape of data, as the actual frames could still be split into different files. Also can't assume a multiframe file is more than a single slice.

I am checking with colleagues if we have some phantom data I can use to create a test case.


moloney:
It seems like the "TestMultiFrameWrapper.test_data_trace" test assumes we will keep any extra-spatial (after 3rd dimension) singular dimensions.  I can keep that behavior if desired but it seems counter intuitive.

moloney:
Closing in favor on #1340

PR #1329, opened 2024-06-26
RF: Prefer using `getlocale()` instead of `getdefaultlocale()`

jhlegarreta:
Prefer using `getlocale()` instead of `getdefaultlocale()`.

Fixes:
```
/home/runner/work/nibabel/nibabel/nibabel/cmdline/dicomfs.py:40:
 DeprecationWarning: 'locale.getdefaultlocale' is deprecated and
 slated for removal in Python 3.15. Use setlocale(), getencoding()
 and getlocale() instead.
      encoding = locale.getdefaultlocale()[1]
```

raised for example at:
https://github.com/nipy/nibabel/actions/runs/9637811213/job/26577586721#step:7:164

### DIFF
--- a/nibabel/cmdline/dicomfs.py
+++ b/nibabel/cmdline/dicomfs.py
@@ -37,7 +37,7 @@ from optparse import Option, OptionParser
 import nibabel as nib
 import nibabel.dft as dft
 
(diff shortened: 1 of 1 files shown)


jhlegarreta:
`doctest` build failures are unrelated.

PR #1331, opened 2024-06-27
RF: Prefer using `"b"` to mark byte literals

jhlegarreta:
Prefer using `"b"` to mark byte literals instead of `np.compat.asbytes`.

Fixes:
```
/home/runner/work/nibabel/nibabel/nibabel/streamlines/tests/test_streamlines.py:9:
 DeprecationWarning: `np.compat`, which was used during the Python 2 to 3 transition,
 is deprecated since 1.26.0, and will be removed
      from numpy.compat.py3k import asbytes
```

raised for example at:
https://github.com/nipy/nibabel/actions/runs/9637811213/job/26577586721#step:7:178


jhlegarreta:
`doctest` build failures are unrelated.

PR #1336, opened 2024-07-05
ENH: Add `.content`, `.text` and `.json` properties to NIfTI extensions

### LINKED ISSUE #1335: NIfTI Extension redesign
Right now the definition of `Nifti1Extension` is (simplified):

```python
class Nifti1Extension:
    def __init__(self, code, content):
        self.code = code
        self._content = self._unmangle(content)

    def get_content():
        return self._content
```

This means that the return type of `get_content()` depends on the subclass, violating the Liskov substitution principle and generally making it difficult to reason about the value of `extension.get_content()` without either inspecting the type of `extension` or `extension.get_content()`.

In #1327 I've found that there are several places in our tests where we've assumed that `get_content()` returns `bytes`, so it is likely that downstream users will also be broken. This seems like an opportunity to redesign.

Instead we could consider following the lead of `requests.Response`/`httpx.Response` and provide a `.content` property that is always `bytes`, a `.text` property that is always `str` (raising an error if decoding fails) and a `.json()` method that decodes the JSON object, which can fail if the contents are not valid JSON.

```python
class Nifti1Extension:
    @property
    def content(self) -> bytes:
        return self._content

    @property
    def text(self) -> str:
        return self._content.decode(getattr(self, 'encoding', 'utf-8'))

    def json(self, **kwargs) -> Any:
        return json.loads(self._content.decode(), **kwargs)
```

This could sit alongside the old, variable-type `get_content()`.

I haven't fully thought through what we want to do for extensions that need to expose complex data types, like pydicom `Dataset`s or `Cifti2Header`s. I'll experiment a little with this with typing to make sure that what I do fits in with typed Python. Probably something like:

```python
class NiftiExtension(ty.Generic[T]):
    _object: T

    def _unmangle(self, contents: bytes) -> T:
        ...

    def _mangle(self, obj: T) -> bytes:
        ...

    def get_content(self) -> T:
        return self._unmangle(self._content)

    # Above stuff...
```

Interested in feedback. Will open a PR with some initial work once I have something that typechecks and tests.

effigies:
The NIfTI-MRS standard (https://github.com/wtclarke/mrs_nifti_standard/) includes a JSON extension. My initial attempt at this (in #1327) created a `NiftiJSONExtension` class, but this had the side-effect of breaking code that already loaded the extension as bytes and manually decoding.

This PR instead includes the following protocol:

```python
class NiftiExtension:
    content: bytes
    text: str
    json: dict[str, Any]
```

This follows the protocol used by requests and httpx and assumes that most extensions will be sufficiently served by one or another of these three attributes (properties, actually), and mangling/unmangling is only needed for more intricate types, such as DICOM.

Nifti1Extension is a non-ideal base class for NIfTI extensions because it assumes that it is safe to use an identity transformation to convert between the on disk and in memory representation of the extension contents, and thus default to `bytes` objects. This makes it difficult to define its typing behavior in a way that allows subclasses to refine the type such that type-checkers understand it.

This patch creates a generic `NiftiExtension` class that parameterizes the "runtime representation" type. Nifti1Extension subclasses with another parameter that defaults to `bytes`, allowing it to be subclassed in turn (preserving the Nifti1Extension -> Nifti1DicomExtension subclass relationship) while still emitting `bytes`. Further, this allows extensions to be created with either bytes or an object (for subclasses that define a particular object type), avoiding unnecessary mangle/unmangle round-trips.

We could have simply made `Nifti1Extension` the base class, but the mangle/unmangle methods need some casts or ignore comments to type-check cleanly. This separation allows us to have a clean base class with the legacy hacks cordoned off into an subclass.

The `Cifti2Extension` needed very little updating.

Closes #1335.


nibabel/nifti1.py
```diff
@@ -427,7 +554,16 @@ class Nifti1DicomExtension(Nifti1Extension):
     header.
     """
 
-    def __init__(self, code, content, parent_hdr=None):
+    code = 2
+    _is_implict_VR: bool = False
```
effigies, line 558:
```suggestion
    _is_implicit_VR: bool = False
```

nibabel/nifti1.py
```diff
@@ -13,12 +13,20 @@
...
+import json
+import sys
+import typing as ty
 import warnings
 from io import BytesIO
 
 import numpy as np
 import numpy.linalg as npl
 
+if sys.version_info <= (3, 12):
```
effigies, line 25:
```suggestion
if sys.version_info < (3, 13):
```

effigies:
A review would be appreciated here. I got back to this because I wanted to play around with serializing xibabel images to NIfTI, optionally dumping "unknown" metadata fields into a JSON extension, and this will make that more convenient.

Does this seem like a reasonable interface?

PR #1338, opened 2024-07-08
MNT: Update coverage config

effigies:
Added global exclusions to target the following cases:

* Typing-specific code that will not be reached by runtime
* `__repr__`. We can test them if it matters, but it usually won't, so don't need to incentivize contributors to add pointless tests.
* Code unreachable in tests (`if 0:`, `if __name__ == '__main__'`)
* Abstract methods (explicitly marked or using `raise NotImplementedError`)

This allows us to remove `# pragma: no cover` comments, which should prevent these cropping up when new chunks of code start with a copy-paste.

Also, I'm removing the exclusion of tests, following https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,19 @@
 [run]
 branch = True
 source = nibabel
-include = */nibabel/*
 omit =
     */externals/*
     */benchmarks/*
     */tests/*
     nibabel/_version.py
+
+[report]
+exclude_also =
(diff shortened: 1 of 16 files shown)


effigies:
I think this should be uncontroversial. Will wait until tomorrow to merge in case anybody wants to look this over.

Issue #977, opened 2020-12-07
Convert DICOM to NIfTI with nibabel


Spenhouet:
I would like to revive this feature request. It was already issued in 2015 here: https://github.com/nipy/nibabel/issues/303
At the time [dcmstack](https://github.com/moloney/dcmstack) was recommended. It combined functions of nibabel to perform this conversion.

dcmstack is no longer maintained (as stated [here](https://github.com/moloney/dcmstack/issues/71#issuecomment-669437175)) and already out-dated. For us it is already missing crucial features (like [enhanced (multiframe) MR DICOMs](https://github.com/moloney/dcmstack/issues/71) or missing Philips headers) which forces us to switch to a temporary solution with dcm2niix (as proposed by nibabel [here](https://nipy.org/nibabel/dicom/dcm2nii_algorithms.html)).

@moloney stated (also [here](https://github.com/moloney/dcmstack/issues/71#issuecomment-669437175)) that this functionality will be added to nibabel.

This seems to be an effort done by @moloney, @matthew-brett, @effigies and @oesteban.
This effort might come into being via the CZI EOSS grant. See the grant application [here](https://github.com/matthew-brett/czi-nibabel).
The milestones outlined [here](https://github.com/matthew-brett/czi-nibabel/blob/master/milestones.md) mention that this functionality should be available 3 months after the grant starts which is January 1, 2021. Not sure if it is reasonable to expect the functionality to be released in nibabel around April 2021?

I would like to have this feature request issue open to track the state of this effort / project.
We would love to have this functionality since it allows us to perform everything in-memory instead of saving conversions to disk as necessary with the dcm2niix solution.

matthew-brett:
Well done for finding the grant application!    We're still finalizing start date - but yes - the plan is to get this done after by 3 months after the start date.  We'll get back to y'all soon on the list when it's clear what the start date will be.

matthew-brett:
By the way, @moloney was also part of the grant discussions, it's just he couldn't easily accept funding with the grant.   We are still planning to work closely with Brendan on the `dcmstack` port.

neurolabusc:
You may want to evaluate [dicom2nifti](https://github.com/icometrix/dicom2nifti). If your remit includes compressed transfer syntaxes, you will need to include some C libraries - dicom2nifti uses gdcmconv for those images. 

[...]

PR #1342, opened 2024-08-12
BF+ENH: Fixes to DICOM scaling, make frame filtering explicit

moloney:
Fixes how we handle DICOM scaling, particularly for Philips and multi-frame files. For Philips data scale factors without defined units should be avoided, and instead a private tag should be used to make image intensities comparable across series. For multi-frame DICOM, it is possible to have different scale factors (potentially coming from different tags) per-frame. We also prefer scale factors from a RealWorldValueMapping provided they have defined units.

The base Wrapper class now has a few new attributes and methods to support this functionality. In particular an attribute `scale_factors` that provides an array of slope/intercept pairs, and a method `get_unscaled_data` that will return the reordered/reshaped data but without the scaling applied. A `vendor` attribute was also added to better support vendor-specific implementation details.

For the MultiFrameWrapper I also added an attribute `frame_order` which exposes the order used to sort the frames, and use this to return the `scale_factors` in sorted order. While implementing this I kept bumping into issues due to the (implicit) frame filtering that was happening in the `image_shape` property, so I made this filtering explicit and configurable and moved it into the class initialization.

### DIFF
--- a/nibabel/nicom/dicomwrappers.py
+++ b/nibabel/nicom/dicomwrappers.py
@@ -13,6 +13,7 @@ than in a property, or property-like thing.
 """
 
 import operator
+import re
 import warnings
 from functools import cached_property
 
@@ -23,6 +24,7 @@ from nibabel.optpkg import optional_package
 from ..openers import ImageOpener
 from . import csareader as csar
 from .dwiparams import B2q, nearest_pos_semi_def, q2bg
+from .utils import Vendor, find_private_section, vendor_from_private
 
 pydicom = optional_package('pydicom')[0]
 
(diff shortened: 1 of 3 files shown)


moloney:
Also, this allows us to get data out of multi-frame files with more than one `StackID`. By default the lowest `StackID` will be used and a warning will be issued, but a user can pass in a frame filter that explicitly selects the `StackID`. I haven't run into this situation yet but it is mentioned [here](https://github.com/nipy/nibabel/issues/977#issuecomment-2110375810).

effigies:
@moloney Thanks for this (and thanks for your patience)!

PR #1349, opened 2024-09-10
BF: Correct viewers _on_mouse, _on_scroll for non RAS matrices

gjpcbecq:
This is the second part of the correction of bugs in viewers.py for matrices with non RAS orientation. 
Corrections have been made on `_on_mouse`and `_on_scroll`. 
I don't know where the original code comes from, but there was an erroneous index positioning a la Matlab style...
There was also confusing ordering of indices and matching from axis to matrices and right values to be passed to the affine transform. 
I think that it can be rewritted to be more easier to understand, but I have no time now. 
Best regards, 
Guillaume

### DIFF
--- a/nibabel/tests/test_viewers.py
+++ b/nibabel/tests/test_viewers.py
@@ -134,3 +134,203 @@ def test_viewer_nonRAS():
     assert_array_equal(sag, data1[6, :, :])
     assert_array_equal(cor, data1[:, :, 32].T)
     assert_array_equal(axi, data1[:, 13, :].T)
+
+
+
+@needs_mpl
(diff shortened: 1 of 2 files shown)


gjpcbecq:
Sorry I did not check for the style... It seems that missing or added spaces and double quotes are not validated. 
I will take this in consideration next time.

effigies:
Just run `ruff check --fix && ruff format`, and commit the results. Should be enough.

gjpcbecq:
> Just run `ruff check --fix && ruff format`, and commit the results. Should be enough.

Doing that asap. 
Many thanks Chris 🙇‍♂️

PR #1350, opened 2024-09-13
BF: `_on_mouse` and `_on_scroll` in viewers for non RAS matrices

gjpcbecq:
I have corrected indices for matrices with flipped dimensions. 
I have also corrected the ordering of RAS dimensions when order of the RAS are different in the data array. 
Perhaps all the code need to be refactored for clarity, using only affine transforms from the viewer to the RAS coordinate system then RAS to ijk, avoiding confusing flipping and reordering. 
Perhaps I will look about that, if I have time.

### DIFF
--- a/nibabel/tests/test_viewers.py
+++ b/nibabel/tests/test_viewers.py
@@ -136,20 +136,19 @@ def test_viewer_nonRAS():
     assert_array_equal(axi, data1[:, 13, :].T)
 
 
-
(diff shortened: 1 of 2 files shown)


effigies (approved):
Thanks for the detailed tests!

gjpcbecq:
You are welcome Chris 🙇‍♂️ thanks for your great job!

PR #1355, opened 2024-09-22
STY: Enforce ruff/flake8-pyi rules (PYI)

DimitriPapadopoulos:


nibabel/openers.py
```diff
@@ -22,6 +22,7 @@
     from types import TracebackType
 
     from _typeshed import WriteableBuffer
+    from typing_extensions import Self
```
effigies, line 25: Need to add this to dependencies. Should guard and import from typing for Python <3.11, and then we can use `typing_extensions >=VER; python_version < '3.11'`.

effigies (approved):

nibabel/openers.py
DimitriPapadopoulos: Add `typing_extensions`.

PR #1356, opened 2024-09-22
STY: Enforce ruff/flake8-pytest-style rules (PT)

DimitriPapadopoulos:


nibabel/tests/test_removalschedule.py
```diff
@@ -125,7 +125,7 @@ def test_module_removal():
     for module in _filter(MODULE_SCHEDULE):
         with pytest.raises(ImportError):
             __import__(module)
-            assert False, f'Time to remove {module}'
+            pytest.fail(f'Time to remove {module}')
```
effigies, line 128: This breaks our test-testing test on L161.

DimitriPapadopoulos: Changed into raise `AssertionError` instead.

nibabel/cmdline/tests/test_roi.py
```diff
@@ -143,7 +143,7 @@ def test_entrypoint(capsys):
         except SystemExit:
             pass
         else:
-            assert False, 'argparse exits on --help. If changing to another parser, update test.'
+            pytest.fail('argparse exits on --help. If changing to another parser, update test.')
```
effigies, line 146: I think the whole `try/except/else` block should be:

```py
with pytest.raises(SystemExit):
    main()
```

nibabel/tests/test_tripwire.py
```diff
@@ -16,9 +16,6 @@ def test_tripwire():
...
         silly_module_name.do_silly_thing
     # Check AttributeError can be checked too
-    try:
+    with pytest.raises(AttributeError) as err:
         silly_module_name.__wrapped__
-    except TripWireError as err:
-        assert isinstance(err, AttributeError)
-    else:
-        raise RuntimeError('No error raised, but expected')
+    assert isinstance(err.value, AttributeError)
```
effigies, line 21: This test was basically verifying that `TripWireError` subclassed `AttributeError`. If we want it to do something different than the `with:` block:

```suggestion
    assert isinstance(err.value, TripWireError)
```

Otherwise I think it would be okay to drop:

```suggestion
```

nibabel/cmdline/tests/test_roi.py
DimitriPapadopoulos: I (think I) applied your suggestion.

effigies:
Failures seem related to https://github.com/scipy/scipy/issues/21623.

PR #1358, opened 2024-09-22
STY: Enforce ruff/flake8-type-checking rules (TCH)

DimitriPapadopoulos:


nibabel/testing/__init__.py
```diff
@@ -26,12 +26,13 @@
 from .helpers import assert_data_similar, bytesio_filemap, bytesio_round_trip
 from .np_features import memmap_after_ufunc
 
+if ty.TYPE_CHECKING:
+    from importlib_resources.abc import Traversable
```
effigies, line 30:
```suggestion
    from importlib.resources.abc import Traversable
```

DimitriPapadopoulos: Indeed. Fixed while rebasing.

PR #1361, opened 2024-09-22
STY: Enforce ruff/Perflint rules (PERF)

DimitriPapadopoulos:


nibabel/tests/test_filehandles.py
```diff
@@ -33,8 +33,7 @@ def test_multiload():
         tmpdir = mkdtemp()
         fname = pjoin(tmpdir, 'test.img')
         save(img, fname)
-        for i in range(N):
-            imgs.append(load(fname))
+        imgs.extend(load(fname) for i in range(N))
```
effigies, line 36: I'm surprised nothing complained about this `i`:

```suggestion
        imgs.extend(load(fname) for _ in range(N))
```

DimitriPapadopoulos: Perhaps because the relevant rules are not (yet) enabled in this specific branch.

PR #1363, opened 2024-09-22
STY: Enforce ruff/pygrep-hooks rules (PGH)

DimitriPapadopoulos:


nibabel/xmlutils.py
```diff
@@ -9,7 +9,7 @@
 """Thin layer around xml.etree.ElementTree, to abstract nibabel xml support"""
 
 from io import BytesIO
-from xml.etree.ElementTree import Element, SubElement, tostring  # noqa
+from xml.etree.ElementTree import Element, SubElement, tostring  # noqa: F401
```
effigies, line 12: I don't think we need this. Just debris from the first pass that nobody looked at hard when linters started complaining.

```suggestion
from xml.etree.ElementTree import Element, tostring
```

DimitriPapadopoulos: It urns out it was neeed after all, to avoid this issue:
```
     File "/home/runner/work/nibabel/nibabel/nibabel/caret.py", line 119, in _to_xml_element
      md = xml.SubElement(metadata, 'MD')
  AttributeError: module 'nibabel.xmlutils' has no attribute 'SubElement'
```

effigies (approved):

PR #1365, opened 2024-09-23
MNT: Apply Repo-Review suggestion MY101

DimitriPapadopoulos:
MY101: MyPy strict mode

Must have `strict` in the mypy config. MyPy is best with strict or nearly strict configuration. If you are happy with the strictness of your settings already, ignore this check or set `strict = false` explicitly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -172,6 +172,7 @@ python_version = "3.11"
 exclude = [
   "/tests",
 ]
+strict = true
 warn_unreachable = true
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Ouchy, that's a lot of errors! Should we address them?

effigies:
If you've got the time and desire, sure.

Tbh, my goal is to make sure checks and suggestions in downstream tools work. I don't have much energy at the moment to fight with mypy.

PR #1369, opened 2024-10-01
STY: Apply ruff preview rules

DimitriPapadopoulos:


nibabel/analyze.py
```diff
@@ -699,7 +699,7 @@ def set_zooms(self, zooms):
         ndim = dims[0]
         zooms = np.asarray(zooms)
         if len(zooms) != ndim:
-            raise HeaderDataError('Expecting %d zoom values for ndim %d' % (ndim, ndim))
```
DimitriPapadopoulos, line 702: Just wondering, does `ndim` really must appear twice in the error message?

nibabel/tests/test_nifti1.py
```diff
@@ -578,12 +578,12 @@ def test_slice_times(self):
         with pytest.raises(HeaderDataError):
             # all None
             hdr.set_slice_times((None,) * len(times))
-        n_mid_times = times[:]
+        n_mid_times = times.copy()
```
DimitriPapadopoulos, line 581: Debatable, as `[:]` has been a Python idiom for such a long time, but I do find the English word `copy` more readable, especially for non-Python programmers.

nibabel/tests/test_round_trip.py
```diff
@@ -108,23 +108,23 @@ def test_round_trip():
     iuint_types = [t for t in iuint_types if t in nifti_supported]
     f_types = [np.float32, np.float64]
     # Expanding standard deviations
-    for i, sd_10 in enumerate(sd_10s):
+    for sd_10 in sd_10s:
```
DimitriPapadopoulos, line 111: `i`, `j`, and `k` are not required, but I believe they appear as “comments” to clarify the intent. Keep them?

nibabel/cmdline/ls.py
```diff
@@ -166,16 +166,16 @@ def proc_file(f, opts):
...
                 row += [f'@l[{np.min(d):.2g}, {np.max(d):.2g}]' if len(d) else '-']
             if opts.counts:
                 items, inv = np.unique(d, return_inverse=True)
                 if len(items) > 1000 and not opts.all_counts:
-                    counts = _err('%d uniques. Use --all-counts' % len(items))
+                    counts = _err(f'{len(items)} uniques. Use --all-counts')
                 else:
                     freq = np.bincount(inv)
-                    counts = ' '.join('%g:%d' % (i, f) for i, f in zip(items, freq))
+                    counts = ' '.join(f'{i:g}:{f:d}' for i, f in zip(items, freq))
```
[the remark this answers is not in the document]
effigies: `g` yes, `d` no.

nibabel/orientations.py
```diff
@@ -322,7 +322,7 @@ def axcodes2ornt(axcodes, labels=None):
            [ 2.,  1.]])
     """
     labels = list(zip('LPI', 'RAS')) if labels is None else labels
-    allowed_labels = sum([list(L) for L in labels], []) + [None]
+    allowed_labels = sum((list(L) for L in labels), []) + [None]
```
DimitriPapadopoulos, line 325: It took me some thinking to understand what this line does. Isn't this code abusing `sum()`? It should at least be moved to a function named `flatten()` or something similar.

Alternatively:
```
    allowed_labels = list(itertools.chain.from_iterable((list(L) for L in labels)))
````
```
    allowed_labels = list(functools.reduce(operator.add, (list(L) for L in labels), []))
````

File: nibabel/cifti2/cifti2_axes.py [hunk 6]
```diff
@@ -373,7 +373,7 @@ def from_mask(cls, mask, name='other', affine=None):
         else:
             raise ValueError(
                 'Mask should be either 1-dimensional (for surfaces) or '
-                '3-dimensional (for volumes), not %i-dimensional' % mask.ndim
```
DimitriPapadopoulos, line 376: As far as I know, Python 3 has dropped support for `%i`.

nibabel/analyze.py
effigies: I think it's reasonable to say you were expecting N zooms for an N-dim image. I think it would also make sense to rewrite something along the lines of `"Zooms must have one value per dimension (ndim={ndim}). Received: {zooms}."`

nibabel/fileslice.py
```diff
@@ -127,7 +127,7 @@ def canonical_slicers(sliceobj, shape, check_inds=True):
             if slicer < 0:
                 slicer = dim_len + slicer
             elif check_inds and slicer >= dim_len:
-                raise ValueError('Integer index %d to large' % slicer)
+                raise ValueError(f'Integer index {slicer} to large')
```
effigies, line 130:
```suggestion
                raise ValueError(f'Integer index {slicer} too large')
```

nibabel/gifti/util.py
```diff
@@ -10,7 +10,7 @@
 from ..volumeutils import Recoder
 
 # Translate dtype.kind char codes to XML text output strings
-KIND2FMT = {'i': '%i', 'u': '%i', 'f': '%10.6f', 'c': '%10.6f', 'V': ''}
+KIND2FMT = {'i': '%d', 'u': '%i', 'f': '%10.6f', 'c': '%10.6f', 'V': ''}
```
effigies, line 13:
```suggestion
KIND2FMT = {'i': '%d', 'u': '%d', 'f': '%10.6f', 'c': '%10.6f', 'V': ''}
```

nibabel/nifti1.py
```diff
@@ -1911,7 +1911,7 @@ def _chk_offset(hdr, fix=False):
             return hdr, rep
         if magic == hdr.single_magic and offset < hdr.single_vox_offset:
             rep.problem_level = 40
-            rep.problem_msg = 'vox offset %d too low for single file nifti1' % offset
+            rep.problem_msg = f'vox offset {offset} too low for single file nifti1'
```
DimitriPapadopoulos, line 1914: I think that's the reason for the current CI error. `offset` is actually a float, but it was printed as in int:
```
  E         - vox offset 10 too low for single file nifti1; setting to minimum value of 352
  E         + vox offset 10.0 too low for single file nifti1; setting to minimum value of 352
```

nibabel/orientations.py
effigies: `sum(list_of_lists, [])` is pretty standard, but this definitely isn't a pretty way of going about it.

If we want to normalize sequences to lists and add (with the marginal benefit of using only builtins):

```suggestion
    allowed_labels = sum(map(list, labels), [None])
```

If we want to chain values and put them in a list at the end:

```suggestion
    allowed_labels = list(chain.from_iterable(labels)) + [None]
```

Definitely don't need to do both.

nibabel/tests/test_round_trip.py
effigies: I don't think the `i`, `j`, `k` provide any additional context here. No need to re-add as comments.

PR #1372, opened 2024-10-08
STY: Apply ruff rules (RUF)

DimitriPapadopoulos:


nibabel/tests/test_analyze.py
```diff
@@ -58,7 +58,7 @@ class TestAnalyzeHeader(tws._TestLabeledWrapStruct):
     header_class = AnalyzeHeader
     example_file = header_file
     sizeof_hdr = AnalyzeHeader.sizeof_hdr
-    supported_np_types = {np.uint8, np.int16, np.int32, np.float32, np.float64, np.complex64}
+    supported_np_types = frozenset((np.uint8, np.int16, np.int32, np.float32, np.float64, np.complex64))
     add_duplicate_types(supported_np_types)
```
effigies, line 62: Right now, `add_duplicate_types` modifies the set in-place, but it could be easily turned into one that accepts a sequence and returns a set. Then you could do:

```
supported np_types = frozenset(expand_duplicate_types((np.uint8, np.int16, np.int32, np.float32, np.float64, np.complex64)))
```

That said, what's the advantage of freezing the set?

nibabel/tests/test_nifti1.py
```diff
@@ -538,7 +538,7 @@ def test_slice_times(self):
         hdr.set_slice_duration(0.1)
         # We need a function to print out the Nones and floating point
         # values in a predictable way, for the tests below.
-        _stringer = lambda val: val is not None and f'{val:2.1f}' or None
+        _stringer = lambda val: (val is not None and f'{val:2.1f}') or None
         _print_me = lambda s: list(map(_stringer, s))
```
effigies, line 542: This is just an old style ternary expression, and we can slide it into the comprehension:

```suggestion
        _print_me = lambda s: [val if val is None else f'{val:2.1f}' for val in s]
```

nibabel/tests/test_volumeutils.py
```diff
@@ -607,7 +607,11 @@ def test_a2f_nanpos():
 
 def test_a2f_bad_scaling():
     # Test that pathological scalers raise an error
-    NUMERICAL_TYPES = sum((sctypes[key] for key in ['int', 'uint', 'float', 'complex']), [])
+    NUMERICAL_TYPES = [
+        x
+        for sublist in (sctypes[key] for key in ('int', 'uint', 'float', 'complex'))
+        for x in sublist
+    ]
```
effigies, line 614: No need for a generator. Might need a rerun of `ruff format`, though...

```suggestion
    NUMERICAL_TYPES = [tp for kind in ('int', 'uint', 'float', 'complex') for tp in sctypes[kind]]
```

nibabel/volumeutils.py
```diff
@@ -35,8 +35,8 @@
     DT = ty.TypeVar('DT', bound=np.generic)
 
 sys_is_le = sys.byteorder == 'little'
-native_code = sys_is_le and '<' or '>'
-swapped_code = sys_is_le and '>' or '<'
+native_code = (sys_is_le and '<') or '>'
+swapped_code = (sys_is_le and '>') or '<'
```
effigies, line 39: Ternary expressions feel cleaner to me. Also seems easier to spot an error if we do this on one line:

```suggestion
native_code, swapped_code = ('<', '>') if sys_is_le else ('>', '<')
```

nibabel/tests/test_analyze.py
DimitriPapadopoulos: That's an error. I hadn't realised `supported_np_types` would be later modified.

PR #1378, opened 2024-10-15
FIX: Restore access to private attr Nifti1Extension._content

effigies:
gh-1336 reused the private attribute ``ext._content`` to exclusively refer to the ``bytes`` representation of the extension contents. This neglected that subclasses might depend on this implementation detail.

Let's be nice to people and rename the attribute to ``_raw`` and provide a ``_content`` property that calls ``self.get_content()``.

Also adds a test to ensure that multiple accesses continue to work as expected.

### DIFF
--- a/nibabel/nifti1.py
+++ b/nibabel/nifti1.py
@@ -326,7 +326,7 @@ class NiftiExtension(ty.Generic[T]):
 
     code: int
     encoding: str | None = None
-    _content: bytes
+    _raw: bytes
     _object: T | None = None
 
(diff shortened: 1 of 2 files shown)


effigies:
@bpinsard Thanks for checking that this works as intended. Would you have time for a code review as well?

bpinsard (approved):
Not 100% familiar with the intricacies of nifti extension but these changes look good to me. Thanks!

effigies:
Thanks!

PR #1380, opened 2024-10-21
FIX: Set MRS type to Nifti1Extension for backwards compatibility

effigies:
cc @wtclarke

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -25,8 +25,8 @@ Eric Larson (EL), Demian Wassermann, Stephan Gerhard and Ross Markello (RM).
 
 References like "pr/298" refer to github pull request numbers.
 
-5.3.2 (Monday 21 October 2024)
-==============================
+5.3.2 (Wednesday 23 October 2024)
+=================================
 
 Bug-fix release in the 5.3.x series.
 
--- a/Changelog
+++ b/Changelog
@@ -25,6 +25,17 @@ Eric Larson (EL), Demian Wassermann, Stephan Gerhard and Ross Markello (RM).
 
 References like "pr/298" refer to github pull request numbers.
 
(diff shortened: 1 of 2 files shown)


wtclarke:
Does `Nifti1Extension` still have the .json() methods? I've just updated the nifti-mrs package to use the new interface. So if the new methods aren't available for this, then I'm happy to just make the changes on my side and not have this fix.

effigies:
Yes, `Nifti1Extension` subclasses `NiftiExtension[bytes]` so `_mangle()` and `_unmangle()` are just identity functions, as was the previous default. The purpose of splitting the classes was so that new subclasses could derive from the base class and declare their types correctly.

wtclarke:
Great, I see now, I was getting confused between `Nifti1Extension` and `Nifti1Extensions`. Thanks for the fix.

effigies:
I guess with the problem solved on your end, we can take a minute to think about how you would like to evolve.

If you would like to offload more of your logic into nibabel and use a `NiftiExtension` directly, then it's possible that creating a `JSONExtension` subclass would make more sense. That would allow `ext.get_object()` to return a cached value that you can mutate. When it's time to save to a new image, that object will override the original `bytes` content.

If read-only access is all you want, then this PR will at least get you back what you had, without users having to upgrade your package if they upgrade nibabel. In practice, though, I expect most people just install the latest versions, and so making this release or not will impact very few people.

effigies:
Well, regardless of the long-term path forward, this has the potential to reduce pain and doesn't make changing in the future harder.

wtclarke:
Hi @effigies, Thanks for this fix. Sorry for the slow reply, four new starters in the group has me rushing about this week. For now I'm happy to keep the logic on my side.

PR #1387, opened 2024-11-20
BF+TST: Fix 'frame_order' for single frame files

moloney:
The `_frame_indices` attribute on the `MultiframeWrapper` was empty for single frame files, causing use of the `frame_order` property to raise an exception.

Also, don't assume StackID is int (it is a str), don't assume DimensionIndexValues has more than one value. Fix how we assign multiple values to DimensionIndexValues when making fake test data (need to pass list instead of tuple).

### DIFF
--- a/nibabel/nicom/dicomwrappers.py
+++ b/nibabel/nicom/dicomwrappers.py
@@ -532,8 +532,8 @@ class Wrapper:
 class FrameFilter:
     """Base class for defining how to filter out (ignore) frames from a multiframe file
 
-    It is guaranteed that the `applies` method will on a dataset before the `keep` method
-    is called on any of the frames inside.
+    It is guaranteed that the `applies` method will called on a dataset before the `keep`
+    method is called on any of the frames inside.
     """
 
     def applies(self, dcm_wrp) -> bool:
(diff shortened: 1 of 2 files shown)


nibabel/nicom/dicomwrappers.py
```diff
@@ -742,14 +749,20 @@ def image_shape(self):
...
+            return (rows, cols)
+        # Initialize array of frame indices
         try:
             frame_indices = np.array(
                 [frame.FrameContentSequence[0].DimensionIndexValues for frame in self.frames]
             )
         except AttributeError:
             raise WrapperError("Can't find frame 'DimensionIndexValues'")
+        if len(frame_indices.shape) == 1:
+            frame_indices = frame_indices.reshape(frame_indices.shape + (1,))
```
effigies, line 765: This should achieve the same result, and I think is a bit more immediately obvious. Might even avoid a copy.

```suggestion
        try:
            frame_indices = np.atleast_2d(
                [frame.FrameContentSequence[0].DimensionIndexValues for frame in self.frames]
            )
        except AttributeError:
            raise WrapperError("Can't find frame 'DimensionIndexValues'")
```

effigies (approved):
LGTM. Small suggestion for your consideration.

PR #1389, opened 2024-11-24
MNT: Update ruff to 0.8.2

DimitriPapadopoulos:


effigies:
@DimitriPapadopoulos Is any of this still undone from #1405?

PR #1399, opened 2024-12-06
chore(rtd): Build API docs

effigies:
Missed in previous PR.


effigies:
That's better: https://nibabel-draft--1399.org.readthedocs.build/en/1399/api.html#api

PR #1404, opened 2025-01-12
ENH: Slightly simplify frame_indices array construction

effigies:
Turning https://github.com/nipy/nibabel/pull/1387#discussion_r1874439909 into a PR.


effigies:
My initial proposal turned out to be wrong, as `atleast_2d` adds dimensions to the front instead of the back. This instead maps `atleast_1d` over the arrays that are scalar or 1d, which feels slightly clearer in intent to me, but not so much that I'd insist on this.

Closing as not worth the churn.

PR #1405, opened 2025-01-12
chore: Update ruff, address complaints

effigies:
The churn machine demands churn.

### DIFF
--- a/nibabel/benchmarks/bench_array_to_file.py
+++ b/nibabel/benchmarks/bench_array_to_file.py
@@ -29,24 +29,25 @@ def bench_array_to_file():
     sys.stdout.flush()
     print_git_title('\nArray to file')
     mtime = measure('array_to_file(arr, BytesIO(), np.float32)', repeat)
-    print('%30s %6.2f' % ('Save float64 to float32', mtime))
+    fmt = '{:30s} {:6.2f}'.format
+    print(fmt('Save float64 to float32', mtime))
     mtime = measure('array_to_file(arr, BytesIO(), np.int16)', repeat)
-    print('%30s %6.2f' % ('Save float64 to int16', mtime))
+    print(fmt('Save float64 to int16', mtime))
     # Set a lot of NaNs to check timing
     arr[:, :, :, 1] = np.nan
     mtime = measure('array_to_file(arr, BytesIO(), np.float32)', repeat)
-    print('%30s %6.2f' % ('Save float64 to float32, NaNs', mtime))
+    print(fmt('Save float64 to float32, NaNs', mtime))
     mtime = measure('array_to_file(arr, BytesIO(), np.int16)', repeat)
-    print('%30s %6.2f' % ('Save float64 to int16, NaNs', mtime))
+    print(fmt('Save float64 to int16, NaNs', mtime))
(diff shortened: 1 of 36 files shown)


effigies:
@DimitriPapadopoulos Care to have a look?

nibabel/benchmarks/bench_array_to_file.py
```diff
@@ -29,24 +29,25 @@ def bench_array_to_file():
     sys.stdout.flush()
     print_git_title('\nArray to file')
     mtime = measure('array_to_file(arr, BytesIO(), np.float32)', repeat)
-    print('%30s %6.2f' % ('Save float64 to float32', mtime))
+    fmt = '{:30s} {:6.2f}'.format
+    print(fmt('Save float64 to float32', mtime))
```
DimitriPapadopoulos, line 33: Have you considered f-strings instead of `str.format`?
```suggestion
    s_width = 30
    f_format = "6.2f"
    print(f'{"Save float64 to float32":{s_width}}, {mtime:{f_format}}')
```
However, f-strings might be more verbose and less readable in this case.

effigies: I did that at first, but preferred to use a format string instead. It's less boilerplate.

nibabel/volumeutils.py
```diff
@@ -338,12 +338,7 @@ def pretty_mapping(
     if getterfunc is None:
         getterfunc = getitem
     mxlen = max(len(str(name)) for name in mapping)
-    fmt = '%%-%ds  : %%s' % mxlen
-    out = []
-    for name in mapping:
-        value = getterfunc(mapping, name)
-        out.append(fmt % (name, value))
-    return '\n'.join(out)
+    return '\n'.join([f'{name:{mxlen}s}  : {getterfunc(mapping, name)}' for name in mapping])
```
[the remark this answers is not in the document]
effigies: Sure, we can remove the `[]`, though it's purely aesthetic. `str.join(...)` is not faster than `str.join([...])` and can be slightly slower since it needs to allocate a generator which then gets turned into a list.

I don't want to change what the function produces in a style PR, so I'll leave the double space.

```suggestion
    return '\n'.join(f'{name:{mxlen}s}  : {getterfunc(mapping, name)}' for name in mapping)
```

DimitriPapadopoulos:
Finally, perhaps these rules be enabled:
```
    "ISC001",
    "ISC002",
```

Indeed, the documentation has been updated under [Conflicting lint rules](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules) in https://github.com/astral-sh/ruff/pull/15123 / https://github.com/astral-sh/ruff/commit/b76d05e2839bf6f0363de694ab6632640d6fb012, in time for 0.9.1:
> [`multi-line-implicit-string-concatenation`](https://docs.astral.sh/ruff/rules/multi-line-implicit-string-concatenation/) (`ISC002`) if used without `ISC001` and `flake8-implicit-str-concat.allow-multiline = false`

The default value of [lint.flake8-implicit-str-concat.allow-multiline](https://docs.astral.sh/ruff/settings/#lint_flake8-implicit-str-concat_allow-multiline) is `true`.

PR #1411, opened 2025-02-16
type: Track type of SpatialImage.affine, test type inference

effigies:
This PR lays some groundwork for further type annotations. The main thing it does is make `SpatialImage` generic over its affine type, which can either be `Affine` (an alias for ndarray that might be refined in the future) or `None`.

This is a somewhat pedantic thing to track, as it is extremely rare not to provide an affine to an image, but there are cases where it is the correct thing to do. This allows for a forthcoming PR to annotate functions that require an affine or can accept an image with `None` in it.

It also introduces tests where we specifically test for what `mypy` says the types of variables are, which lets us verify not only that our types are internally valid, but that they expose what we expect to downstream users.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,22 +239,18 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        check: ['style', 'doctest', 'typecheck', 'spellcheck']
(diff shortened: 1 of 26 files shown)


effigies:
This is ready to go. Reviews welcome, or else will merge on Thursday, Feb 20.

jhlegarreta (approved):

PR #1415, opened 2025-05-14
Support ushort for the mgh/mgz reader

bthyreau:
Hi,

nibabel currently crashes when loading an mgh/mgz file of type unsigned short.
As an consequence, recent FreeSurfer 7.4/8 clinical pipelines (that uses nibabel) now crashes on uint16 input MRIs (of any format, e.g nifti1) with an unfriendly python traceback, _because it can't read back the mgz it just converted its input to_.

I noted a bit of conflicting opinion on whether ushort should be supported for the MGH/MGZ file format.
- Nibabel's freesurfer explicitly (e.g test_mghformat.py:bad_dtype_mgh) forbids it for new object creation. And ANTs ITK-based tools refuse both to read and to write ushort mgz.
- But on the other hand, Freesurfer's own "mri_info" fully handle it, as so do other tools. In particular "mri_convert" obviously too (though not for explicit cast, ie. the -odt option!) considering FreeSurfer itself may create such ushort mgz out of well-supported regular ushort nifti files - without user's easily actionable workaround.

Therefore, officially supported or not, i suggest there is little drawback in nibabel not crashing when loading such ushort mgz.
The attached one-liner do that.

### DIFF
--- a/nibabel/freesurfer/mghformat.py
+++ b/nibabel/freesurfer/mghformat.py
@@ -62,6 +62,7 @@ _dtdefs = (  # code, conversion function, dtype, bytes per voxel
     (4, 'int16', '>i2', '2', 'MRI_SHORT', np.int16, np.dtype('i2'), np.dtype('>i2')),
     (1, 'int32', '>i4', '4', 'MRI_INT', np.int32, np.dtype('i4'), np.dtype('>i4')),
     (3, 'float', '>f4', '4', 'MRI_FLOAT', np.float32, np.dtype('f4'), np.dtype('>f4')),
+    (10, 'int16', '>i2', '2', 'MRI_SHORT', np.int16, np.dtype('i2'), np.dtype('>i2')),
 )
 
 # make full code alias bank, including dtype column
--- a/nibabel/freesurfer/mghformat.py
+++ b/nibabel/freesurfer/mghformat.py
@@ -57,6 +57,10 @@ hf_dtype = np.dtype(header_dtd + footer_dtd)
 
 # caveat: Note that it's ambiguous to get the code given the bytespervoxel
 # caveat 2: Note that the bytespervox you get is in str ( not an int)
+# FreeSurfer historically defines codes 0-10 [1], but only a subset is well supported.
+# Here we use FreeSurfer's MATLAB loader [2] as an indication of current support.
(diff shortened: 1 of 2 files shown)


effigies:
Thanks! I agree, if FreeSurfer creates it, we should be able to read it.

Do you have a link to FreeSurfer source that we can cross-reference? If we add it in a comment, it will make it easier to look up in the future, if they add more dtypes (however unlikely that may be).

effigies:
Looks like:

* https://github.com/freesurfer/freesurfer/blob/3cc9069/include/mri.h#L53-L63
* https://github.com/freesurfer/freesurfer/blob/3cc9069/matlab/load_mgh.m#L195-L207

I guess I would follow the load_mgh.m fields, since others don't seem to be used.

bthyreau:
Thanks for the quick reply and fixing my botched commit. Indeed, it seems that FreeSurfer introduced the USHORT into mgz only in 2021, in freesurfer/freesurfer@b82c3a8dfeb283260ba6b1be6cb088194833116b, which may explain why it is not consistently mentionned everywhere. 
I agree with you that supporting only the 5 types is probably the best for now, as the other types defined in mri.h are rarely used, if at all.

nibabel/freesurfer/mghformat.py
```diff
@@ -59,9 +59,10 @@
 # caveat 2: Note that the bytespervox you get is in str ( not an int)
```
effigies, line 59:
```suggestion
# caveat 2: Note that the bytespervox you get is in str ( not an int)
# FreeSurfer defines codes 0-10, but only seems to support a subset, and we will use their MATLAB
# loader as an indication of intended support.
# Full definitions: https://github.com/freesurfer/freesurfer/blob/3cc9069/include/mri.h#L53-L63
# Support: https://github.com/freesurfer/freesurfer/blob/3cc9069/matlab/load_mgh.m#L195-L207
```

effigies (commented):
Adding a comment so we can find these links again easily. Can you remove the now-failing test, or select another unsupported datatype (like `float16`)?

bthyreau:
Thanks a lot. I updated the comment in the code. 
To update the failure-case test, i picked float64 (double) that is popular with FSL Nifti, yet poorly supported by FreeSurfer in general and fully unsupported by its mgz format

PR #1420, opened 2025-06-30
Add ability to read/write freesurfer surface files in scanner ras

jduemose:
Previously, only freesurfer surface files with coordinates in unknown or surface ras/tkregister space could be read. Following #1419, this PR adds the ability to read surface files whose coordinates are stored in scanner ras as well.

Let me know if calling a freesurfer function via `subprocess` as part of the test is a problem - but I wasn't sure how to test this feature otherwise.

### DIFF
--- a/nibabel/freesurfer/tests/test_io.py
+++ b/nibabel/freesurfer/tests/test_io.py
@@ -7,6 +7,7 @@ import unittest
 from os.path import isdir
 from os.path import join as pjoin
 from pathlib import Path
+import subprocess
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


nibabel/freesurfer/tests/test_io.py
```diff
@@ -90,6 +91,13 @@ def test_geometry():
         assert any('volume information contained' in str(ww.message) for ww in w)
         assert any('extension code' in str(ww.message) for ww in w)
 
+        # Test reading/writing a surface file in scanner RAS
+        cmd = f"mris_convert --to-scanner {pjoin(data_path, 'surf', 'lh.inflated')} {surf_path}"
```
effigies, line 95: We don't install FreeSurfer to run tests. Could you contribute the converted file to https://codeberg.org/nipy-data/nitest-freesurfer?

PR #1421, opened 2025-07-05
ENH: Allow path-like objects for tractogram load/saving

jhlegarreta:
Allow path-like objects for tractogram load/saving.

Crucially, this patch set uses the capabilities of `pathlib.Path` to get the suffix of a filename when detecing the format instead of relying on the `splitext` function of the `os.path` module, which requires a `str` instance.

Add the corresponding tests.

Change the docstrings to mark the types of the filenames as `path-like` objects instead of only `str` objects.


jhlegarreta:
Style errors predated this PR. Addressed in PR #1423.

jhlegarreta:
@effigies this is ready to be merged.

effigies:
> relying on the splitext function of the os.path module, which requires a str instance.

```py
In [3]: import os.path

In [4]: from pathlib import Path

In [5]: os.path.splitext('/path/to/streamline.tck')
Out[5]: ('/path/to/streamline', '.tck')
```

`os.path` was retrofitted to allow `PathLike`s in almost all of its functions. I don't think this change is necessary, but documenting and testing the support for `PathLike`s is welcome. Could you try reverting the splitext change and see if the tests still pass?

PR #1422, opened 2025-07-05
ENH: Use appropriate message when reporting TCK `file` attribute error

jhlegarreta:
Use appropriate message when reporting `file` attribute error in TCK: the previous message was mistakenly reporting the non-existence of the `file` attribute, which indeed has to exist to be able to do the `hdr['file']` without getting a `KeyError`. This patch set uses the message that was not being used when the check was not being successful.


jhlegarreta:
@effigies this is ready to be merged.

effigies (approved):
Good catch!

PR #1423, opened 2025-07-05
STYLE: Call `write` once instead of looping

jhlegarreta:
Call `write` once instead of looping by concatenating all strings that need to be written.

Avoid using `writelines` as `ImageOpener` objects do not have that attribute.

Fixes:
```
-                for val in bvals:
-                    fid.write(f'{val} ')
+                fid.writelines(f'{val} ' for val in bvals)
```

and other similar errors raised for example in:
https://github.com/nipy/nibabel/actions/runs/16092302666/job/45410655457?pr=1422#step:7:23

Documentation:
https://docs.astral.sh/ruff/rules/for-loop-writes/

### DIFF
--- a/nibabel/cmdline/parrec2nii.py
+++ b/nibabel/cmdline/parrec2nii.py
@@ -358,8 +358,7 @@ def proc_file(infile, opts):
             )
             with open(basefilename + '.bvals', 'w') as fid:
                 # np.savetxt could do this, but it's just a loop anyway
-                for val in bvals:
-                    fid.write(f'{val} ')
+                fid.write(' '.join(str(val) for val in bvals) + ' ')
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Typecheck issues being addressed in PR #1424.

@effigies This is ready to be merged.

effigies:
@jhlegarreta Thanks for your patience. I've looked into this. While the bvals one is almost certainly a trivial increase in memory usage, the purpose of writing the blocks of zeros using a loop was to avoid increasing memory usage. The `writelines()` change recommended by ruff does not concatenate internally, instead making multiple `os.write()` calls, so passing a generator expression should not cause a slowdown or increase in memory use.

Applied in #1433.

Issue #1430, opened 2025-09-04
PAR/REC Changes for Philips Software R11


alexdaniel654:
My site has just upgraded our scanner from software release R5.9 to R11. When I've tried to read in the test data I acquired I get the following error:
```
File "c:\Users\ppzajd\Desktop\Temp\nibabel_testing\nibabel_r11_test.py", line 5, in <module>
    r11_img = nib.load('R11_T2W_TSE.PAR')
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ppzajd\Git\nibabel\nibabel\loadsave.py", line 114, in load
    img = image_klass.from_filename(filename, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ppzajd\Git\nibabel\nibabel\parrec.py", line 1332, in from_filename
    return klass.from_file_map(
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ppzajd\Git\nibabel\nibabel\parrec.py", line 1295, in from_file_map
    hdr = klass.header_class.from_fileobj(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ppzajd\Git\nibabel\nibabel\parrec.py", line 767, in from_fileobj
    info, image_defs = parse_PAR_header(fileobj)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ppzajd\Git\nibabel\nibabel\parrec.py", line 524, in parse_PAR_header
    general_info = _process_gen_dict(gen_dict)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\ppzajd\Git\nibabel\nibabel\parrec.py", line 351, in _process_gen_dict
    props = _hdr_key_dict[key]
            ~~~~~~~~~~~~~^^^^^
KeyError: 'Max. number of Diffusion values'
```
Having done a bit of debugging it seems that Philips have _helpfully_ capitalised the D of diffusion in the general information key of the PAR files in this software release i.e. the old PAR files was
```
# === GENERAL INFORMATION ========================================================
#
.    Patient name                       :   R_5_Data

...

.    Diffusion echo time [ms]           :   0.0000
.    Max. number of diffusion values    :   1
.    Max. number of gradient orients    :   1
```
and is now
```
[...]

PR #1432, opened 2025-09-15
BF+TST: Philips R11 PAR/REC Changes

alexdaniel654:
This should fix the issue that nibabel can't read PAR/REC files produced from Philips software R11. More information is in #1430


effigies (approved):
LGTM. Thanks!

Will merge after tests pass.

Issue #1426, opened 2025-08-11
pylance warning reportPrivateImportUsage


pcasl:
import nibabel 
data_path = 'path/to/your/data.nii'
img = nibabel.load(data_path)

It raises up warning around the load. Any suggestions? Thanks

pcasl:
python3.13 nibabel 5.3.2

effigies:
I would ask the pylance developers how to reproduce this more directly. vscode is too complex to be a useful reproduction.

shnizzedy:
I think adding an `__all__` to [`__init__.py`](https://github.com/nipy/nibabel/blob/4ee2a49cccbd317fedac4be9d21308782d552edf/nibabel/__init__.py) would resolve this warning (by explicitly making the intended symbols public), but I don't know if there's a design reason to not set `__all__` here?

PR #1436, opened 2025-09-24
PL: Add stub file for top-level `__init__.py`

shnizzedy:
Related to #1426 by @pcasl
Related to #1220 by @effigies 

## Description

This PR adds a stub file for the top-level `__init__.py`, including an `__all__` to expose the types of all symbols imported or defined therein.

### Without the stub file

warning: &quot;load&quot; is not exported from module &quot;nibabel&quot;

### With the stub file

no warning

### DIFF
--- a/nibabel/__init__.pyi
+++ b/nibabel/__init__.pyi
@@ -44,11 +44,6 @@ from nibabel.pkg_info import get_pkg_info as _get_pkg_info
 from nibabel.spm2analyze import Spm2AnalyzeHeader, Spm2AnalyzeImage
 from nibabel.spm99analyze import Spm99AnalyzeHeader, Spm99AnalyzeImage
 
-try:
-    from pytest import ExitCode  # type: ignore[import]  # noqa: PT013
-except (ImportError, ModuleNotFoundError):
-    class ExitCode(int): ...  # type: ignore[no-redef]
-
 def get_info() -> dict[str, str]: ...
 def test(
     label: Any = None,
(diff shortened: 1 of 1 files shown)


nibabel/__init__.pyi
```diff
@@ -0,0 +1,117 @@
...
+    verbose: int = 1,
+    extra_argv: list[Any] | None = None,
+    doctests: bool = False,
+    coverage: bool = False,
+    raise_warnings: Any = None,
+    timer: Any = False,
+) -> ExitCode: ...
+def bench(
+    label: Any = None, verbose: int = 1, extra_argv: list[Any] | None = None
+) -> ExitCode: ...
```
effigies, line 64: Is there a reason not to just do `int`?

nibabel/__init__.pyi
```diff
@@ -0,0 +1,117 @@
...
+
+__all__ = [
```
effigies, line 66: I'm curious if the `__all__` is necessary in a `.pyi` file. A stub is purely for external use, so I would have thought a symbol showing up here implies that it's exported.

nibabel/__init__.pyi
effigies, line 64:
```suggestion
def get_info() -> dict[str, str]: ...
def test(
    label: Any = None,
    verbose: int = 1,
    extra_argv: list[Any] | None = None,
    doctests: bool = False,
    coverage: bool = False,
    raise_warnings: Any = None,
    timer: Any = False,
) -> int: ...
def bench(
    label: Any = None, verbose: int = 1, extra_argv: list[Any] | None = None
) -> int: ...
```

nibabel/__init__.pyi
```diff
@@ -0,0 +1,112 @@
...
+    verbose: int = 1,
+    extra_argv: list[Any] | None = None,
+    doctests: bool = False,
+    coverage: bool = False,
+    raise_warnings: Any = None,
+    timer: Any = False,
+) -> int: ...
+def bench(
+    label: Any = None, verbose: int = 1, extra_argv: list[Any] | None = None
+) -> int: ...
```
effigies, line 59:
```suggestion
def bench(label: Any = None, verbose: int = 1, extra_argv: list[Any] | None = None) -> int: ...
```

PR #1005, opened 2021-03-23
NF: Add zstd compression support

vanandrew:
Some stuff I was playing around with... This PR adds zstd compression/decompression support for the `Opener` class.

[zstd (Zstandard)](https://facebook.github.io/zstd/) is a relatively new compression library that boasts significant improvements in compression/decompression speeds over zlib with minimal differences in compression ratios.

Results from `bench_fileslice.py`  with `zst` added:
```
*************************
Bytes slice - raw (ratio)
*************************

+-----------------+--------------+--------------+--------------+--------------+
|                 |   mid int    |    step 1    | half step 1  | step mid int |
+=================+==============+==============+==============+==============+
| axis 0, len 64  | 0.73 (4.40)  | 0.23 (1.38)  | 0.21 (1.25)  | 0.19 (1.15)  |
| axis 1, len 64  | 0.02 (0.12)  | 0.17 (1.01)  | 0.13 (0.81)  | 0.02 (0.14)  |
| axis 2, len 32  | 0.01 (0.04)  | 0.12 (0.73)  | 0.13 (0.77)  | 0.02 (0.14)  |
| axis 3, len 100 | 0.00 (0.00)  | 0.15 (0.90)  | 0.09 (0.56)  | 0.00 (0.03)  |
+-----------------+--------------+--------------+--------------+--------------+
Base time: 0.17

************************
File slice - raw (ratio)
************************

+-----------------+--------------+--------------+--------------+--------------+
|                 |   mid int    |    step 1    | half step 1  | step mid int |
+=================+==============+==============+==============+==============+
| axis 0, len 64  | 1.19 (10.13) | 0.16 (1.40)  | 0.16 (1.36)  | 0.11 (0.96)  |
| axis 1, len 64  | 0.03 (0.24)  | 0.12 (1.04)  | 0.17 (1.44)  | 0.07 (0.59)  |
| axis 2, len 32  | 0.01 (0.07)  | 0.17 (1.42)  | 0.15 (1.28)  | 0.02 (0.15)  |
| axis 3, len 100 | 0.00 (0.01)  | 0.15 (1.31)  | 0.07 (0.60)  | 0.01 (0.04)  |
+-----------------+--------------+--------------+--------------+--------------+
Base time: 0.12

**********************
gz slice - raw (ratio)
**********************

+-----------------+--------------+--------------+--------------+--------------+
|                 |   mid int    |    step 1    | half step 1  | step mid int |
+=================+==============+==============+==============+==============+
[...]

PR #1444, opened 2025-11-26
Use `compression.zstd` or `backports.zstd` in preference to `pyzstd`

pauldmccarthy:
Hi @effigies, I hope you're well! Also pinging @vanandrew, as this builds on the work you did in #1005.

This PR adjusts `nibabel` so that it will use one of  `compression.zstd`, `backports.zstd`, or `pyzstd`, in that order of preference, for handling zstandard-compressed files.

I've been playing around with adding zstandard support to the core FSL tools, and noticed that `nibabel` currently relies on `pyzstd`. Support for this format is now built into Python as of 3.14 via `compression.zstd`, and the recommendation for older Python versions is to use the `backports.zstd` package, as outlined in this handy migration guide:

https://pyzstd.readthedocs.io/en/stable/

1. To handle conditional import of one of several optional modules, I've adjusted the `nibabel.optpkg.optional_package` function to accept either a single module name, or a sequence of names - it will try to import each in order, returning the first one that succeeds.

2. I moved `_zstd_open`, `_gzip_open`, and `DeterministicGzipFile`  from `nibabel.openers` into `nibabel._compression`, as they seem like a better fit there.

3. The `compression.zstd.ZstdFile` API differs very slightly from the `pyzstd.ZstdFile` API - the latter expects an overloaded `level_or_option` parameter, whereas the former expects separate `level` / `option` parameters. I've adjusted nibabel so that it will accept all three, and will pass the appropriate parameter to the underlying library.

4. I've tried to update the metadata files (`uv.lock`, `tox.ini` etc), but am not particularly familiar with how the project is managed these days, so let me know if there's anything I ned to fix, or anything else that is needed.

Thanks!

### DIFF
--- a/nibabel/_compression.py
+++ b/nibabel/_compression.py
@@ -183,19 +183,14 @@ def zstd_open(
         alert_future_error(
             'The level_or_option parameter will be removed in a future version of nibabel',
             '7.0',
-            warning_rec='This warning can be silenced by using the separate '
-            'level/option parameters',
+            warning_rec='This warning can be silenced by using the separate level/option parameters',
             error_rec='Future errors can be avoided by using the separate level/option parameters',
             error_class=TypeError,
         )
-        level_or_option_provided = sum(
-            (level_or_option is not None, level is not None, options is not None)
-        )
-        if level_or_option_provided > 1:
+        if level is not None or options is not None:
             raise ValueError('Only one of level_or_option, level or options may be specified')
-        if level_or_option is not None:
-            if isinstance(level_or_option, int):
-                level = level_or_option
-            else:
-                options = level_or_option
+        if isinstance(level_or_option, int):
+            level = level_or_option
+        else:
+            options = level_or_option
     return zstd.ZstdFile(filename, mode, level=level, options=options, zstd_dict=zstd_dict)


nibabel/arrayproxy.py
```diff
@@ -36,6 +36,7 @@
 import numpy as np
 
 from . import openers
+from ._compression import HAVE_INDEXED_GZIP
```
effigies:
```suggestion
from . import _compression
```

pauldmccarthy:
I found an unrelated issue while testing locally:

```
___________________________________________________________ TestRunAllTests.test_first ____________________________________________________________

self = <nibabel.tests.test_api_validators.TestRunAllTests object at 0x7dcc2a3d3ca0>

    def meth(self):
        for imaker, params in self.obj_params():
>           validator(self, imaker, params)

nibabel/tests/test_api_validators.py:22:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <nibabel.tests.test_api_validators.TestRunAllTests object at 0x7dcc2a3d3ca0>, obj = 1, param = 2

    def validate_first(self, obj, param):
>       self.run_tests.add('first')
E       AttributeError: 'dict' object has no attribute 'add'

nibabel/tests/test_api_validators.py:108: AttributeError
``` 

It looks like `run_tests` should be a `set()`?

effigies (approved):
LGTM. Small logic simplification.

Issue #1460, opened 2026-02-02
`nibabel.orientations.aff2axcodes` potentially returns incorrect results


dkuegler:
Hello,

I am using `aff2axcodes` to determine the `axcodes` of an affine matrix and the results are a bit convoluted.

Specifically, this is about affine matrices, that are rotated around 45 degrees in multiple directions. I do not even know what the "correct" results really should be...

Let us consider the following affine matrix:

```python
affine = \
  [[-0.585182553995787, 0.5048269789762401, -0.6345952251606463, -2.218487624719689],
   [-0.5327455539210799, 0.35065247835655966, 0.7702110192666194, 2.028722778552794], 
   [-0.6113456904863974, -0.7887918361140193, -0.06374861569935834, 4.227025896592773], 
   [0.0, 0.0, 0.0, 1.0]]
```
This matrix is orthonormal and valid (generated with `scipy.spatial.transform.Rotation.from_rotvec().as_matrix()`):
```python
np.linalg.inv(affine[:3,:3]) - affine[:3, :3].T
# test for orthonormal, approx zero
# [[ 2.22044605e-16  3.33066907e-16  1.11022302e-16], [-3.33066907e-16 -2.22044605e-16  2.22044605e-16], [ 3.33066907e-16 -5.55111512e-16  1.66533454e-16]]
np.linalg.norm(affine[:3,:3], axis=0) , np.linalg.norm(affine[:3,:3], axis=1)
# test for length of vectors: ([1. 1. 1.], [1. 1. 1.])
```

an even more extreme affine is
```python
affine = \
  [[0.14794184343060812, 0.4876343794531119, -0.6167057020473639, -0.38513162667728795],
   [0.5617554027693318, -0.5045939256658585, -0.26422686774925774, -0.13797634685442106],
   [0.5500400698329341, 0.38418443768129973, 0.4357272534759381, 0.652584993353186],
   [0.0, 0.0, 0.0, 1.0]] # scale factors are 0.8, 0.8, 0.8
```
which following this paradigm is either `'asl'` or `'ars'` --> the second vector may be least aligned with dimension 3, but nonetheless, vectors 1 and 3 are more aligned with the other two directions still.

Running `nibabel.orientations.aff2axcodes`:
```python
"".join(nib.orientations.aff2axcodes(affine, ("lr", "pa", "is"))) # 'ira'
```

The problem here is, that it really should be `'lia'`!, why:

[...]

PR #1461, opened 2026-02-02
Fix nibabel.orientations.io_orientation

dkuegler:
This PR fixes the behavior of orientations.io_orientation, which extracts the `ornt`-array from an affine. Thus far, it always prioritizes the primary direction of the first column vector over other more aligned column vectors of the affine. Now, it ranks how well the axes align with the coordinate axes and iteratively selects the most aligned axis, eliminating that axis from further consideration. Currently, there are cases where the order of the column vectors impact results, now, changing the order of column vectors is a stable operation.

E.g. previously, an affine might have returned ASL, but after "singular" as_closest_canonical alignment does not return RAS (because of this instability).

This solution fixes several problems in my code, and would impact other functions using `io_orientation` such as `as_closest_canonical` and others (many indirectly).

I described the issue in #1460 as well.

### DIFF
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -70,27 +70,18 @@ def io_orientation(affine, tol=None):
     # the matrix R is such that np.dot(R,R.T) is projection onto the
     # columns of P[:,keep] and np.dot(R.T,R) is projection onto the rows
     # of Qs[keep].  R (== np.dot(R, np.eye(p))) gives rotation of the
-    # unit input vectors to output coordinates.  Therefore, the row
-    # index of abs max R[:,N], is the output axis changing most as input
-    # axis N changes.  In case there are ties, we choose the axes
-    # iteratively, removing used axes from consideration as we go
+    # unit input vectors to output coordinates.  Therefore, we iteratively
+    # select the column vector with the highest alignment with any coordinate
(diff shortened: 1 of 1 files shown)


dkuegler:
Unfortunately, I copied the code from the wrong file and only had the untested version...
--> I fixed it now :)

Here is my test
```python
@pytest.mark.parametrize(
    argnames=["affine", "axcode"],
    argvalues=[
        [np.diag([2, 3, 4, 1]), "RAS"],
        [np.diag([-2, 3, -4, 1])[:, [1, 0, 2, 3]], "ALI"],
        [np.diag([2, -3, 4, 1])[:, [0, 2, 1, 3]], "RSP"],
        [np.diag([-2, -3, -4, 1])[:, [2, 0, 1, 3]], "ILP"],
    ],
)
def test_aff2axcodes(affine: AffineMatrix4x4, axcode: StrictOrientationType):
    """Test whether aff2axcodes works as expected."""
    actual = "".join(aff2axcodes(affine, ("lr", "pa", "is")))
    expected = axcode.lower()
    assert actual == expected, "aff2axcodes did not return the expected axcode!"
```

My second test is a bit more complex and involves other function that are part of FastSurfer and not nibabel...

PR #1467, opened 2026-02-26
[MAINT] Bump pre-commit hooks, ignore some ruff rules, fix others

Remi-Gau:
- WIP PR
- bumping hooks to see what fails and to discuss what can be auto-fixed, ignored...

PR #1468, opened 2026-02-26
[MAINT] silence some ruff errors

Remi-Gau:
- related to #1467
- simply run pre-commit on all files 
- mostly adds an explicit strict to all zip()
- ignore a couple of other errors that would require manual fixes
- adapts a conditional import that is always true since we are always on >= 3.10

### DIFF
--- a/nibabel/orientations.py
+++ b/nibabel/orientations.py
@@ -279,7 +279,7 @@ def ornt2axcodes(ornt, labels=None):
     ('F', 'L', 'U')
     """
     if labels is None:
-        labels = list(zip('LPI', 'RAS'))
+        labels = list(zip('LPI', 'RAS', strict=False))
     axcodes = []
     for axno, direction in np.asarray(ornt):
         if np.isnan(axno):
@@ -324,7 +324,7 @@ def axcodes2ornt(axcodes, labels=None):
            [ 0., -1.],
            [ 2.,  1.]])
     """
-    labels = list(zip('LPI', 'RAS')) if labels is None else labels
+    labels = list(zip('LPI', 'RAS', strict=False)) if labels is None else labels
     allowed_labels = sum(map(list, labels), [None])
     if len(allowed_labels) != len(set(allowed_labels)):
         raise ValueError(f'Duplicate labels in {allowed_labels}')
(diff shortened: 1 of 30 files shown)


nibabel/viewers.py
```diff
@@ -175,7 +175,7 @@ def __init__(self, data, affine=None, axes=None, title=None):
                 ['center', 'top'],
                 ['right', 'center'],
             ]
-            for pos, anchor, lab in zip(poss, anchors, label):
+            for pos, anchor, lab in zip(poss, anchors, label, strict=False):
```
effigies:
```suggestion
            for pos, anchor, lab in zip(poss, anchors, label, strict=True):
```

nibabel/spaces.py
```diff
@@ -83,7 +83,7 @@ def vox2out_vox(mapped_voxels, voxel_sizes=None):
         if not np.all(np.array(voxel_sizes) > 0):
             raise ValueError('voxel sizes should all be positive')
         out_vox[:n_axes] = voxel_sizes
-    in_mn_mx = zip([0, 0, 0], np.array(in_shape) - 1)
+    in_mn_mx = zip([0, 0, 0], np.array(in_shape) - 1, strict=False)
```
effigies:
```suggestion
    in_mn_mx = zip([0, 0, 0], np.array(in_shape) - 1, strict=True)
```

nibabel/orientations.py
```diff
@@ -279,7 +279,7 @@ def ornt2axcodes(ornt, labels=None):
     ('F', 'L', 'U')
     """
     if labels is None:
-        labels = list(zip('LPI', 'RAS'))
+        labels = list(zip('LPI', 'RAS', strict=False))
```
effigies:
```suggestion
        labels = list(zip('LPI', 'RAS', strict=True))
```

effigies (commented):
I suppose that we should prefer `strict=True` when there is no good reason for the lengths to differ? Or is it more efficient to use `strict=False`, and so we only prefer `strict=True` if they can differ and should error?

I strongly suspect that most could be strict, but here are a few that I have verified manually.

PR #1475, opened 2026-03-06
chore/doc: Run style checks with pre-commit, update docs to match

### LINKED ISSUE #1473: update how to apply style guide
as kind of a follow up to this and to setting up pre-commit ci, should there be a clean up of tox to remove recipes related to [style](https://github.com/nipy/nibabel/blob/1b9ed44d26c05f44e62bc7ce36f4b09dfda1803e/tox.ini#L138) (same for the [documentation](https://nipy.org/nibabel/devel/devguide.html#style-guide) and maybe [pyproject.toml](https://github.com/nipy/nibabel/blob/1b9ed44d26c05f44e62bc7ce36f4b09dfda1803e/pyproject.toml#L86))

_Originally posted by @Remi-Gau in https://github.com/nipy/nibabel/issues/1467#issuecomment-4002842672_

effigies:
This makes `tox -e style|spellcheck` simply call equivalent `pre-commit` commands, which removes the need for pinning.

I think pinning all dependencies in type checking is too important to only pin mypy, so I switched that to a local pre-commit hook.

Rendered docs: https://nibabel-draft--1475.org.readthedocs.build/en/1475/devel/devguide.html#testing

Closes #1473.


Remi-Gau (approved):
makes sense to me

Issue #1392, opened 2024-11-26
XA61A localizer crash


bpinsard:
I got my hands on Siemens XA61A data (Cima.X), and the localizer causes heudiconv (@yarikoptic FYI) to crash in nibabel code.

```
Traceback (most recent call last):                                             
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/bin/heudiconv", line 8, in <module>
    sys.exit(main())                                                           
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/heudiconv/cli/run.py", line 30, in main
    workflow(**kwargs)                                                         
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/heudiconv/main.py", line 410, in workflow
    study_sessions = get_study_sessions(        
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/heudiconv/parser.py", line 221, in get_study_sessions
    seqinfo_dict = group_dicoms_into_seqinfos(
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/heudiconv/dicoms.py", line 342, in group_dicoms_into_seqinfos
    mwinfo = validate_dicom(filename, dcmfilter)                    
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/heudiconv/dicoms.py", line 179, in validate_dicom
    del mw.series_signature[sig]                                                                                                                              
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/nibabel/nicom/dicomwrappers.py", line 886, in series_signature
    signature['image_shape'] = (self.image_shape, eq)
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)                                                  
  File "/home/bpinsard/.virtualenvs/heudiconv_dev/lib/python3.10/site-packages/nibabel/nicom/dicomwrappers.py", line 775, in image_shape
    raise WrapperError("Number of slice indices and positions don't match")
nibabel.nicom.dicomwrappers.WrapperError: Number of slice indices and positions don't match
```

This is a classic localizer with 3 slices in each plane, but packed into a single enhanced dicoms.
Possibly Siemens is not following some dicoms specs.

[...]

PR #1485, opened 2026-03-23
ENH: Multi frame filter to handle stack with multiple orientations

octomike:
Some vendors produce enhanced DICOMs with (logical) frame groups of different orientations sharing the same Stack ID - this breaks in `image_shape`.

This filter groups frames by orientation and returns only the frames of the first logical stack.

Relevant for #1392
---

I basically hit #1392 with some XA70A localizer data from a recent Siemens scanner. Unfortunately filtering in downstream applications (heudiconv) is not possible for us, because we rely on the localizer name to set a session ID (reproin logic).

I'm aware this PR does not solve the issue. Still, it's super helpful for us and maybe to others as well. The code is emitting a warning that frames have been discarded, similar to the two existing filters. Also, I have trouble interpreting the [definition](https://dicom.nema.org/medical/dicom/current/output/html/part03.html#table_C.7.6.16-3) of the Stack ID and I'm still not 100% certain this is actually allowed in DICOM or if our data is non-conformant.

Be aware, that the method was drafted by claude/AI with me reviewing and testing it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "nibabel-data/dcm_qa_xa30"]
 	path = nibabel-data/dcm_qa_xa30
 	url = https://github.com/neurolabusc/dcm_qa_xa30.git
+[submodule "nibabel-data/dcm_qa_xa60"]
+	path = nibabel-data/dcm_qa_xa60
+	url = https://github.com/neurolabusc/dcm_qa_xa60.git
--- a/nibabel/nicom/dicomwrappers.py
+++ b/nibabel/nicom/dicomwrappers.py
@@ -597,7 +597,38 @@ class FilterDwiIso(FrameFilter):
         return frame.MRDiffusionSequence[0].DiffusionDirectionality != 'ISOTROPIC'
 
 
-DEFAULT_FRAME_FILTERS = (FilterMultiStack(), FilterDwiIso())
+class FilterMultiOrient(FrameFilter):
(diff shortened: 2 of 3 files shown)


effigies (commented):
I think I understand this, and it makes reasonable sense to me.

@octomike Can you merge your commit from #1486 into this? I think the shape should be `(512, 512, 7)`, not `(512, 512, 3)`.

It would also be a good use of tests to demonstrate how the feature should actually be used. Here's what I was able to figure out:

```py
    @dicom_test
    @needs_nibabel_data('dcm_qa_xa60')
    def test_data_multi_orient_extract_all(self):
        # Test that a file with 3 orientations sharing the same StackID
        with ImageOpener(DATA_FILE_XA60_MULTI_ORIENT) as fobj:
            dcm_data = pydicom.dcmread(fobj)
        for keep_group, nslices in enumerate([7, 3, 3]):
            dw = didw.wrapper_from_data(
                dcm_data, frame_filters=[didw.FilterMultiOrient(keep_group=keep_group)]
            )
            assert dw.image_shape == (512, 512, nslices)
```

If other ergonomics are intended, please use those.

octomike:
@effigies done and I included your test - makes total sense to test that feature as well

octomike:
Rebased on master (and included the typo fix for `DEFUALT_FRAME_FILTERS`)

effigies (approved):
Great, these tests make sense to me.

@bpinsard @moloney Any concerns on your end?

bpinsard:
No concern on my end.

PR #1486, opened 2026-03-24
TST: fail on multiorientation multiframe localizer

octomike:
As request in #1485

There is no need to add a test file to nitest-dicom - we already have one available in dcm2niix' dcm_qa_xa60.

Is that what you had in mind, @effigies ?


effigies:
This can be closed. Just to be clear about the way I go about these, I would figure out what I want the final behavior to be, write a test, commit and push the test so that everybody can see that it fails. Then add the feature/fix and that makes the test pass, and push that to the same PR.

Issue #927, opened 2020-06-30
MRTrix MIH/MIF format


effigies:
Worth looking into: https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#mrtrix-image-formats

Convertible to a NIfTI-2, but doesn't seem to be entirely volumetric, so might be worth thinking about a non data/affine/header API.

cc @mattcieslak

Lestropie:
I am intending to assign an intern to this project in the near future; if anyone has attempted to make progress on this enhancement but has not indicated such here, please let me know.

mattcieslak:
Perfect timing @Lestropie! we are about to start on this too. Do you know if [this code](https://github.com/MRtrix3/mrtrix3/blob/master/matlab/read_mrtrix.m) still works?

Lestropie:
The MatLab read/write functions should be working on `master`.

What I've not yet looked into is the extent to which the NiBabel API is or is not compatible with the concept of arbitrary strides ([documentation](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#strides); [manuscript](https://www.sciencedirect.com/science/article/pii/S1053811919307281#appsec1)), as this would influence whether or not the data can be mapped directly.

Is this something that you've already allocated resources to ("are" v.s. "were")? Because if so, that would have consequences for my planning of the relevant internship. Otherwise, my intern would be in contact on here for any aspects of NiBable for which I don't have the relevant expertise.

matthew-brett:
On a quick scan, I don't think the strides will be a problem - the trick will be to use the strides in creating a suitable array proxy object - see https://github.com/nipy/nibabel/blob/master/nibabel/arrayproxy.py for the basic definition, and  https://github.com/nipy/nibabel/blob/master/nibabel/brikhead.py#L210 for a customized version for another format.

Lestropie:
@mattcieslak Can I get clarification on whether you "are about to start" as in *intending* to start, or *actually* starting the work? If the latter, I need to alter the project contents for my intern. (Assigning the issue to a user would be a good indicator either way)

mattcieslak:
I was planning in the next few weeks to translate that matlab code to python for a project that reads fixel data from mif files. I wasn't planning on handling all the different strides scenarios though. If you have an employee who would work on this full time that might be the best route for nibabel

Lestropie:
[...]

PR #1489, opened 2026-03-28
Add support for MRtrix MIF file format

tsalo:
Related to #927 (does not close because this doesn't include MIH support). I used Claude to write this for [`ModelArrayIO`](https://github.com/PennLINC/ModelArrayIO), but I don't have much experience with the file format so I'm happy to make any changes.

I don't have any small MIF files that would be appropriate for nibabel's tests on hand, but I can hunt some down.

### DIFF
--- a/nibabel/mif.py
+++ b/nibabel/mif.py
@@ -13,14 +13,15 @@ MIF is the MRtrix imaging format.
 
 from __future__ import annotations
 
+import io
 import sys
 from copy import deepcopy
 
 import numpy as np
 
 from .arrayproxy import ArrayProxy, is_proxy
-from .filebasedimages import FileBasedHeader, SerializableImage
-from .spatialimages import SpatialImage
+from .filebasedimages import SerializableImage
+from .spatialimages import SpatialHeader, SpatialImage
 
 
 def _readline(fileobj) -> bytes:
@@ -179,7 +180,7 @@ def _mif_apply_layout_for_write(data: np.ndarray, layout: list[int]) -> np.ndarr
     return np.ascontiguousarray(data)
 
 
-class MifHeader(FileBasedHeader):
+class MifHeader(SpatialHeader):
     """Header for MIF (.mif / .mif.gz) image files.
 
     The MIF format uses a text header with ``key: value`` pairs followed by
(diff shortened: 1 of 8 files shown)


nibabel/mif.py
```diff
@@ -0,0 +1,617 @@
...
+        raise ValueError(f'Cannot represent numpy dtype {dtype!r} in MIF format')
+    if dtype.itemsize == 1:
+        return base_name
+
+    byte_order = dtype.byteorder
+    if byte_order == '=':
+        byte_order = '<' if sys.byteorder == 'little' else '>'
+    elif byte_order == '|':
+        return base_name
+    return base_name + ('LE' if byte_order == '<' else 'BE')
```
effigies, line 109: A lot of this work is already done in `nibabel.volumeutils`. Building off your first structure:

```python
_dtdefs = [
  (f'{mift}{end}', (label := f'{end and recode[end]}{dt}'), np.dtype(label).type)
  for mift, dt in _MIF_DTYPE_MAP.items()
  for end in ('', 'BE', 'LE')
]

data_type_codes = nb.volumeutils.make_dt_codes(_dtdefs)
```

This allows you to replace your functions:

```python
_mif_dtype_to_str = lambda dtype: data_type_codes[dtype]
_mif_parse_dtype = lambda dtype_str: data_type_codes.dtype[dtype_str]
```

See, for example:

```py
>>> data_type_codes.dtype['Float32']
dtype('float32')

>>> data_type_codes.dtype['Float32BE']
dtype('>f4')

>>> data_type_codes.dtype['Float32LE']
dtype('float32')

>>> data_type_codes[np.dtype('f8')]
'Float64LE'
```

This doesn't do the same validation. Bad names will show up as `KeyError`s, instead of custom-raised `ValueError`s, and you could theoretically have a file with a bad name that matches a numpy dtype string, and that would be accepted.

tsalo:
@mattcieslak any chance you have some very small MIF files (< 50K) that we can use for testing here?

effigies (commented):
I didn't get very far in reviewing this, but I had written a comment. I can try to look at this again on Wednesday, but hitting post in case this is helpful to you now.

nibabel/mif.py
tsalo: Thanks so much!

PR #1496, opened 2026-04-11
ubuntu arm support for unittests in gh actions

### LINKED ISSUE #1410: Add ubuntu-24.04-arm runners
We can now start testing on arm64 for Linux: https://github.com/orgs/community/discussions/148648

leoyala:
This PR adds new ubuntu arm images to the matrix that runs the unittests using tox.

Closes #1410

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -111,7 +111,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


effigies:
Sorry for the direct pushes instead of suggestions. GitHub PR threads are broken right now, as you might have seen.

Thanks for pushing on this! I went ahead and did some stylistic cleanups here, since editing this file is annoying. Hopefully it's a little nicer to edit (and more readable, though that's subjective) in a more compact format.

PR #1509, opened 2026-05-22
Set the compression level to match one of the running Python's

befeleme:
Attempt at fixing #1477
Ran the test suite in Fedora with both Python 3.14 and 3.15 and both runs were successful.


effigies:
Perfect! Would you care to add yourself as a contributor to the CITATION.cff file?

effigies (approved):

PR #1522, opened 2026-07-02
FIX: Set `patient_birth_date` dtype in ecat header to signed int32 (gh-1434)

### LINKED ISSUE #1434: patient_birth_date dtype should be signed integer in nibabel.ecat
Hi,  

I noticed that in [`nibabel/ecat.py`](https://github.com/nipy/nibabel/blob/master/nibabel/ecat.py), the `patient_birth_date` field is defined as:  

```python
('patient_birth_date', np.uint32),
```  

This causes incorrect date interpretations, for example producing birth dates in the future rather than before the Unix epoch. It seems this field should be a **signed integer** (`np.int32`) to correctly handle dates.  

At least in the ECAT files I have, `patient_birth_date` **must be signed integers** to interpret dates correctly. Tools like Vinci read these files correctly when treated as signed.  

**Example:**  
```python
import numpy as np
import datetime, calendar

# 19 Dec 1915 at midnight UTC
ts = calendar.timegm(datetime.datetime(1915, 12, 19).timetuple())

# Structured array with wrong dtype
dtype_wrong = [('patient_birth_date', np.uint32)]
arr_wrong = np.array([(ts,)], dtype=dtype_wrong)

# Structured array with correct dtype
dtype_right = [('patient_birth_date', np.int32)]
arr_right = np.array([(ts,)], dtype=dtype_right)

# Convert back to datetime
dt_wrong = datetime.datetime.utcfromtimestamp(arr_wrong['patient_birth_date'][0])
dt_right = datetime.datetime.utcfromtimestamp(arr_right['patient_birth_date'][0])

print("Unsigned (wrong):", arr_wrong['patient_birth_date'][0], "->", dt_wrong)
print("Signed (correct):", arr_right['patient_birth_date'][0], "->", dt_right)
```  

**Example output:**  
```
Unsigned (wrong): 2589690496 -> 2052-01-24 06:28:16
Signed (correct): -1705276800 -> 1915-12-19 00:00:00
```  

Changing it to `np.int32` fixes this issue. I can also prepare a **PR** for this update if that’s helpful.  

Thanks!

CedricConday:
## Summary

Fixes #1434.

`nibabel/ecat.py` types `patient_birth_date` in the main header as `np.uint32`. The field is a `time_t` (seconds since the Unix epoch), which is signed — negative for dates before 1970. With the unsigned type, any pre-1970 birth date is reinterpreted as a far-future date:

```python
# 19 Dec 1915  ->  ts = -1704153600
uint32(int32(ts)) -> 2590813696  -> 2052-02-06   # bug
int32(ts)         -> -1704153600 -> 1915          # correct
```

The issue reporter confirmed against 100+ ECAT v7.0 and v7.3 files that unsigned is never correct for their data, and other tools (e.g. Vinci) read the field as signed. There is a single `main_header_dtd` (no version-specific variants), so the change applies uniformly.

## Changes

- `nibabel/ecat.py`: `patient_birth_date` dtype `np.uint32` -> `np.int32`, with a comment noting the `time_t` semantics.
- `nibabel/tests/test_ecat.py`: add `test_patient_birth_date_signed` — asserts the field dtype is signed and that a pre-1970 timestamp round-trips through the header unchanged.

## Testing

```
$ pytest nibabel/tests/test_ecat.py -q
32 passed
```

### DIFF
--- a/nibabel/ecat.py
+++ b/nibabel/ecat.py
@@ -90,7 +90,9 @@ main_header_dtd = [
     ('patient_age', np.float32),
     ('patient_height', np.float32),
     ('patient_weight', np.float32),
-    ('patient_birth_date', np.uint32),
+    # time_t (seconds since the Unix epoch); signed so that birth dates
+    # before 1970 are representable as negative values (see gh-1434)
+    ('patient_birth_date', np.int32),
     ('physician_name', '32S'),
     ('operator_name', '32S'),
     ('study_description', '32S'),
--- a/nibabel/tests/test_ecat.py
+++ b/nibabel/tests/test_ecat.py
@@ -77,7 +77,9 @@ class TestEcatHeader(tws._TestWrapStructBase):
         # patient_birth_date is a time_t (seconds since the Unix epoch) and
         # must be signed so that pre-1970 birth dates survive as negative
         # values rather than wrapping to a far-future unsigned value (gh-1434).
-        assert np.issubdtype(self.header_class.template_dtype['patient_birth_date'], np.signedinteger)
+        assert np.issubdtype(
+            self.header_class.template_dtype['patient_birth_date'], np.signedinteger
+        )
         hdr = self.header_class()
         # 19 Dec 1915, i.e. -1704153600 seconds from the epoch
         hdr['patient_birth_date'] = -1704153600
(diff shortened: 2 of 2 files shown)


effigies (approved):
LGTM, but can you run `ruff format`?

effigies:
Also, please feel free to add yourself to https://github.com/nipy/nibabel/blob/master/CITATION.cff (at the end is fine, the order will be updated during release).

CedricConday:
Done — pushed the `ruff format` pass. Thanks for the review!

effigies:
Thanks!

PR #1534, opened 2026-07-26
BF: Resolve slice starts against axis length in slice_affine

### LINKED ISSUE #1533: img.slicer produces a wrong affine for negative starts and reversed slices
`SpatialFirstSlicer.slice_affine` takes the slice's `start` and puts it straight into the translation column of the voxel-to-voxel transform:

```python
# nibabel/spatialimages.py:459
transform[i, i] = subslicer.step if subslicer.step is not None else 1
transform[i, 3] = subslicer.start or 0
```

That holds only when `start` is a non-negative integer and the step is positive. Two cases break it:

1. **A negative start is written verbatim.** `img.slicer[-2:]` on a 10-voxel axis puts `-2` into the affine where voxel `8` belongs. `check_slicing` runs the slicer through `canonical_slicers`, which normalises negative *integer indices* but leaves `start` and `stop` inside `slice` objects untouched.

2. **A negative step with no explicit start.** The first voxel is then `n-1`, but `subslicer.start or 0` yields `0`, so `img.slicer[::-1]` gets the origin wrong.

Either way the returned image holds exactly the right data with an affine that places it somewhere else in world space. Nothing warns or raises.

## Reproduction

```python
import numpy as np
import nibabel as nib
from nibabel.affines import apply_affine

shape = (10, 4, 4)
data = np.arange(int(np.prod(shape)), dtype=np.int32).reshape(shape)
affine = np.diag([-2.0, 2.0, 2.0, 1.0])
affine[:3, 3] = [90.0, -126.0, -72.0]
img = nib.Nifti1Image(data, affine)
lookup = {int(v): ijk for ijk, v in np.ndenumerate(data)}

def check(label, sl):
    simg = img.slicer[sl]
    sdata = np.asanyarray(simg.dataobj)
    worst = 0.0
    for ijk, v in np.ndenumerate(sdata):
        got = apply_affine(simg.affine, ijk)
        want = apply_affine(img.affine, lookup[int(v)])
        worst = max(worst, float(np.max(np.abs(got - want))))
    print(f'  {"ok  " if worst < 1e-9 else "FAIL"}  {label:<20} max RAS+ error {worst:7.3f} mm')
    return sdata

a = check('img.slicer[8:]',  (slice(8, None),))
b = check('img.slicer[-2:]', (slice(-2, None),))
print('  identical data:', np.array_equal(a, b))

c = check('img.slicer[9::-1]',  (slice(9, None, -1),))
d = check('img.slicer[::-1]',   (slice(None, None, -1),))
e = check('img.slicer[-1::-1]', (slice(-1, None, -1),))
print('  identical data:', np.array_equal(c, d) and np.array_equal(d, e))
```

Output on `main` (`4704d904b`):

```
  ok    img.slicer[8:]       max RAS+ error   0.000 mm
  FAIL  img.slicer[-2:]      max RAS+ error  20.000 mm
  identical data: True
  ok    img.slicer[9::-1]    max RAS+ error   0.000 mm
  FAIL  img.slicer[::-1]     max RAS+ error  18.000 mm
  FAIL  img.slicer[-1::-1]   max RAS+ error  20.000 mm
  identical data: True
```

`img.slicer[2:-2]` and `img.slicer[::2]` both pass, so a negative *stop* is fine. It is the start that goes wrong.

This also reaches `nib-roi`, which advertises `-i I1:I2[:-1]  Start/stop [flip] along first axis` and ships a `sanitize()` helper so argparse will accept negative starts.

## Why this looks like a bug rather than intended behaviour

Three spellings of the same operation return byte-identical data and three different affines. Only `img.slicer[9::-1]` matches ground truth. The same goes for `img.slicer[8:]` against `img.slicer[-2:]`.

The documented contract is RAS+ invariance. From `doc/source/nibabel_images.rst`:

> The `slicer` attribute provides an array-slicing interface to produce new images with an appropriately adjusted header, such that the data at a given RAS+ location is unchanged.

and `slice_affine`'s own docstring says it "adjusts the intercept to account for any cropping".

The worked example further down that same page is affected. Under the heading "an image can be flipped along an axis, maintaining an appropriate affine matrix", it runs `ras = img.slicer[::-1]` and prints an X translation of `117.86`, which is the same value as the unflipped image. The correct value is `-136.14`, so the affine printed in the docs is off by 254 mm.

## Why the test suite does not catch it

`nibabel/cmdline/tests/test_roi.py` does exercise `-j -1:1:-1`, which has both a negative start and a negative step, but it asserts `np.allclose(in_sliced.affine, out_img.affine)`. That compares the implementation against itself, so it passes either way.

The randomised slicing test in `test_spatialimages.py` checks data equality only and never looks at the affine, and its `slice_elems` list has no negative start and no negative step. The affine assertions nearby use non-negative starts.

## Suggested fix

Resolve each spatial subslicer against its axis length before building the transform, so `subslicer.indices(self.img.shape[i])` supplies a concrete start. That keeps the change local.

Making `canonical_slicers` normalise negative bounds inside `slice` objects would fix it further upstream, and its docstring arguably promises that already, but it would also change `predict_shape` and `fileslice`, so it seems like the riskier option.

The doctest at `doc/source/nibabel_images.rst` needs updating in the same change, since the affine it prints is the current incorrect value.

I have a regression test ready that asserts, voxel by voxel, that `simg.affine @ [i,j,k,1]` equals `img.affine @ [orig_i,orig_j,orig_k,1]`, parametrised over the equivalent spellings above. It fails on the negative-start and bare `::-1` cases today. Happy to open a PR if this approach looks right to you.

## Versions

Reproduced on `main` at `4704d904b`, and the same line is unchanged in 5.3.2, 5.2.1 and 4.0.2. It has been there since `slicer` landed in #550.

```
nibabel 5.5.0.dev117+g4704d904b
numpy 2.5.1
Python 3.12
```

wterrr:
`SpatialFirstSlicer.slice_affine` copied `subslicer.start` straight into the affine translation. Two cases were therefore wrong:

* A negative start (``img.slicer[-2:]``) put the negative value into the affine instead of the equivalent positive index.
* A negative step with an implicit start (``img.slicer[::-1]``) used 0 as the origin, but the first voxel of the result is the *last* voxel of the axis.

In both cases the returned data was correct while the affine was not, so the image was silently shifted in RAS+ space. ``img.slicer[-2:]`` and ``img.slicer[8:]`` select the same voxels but produced affines 20 mm apart. `nib-roi` is affected too, since it accepts negative starts and ``:-1`` steps and calls `slice_affine` directly.

Resolve each spatial subslicer with `slice.indices()` so the transform is built from the concrete start and step that `dataobj[slicer]` uses.

Closes gh-1533.


Arthur031221:
@wterrr The three follow-on pieces are up as wterrr/nibabel#1 against this branch, opened on 27 July: the doctest update, an independent affine assertion in `test_nib_roi`, and one line covering `slice_affine` directly.

Flagging them here because two of them affect what this PR leaves behind:

`doc/source/nibabel_images.rst` line 352 prints the X translation for `img.slicer[::-1]` as `117.86`, which is the pre-fix value. Once this merges the correct value is `-136.14` and that doctest starts failing. The doctest job in `test.yml` runs with `continue-on-error: true`, so CI would stay green and the wrong affine would sit in the docs.

`test_nib_roi` asserts `np.allclose(in_sliced.affine, out_img.affine)`, which compares the slicer against itself. It already exercises `-j -1:1:-1`, a negative start with a negative step, and it passes either way. That is why nothing caught this. Computing the expected affine independently makes it catch it: with the fix here reverted, the old assertion still gives 12 passed and the new one gives 2 failed.

The third is one line asserting `slice_affine` directly rather than only through `img.slicer`, since the affine calculator and the slicer were asked for separately in the issue.

Whichever is least work for you. If you would rather not take a PR into your branch, I can open the doctest and `test_roi` changes as a separate PR against master to merge after this one. They cannot go in before, since the doctest only becomes correct once this fix is in. The `slice_affine` line sits inside the test added here, so it belongs either on this PR or on a follow-up.

No rush from my side. I just did not want the doctest to slip through quietly.

PR #1538, opened 2026-08-06
docs: clarify filename parameter types in load docstring

meryemozubek:


effigies:
Thanks for this proposal. I don't think we can accept it as-is, since we don't accept `bytes` objects. I'm not sure what's confusing about the current text that is motivating this, as the variable is named `filename` and it can be either a `str` or an `os.PathLike`.

PR #1539, opened 2026-08-20
BF: Handle frame indices where the slice index is not first (gh-1388)

### LINKED ISSUE #1388: nicom MultiframeWrapper fails to return image_shape attribute with multi-volume data
Does the nibabel.nicom dicom wrapper support multi-frame dicoms with multiple volumes?

I have a dicom with frame index ordering: "Temporal Position Index", "Stack ID", "In Stack Position Number".  Loading the dicom and attempting to access the image_shape attribute
```python
mw = dw.wrapper_from_file(filepath, force=True, stop_before_pixels=True)
mw.image_shape
```
fails with error:
```
  File "/home/akuurstr/.config/JetBrains/PyCharm2024.2/scratches/heudiconv_bruker_dicom_conversion.py", line 65, in <module>
    mw.image_shape
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/virtualization/python_virtualenvs/heudiconv/lib/python3.10/site-packages/nibabel/nicom/dicomwrappers.py", line 770, in image_shape
    raise WrapperError('Non-singular index precedes the slice index')
nibabel.nicom.dicomwrappers.WrapperError: Non-singular index precedes the slice index
```

CedricConday:
Closes #1388.

## Problem

`MultiframeWrapper.image_shape` rejects any file whose `DimensionIndexSequence`
places a non-singular index before `InStackPositionNumber`:

```
WrapperError: Non-singular index precedes the slice index
```

The Bruker enhanced DICOM in #1388 has frame index order
`TemporalPositionIndex`, `StackID`, `InStackPositionNumber`, so the series
cannot be read at all.

That layout is conformant. DICOM PS3.3
[C.7.6.17](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.17.html)
ties the value order to the sequence order — *"The ordering of the Items in the
Sequence defines the ordering in the Dimension Index Values Attribute"* — and
leaves that choice to the writer: *"It is up to the generating applications to
decide what Attributes are important to describe the multi-frame dimensions"*,
and *"The actual order of the Frames in the object is up to the generating
application."* Nothing requires `InStackPositionNumber` to come first.

## Why the restriction exists

There is a real invariant behind it:

* the shape gains one axis per surviving column of `frame_indices`, **in column
  order**;
* `frame_order` sorts with `np.lexsort(self._frame_indices.T)`, where the
  **last** column is the primary key;
* `get_unscaled_data` reshapes with `order='F'`, so frames must sort
  slice-fastest.

Those agree only when the slice column comes first, which today happens by
accident: in the layouts that work, the columns preceding the slice index are
singular and get dropped.

This is worth stating because it constrains the fix. Simply relaxing the
`len(shape) > 2` check returns the correct shape with **silently transposed
volumes**, since the lexsort would then take the slice column as its primary
key. That is worse than the exception.

## Fix

The shape calculation moves into `_calc_shape`, and `image_shape` retries it
with the slice column moved to the front **only if** the original column order
raises. The reorder keeps the shape axes aligned with the index columns, so the
retried files also unpack in the right order.

Scoping it as a retry rather than an unconditional reorder is deliberate: a file
that is handled today never raises, so it never reaches the fallback, and its
shape, frame order and data are bit-identical to before. I first wrote this as
an unconditional reorder and it caused three regressions — a file with uneven
`InStackPositionNumber` values became a hard `WrapperError: Missing slices`, a
single-slice multiframe gained a spurious singleton axis, and one layout was
silently reinterpreted. All three now match `master` exactly.

## Behaviour change

`test_shape` previously asserted that this layout raises:

```python
# Check non-singular dimension preceding slice dim raises
```

It now asserts `(32, 64, 3, 2)`. That is the point of the issue, and matches
@moloney's reading on the thread:

> The code currently assumes we don't need to reorder the DimensionIndexValues,
> mostly out of expediency and lack of a clear need. Your dataset appears to
> show that we do need to check for and allow this.

The class docstring is updated, since the remaining axes are no longer strictly
"order preserved". No `Changelog` entry — that file looks to be maintained by
the release manager rather than per-PR; happy to add one if you'd prefer.

## Tests

The issue stalled waiting on a sample dataset, so this uses the existing
`fake_shape_dependents` harness and needs no external data:

* `test_shape` — the previously rejected layout returns `(32, 64, 3, 2)`.
* `test_data_fake_slice_dim_not_first` — checks the **data**, for a 4D volume
  with the temporal index ahead of the slice index, in acquisition order and
  with the frames scrambled.
* `test_data_fake_slice_dim_not_first_5d` — two differently sized non-slice
  dimensions after the slice index, so misordering *among* the non-slice axes is
  caught too.

Verified failing on `master` and passing here. I also sabotaged the fix to
confirm the tests bite: reversing the non-slice column order and disabling the
retry each fail the suite. `nibabel/nicom` is green serially and under
`--parallel-threads 16`; the full suite is 5288 passed, 270 skipped.

---

_AI-assisted, human-reviewed._

### DIFF
--- a/nibabel/nicom/dicomwrappers.py
+++ b/nibabel/nicom/dicomwrappers.py
@@ -33,6 +33,14 @@ class WrapperError(Exception):
     pass
 
 
+class SliceIndexOrderError(WrapperError):
+    """Raised when a non-singular index precedes the slice index
+
+    Internal to `MultiframeWrapper.image_shape`, which retries with the slice
+    index moved to the front.
+    """
+
+
 class WrapperPrecisionError(WrapperError):
     pass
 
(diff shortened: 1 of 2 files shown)


CedricConday:
The red check here isn't the diff — `test-package (archive)` fails at packaging:

```
error: Source distribution `archive/nibabel-archive.tgz` has a non-PEP 625-compliant filename; only `.tar.gz` and `.zip`
```

`sdist` and `wheel` are then cancelled as a result. The same failure is on #1540, which is an unrelated change, so it looks like the workflow's archive step rather than either branch. Happy to rebase once that's sorted.

PR #1540, opened 2026-08-20
BF: Preserve MGH tag data across an unmodified round-trip (gh-1402)

### LINKED ISSUE #1402: nibabel copying mgz header incorrectly
I am currently using nibabel 5.3.2.  I wrote a script to makes an edit to an mgz image. After running the script I noticed that nibabel adds what looks like rounding error to some of the mgz header. Below is my whole script. 

```python
import nibabel as nb
import numpy as np
import sys

oldfilledmgz = nb.freesurfer.mghformat.load(sys.argv[1])
newfilledmgz = nb.freesurfer.mghformat.load(sys.argv[2])
WMmgz        = nb.freesurfer.mghformat.load(sys.argv[3])

oldfilled=np.asanyarray(oldfilledmgz.dataobj)
newfilled=np.asanyarray(newfilledmgz.dataobj)
WM=np.asanyarray(WMmgz.dataobj)

if not np.equal(oldfilled,newfilled):
    WM[(oldfilled == 0) & ((newfilled == 127) | (newfilled==255)) ] = 255
    WM[(newfilled == 0) & ((oldfilled == 127) | (oldfilled==255)) ] = 1
    WMnew = nb.freesurfer.mghformat.MGHImage(WM, WMmgz.affine, header=WMmgz)
    WMnew.set_filename(sys.argv[3])
    nb.save(WMnew, sys.argv[3])
```

CedricConday:
Closes #1402.

## Problem

FreeSurfer may write variable-length tag data after the MGH footer —
`TAG_CMDLINE` history, `TAG_AUTO_ALIGN`, `TAG_MRI_FRAME`, embedded colour
tables. `MGHHeader.from_fileobj` reads exactly `footer_dtype.itemsize` bytes and
`writeftr_to` writes back only that struct, so everything after the footer is
silently dropped on a load/save round-trip.

This is reproducible with the file already in the repo:

```python
import gzip, tempfile, nibabel as nib
d = tempfile.mkdtemp()
img = nib.load('nibabel/tests/data/test.mgz')
img.to_filename(f'{d}/rt.mgz')
len(gzip.open('nibabel/tests/data/test.mgz', 'rb').read())  # 23215
len(gzip.open(f'{d}/rt.mgz', 'rb').read())                  # 784  -> 22431 bytes lost
```

That file's tags are `[(41, 7), (42, 22400)]` — `TAG_PEDIR` and `TAG_MRI_FRAME`.

It presents confusingly, which is probably why the original thread stalled: the
header compares as identical in both nibabel and `mri_info`, and only the file
size differs. That is exactly what the reporter described — *"Both nibabel and
FreeSurfer's mri_info says that both images have identical headers. However
after decompressing the mgz to mgh I noticed the size of both files are not the
same."* — and it is consistent with their `recon-all` failure.

## Fix

Retain the trailing bytes on read and write them back after the footer. They are
not interpreted.

**Tags are not portable between volumes.** `TAG_MRI_FRAME` stores per-frame
acquisition parameters — 11200 bytes per frame in the bundled file — and
`TAG_AUTO_ALIGN` holds a geometry-specific transform. Writing them next to a
different volume produces a file FreeSurfer will misparse: it would read
per-frame state for frames that no longer exist. So the bytes are recorded
together with the header state they were read with, and written back only while
that header is unchanged:

```python
img = nib.load('.../test.mgz')          # nframes 2, 22431 tag bytes
nib.funcs.four_to_three(img)[0]         # nframes 1 -> tags dropped, not rewritten
img.slicer[:2, :2, :2]                  # tags dropped
img.to_filename(...)                    # unchanged header -> all 22431 bytes preserved
```

Editing voxel data while leaving the header alone — the case in the issue — keeps
the tags.

Two supporting changes:

* The read is bounded by `MAX_TAG_BYTES`. A file whose `dims` field understates
  the data makes the voxel data itself look like tag data; unbounded, a 597 KB
  `.mgz` can expand to over a gigabyte resident, on what is otherwise a lazy
  header read. The bound is deliberately generous (64 MiB) because
  `TAG_MRI_FRAME` grows with the frame count — a long BOLD run legitimately
  carries megabytes of tags. Over the bound, it warns and keeps none.
* `writeftr_to` truncates. Writing a smaller image over a reused stream (e.g.
  `to_file_map` with a `BytesIO`) previously left bytes from the larger image
  behind; harmless before, but they would now be read back as tag data and
  become sticky. `ImageOpener` gains a `truncate` passthrough for this.

## Tests

* `test_tag_data_roundtrip` — byte-exact round-trip; tags survive a data-only
  edit; tagless files are untouched.
* `test_tag_data_not_reused_across_volumes` — uses the bundled `test.mgz`, and
  asserts tags are dropped for `four_to_three`, a reshape and a dtype change.
* `test_tag_data_bounded` — over-long trailing region warns and keeps no tags.
* `test_tag_data_not_laundered_from_reused_stream` — the truncation case.

Verified against `master`: the round-trip above loses 22431 bytes there and 0
here. Full suite 5290 passed, 270 skipped; `nibabel/freesurfer` is also green
under `--parallel-threads 16`.

One judgement call worth flagging: because the fingerprint is the header, editing
only voxel data keeps the tags. That is right for `TAG_MRI_FRAME` and
`TAG_CMDLINE`, but a segmentation whose labels are rewritten in place would keep
a colour table that no longer matches. Distinguishing those needs the tags to be
parsed, which is a larger change than this. Happy to go further if you'd prefer.

---

_AI-assisted, human-reviewed._

### DIFF
--- a/nibabel/freesurfer/mghformat.py
+++ b/nibabel/freesurfer/mghformat.py
@@ -11,6 +11,8 @@
 Author: Krish Subramaniam
 """
 
+import warnings
+from io import UnsupportedOperation
 from os.path import splitext
 
 import numpy as np
@@ -29,6 +31,12 @@ from ..wrapstruct import LabeledWrapStruct
 # mgh header
 # See https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/MghFormat
 DATA_OFFSET = 284
+# Upper bound on the tag data retained after the footer.  This has to be generous:
+# TAG_MRI_FRAME stores per-frame acquisition parameters, so a long BOLD run
+# legitimately carries megabytes of tags.  The bound exists only to stop a file whose
+# `dims` field understates the data -- which makes the voxel data itself look like
+# tag data -- from pulling an unbounded amount into memory.
+MAX_TAG_BYTES = 2**26
 # Note that mgh data is strictly big endian ( hence the > sign )
 # fmt: off
 header_dtd = [
@@ -98,6 +106,19 @@ class MGHHeader(LabeledWrapStruct, SpatialHeader):
 
     The header also consists of the footer data which MGH places after the data
     chunk.
+
+    FreeSurfer may write additional, variable-length tag data after the footer
+    (e.g. ``TAG_CMDLINE`` history, auto-align matrices, embedded colour tables).
+    That data is not interpreted here, but is retained verbatim in ``_tags`` so
+    that an otherwise unmodified load/save round-trip does not discard it.
+
+    Tags are *not* portable between volumes.  Several are tied to the volume they
+    were written with -- ``TAG_MRI_FRAME`` carries per-frame state sized by the
+    frame count, and ``TAG_AUTO_ALIGN`` holds a geometry-specific transform -- so
+    writing them alongside a different volume yields a file FreeSurfer will
+    misparse.  The bytes are therefore recorded together with the header state they
+    were read with (``_tags_binaryblock``) and are only written back while the
+    header is unchanged; see ``_tags_valid``.
     """
 
     # Copies of module-level definitions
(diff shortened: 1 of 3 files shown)


CedricConday:
Same as on #1539 — `test-package (archive)` fails on the archive filename:

```
error: Source distribution `archive/nibabel-archive.tgz` has a non-PEP 625-compliant filename; only `.tar.gz` and `.zip`
```

Not from this diff, which doesn't touch packaging. Happy to rebase whenever it's convenient.

